PR #4, opened 2016-08-23
Add dependency on the compiler in //tools/modules:compiler_info

endobson:
This fixes the issue I was running into where `tools/modules/update.sh` wasn't working.


bzz:
Just FYI - according to the project's [docs on contributing](http://www.kythe.io/contributing#code-review):

> All changes to Kythe must go through code review before being submitted,
> The Kythe team has chosen to use a Phabricator instance located at https://kythe.io/phabricator for code reviews.

to get it reviewd you might need to close this PR and submit changes to [Phabrica†or]( https://kythe.io/phabricator) instead.

shahms:
It doesn't look like this is still relevant?

PR #8, opened 2017-02-21
Add linkopts=[-lpthread] for gflag and glog

j3parker:
This matches the change in gflags' BUILD files: https://github.com/gflags/gflags/pull/189
It was required for Kythe to build on my desktop (Ubuntu 16.10)

Aside: gflags natively supports Bazel. Would it be reasonable for a follow-up PR to switch to using it directly via a `git_repository` rule vs. the custom BUILD file in this repo?


j3parker:
Just saw the note about Phabricator. Will submit there.

PR #18, opened 2017-12-11
Update the build rule for pkg-config

yuikns:
For the purpose of make it works better in macOS, I wish to submit this
patch.

1. symlinks

In libmongocxx, the structure of the headers are seems as follow:

```
.
├── bsoncxx
│   └── v_noabi
│       └── bsoncxx
└── mongocxx
    └── v_noabi
        └── mongocxx
```

And the cflags are like this:

```
$ pkg-config --cflags libmongocxx
-I/usr/local/Cellar/mongo-cxx-driver/3.1.3/include/mongocxx/v_noabi \
  -I/usr/local/Cellar/mongo-c-driver/1.8.2/include/libmongoc-1.0 \
  -I/usr/local/Cellar/mongo-cxx-driver/3.1.3/include/bsoncxx/v_noabi \
  -I/usr/local/Cellar/mongo-c-driver/1.8.2/include/libbson-1.0
```

There are 2 header directory using basename `v_noabi` but contains
different subdirectories.

This patch will check the basename, and add a suffix for duplicated
basenames.

Besides, it will change nothing if there is no conflict basenames.

2. linkopts

If we try to parse the static libs in macOS, we may get result as
follow:

```
$ pkg-config --static --libs libmongocxx
-L/usr/local/Cellar/mongo-cxx-driver/3.1.3/lib \
  -L/usr/local/Cellar/mongo-c-driver/1.8.2/lib \
  -L/usr/local/Cellar/mongo-cxx-driver/3.1.3/lib \
  -L/usr/local/Cellar/mongo-c-driver/1.8.2/lib \
  -lmongocxx -lmongoc-1.0 -lsasl2 \
  -framework Security \
  -framework CoreFoundation \
  -lz -lbsoncxx -lbson-1.0
```

Note: the option '-framework' and 'Security/CoreFoundation' should be
connected in macOS.

### DIFF
--- a/tools/build_rules/config/pkg_config.bzl
+++ b/tools/build_rules/config/pkg_config.bzl
@@ -82,7 +82,13 @@ def _linkopts(repo_ctx, pc_args):
   if result.return_code != 0:
     return _fail(repo_ctx, "Unable to determine linkopts", result.stderr)
   stdout = result.stdout
-  return success([arg for arg in stdout.strip().split(" ") if arg])
+  args = []
+  for arg in [arg for arg in stdout.strip().split(" ") if arg]:
+    if arg[0] == "-" or len(args) == 0:
+      args += [arg]
+    else:
+      args[-1] += (" " + arg)
+  return success(args)
 
 def _extract_prefix(flags, prefix):
   stripped, remain = [], []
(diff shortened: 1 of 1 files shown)


yuikns:
I signed it!

creachadair:
Since we no longer have this file in the repo, I'm going to go ahead and close this PR, but thanks for offering the change.

PR #20, opened 2018-05-08
Reduce JarDetails redundancy by referring to each enclosing jar by index

schroederc:


kythe/java/com/google/devtools/kythe/analyzers/jvm/ClassFileIndexer.java
```diff
@@ -143,16 +143,37 @@ private static void analyzeCompilation(
         }
       }
     }
-    Map<FileInfo, VName> map = new HashMap<>();
+    List<VName> index = new ArrayList<>();
```
danielmoy-google: If it doesn't need to mutate later, consider ImmutableList.Builder.  I didn't see it mutating, but I also didn't exhaustively check the return usage...

I don't know how we feel about unspeakable lambdas here, but that's also an option: return jarDetails.getJarList().stream().map(j -> j.getVName()).collect(ImmutableList.toImmutableList());

PR #27, opened 2018-05-22
Refactor extractor interface to use a container.

danielmoy-google:
Feedback from previous phabricator review D2323.

### DIFF
--- a/kythe/go/extractors/config/extractor.go
+++ b/kythe/go/extractors/config/extractor.go
@@ -29,17 +29,16 @@ import (
 // kytheConfigFileName The name of the Kythe extraction config
 const kytheExtractionConfigFile = ".kythe-extraction-config"
 
-// Repo is a loose container of input/output parameters for doing extraction on
-// remote repositories.
+// Repo is a container of input/output parameters for doing extraction on remote
+// repositories.
 type Repo struct {
-	// URI is the remote location of the repo itself.
+	// The remote location of the repo itself.
 	URI string
-	// OutputPath is where to write any relevant output from an extraction
-	// call.
+	// Where to write from an extraction.
 	OutputPath string
-	// ConfigPath is an optional path to a
-	// kythe.proto.ExtractionConfiguration object that will detail how
-	// extraction is performed.
+	// An optional path to a file containing a
+	// kythe.proto.ExtractionConfiguration encoded as JSON that details how
+	// to perform extraction.
 	ConfigPath string
 }
 
--- a/kythe/go/extractors/config/extractor.go
+++ b/kythe/go/extractors/config/extractor.go
@@ -29,13 +29,27 @@ import (
 // kytheConfigFileName The name of the Kythe extraction config
 const kytheExtractionConfigFile = ".kythe-extraction-config"
 
+// Repo is a loose container of input/output parameters for doing extraction on
+// remote repositories.
+type Repo struct {
+	// URI is the remote location of the repo itself.
+	URI string
+	// OutputPath is where to write any relevant output from an extraction
+	// call.
+	OutputPath string
+	// ConfigPath is an optional path to a
+	// kythe.proto.ExtractionConfiguration object that will detail how
+	// extraction is performed.
+	ConfigPath string
+}
+
 // Extractor is the interface for handling kindex generation on repos.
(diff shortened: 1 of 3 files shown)


kythe/go/extractors/config/extractor.go
```diff
@@ -29,21 +29,35 @@ import (
 // kytheConfigFileName The name of the Kythe extraction config
 const kytheExtractionConfigFile = ".kythe-extraction-config"
 
+// Repo is a loose container of input/output parameters for doing extraction on
```
creachadair: s/loose//

kythe/go/extractors/config/extractor.go
```diff
@@ -29,21 +29,35 @@ import (
...
+// remote repositories.
+type Repo struct {
+	// URI is the remote location of the repo itself.
```
creachadair: [optional] It's not essential to use the "starts with name" convention for struct fields (though it is fine if you prefer it).

kythe/go/extractors/config/extractor.go
```diff
@@ -29,21 +29,35 @@ import (
...
+	URI string
+	// OutputPath is where to write any relevant output from an extraction
```
creachadair:
```
// Where to write output from an extraction
```

kythe/go/extractors/config/extractor.go
```diff
@@ -29,21 +29,35 @@ import (
...
+	// call.
+	OutputPath string
+	// ConfigPath is an optional path to a
```
creachadair:
```
// An optional path to a file containing a kythe.proto.ExtractionConfiguration
// encoded as JSON that details how to perform extraction.
```

(exact wording not crucial: Key point is to mention the file format, and that this is a file path rather than the object itself)

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -133,16 +134,21 @@ func (g harness) TestRepo(repo string) (Result, error) {
...
 
 	repoDir, err := ioutil.TempDir("", repoName)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create temp dir for repo %s: %v", repoURL, err)
+		return nil, fmt.Errorf("failed to create temp dir for repo %s: %v", repoURI, err)
 	}
 	defer os.RemoveAll(repoDir)
 
-	if err = g.repoFetcher.Fetch(repoURL, repoDir); err != nil {
+	repo := config.Repo{
```
creachadair:
```
	if err = g.repoFetcher.Fetch(config.Repo{
		URI:        repoURI,
		OutputPath: repoDir,
	}); err != nil {
		return nil, err
	}
```

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -163,16 +169,20 @@ func (g harness) filenamesFromRepo(repoURL string) (map[string]bool, error) {
...
 
-	err = g.extractor.ExtractRepo(repoURL, tmpOutDir, g.configPath)
 	ret := map[string]bool{}
+	err = g.extractor.ExtractRepo(config.Repo{
+		URI:        repoURI,
+		OutputPath: tmpOutDir,
+		ConfigPath: g.configPath,
+	})
 	if err != nil {
 		return ret, err
```
creachadair: Can this be more simply `return nil, err` given that `ret` has nothing in it? (A `nil` map behaves like an empty map for reads)

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: Ah cool for some reason I figured the newline breaking in the if statement wasn't kosher.

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -177,15 +175,15 @@ func (g harness) filenamesFromExtraction(repoURI string) (map[string]bool, error
 	}
 	defer os.RemoveAll(tmpOutDir)
 
-	ret := map[string]bool{}
 	err = g.extractor.ExtractRepo(config.Repo{
```
creachadair: `if err := …; err != nil {`

PR #28, opened 2018-05-22
Add package cmdutil.

creachadair:
This package contains shared logic for setting up command-line tools using the
subcommands package. The cmdutil.Info type handles reporting name, synopsis,
and usage information, and provides some default behaviour for other methods of
the required interface.

### DIFF
--- a/kythe/go/util/cmdutil/cmdutil_test.go
+++ b/kythe/go/util/cmdutil/cmdutil_test.go
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


kythe/go/util/cmdutil/cmdutil_test.go
```diff
@@ -0,0 +1,50 @@
+/*
```
salguarnieri: Is this a test or an example? If an example, I think we should rename the file and add a file/package level doc explaining that.

creachadair: > Is this a test or an example?

It's both. The `testing` package also supports testable examples, which have the dual benefit of testing the code (or at least exercising it) and demonstrating its use. 

https://blog.golang.org/examples

PR #34, opened 2018-05-22
Fix build

danielmoy-google:


kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -85,12 +85,16 @@ func NewGitTestingHarness(configPath string) Tester {
...
-// single bit representing extraction/download success, consider file count (or
-// if we've done indexing, symbolc ount).
+// TODO(danielmoy): consider better metrics here. For example consider having
+// the smoke test harness try to run a kythe indexer in addition to an
+// extraction and see how much symbol coverage we have.  This might be out of
+// scope for a simple smoke test harness though.
 type Result struct {
 	// Whether the repo was successfully downloaded or extracted.
 	Downloaded, Extracted bool
+	// The number of downloaded and extracted files.
```
jrtom: Are there ever times in which there are 0 downloaded or extracted files but we want to consider the download or extraction to have been successful?

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -99,23 +103,19 @@ func (g harness) TestRepo(repo string) (Result, error) {
 	fromRepo, err := g.filenamesFromRepo(repo)
 	if err != nil {
 		log.Printf("Failed to read repo from remote: %v", err)
-		return Result{false, false, 0.0}, nil
+		return Result{false, false, len(fromRepo), 0, 0.0}, nil
```
jrtom: is len(fromRepo) ever going to be nonzero here?

danielmoy-google:
Abandoning this because I accidentally jammed unrelated things into the same request.

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: Yes, empty repos are technically speaking successes (/futurama pedant meme)

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: Probably not, but this future-proofs it in case we ever start returning partial errors that we can recover from.

PR #38, opened 2018-05-22
Implement MarkedSource rendering once in the codebase per language.

zrlk:
QualifiedNameExtractor doesn't render MarkedSource the same way as
MarkedSourceRenderer (or the C++ renderer). To keep it from diverging
any further, teach MarkedSourceRenderer how to render plain text and
redirect QualifiedNameExtractor to use it.

Also remove a non-standard feature that uses "." as an implicit
separator in certain circumstances.


zrlk:
Looks like the Go implementation is doing something similar to the old Java implementation, complete with the odd `firstMatching` search (and nonstandard default context delimiter).

PR #39, opened 2018-05-23
Implement a kcd.Reader around a kzip file.

creachadair:
Package kzipdb provides an implementation of the kcd.Reader interface that
consumes data from a kzip file. This implementation does not allow writing.

### DIFF
--- a/kythe/go/platform/kcd/kzipdb/kzipdb.go
+++ b/kythe/go/platform/kcd/kzipdb/kzipdb.go
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package kzipdb implements kcd.Reader using a kzip file as its backing
+// store. See also: http://www.kythe.io/docs/kythe-index-pack.html.
+package kzipdb
+
+import (
+	"context"
+
+	"github.com/golang/protobuf/proto"
+	"kythe.io/kythe/go/platform/kcd"
+	"kythe.io/kythe/go/platform/kcd/kythe"
+	"kythe.io/kythe/go/platform/kzip"
+)
+
+// DB implements kcd.Reader using a kzip.Reader as its backing store.
+type DB struct {
+	Reader *kzip.Reader
+}
+
+// Revisions implements a method of kcd.Reader.
+func (db DB) Revisions(ctx context.Context, want *kcd.RevisionsFilter, f func(kcd.Revision) error) error {
+	revisionMatches, err := want.Compile()
+	if err != nil {
+		return err
+	}
+	seen := make(map[kcd.Revision]bool)
+	return db.Reader.Scan(func(unit *kzip.Unit) error {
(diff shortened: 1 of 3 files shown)


kythe/go/platform/kcd/kzipdb/kzipdb.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, digest := range unitDigests {
+		unit, err := db.Reader.Lookup(digest)
+		if err != nil {
+			return err
+		}
+		data, err := proto.Marshal(unit.Proto)
+		if err != nil {
+			return err
+		}
+		if err := f(unit.Digest, kythe.Format, data); err != nil {
```
salguarnieri: What is kythe.Format and is it safe to hardcode here?

kythe/go/platform/kcd/kzipdb/kzipdb_test.go
```diff
@@ -0,0 +1,198 @@
...
+		db.Revisions(ctx, &kcd.RevisionsFilter{Corpus: "bogus"}, func(rev kcd.Revision) error {
+			t.Errorf("Unexpected revision found: %v", rev)
+			return nil
+		})
+	})
+	t.Run("MatchSomeRevs", func(t *testing.T) {
+		want := []kcd.Revision{
+			{Revision: "123", Corpus: "kythe"},
+			{Revision: "456", Corpus: "kythe"},
+			{Revision: "789", Corpus: "kythe"},
```
salguarnieri: Can you add a revision that doesn't have a kythe corpus so at least one revision doesn't match here?

kythe/go/platform/kcd/kzipdb/kzipdb_test.go
```diff
@@ -0,0 +1,198 @@
...
+			return nil
+		})
+		if numFound != 2 {
+			t.Errorf("Find: got %d compilations, want 2", numFound)
+		}
+	})
+	t.Run("ReadUnits", func(t *testing.T) {
+		i := 0
+		db.Units(ctx, testInput.units, func(digest, key string, data []byte) error {
+			if want := testInput.units[i]; digest != want {
```
salguarnieri: Is it safe to assume the order of processing will be the same as the test input array order?

kythe/go/platform/kcd/kzipdb/kzipdb.go
creachadair: It's a constant defined here: http://godoc.org/kythe.io/kythe/go/platform/kcd/kythe#Format

And yes, it's safe to hardcode. I'm actually planning to get rid of the format key entirely, as soon as is practical. It was a kluge to support multiple compilation formats when we were migrating the internal implementation, but it's no longer needed.

kythe/go/platform/kcd/kzipdb/kzipdb_test.go
creachadair: These are the expected values, not the input values—the input values above already contain some non-matching values.

kythe/go/platform/kcd/kzipdb/kzipdb_test.go
creachadair: In this implementation yes. It's not a required property of the interface, but since this test is specific to this format it's safe.

kythe/go/platform/kcd/kzipdb/kzipdb_test.go
salguarnieri: Aren't the inputs values:
{Unit: newUnit("A", "go", "kythe"), Index: newIndex("123")},
+			{Unit: newUnit("B", "c++", "kythe"), Index: newIndex("456", "789")},
+			{Unit: newUnit("C", "go", "chromium"), Index: newIndex("123")},
+			{Unit: newUnit("D", "protobuf", "kythe"), Index: newIndex("789")},

if so, there aren't any revisions that aren't 123, 456, or 789

creachadair: Ah, I see what you mean. I thought you wanted another output value here. 👍

Note that we're comparing "whole revisions", i.e., revision + corpus, so a mismatch should not be possible. Nevertheless, I added an explicit case to guard against this. Thanks!

PR #42, opened 2018-05-23
Add scaffolding for Apache Beam pipeline implementation

schroederc:


kythe/go/serving/tools/write_tables/write_tables.go
```diff
@@ -54,6 +57,8 @@ var (
 		"Size of the reading/writing buffers for the intermediary data shards.")
 
 	verbose = flag.Bool("verbose", false, "Whether to emit extra, and possibly excessive, log messages")
+
+	experimentalNewPipeline = flag.Bool("experimental_new_pipeline", false, "Whether to use the new experimental pipeline implementation")
```
creachadair: "New" has an amazing tendency to bit-rot. Can we say "Beam" instead?

schroederc: ✓

PR #43, opened 2018-05-23
Move to csv for repo testing.

danielmoy-google:
Now that we're running this on many, many more repositories, prefer a
different format.

### DIFF
--- a/kythe/go/platform/tools/extraction/repotester.go
+++ b/kythe/go/platform/tools/extraction/repotester.go
@@ -16,7 +16,7 @@
 
 // Binary repotester tests the high level validity of repo extraction. It tries
 // to run similar logic to sibling binary extractrepo on a specified target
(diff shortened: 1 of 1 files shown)


kythe/go/platform/tools/extraction/repotester.go
```diff
@@ -16,7 +16,7 @@
 
 // Binary repotester tests the high level validity of repo extraction. It tries
 // to run similar logic to sibling binary extractrepo on a specified target
-// repositories, and reports results.
+// repositories, and reports results in csv format.
```
creachadair: If we really want to use CSV, we should probably use https://godoc.org/encoding/csv so that things get quoted correctly if there are internal punctuation/whitespace.

danielmoy-google:
Updated

PR #44, opened 2018-05-23
Add Beam *spb.Entry stream reader

schroederc:


kythe/go/serving/pipeline/beamio/BUILD
```diff
@@ -0,0 +1,29 @@
+load("//tools:build_rules/shims.bzl", "go_library", "go_test")
+
+package(default_visibility = ["//kythe:default_visibility"])
+
+go_library(
+    name = "beamio",
+    srcs = glob(
```
creachadair: There's only one! Please let's just list it.

kythe/go/serving/pipeline/beamio/entries.go
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package beamio
+
+import (
+	"context"
+
+	"kythe.io/kythe/go/storage/stream"
+
+	spb "kythe.io/kythe/proto/storage_go_proto"
```
creachadair: I think we wound up putting proto imports at the end?

schroederc: ✓

kythe/go/serving/pipeline/beamio/BUILD
schroederc: ✓

PR #45, opened 2018-05-23
Bug fix: Avert a crash reading a nil proto field.

creachadair:
Since the index field of an indexed compilation may legally be unset, use the
accessor to fetch its revisions list.


kythe/go/platform/kcd/kzipdb/kzipdb.go
```diff
@@ -41,7 +41,7 @@ func (db DB) Revisions(ctx context.Context, want *kcd.RevisionsFilter, f func(kc
 	seen := make(map[kcd.Revision]bool)
 	return db.Reader.Scan(func(unit *kzip.Unit) error {
 		corpus := unit.Proto.GetVName().GetCorpus()
-		for _, revision := range unit.Index.Revisions {
+		for _, revision := range unit.Index.GetRevisions() {
```
salguarnieri: For extra safety, does it also make sense to change Index to GetIndex()?

creachadair: It would, but because `*Unit` is not a protobuf generated type, we can't.

PR #48, opened 2018-05-24
Support timeout for repo testing.

### LINKED ISSUE #47: staticanalysis smoke test harness should have hard timeout
Right now there's a few test repos we're using that hang indefinitely.  We should have a timeout before we automatically fail and move on, since the primary purpose here is breadth of repos.

danielmoy-google:
fixes #47.

### DIFF
--- a/kythe/go/extractors/config/smoke/smoke.go
+++ b/kythe/go/extractors/config/smoke/smoke.go
@@ -43,21 +43,21 @@ import (
 //
 // TODO(danielmoy): support more than just java.
 type Tester interface {
-	TestRepo(repo string) (Result, error)
+	TestRepo(ctx context.Context, repo string) (Result, error)
 }
 
 // Fetcher is a thin wrapper over something which fetches a given repo and
 // writes it to an output directory. Note that the ConfigPath parameter from
 // config.Repo does not affect Fetch at all.
 type Fetcher interface {
-	Fetch(repo config.Repo) error
+	Fetch(ctx context.Context, repo config.Repo) error
 }
 
 type gitCommandlineFetcher struct{}
 
-func (g gitCommandlineFetcher) Fetch(repo config.Repo) error {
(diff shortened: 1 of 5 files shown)


kythe/go/extractors/config/smoke/smoke_test.go
```diff
@@ -159,7 +160,7 @@ func verifyRepoCoverage(t *testing.T, te testExtractor, expectedDownloadCount, e
 		repoFetcher: te,
 	}
 
-	r, err := h.TestRepo("foo")
+	r, err := h.TestRepo(context.TODO(), "foo")
```
creachadair: Can you not pass along the caller's `ctx` here? (If not, please add a comment to say why)

danielmoy-google: Sorry what caller's `ctx`? In the unit test I'm not sure what the deal is.

creachadair: Sorry, I didn't expand the clicky, and misread the code.

danielmoy-google: Ok to resolve this using `s/TODO/Background/`

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -187,7 +187,7 @@ func (g harness) filenamesFromExtraction(repoURI string) (map[string]bool, error
 	ret := map[string]bool{}
 	err = filepath.Walk(tmpOutDir, func(path string, info os.FileInfo, err error) error {
 		if err == nil && filepath.Ext(path) == ".kindex" {
-			cu, err := kindex.Open(context.Background(), path)
```
danielmoy-google: (as a side effect we fix this problematic context.Background() I stuffed in there before too)

PR #50, opened 2018-05-24
Combine *spb.Entry protos into *ppb.Nodes

schroederc:
Nodes will be the basic input for the rest of the Kythe pipeline.  While
combining entries, duplicate edges/fact_values are discarded and known
schema strings are stored as enums.

### DIFF
--- a/kythe/go/services/graphstore/compare/compare.go
+++ b/kythe/go/services/graphstore/compare/compare.go
@@ -20,6 +20,8 @@ package compare
 
 import (
 	"bytes"
+	"fmt"
+	"strings"
 
 	spb "kythe.io/kythe/proto/storage_go_proto"
 )
@@ -42,6 +44,16 @@ func (s *ByEntries) Pop() interface{} {
 	return out
 }
 
+// ToOrder returns LT if c < 0, EQ if c == 0, or GT if c > 0.
+func ToOrder(c int) Order {
+	if c < 0 {
+		return LT
+	} else if c > 0 {
+		return GT
+	}
+	return EQ
+}
+
 // An Order represents an ordering relationship between values.
 type Order int
 
@@ -52,17 +64,27 @@ const (
 	GT Order = 1  // lhs > rhs
 )
 
-// Strings returns LT if s < t, EQ if s == t, or GT if s > t.
-func Strings(s, t string) Order {
-	switch {
-	case s < t:
-		return LT
-	case s > t:
-		return GT
+func (o Order) String() string {
+	switch o {
+	case LT:
+		return "LT"
+	case GT:
+		return "GT"
+	case EQ:
+		return "EQ"
 	default:
-		return EQ
+		return fmt.Sprintf("Order(%d)", o)
 	}
 }
+
+// Strings returns LT if s < t, EQ if s == t, or GT if s > t.
+func Strings(s, t string) Order { return Order(strings.Compare(s, t)) }
+
+// Bytes returns LT if s < t, EQ if s == t, or GT if s > t.
+func Bytes(s, t []byte) Order { return Order(bytes.Compare(s, t)) }
+
+// Ints returns LT if a < b, EQ if a == b, or GT if a > b.
+func Ints(a, b int) Order { return ToOrder(b - a) }
 
 var (
 	emptyVName = new(spb.VName)
@@ -127,7 +149,7 @@ func ValueEntries(e1, e2 *spb.Entry) Order {
 	if c := Entries(e1, e2); c != EQ {
 		return c
 	}
-	return Order(bytes.Compare(e1.FactValue, e2.FactValue))
+	return Bytes(e1.FactValue, e2.FactValue)
 }
 
 // EntriesEqual reports whether e1 and e2 are equivalent, including their fact
--- a/kythe/go/serving/pipeline/nodes/nodes.go
+++ b/kythe/go/serving/pipeline/nodes/nodes.go
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package nodes provides Beam transformations over *ppb.Nodes.
+package nodes
+
+import (
+	"reflect"
+	"sort"
+
+	"kythe.io/kythe/go/services/graphstore/compare"
+	"kythe.io/kythe/go/util/schema"
+	"kythe.io/kythe/go/util/schema/edges"
+	"kythe.io/kythe/go/util/schema/facts"
+
+	"github.com/apache/beam/sdks/go/pkg/beam"
+
+	ppb "kythe.io/kythe/proto/pipeline_go_proto"
+	scpb "kythe.io/kythe/proto/schema_go_proto"
+	spb "kythe.io/kythe/proto/storage_go_proto"
+)
+
+func init() {
+	beam.RegisterFunction(embedSourceKey)
+	beam.RegisterFunction(entryToNode)
+
+	beam.RegisterType(reflect.TypeOf((*combineNodes)(nil)).Elem())
+}
+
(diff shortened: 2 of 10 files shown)


kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -0,0 +1,178 @@
...
+// FromEntries transforms a PCollection of *ppb.Entry protos into *ppb.Nodes.
+func FromEntries(s beam.Scope, entries beam.PCollection) beam.PCollection {
+	return beam.ParDo(s, embedSourceKey,
+		beam.CombinePerKey(s, &combineNodes{},
+			beam.ParDo(s, entryToNode, entries)))
+}
+
+func entryToNode(e *spb.Entry) (*spb.VName, *ppb.Node) {
+	n := &ppb.Node{}
+	if e.EdgeKind == "" {
```
jrtom: Does an Entry ever simultaneously encode (a) facts for a node and (b) edge data associated with that node?  This code implicitly assumes that it doesn't.  If it never should, this code should probably verify that it doesn't; if it could, then this logic needs to be structured a bit differently.

kythe/proto/pipeline.proto
```diff
@@ -0,0 +1,65 @@
...
+
+package kythe.proto.pipeline;
+
+option java_package = "com.google.devtools.kythe.proto";
+
+import "kythe/proto/schema.proto";
+import "kythe/proto/storage.proto";
+
+// A Node represents a Kythe graph node with all its known facts and edges.
+message Node {
```
jrtom: The term that we're using for this concept internally is "Source".  Calling it "Node" is confusing IMO because it's not a node per se, but effectively a union that can represent either a node and its facts, or a node and its associated edge data (although see my comment on `nodes.go` about that).  Any reason not to go with the existing internal name?

Please add documentation that specifies what constraints, if any, should be imposed on which sets of fields are expected to be simultaneously populated.  (The existing code in `nodes.go` implies that either `fact` and the `*kind` fields will be populated, or `edge` will be populated, but not both.)

kythe/go/services/graphstore/compare/compare.go
```diff
@@ -52,18 +64,28 @@ const (
...
+	switch o {
+	case LT:
+		return "LT"
+	case GT:
+		return "GT"
+	case EQ:
+		return "EQ"
 	default:
-		return EQ
+		return fmt.Sprintf("Order(%d)", o)
```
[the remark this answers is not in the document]
schroederc: It could go either way, but usually Go idioms stray away from panics if possible.  This is just used for test messages anyway.

kythe/go/serving/pipeline/nodes/nodes.go
schroederc: It is the case that an Entry is either a fact or edge, but not both, based on which fields are set: (source, edge_kind, target) or (source, fact_name, fact_value).  There's some legacy around setting the fact_name to "/" for edges, but that's not required anymore.

I'm not sure what you mean by verify.  There are a couple of options in Beam: ignore, crash, or log.  Currently chosen was "ignore".  Generally, the Kythe design prefers to not crash under any circumstance and provide lossy data over no data, but this could be considered a critical error so I'll go with that for now.

kythe/proto/pipeline.proto
schroederc: The Source proto is also external.

But no, it is Node in the Kythe graph.  A Kythe node is composed of all of its edges and its facts (not either).  This exactly the same as the Source proto without the special-cases and enums.  There's not a whole lot of constraints on this proto as it will be filtered and massaged throughout the pipeline.

jrtom: You're right, the `Source` proto is also external, I got briefly confused because it's in `internal.proto`.  :P

So if this is intended to fill the same ecological niche as `Source`, would it make sense to augment it to include the special cases and enums rather than adding a new message?

Thanks for the clarification on this proto's use.  Would it make sense to add an enum that would specify what type of use case a given message is satisfying?  (All the current uses look like they're either storing facts or edges but not both, but I understand that there will be other uses later.)  That would make validation possible.

kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -0,0 +1,193 @@
...
+		f.Name = &ppb.Fact_GenericName{e.FactName}
+	} else {
+		f.Name = &ppb.Fact_KytheName{name}
+	}
+	return f
+}
+
+type combineNodes struct{}
+
+func (combineNodes) CreateAccumulator() *ppb.Node { return &ppb.Node{} }
```
jrtom: I don't recognize this declaration syntax: what is `combineNodes` here, since it appears not to be in the form of a receiver?  I mean, since in this case you don't need a variable to refer to it in the method body, I can see why Go might let you skip that part of the declaration, but if there's something else going on here, I'd like to understand it.

kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -0,0 +1,193 @@
...
+	name := schema.FactName(e.FactName)
+	if name == scpb.FactName_UNKNOWN_FACT_NAME {
+		f.Name = &ppb.Fact_GenericName{e.FactName}
+	} else {
+		f.Name = &ppb.Fact_KytheName{name}
+	}
+	return f
+}
+
+type combineNodes struct{}
```
jrtom: Please add documentation for this type and for its public methods below.

kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (combineNodes) CreateAccumulator() *ppb.Node { return &ppb.Node{} }
+
+func (c *combineNodes) MergeAccumulators(accum, n *ppb.Node) *ppb.Node {
```
jrtom: Is it fair to assume that we would consider it an error if we tried merging `accum` with `n`, both had a populated `[Sub]Kind` field, and they didn't match?

If so, please add a check (logging+counter?) for this case.

If not, please make sure the documentation clarifies that this is OK.

(Another way to handle this might be to populate the *kind fields and the fact/edge fields separately, since merging makes sense for facts/edges but not for *kinds.)

kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, f := range n.Fact {
+		accum.Fact = append(accum.Fact, f)
+	}
+	for _, e := range n.Edge {
+		accum.Edge = append(accum.Edge, e)
+	}
+	return accum
+}
+
+func (c *combineNodes) AddInput(accum, n *ppb.Node) *ppb.Node { return c.MergeAccumulators(accum, n) }
```
jrtom: Why provide both `AddInput` and `MergeAccumulators`?

kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (c *combineNodes) ExtractOutput(n *ppb.Node) *ppb.Node {
+	// TODO(schroederc): deduplicate earlier during combine
+	if len(n.Fact) > 1 {
+		sort.Slice(n.Fact, func(i, j int) bool { return compareFacts(n.Fact[i], n.Fact[j]) == compare.LT })
```
jrtom: nano-nit: I wouldn't use i and j both here and farther down in the method body (same for the Edge-related code below); using distinct variables would improve readability IMO.

kythe/proto/pipeline.proto
schroederc: I'd rather make a clean break from the current system.  It removes the need for churn elsewhere.

The current use is storing both facts and edges.  The other "types" of uses are really just filtered views, not much to tag an enum on.

kythe/go/serving/pipeline/nodes/nodes.go
jrtom: OK.

kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -112,6 +112,10 @@ func entryToFact(e *spb.Entry) *ppb.Fact {
 	return f
 }
 
+// combineNodes is a Beam combiner for *ppb.Nodes.  All facts and edges are
+// merged into a single *ppb.Node.  If a fact has multiple values, a single is
```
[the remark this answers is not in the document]
jrtom: also: how is the single value chosen?  arbitrarily?  first written?  last written?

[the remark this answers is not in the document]
schroederc: It's arbitrary because we have no sense of first/last written.

kythe/go/serving/pipeline/nodes/nodes.go
schroederc: Yes, it is a required combiner method.

PR #51, opened 2018-05-25
Java indexer: avoid infinite loop while generating TypeVar signature

schroederc:


kythe/java/com/google/devtools/kythe/platform/java/helpers/SignatureGenerator.java
```diff
@@ -455,10 +455,12 @@ public Void visitTypeVar(TypeVar t, StringBuilder sbout) {
       sbout.append(t.tsym.name);
     } else {
       if (!visitedTypes.containsKey(t)) {
+        boundedVars.add(t);
```
jrtom: I don't understand what effect adding and removing `t` from `boundedVars` has.  What am I missing?

schroederc: The entire point of that set is avoid infinite recursion.  The javac AST has cycles that between method nodes and their declared type variables.  The set is checked just above to break that cycle.

jrtom: ah, I see, the `accept()` call on line 460 ends up calling this method again, got it.

PR #69, opened 2018-05-30
Empty shell for proposed index-after-extract.

danielmoy-google:
Idea in subsequent submit will be to index any successful github repo
extractions, and report on them as well.


kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -38,8 +38,7 @@ import (
 // automatically. The current implementation is to simply check the fraction of
 // files covered by extractor output.
 //
-// TODO(danielmoy): hook up an indexing step and check actual semantic output
-// instead of simple files.
+// Also optionally supports indexing the extractors output.
 //
 // TODO(danielmoy): support more than just java.
 type Tester interface {
```
creachadair: I'm not really convinced this needs to be an interface—it seems like it could be a plain function

    type Tester func(ctx context.Context, repo config.Repo) (Result, error)

Similarly with `Indexer` and `Fetcher` below. At some point it might make sense to make these configuration values on the smoke test constructor, see below. That makes it a little nicer to provide simple default implementations, e.g.,

```
func GitFetch(ctx context.Context, repo config.Repo) error {
  return exec.CommandContext(ctx, "git", "clone", repo.URI, repo.OutputPath).Run()
}
```

See also my comment at the constructor.

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -72,12 +87,20 @@ type harness struct {
 //
 // An extraction config can be optionally read from a specified file.  The
 // format follows kythe.proto.ExtractionConfiguration.
-func NewGitTestingHarness(configPath string) Tester {
-	return harness{
+//
+// Finally, the testing harness can optionally try to index the extraction
+// output.
+func NewGitTestingHarness(configPath string, doIndex bool) Tester {
```
creachadair: It might be worth breaking out a settings type here, e.g., 

```
type HarnessOptions struct {
	Indexer   Indexer
	Fetcher   Fetcher
	Extractor Extractor
}
```

This makes it easy to provide defaults:

```
func (o *HarnessOptions) extractor() Extractor {
	if o == nil || o.Extractor == nil {
		return config.DefaultExtractor{}
	}
	return o.Extractor
}
// etc.
```

And then the constructor has a simpler form:

```
func NewTestingHarness(configPath string, opts *HarnessOptions) Tester {
	h := &harness{
		extractor:  opts.extractor(),
		configPath: configPath,
		fetcher:    opts.fetcher(),
		indexer:    opts.indexer(),
	}
	return h.TestRepo
}
```

Then the user can do what they want without having to write a bunch of separate constructors:

```
	h := smoke.NewTestingHarness("github.com/google/kythe.git", &smoke.HarnessOptions{
		Fetcher: smoke.GitFetch,
		Indexer: indexers.Java,
	})
```

There's more in this than you might want to do in a single change, but I'm a little concerned by the proliferation of interfaces without much generality. Though even if we keep those, I think the options-type has potential to simplify some of this.

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -103,15 +127,31 @@ func (g harness) TestRepo(ctx context.Context, repo string) (Result, error) {
...
 		// returning separate false results if either err != nil.
-		return Result{true, false, len(fromRepo), len(fromExtraction), 0.0}, nil
+		return Result{
+			Downloaded:    true,
+			Extracted:     false,
+			Indexed:       false,
+			DownloadCount: len(fromRepo),
+			ExtractCount:  len(fromExtraction),
+			IndexCount:    0,
+			FileCoverage:  0.0,
```
creachadair: `FileCoverage: 0`

Untyped literals get concrete type from their use, it's not necessary to decorate them lexically (in most cases).

https://blog.golang.org/constants, if you're interested in more.

danielmoy-google: Ah right, forgot about that.

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: Interesting.  Why not just use the HarnessOptions as the only type thing there, and stuff the methods on it?  (See commi)

danielmoy-google:
Addressed comments, though unclear how we want to leave HarnessOptions.  Kinda leaning towards not having a separate private harness struct.

danielmoy-google:
Hopefully updated in face of merge conflict.

kythe/go/extractors/config/smoke/smoke.go
creachadair: Yeah, that's also a viable approach. I often separate the types if the underlying type has mutable state, so that the options don't get mixed up with hidden details. But here it'd arguably be fine to export the settings directly, and just implement everything as methods as you say.

In that case, you could even maybe do away with the constructor,

```
// A Harness contains the settings necessary to test a repository.
// The ConfigPath is required, all the other fields are optional and use
// defaults as specified.
type Harness struct {
  ConfigPath string
   Extractor Extractor  // defaults to blah
  ...
}
```

kythe/go/extractors/config/smoke/smoke.go
creachadair: It depends. It's generally idiomatic to rely on default zero values for things—and in fact it's a good idea to choose field names so that the zero _does_ make a sensible default, where practical. But sometimes it can be worthwhile for documentation. Here, I probably wouldn't bother except maybe to document that the default coverage allows anything to be "ok".

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -38,46 +38,77 @@ import (
 // automatically. The current implementation is to simply check the fraction of
 // files covered by extractor output.
 //
-// TODO(danielmoy): hook up an indexing step and check actual semantic output
-// instead of simple files.
+// Also optionally supports indexing the extractors output.
 //
 // TODO(danielmoy): support more than just java.
```
creachadair: I think this comment is no longer needed?

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -38,46 +38,77 @@ import (
...
-type Tester interface {
-	TestRepo(ctx context.Context, repo string) (Result, error)
-}
+type Tester func(ctx context.Context, repo string) (Result, error)
 
 // Fetcher is a thin wrapper over something which fetches a given repo and
```
creachadair: s/thin wrapper over something which//

```
// A Fetcher fetches the contents of the given repo and writes it to an
// output directory. The implementation may ignore the repo.ConfigPath.

```

creachadair: Not sure if the comment about `ConfigPath` is meant to be a constraint on the implementation, or a description of the current behaviour.

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -38,46 +38,77 @@ import (
...
+// TODO(danielmoy): kzip?  This is generally a thing that needs supporting in
+// smoke.go and related files now that I think about it.
+type Indexer func(ctx context.Context, inputDir, outputDir string) error
+
+func EmptyIndexer(ctx context.Context, inputDir, outputDir string) error {
+	return nil
 }
 
-// NewGitTestingHarness creates a simple Tester which uses
+type HarnessOptions struct {
```
creachadair: Please document this.

    // HarnessOptions …

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: ah I think that's still needed

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: yea probably can ommit the later bit.  Really I shouldn't be re-using the same config.Repo since the interface is only "mostly what I want", but mostly just doing it for expediency.

kythe/go/extractors/config/smoke/smoke.go
danielmoy-google: done

kythe/go/extractors/config/smoke/smoke.go
creachadair: Maybe—but maybe somewhere else? It's not really part of the `Tester` documentation.

kythe/go/extractors/config/smoke/smoke.go
creachadair: :)

danielmoy-google:
ah yea there's another TODO for java already I suppose

danielmoy-google:
ok last round maybe

PR #72, opened 2018-05-31
Tweaks and additions to the 0.0.27 pre-release notes.

creachadair:
Add mention of Objective C and TypeScript. Fix various cosmetic capitalization
and punctuation and formatting bits.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,68 +4,72 @@
 
 Due to the period of time between this release and v0.0.26, many relevant
 changes and fixes may not appear in the following list.  For a complete list of
-changes, please check: https://github.com/google/kythe/commits/
+changes, please check the commit logs:
+https://github.com/google/kythe/search?q=committer-date%3A2016-11-01..2018-05-31&type=Commits
 
-Notable changes:
- - remove GRPC server and client code
+### Notable additions
+
+ - First release of Go indexer and Go extractors.
+ - Objective C is now supported in the C++ indexer.
+ - First release of TypeScript indexer.
+ - identifier.proto: adds a new `IdentifierService` API.
+ - Runtime plugins can be added to the Java indexer.
+
+### Notable changes
+
+ - Remove gRPC server and client code.
  - Schema:
-   - anchors no longer have `childof` edges to their parent `file`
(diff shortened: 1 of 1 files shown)


creachadair:
Do we want to mention also some of the Bazel changes, e.g., that nearly-all the dependencies are now configured as external repositories? This doesn't affect users of the release tarball, but is a useful point of interest for someone checking out at this tag.

RELEASES.md
```diff
@@ -4,68 +4,72 @@
 
 Due to the period of time between this release and v0.0.26, many relevant
 changes and fixes may not appear in the following list.  For a complete list of
-changes, please check: https://github.com/google/kythe/commits/
+changes, please check the commit logs:
+https://github.com/google/kythe/search?q=committer-date%3A2016-11-01..2018-05-31&type=Commits
```
creachadair: We will need to update the dates here, once we pick a release date. :)

schroederc: `https://github.com/google/kythe/compare/v0.0.26...v0.0.27`

RELEASES.md
```diff
@@ -4,68 +4,72 @@
...
 
-Notable changes:
- - remove GRPC server and client code
+### Notable additions
```
schroederc: Please don't change the format yet.  It doesn't match at all with the rest of the log.  We're just pushing out a release to catch-up.

RELEASES.md
```diff
@@ -4,68 +4,72 @@
...
+
+ - First release of Go indexer and Go extractors.
+ - Objective C is now supported in the C++ indexer.
+ - First release of TypeScript indexer.
```
schroederc: This is not included in the release.

creachadair: Should it be?

RELEASES.md
creachadair: Oh, nice, thanks. Done.

RELEASES.md
creachadair: Reverted the header tags.

RELEASES.md
schroederc: I was pushing it the next release to simplify.

creachadair: Makes sense—I removed it.

PR #75, opened 2018-05-31
Update generated protocol buffer files

schroederc:


creachadair:
> Don't suppose these can be autogenerated?

Not really. Bazel can't write to the source directory. But I'm planning to add tests that will fail if we forget to update them (cf. https://github.com/google/kythe/pull/55).

PR #78, opened 2018-06-01
Add tests to verify that protos are kept in sync.

creachadair:
Extend our go_kythe_proto macro to generate a test that will fail if the
checked-in version of the proto we supply for "go get" differs from the version
generated by Bazel.

The test can't actually fix it, but will prompt the user to do so.

### DIFF
--- a/kythe/proto/go.bzl
+++ b/kythe/proto/go.bzl
@@ -1,8 +1,37 @@
 """This module provides rules for building protos for golang."""
 
 load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
+load("@io_bazel_rules_go//go:def.bzl", _GoSource = "GoSource")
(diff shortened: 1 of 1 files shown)


creachadair:
> [I] wonder if we should have a canonical doc somewhere that tells people to run tests before merge.

We don't have one right now. It's still possible to run `arc lint` and `arc unit` if you have them installed, which will check for common things like the tests passing. Some of us have a pre-push hook that does this check, which can be as simple as

```
$ cat > .git/hooks/pre-push <<EOF
#!/bin/sh
arc lint
arc unit
EOF
```

but nothing forces it to happen otherwise. Longer term, we should get back to some kind of CI that we can use for pre-submit testing.

danielmoy-google:
Welp I should add such a presubmit hook then.

PR #79, opened 2018-06-01
Fixing lint errors.

danielmoy-google:


kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -111,8 +113,10 @@ type Result struct {
 	FileCoverage float64
 }
 
-func (g Harness) TestRepo(ctx context.Context, repo string) (Result, error) {
-	fromRepo, err := g.filenamesFromRepo(ctx, repo)
+// TestRepo compares the result of extraction and optional indexing with
+// expected output from a given repo.
+func (h Harness) TestRepo(ctx context.Context, repo string) (Result, error) {
```
creachadair: I didn't notice this the last time, but it's better to use the same type for all the methods, unless you have a specific reason to do otherwise (here, some are on the pointer, some on the base type—either is probably fine, but it would be good to be consistent).

danielmoy-google: Ah yea I mean to consolidate on not-pointer since it's not being mutated and doesn't need it.  (Or at least that's my vague understanding)

kythe/go/extractors/config/smoke/smoke.go
```diff
@@ -72,21 +72,21 @@ type Harness struct {
 	ConfigPath string           // defaults to ""
 }
 
-func (h *Harness) fetcher() Fetcher {
+func (h Harness) fetcher() Fetcher {
```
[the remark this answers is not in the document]
danielmoy-google: ah yea, my incompatible bazel .14

WORKSPACE
```diff
@@ -4,7 +4,7 @@ load("//:version.bzl", "check_version")
 
 # Check that the user has a version between our minimum supported version of
 # Bazel and our maximum supported version of Bazel.
-check_version("0.13.0", "0.13.1")
+check_version("0.13.0", "0.14.0")
```
creachadair: I don't think this change is safe. We are strictly _not_ yet compatible with 0.14.x yet. :(

danielmoy-google:
Indeed it's not :(

https://github.com/google/kythe/issues/80

Wishful thinking that one day we'll have a bazel release that doesn't break us.

PR #83, opened 2018-06-01
Fix go_kythe_proto generated file_diff_tests

schroederc:


kythe/proto/go.bzl
```diff
@@ -58,7 +58,7 @@ def go_kythe_proto(proto=None, deps=[]):
     deps: the deps for the proto lib
   """
   base = proto.rsplit(":", 2)[-1]
-  filename = base.split("_", 2)[0] + ".pb.go"
+  filename = '_'.join(base.split("_")[:-1]) + ".pb.go"
```
creachadair: This makes sense but it's weird that it didn't show up for me when I ran all the tests before merging. Sorry about that.

PR #84, opened 2018-06-01
Add imputes edge kind to Java analyzers

salguarnieri:


kythe/java/com/google/devtools/kythe/analyzers/base/EdgeKind.java
```diff
@@ -38,6 +38,7 @@
   EXPORTS("exports"),
   EXTENDS("extends"),
   GENERATES("generates"),
+  IMPUTES("imputes"),
```
schroederc: This should go above as one of the anchor edges.

salguarnieri: done

kythe/java/com/google/devtools/kythe/analyzers/base/EdgeKind.java
```diff
@@ -23,6 +23,8 @@
   DOCUMENTS(true, "documents"),
   TAGGED(true, "tagged"),
 
+  // Edges from an anchor.
+  IMPUTES("imputes"),
```
schroederc: It needs to use the other constructor, that's what signifies it as an anchor edge: `IMPUTES(true, "imputes")`.

salguarnieri: done

PR #85, opened 2018-06-01
Update //third_party:licenses

schroederc:


third_party/jsr250_api/BUILD
```diff
@@ -0,0 +1,3 @@
+licenses(["notice"])
```
zrlk: CDDL 1.0 is reciprocal

third_party/libuuid/BUILD
```diff
@@ -0,0 +1,5 @@
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])
```
zrlk: gpl2 is restricted

third_party/jsr250_api/BUILD
schroederc: ✓

third_party/libuuid/BUILD
schroederc: ✓

PR #93, opened 2018-06-01
Clang tidy fixes.

danielmoy-google:
I am ambivalent about whether we actually do this.


danielmoy-google:
Sure we can go the other way instead

PR #94, opened 2018-06-01
typescript: index 'extends' clauses in interfaces

evmar:
The argument to an 'extends' clause is a type, not a value.
This matters when Interface1 extends Interface2, as the latter
is *only* a type and not a value.


kythe/typescript/indexer.ts
```diff
@@ -392,6 +392,22 @@ class Vistor {
...
   }
 
+  /**
+   * visitHeritage visits the X found in an 'extends X' or 'implements X'.
+   *
+   * These are subtle in an interesting way.  When you have
+   *   interface X extends Y {}
+   * that is referring to the *type* Y (because interfaces are types, not
+   * values).  But it's also legal to write
+   *   class X extends (class Z { ... }) {}
```
zrlk: Could you add a test for this case?

schroederc:
Any progress on this?

evmar:
In https://github.com/google/kythe/pull/2838 I add a test with some class expressions, and in this change (based on that) I added some more tests.

kythe/typescript/indexer.ts
evmar: Done

PR #98, opened 2018-06-01
Skip Go test when not ran using Bazel

schroederc:


kythe/go/util/markedsource/markedsource_test.go
```diff
@@ -83,6 +83,9 @@ func runOracle(t *testing.T, ms *cpb.MarkedSource) *oracleResults {
 // renderer.  Each test parses the C++ doc utility output and compares the
 // results with the native Go implementations.
 func TestInteropt(t *testing.T) {
+	if os.Getenv("TEST_WORKSPACE") != "io_kythe" {
```
salguarnieri: Can you add a comment why this has to be skipped outside of bazel?

schroederc: ✓

PR #99, opened 2018-06-01
Fix go vet issue

schroederc:


kythe/go/util/schema/mkdata/mkdata.go
```diff
@@ -193,6 +193,7 @@ func FactNameString(f scpb.FactName) string { return factNamesRev[f] }
 
 // SubkindString returns the string representation of the given subkind.
 func SubkindString(k scpb.Subkind) string { return subkindsRev[k] }
+
```
salguarnieri: Was this line addition intentional?

PR #117, opened 2018-06-05
Allow JSON initialization to run multiple times.

benjyw:
When initializing in main(), we call`usingTypeRegistry()`, which throws `IllegalArgumentException` if called a second time.

However we may in fact call main multiple times, e.g., if running kythe in a nailgun server.

This change works around the problem by short-circuiting the `usingTypeRegistry()` call if the same type registry is already in use. It will trigger the same `IllegalArgumentException` however, if you try to change the type registry to some other one.


benjyw:
All javatests pass. `google-java-format` run on the changes.

PR #118, opened 2018-06-05
Fix generic class member JVM names.

benjyw:
Previously, the Java indexer would emit a "generated" JVM node with a signature like `foo.bar.Generic<T>.member` while the JVM indexer would emit the same node with the signature `foo.bar.Generic.member`.  

This commit fixes this by modifying the Java indexer to emit the unadorned name that the JVM indexer emits.

This seems like the right direction to change, because generics are entirely a language feature, and don't exist at the JVM level.  Also, the Java indexer already emits the unadorned name for the class itself. It was just the members that were mismatched.


benjyw:
All javatests pass. `google-java-format` run on the changes.

schroederc:
Thanks!  I agree that this is the direction it should move and this was a bug.

PR #119, opened 2018-06-05
Move services/graphstore/compare to more general util/compare package

schroederc:
The added Compare/Seq/(Order).AndThen functions allow for easier chaining of comparisons over arbitrary types.


kythe/go/util/compare/compare.go
```diff
@@ -0,0 +1,272 @@
...
+// fact values (if any).
+//
+// The ordering for entries is defined by lexicographic comparison of
+// [source, edge kind, fact name, target].
+func Entries(e1, e2 *spb.Entry) Order {
+	return Seq(e1, e2,
+		ByEntrySource, ByEntryEdgeKind, ByEntryFactName, ByEntryTarget)
+}
+
+// ByEntries is a min-heap of entries, ordered by Entries.
```
salguarnieri: What makes this a min-heap? I don't think I see the bits that guarantee min-heap properties.

kythe/go/util/compare/compare.go
```diff
@@ -0,0 +1,272 @@
...
+// Compare returns the Order between two arbitrary values of the same type.
+//
+// Only the following types are currently supported: {string, int, []byte}.
+//
+// Options may be provided to change the semantics of the comparison.  Other
+// types may be compared if an appropriate By Option transforms the values into
+// a supported type or a With Option is provided for the types given.
+//
+// Note: this function panics if a and b are different types
+func Compare(a, b interface{}, opts ...Option) (o Order) {
```
salguarnieri: It might be nice to note that only the final With clause is honored.

kythe/go/util/compare/compare.go
schroederc: It's a weird idiom.  The type implements this interface: https://godoc.org/container/heap#Interface.  As long as you interact with the type using the accompanying Push/Pop functions, it all works out.  This is type wasn't added in this commit, it just got carried over from the previous package.

kythe/go/util/compare/compare.go
schroederc: I added the note to the With doc.

PR #120, opened 2018-06-05
Represent type params by their erasure in JVM names.

benjyw:
Previously we hard-coded j.l.Object in the source indexer, which did not match what the binary indexer emits.

Tests that this works for both simple and multiple-bound generics (in the latter case, the first bound is the erasure, by definition).


benjyw:
All javatests pass. Ran `google-java-format` on the changes.

PR #125, opened 2018-06-06
disksort: properly flush snappy.Writer

schroederc:
Since the Snappy Reader/Writers have a static buffer size, remove the IOBufferSize option and stick with what Snappy uses as a buffer size.


kythe/go/util/disksort/disksort.go
```diff
@@ -177,6 +166,8 @@ type mergeIterator struct {
 	workDir   string
 }
 
+const ioBufferSize = 2 << 15
```
creachadair: I think this is more typically `1<<16` but I guess it doesn't matter. :)

PR #126, opened 2018-06-06
Add modified 'direct' Go Apache Beam runner that uses disksort GBK

schroederc:
Using a disksort for direct CoGBK operations allows the pipeline to
support much larger input sizes without running out of memory while
buffering values locally.


third_party/beam/README.google
```diff
@@ -0,0 +1,6 @@
+URL: https://github.com/apache/beam
+Version: 72cbd99d6b62bc7ed16dbd1288cd61d54e8bda37
+License: Apache License 2.0: https://opensource.org/licenses/Apache-2.0
+Local Modifications:
+  - Narrow checkout of sdks/go/pkg/beam/runners/direct
+  - Modified package to use disksort for CoGBK implementation
```
danielmoy-google: do we need to talk at all about the BUILD file or does that not count here?

third_party/beam/sdks/go/pkg/beam/runners/direct/gbk.go
```diff
@@ -57,88 +108,248 @@ func (n *CoGBK) StartBundle(ctx context.Context, id string, data exec.DataManage
...
+	val, err := i.n.valDec[i.idx].Decode(bytes.NewBuffer(iv.Value))
+	if err != nil {
+		return exec.FullValue{}, fmt.Errorf("error decoding value: %v", err)
+	}
+	return val, nil
+}
+
+func (i *iterStream) next() (*iterValue, error) {
+	if i.prev != nil {
+		var prevItems int
```
danielmoy-google: unused prevItems?

I mean I see you incrementing it, but then... nothing

third_party/beam/sdks/go/pkg/beam/runners/direct/gbk.go
```diff
@@ -18,37 +18,88 @@ package direct
 import (
```
danielmoy-google: this file with all the changes have a test upstream somewhere?

danielmoy-google:
Looks like it buried two of my comments: unused prevItems int variable, and general question on whether gbk.go has a test or not

third_party/beam/sdks/go/pkg/beam/runners/direct/gbk.go
schroederc: I added some tests in gbk_test.go

third_party/beam/sdks/go/pkg/beam/runners/direct/gbk.go
schroederc: ✓

third_party/beam/README.google
schroederc: `¯\_(ツ)_/¯`

PR #134, opened 2018-06-07
extract_kindex: Add support for writing kzip files as output.

creachadair:
As part of the migration from kindex files, update the extract_kindex tool to
support writing kzip output. We will eventually want to rename the binary too,
but that can be a separate step; for now, it continues to support kindex also.


creachadair:
My PR feels very approved-of. Thanks!

PR #137, opened 2018-06-07
Add support for local copying instead of git cloning.

danielmoy-google:
pull/136 is in a bad state somehow.  git cherry-pick and try again?

### DIFF
--- a/kythe/go/extractors/config/extractor.go
+++ b/kythe/go/extractors/config/extractor.go
@@ -19,6 +19,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"os"
@@ -48,6 +49,59 @@ func GitCopier(repoURI string) func(ctx context.Context, outputDir string) error
 	return func(ctx context.Context, outputDir string) error {
 		// TODO(danielmoy): strongly consider go-git instead of os.exec
 		return exec.CommandContext(ctx, "git", "clone", repoURI, outputDir).Run()
+	}
+}
+
+// LocalCopier returns a function that copys a local repository.
+func LocalCopier(repoPath string) func(ctx context.Context, outputDir string) error {
+	return func(ctx context.Context, outputDir string) error {
+		gitDir := filepath.Join(repoPath, ".git")
+		// TODO(danielmoy): consider extracting all or part of this
+		// to a more common place.
+		return filepath.Walk(repoPath, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if repoPath == path {
+				// Intentionally do nothing for base dir.
+				return nil
+			}
+			if filepath.HasPrefix(path, gitDir) {
+				// Also skip all that .git stuff.
+				return nil
+			}
+			rel, err := filepath.Rel(repoPath, path)
+			if err != nil {
+				return err
+			}
+			outPath := filepath.Join(outputDir, rel)
+			if info.IsDir() {
+				if _, err := os.Stat(outPath); err != nil {
+					if err := os.Mkdir(outPath, 0777); err != nil {
+						return err
+					}
+				}
+				return nil
+			}
+			if !info.Mode().IsRegular() {
+				// Notably in here are any links or other odd things.
+				return fmt.Errorf("Can't handle the truth")
+			}
+			inf, err := os.Open(path)
+			if err != nil {
+				return err
+			}
+			defer inf.Close()
+			of, err := os.Create(outPath)
+			if err != nil {
+				return err
+			}
+			if _, err := io.Copy(of, inf); err != nil {
+				of.Close()
+				return err
+			}
+			return of.Close()
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


kythe/go/extractors/config/extractor.go
```diff
@@ -51,6 +52,59 @@ func GitCopier(repoURI string) func(ctx context.Context, outputDir string) error
 	}
 }
 
+// LocalCopier returns a function that copys a local repository.
```
creachadair: copys → copies

kythe/go/extractors/config/extractor.go
```diff
@@ -51,6 +52,59 @@ func GitCopier(repoURI string) func(ctx context.Context, outputDir string) error
...
+	return func(ctx context.Context, outputDir string) error {
+		gitDir := filepath.Join(repoPath, ".git")
+		// TODO(danielmoy): consider extracting all or part of this
+		// to a more common place.
+		return filepath.Walk(repoPath, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if repoPath == path {
+				// Intentionally do nothing for base dir.
```
creachadair: Please mention in the doc comment that this assumes the output directory has already been created.

kythe/go/extractors/config/extractor.go
```diff
@@ -51,6 +52,59 @@ func GitCopier(repoURI string) func(ctx context.Context, outputDir string) error
...
+				return nil
+			}
+			if filepath.HasPrefix(path, gitDir) {
```
creachadair: if path == gitDir {
        return filepath.SkipDir
    }

https://godoc.org/path/filepath#SkipDir

kythe/go/extractors/config/extractor.go
```diff
@@ -51,6 +52,59 @@ func GitCopier(repoURI string) func(ctx context.Context, outputDir string) error
...
+				if _, err := os.Stat(outPath); err != nil {
+					if err := os.Mkdir(outPath, 0777); err != nil {
+						return err
+					}
+				}
+				return nil
+			}
+			if !info.Mode().IsRegular() {
+				// Notably in here are any links or other odd things.
+				return fmt.Errorf("Can't handle the truth")
```
creachadair: Maybe we should just log and skip these?

kythe/go/extractors/config/extractor.go
```diff
@@ -51,6 +52,59 @@ func GitCopier(repoURI string) func(ctx context.Context, outputDir string) error
...
+				// Also skip all that .git stuff.
+				return nil
+			}
+			rel, err := filepath.Rel(repoPath, path)
+			if err != nil {
+				return err
+			}
+			outPath := filepath.Join(outputDir, rel)
+			if info.IsDir() {
+				if _, err := os.Stat(outPath); err != nil {
```
creachadair: `if err := os.MkdirAll(filepath.Dir(outPath), 0755); err != nil {`

kythe/go/extractors/config/extractor.go
danielmoy-google: oh that's awesome

kythe/go/extractors/config/extractor.go
danielmoy-google: ah replacing both lines? sure that makes sense

kythe/go/extractors/config/extractor.go
danielmoy-google: Sure

kythe/go/extractors/config/extractor.go
creachadair: Yeah—`os.MkdirAll` creates all the intermediate directories, and succeeds silently if they all already exist.

The other operative point was to create only the directory, not the file. Or is `outPath` just the directory already? If so: maybe `outputBase` for the base directory and `outputDir` for the concatenated one?

kythe/go/extractors/config/extractor.go
```diff
@@ -67,25 +68,23 @@ func LocalCopier(repoPath string) func(ctx context.Context, outputDir string) er
...
-				// Also skip all that .git stuff.
-				return nil
+				return filepath.SkipDir
 			}
 			rel, err := filepath.Rel(repoPath, path)
 			if err != nil {
 				return err
 			}
 			outPath := filepath.Join(outputDir, rel)
 			if info.IsDir() {
```
creachadair: Another approach we could take here is to skip directories entirely, and just use `os.MkdirAll` on any regular file path:

```
if info.Mode().IsRegular() {
  if err := os.MkdirAll(filepath.Dir(outPath), 0755); err != nil {
    return …
  }
  …copy the file…
} else if !info.IsDir() {
  log.Printf("Skipping non-regular file %q", path)
}
return nil
```

or words to that effect?

danielmoy-google: Ah yea good point.  That makes it a bit cleaner.

kythe/go/extractors/config/extractor.go
```diff
@@ -75,31 +75,30 @@ func LocalCopier(repoPath string) func(ctx context.Context, outputDir string) er
...
+				}
+				if _, err := io.Copy(of, inf); err != nil {
+					of.Close()
+					return fmt.Errorf("failed to copy repo file: %v", err)
+				}
+				return of.Close()
+			} else if !info.IsDir() {
 				// Notably in here are any links or other odd things.
 				log.Printf("Unsupported file %s with mode %s\n", path, info.Mode())
 				return nil
```
creachadair: You can drop either this one (to fall through), or the one at line 101 (the compiler will see that both branches return).

I wouldn't bother keeping both.

danielmoy-google: done

PR #138, opened 2018-06-07
Add target definitions to Beam-produced file decorations

schroederc:


kythe/go/serving/pipeline/beam.go
```diff
@@ -367,3 +418,23 @@ func moveSourceToKey(n *ppb.Node) (*spb.VName, *ppb.Node) {
...
 	}
 }
+
+func (k *KytheBeam) directDefinitions() beam.PCollection {
+	s := k.s.Scope("DirectDefinitions")
+	return beam.ParDo(s, toDefinition, k.References())
+}
+
+func toDefinition(r *ppb.Reference, emit func(*spb.VName, *srvpb.ExpandedAnchor)) error {
+	if !edges.IsVariant(refKind(r), edges.Defines) {
```
jrtom: suggest removing the negation to make this a bit simpler:

    if edges.IsVariant(refKind(r), edges.Defines) {
      emit(r.Source, r.Anchor)
    }
    return nil

PR #139, opened 2018-06-08
Fix the JVM signatures of methods with generic parameters.

benjyw:
The JVM indexer emits, e.g., `pkg.Jvm.Generic.tlistmethod(Ljava/util/List;)V`, 
while prior to this change the source indexer emitted, e.g., `pkg.Jvm.Generic.tlistmethod(Ljava/util/List<T>;)V`.

This change creates a `referenceType(Type referent)` helper method in `KytheTreeScanner`, instead of having to remember how to get the underlying `tsym` everywhere.


benjyw:
google-java-format ran, all javatests pass.

schroederc:
Can you please upgrade `google-java-format` to version 1.6?  It better handles the Flogger statements and means won't churn that part of the code.

benjyw:
Done.

benjyw:
I notice that 1.6 does churn the test files, because it wants to turn 

`//- @foo defines/binding Foo`

to
`// - @foo defines/binding Foo`

So I have not run it on the test file.

benjyw:
Closed in favor of https://github.com/google/kythe/pull/144.

PR #141, opened 2018-06-08
Preemptively support all Bazel 0.14.* patch versions

schroederc:


schroederc:
PTAL; I reversed the `min_required` check

PR #143, opened 2018-06-08
Add file tree directories and corpus-roots to Beam pipeline

schroederc:


kythe/go/serving/pipeline/filetree_test.go
```diff
@@ -0,0 +1,117 @@
...
+			Root:   []string{"", "root"},
+		}, {
+			Corpus: "corpus2",
+			Root:   []string{""},
+		}},
+	}}
+
+	p, s, nodes := ptest.CreateList(testNodes)
+	cr := beam.DropKey(s, FromNodes(s, nodes).CorpusRoots())
+	debug.Print(s, cr)
```
jrtom: Is it common in Go to leave debug statements in tests like this (and below)?

kythe/go/serving/pipeline/filetree.go
```diff
@@ -0,0 +1,166 @@
...
+func (k *KytheBeam) Directories() beam.PCollection {
+	s := k.s.Scope("Directories")
+	files := k.getFileVNames()
+	return beam.CombinePerKey(s, &combineDirectories{}, beam.ParDo(s, fileToDirectories, files))
+}
+
+func addCorpusRootsKey(val beam.T) (string, beam.T) { return "dirs:corpusRoots", val }
+
+func fileToDirectories(file *spb.VName, emit func(string, *srvpb.FileDirectory)) {
+	path := filepath.Clean(filepath.Join("/", file.GetPath()))[1:]
```
jrtom: why the slice starting at 1 rather than 0?  please document

kythe/go/serving/pipeline/filetree.go
```diff
@@ -0,0 +1,166 @@
...
+	dir := &spb.VName{
+		Corpus: file.Corpus,
+		Root:   file.Root,
+		Path:   filepath.Dir(path),
+	}
+	if dir.Path == "." {
```
jrtom: Do we have any interest in supporting ".." (or for that matter "...") in this context?

Whether we do or not, consider making this conversion a function so that we can use it again below (line 93).

kythe/go/serving/pipeline/filetree.go
```diff
@@ -0,0 +1,166 @@
...
+// Directories returns a Kythe *srvpb.FileDirectory table for the Kythe FileTree
+// service.  The beam.PCollection has elements of type KV<string,
+// *srvpb.FileDirectory>.
+func (k *KytheBeam) Directories() beam.PCollection {
+	s := k.s.Scope("Directories")
+	files := k.getFileVNames()
+	return beam.CombinePerKey(s, &combineDirectories{}, beam.ParDo(s, fileToDirectories, files))
+}
+
+func addCorpusRootsKey(val beam.T) (string, beam.T) { return "dirs:corpusRoots", val }
```
jrtom: please add minimal documentation for private functions/methods

kythe/go/serving/pipeline/filetree_test.go
schroederc: It's for the test log.  Makes is easier to see the PCollection differences; the passert package is not very detailed yet.

kythe/go/serving/pipeline/filetree.go
schroederc: The above `Clean` handles those.

kythe/go/serving/pipeline/filetree.go
schroederc: ✓

kythe/go/serving/pipeline/filetree.go
schroederc: ✓

kythe/go/serving/pipeline/filetree.go
jrtom: Clean handles ".." and ".", but not "...".  Not sure whether we would ever encounter the latter in this context, though, so it may be a non-issue.

I'd still recommend using a function for this conversion, so that the code will be more readable/clean:

    func currentAsEmpty(string path) string {
      if path == "." {
        return ""
      }
      return path
    }
    ...
    dir.Path = currentAsEmpty(filePath.Dir(dir.Path))
    // etc.

but I don't insist on it.

schroederc: ✓

PR #144, opened 2018-06-10
Obtain JVM (aka "external") names in a principled way.

benjyw:
Instead of hand-crafted logic, this defers to the compiler's own data structures, which do the right thing in multiple cases, such as type erasure, synthetic inner class ctor params, generics in method signatures, etc.

This change adds several test cases.

This supersedes https://github.com/google/kythe/pull/139 and replaces logic previously implemented in https://github.com/google/kythe/pull/120 and https://github.com/google/kythe/pull/118.


benjyw:
javatests pass, style checked.

PR #145, opened 2018-06-11
Fix JVM names of inner classes.

benjyw:
Previously we emitted `pkg.Outer.Inner`, but at the JVM level
there are no inner classes, and `Inner` compiles to a top-level
class in pkg `pkg` named `Outer$Inner`. So we should emit accordingly.

The new `methodWithInnerParam` test exposes the issue - prior to this
change, the source and binary indexers would have emitted different
method signatures.


benjyw:
javatests pass, style checked.

PR #149, opened 2018-06-12
Fix bug.  I should not write code while eating and also watching TV.

danielmoy-google:


kythe/go/extractors/config/extractor.go
```diff
@@ -208,7 +208,7 @@ func findConfig(configPath, repoDir string) (*ecpb.ExtractionConfiguration, erro
 	if os.IsNotExist(err) {
 		// TODO(danielmoy): This needs to be configurable by builder, language, etc.
 		return Load(mvn.DefaultConfig())
-	} else {
+	} else if err != nil {
```
creachadair: Oops! My bad for missing that. :)

PR #151, opened 2018-06-12
Add tests to ensure all used proto types and functions are registered with Beam

schroederc:
Unregistered types/functions become runtime panics on remote Beam runners.

### DIFF
--- a/kythe/go/serving/pipeline/BUILD
+++ b/kythe/go/serving/pipeline/BUILD
@@ -46,6 +46,7 @@ go_test(
     srcs = ["beam_test.go"],
     library = ":pipeline",
     deps = [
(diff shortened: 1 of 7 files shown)


danielmoy-google:
Well I wrote a comment but somehow I deleted it.  Brief question whether the failure message will be sufficiently descriptive to tell people what to update and where.  I guess probably it is?

schroederc:
Yeah, the error message contains the type/function that needs to be registered.

danielmoy-google:
Yea that was what my deleted comment pointed out.  I was contemplating whether a specific message like "you need to update beam.go to include missing stuff %v" or whatever might be necessary.

But maybe the fact that beam_test.go is the failing test, people should take the hint lol.

PR #153, opened 2018-06-13
Add PagedEdgeSet to Beam pipeline output

schroederc:


kythe/go/serving/pipeline/beam.go
```diff
@@ -503,3 +513,118 @@ func refKind(r *ppb.Reference) string {
...
+			Kind:    e.Kind,
+			Ordinal: e.Ordinal,
+		})
+	}
+}
+
+// reverseEdge emits the reverse of each *ppb.Edge, embedding the associated TargetNode.
+func reverseEdge(src *spb.VName, nodeStream func(**ppb.Node) bool, edgeStream func(**ppb.Edge) bool, emit func(*spb.VName, *ppb.Edge)) {
+	var node *ppb.Node
+	for nodeStream(&node) {
```
jrtom: could you express this as follows?

    if !nodeStream(&node) {
      node = &ppb.Node{Source: src}
    }

That seems a bit more concise.

kythe/go/serving/pipeline/beam.go
```diff
@@ -503,3 +513,118 @@ func refKind(r *ppb.Reference) string {
...
+}
+
+// groupEdges emits *srvpb.PagedEdgeSets and *srvpb.EdgePages for a node and its forward/reverse
+// edges.
+func groupEdges(src *spb.VName, nodeStream func(**ppb.Node) bool, edgeStream, revStream func(**ppb.Edge) bool, emitSet func(string, *srvpb.PagedEdgeSet), emitPage func(string, *srvpb.EdgePage)) {
+	set := &srvpb.PagedEdgeSet{}
+	// TODO(schroederc): paging
+
+	var node *ppb.Node
+	for nodeStream(&node) {
```
jrtom: AFAICT you're calling `convertPipelineNode()` for every element of `nodeStream`, which means that `set.Source` will be initialized based on the last element of `nodeStream`, rather than the first (as in `reverseEdge()`, above).

Could you instead do this?

    if nodeStream(&node) {
        node.Source = src
        set.Source = convertPipelineNode(node)
    } else {
       set.Source = &srvpb.Node{...}
    }

Or is there some reason you prefer to use the last rather than the first element?  (Even if there is, I would think you'd want to avoid calling `convertPipelineNode()` for each element.)

kythe/go/serving/tools/write_tables/write_tables.go
```diff
@@ -141,12 +141,14 @@ func runExperimentalBeamPipeline(ctx context.Context) error {
...
 	k := pipeline.FromEntries(s, entries)
 	shards := 8 // TODO(schroederc): better determine number of shards
+	edgeSets, edgePages := k.Edges()
 	xrefSets, xrefPages := k.CrossReferences()
 	beamio.WriteLevelDB(s, *tablePath, shards,
 		k.CorpusRoots(),
 		k.Directories(),
 		k.Decorations(),
 		xrefSets, xrefPages,
+		edgeSets, edgePages,
```
jrtom: If there are more items to add to what's being written out here, please add a TODO.

kythe/go/serving/pipeline/beam.go
```diff
@@ -45,11 +45,15 @@ func init() {
 	beam.RegisterFunction(defToDecorPiece)
 	beam.RegisterFunction(fileToDecorPiece)
 	beam.RegisterFunction(groupCrossRefs)
+	beam.RegisterFunction(groupEdges)
```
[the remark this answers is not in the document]
schroederc: Generally, all DoFns, CombineFns, and proto types need to be registered.

kythe/go/serving/pipeline/beam.go
```diff
@@ -503,3 +513,118 @@ func refKind(r *ppb.Reference) string {
...
+	rev := beam.ParDo(s, nodeToReverseEdges, k.nodes)
+
+	return beam.ParDo2(s, groupEdges, beam.CoGroupByKey(s, nodes, edges, rev))
+}
+
+// nodeToReverseEdges emits an *ppb.Edge with its SourceNode populated for each of n's edges.  The
+// key for each *ppb.Edge is its Target VName.
+func nodeToReverseEdges(n *ppb.Node, emit func(*spb.VName, *ppb.Edge)) {
+	for _, e := range n.Edge {
+		emit(e.Target, &ppb.Edge{
```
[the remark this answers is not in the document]
schroederc: I'm not convinced.  This would be the equivalent:

```
edge := *e // shallow copy (or create a newEdge func)
edge.Source = nil
edge.SourceNode = n
emit(e.Target, &edge)
```

I find that the existing code is more readable:

```
emit(e.Target, &ppb.Edge{
  SourceNode: n,
  Target:     e.Target,
  Kind:       e.Kind,
  Ordinal:    e.Ordinal,
})
```

kythe/go/serving/tools/write_tables/write_tables.go
schroederc: ✓

kythe/go/serving/pipeline/beam.go
schroederc: The input PCollection should have singletons per key so it shouldn't matter either way, but I changed it to match `reverseEdge`.

kythe/go/serving/pipeline/beam.go
schroederc: ✓

kythe/go/serving/pipeline/beam.go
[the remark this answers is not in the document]
jrtom: Ah, hadn't noticed that there's a `Source` field that needs to be nil-ed out.  That could be done inside the `newEdge()` function, but I understand why you might prefer the status quo.

kythe:
[Buildbot bazel-kythe status](http://localhost:8010/#/builders/1/builds/3) for https://github.com/google/kythe/commit/49f63b9605c66652b2f05a4770d47c62de5bb5e6: `failed 'bazel test ...' (failure)`

kythe:
[Buildbot go-kythe status](http://localhost:8010/#/builders/1/builds/4) for https://github.com/google/kythe/commit/49f63b9605c66652b2f05a4770d47c62de5bb5e6: `build successful`

schroederc:
I opened #157 for the test failure.  It seems to be a flake and not relevant to this change.

PR #167, opened 2018-06-18
Add missing runtime dependency for Java Truth library

schroederc:


creachadair:
There's no truth in Beauty,
and no beauty in Truth.

Issue #156, opened 2018-06-14
Docker in Docker - probably harmful?


danielmoy-google:
I'll preface this by saying I have precious little experience with Docker, so I might be missing things / misstating things / etc.  Mostly just writing this down because it's taking me longer than I was expecting to resolve.

We're currently looking for a clean way to package up extractrepo for deployment.  An obvious choice there is docker.  All well and good, except, oh wait, extractrepo itself builds and runs docker images:

$ bazel build kythe/go/platform/tools/extraction/extractrepo:extractrepo
$ cp ^ that thing -> some directory with a Dockerfile that just copies in the binary and runs it, along with any deps it needs (git commandline, kythe extraction scripts, etc)
$ docker build --tag test-extract aforementioned-directory

^ The above 3 steps can be replaced with a docker file that handles pulling in kythe, running tools/modules/update, building the extractrepo binary, etc.  But that (a) takes a long time, and (b) results in a much larger binary.

Anyways that's not the difficulty, the problem is then:

$ docker run -i -t test-extract -local ~/some/repo ~/some/output/dir
... exec: "docker": executable file not found in $PATH

There's various posts on how to do Docker in Docker, indicating this is technically possible to do:

https://github.com/jpetazzo/dind

But apparently it's not a great idea for all situations:
https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/
https://stackoverflow.com/questions/27879713/is-it-ok-to-run-docker-from-inside-docker

Particularly damning is that link from jpetazzo, since they're the one who came up with docker-in-docker in the first place.

I suppose one other alternative might be docker API?

https://godoc.org/github.com/docker/docker/client

danielmoy-google:
astroilov points out correctly that the other alternative approach here is to excise the internal docker bits from current extractrepo binary.

Right now it's [fairly baked in](https://github.com/google/kythe/blob/master/kythe/go/extractors/config/config_parser.go#L71), and pulling it out wouldn't be pretty.  In particular given the dynamic generation of the ephemeral dockerfile it produces, we'd need to have separate running steps there: running docker image that just takes config as input and generates an ephemeral dockerfile, and then something to run that resulting dockerfile.

I don't see a super clean approach either way.

danielmoy-google:
Behind door A, investigating the extent to which dind will work for us:

```
FROM docker:dind

ADD extractrepo extractrepo

ENTRYPOINT ["/extractrepo"]
```

[...]

PR #170, opened 2018-06-18
Example docker-in-docker file for kythe mvn extraction.

danielmoy-google:


danielmoy-google:
Ehhh I should fill out more detail in the Dockerfile comments on how to build and run it.

danielmoy-google:
Note this is also technically a backwards step on #156, but intentional.

PR #172, opened 2018-06-19
Add Go Riegeli file format Reader and Writer

schroederc:
This is simplified implementation of the C++ library at
https://github.com/google/riegeli/.

The Reader/Writer currently only support uncompressed, sequential access
of standard record chunks, but the format is 100% compatible with the
C++ library.


kythe/go/util/riegeli/block_test.go
```diff
@@ -0,0 +1,224 @@
...
+	var buf bytes.Buffer
+	w := &blockWriter{w: &buf}
+
+	// Write a full block
+	chunk := bytes.Repeat([]byte{0}, usableBlockSize)
+	n, err := w.Write(chunk)
+	if err != nil {
+		t.Fatal(err)
+	} else if expected := blockSize; n != expected {
+		t.Fatalf("Unexpected write size: expected: %d; found: %d", expected, n)
```
creachadair: Go conventionally orders these as got, want (i.e., found, expected).

kythe/go/util/riegeli/reader.go
```diff
@@ -0,0 +1,180 @@
...
+	return val, nil
+}
+
+func (r *Reader) readNextRecordChunk() error {
+	for r.valsReader == nil || r.numRecords == 0 {
+		c, err := r.r.Next()
+		if err != nil {
+			return err
+		} else if c.Header.ChunkType != recordChunkType {
+			// TODO(schroederc): support non-record chunk types (especially RecordsMetadata)
```
creachadair: Do these occur often enough to be worth logging?

kythe/go/util/riegeli/reader.go
```diff
@@ -0,0 +1,180 @@
...
+			continue
+		}
+
+		rec, err := decodeRecordChunk(c)
+		if err != nil {
+			return fmt.Errorf("error decoding record chunk: %v", err)
```
creachadair: s/error //

kythe/go/util/riegeli/reader.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return nil
+}
+
+type blockReader struct {
+	r   io.Reader
+	buf *bytes.Reader
+}
+
+// Read implements the io.Reader interface by skipping over the interleaven
```
creachadair: This seems a slightly odd interface to implement. What's the use case for reading the data stream without the record boundaries?

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+	"io"
+)
+
+func (w *Writer) ensureFileHeader() error {
+	if w.fileHeaderWritten {
+		return nil
+	}
+
+	// https://github.com/google/riegeli/blob/master/doc/riegeli_records_file_format.md#file-signature
+	_, err := (&chunk{Header: chunkHeader{ChunkType: fileSignatureChunkType}}).WriteTo(w.w, w.w.pos)
```
creachadair: I think this'd be a little easier to write if you separate the value. Also, given this is static, maybe we can just keep one around?

```
var fileHeader = &chunk{
   Header: chunkHeader{ChunkType: fileSignatureChunkType},
}
…
_, err := fileHeader.WriteTo(w.w, w.w.pos)
```

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+	buf[24] = byte(h.ChunkType)
+	binary.LittleEndian.PutUint64(buf[25:33], h.NumRecords) // overwrite buf[32] below
+	binary.LittleEndian.PutUint64(buf[32:40], h.DecodedDataSize)
+	hash := hashBytes(buf[8:])
+	binary.LittleEndian.PutUint64(buf[:8], hash)
+	return w.Write(buf[:])
+}
+
+// WriteTo writes the chunk to w, given its starting position within w.
+func (c *chunk) WriteTo(w io.Writer, pos int) (int, error) {
```
creachadair: This name collides semantically with `io.WriterTo`, but the latter doesn't take an offset. How about `WriteToOffset` or something like that?

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+	binary.LittleEndian.PutUint64(c.Header.DataHash[:], hashBytes(c.Data))
+	var buf bytes.Buffer
+	if _, err := c.Header.WriteTo(&buf); err != nil {
+		return 0, err
+	}
+	(&buf).Write(c.Data)
```
creachadair: `buf.Write(c.Data)`

It will take the address for you.

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+	padding := paddingSize(pos, &c.Header)
+	for i := 0; i < padding; i++ {
+		(&buf).WriteByte(0)
```
creachadair: 193, ibid.

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+	binary.LittleEndian.PutUint64(buf[32:40], h.DecodedDataSize)
+	hash := hashBytes(buf[8:])
+	binary.LittleEndian.PutUint64(buf[:8], hash)
+	return w.Write(buf[:])
+}
+
+// WriteTo writes the chunk to w, given its starting position within w.
+func (c *chunk) WriteTo(w io.Writer, pos int) (int, error) {
+	binary.LittleEndian.PutUint64(c.Header.DataHash[:], hashBytes(c.Data))
+	var buf bytes.Buffer
```
creachadair: Maybe for future consideration: It's probably worth keeping a buffer around for this that can be reset. The allocations add up quickly under load.

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+
+	buf.Write(r.CompressedSizes)
+	buf.Write(r.CompressedValues)
+
+	return buf.Bytes()
+}
+
+// WriteTo implements the io.WriterTo interface for chunkHeaders.
+func (h *chunkHeader) WriteTo(w io.Writer) (int, error) {
+	var buf [chunkHeaderSize]byte
```
creachadair: Can you include a little comment with the layout?

```
	// Bytes  :: Description
	// 0..7   -- header checksum (format?)
	// 8..15  -- data size
	// 16..23 -- data checksum (format?)
	// 24     -- chunk type
	// 25..31 -- record count (56 bits)
	// 32..39 -- decoded data size
```

I know this is part of the spec, but with all the indices it's a little tricky to follow.

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+	binary.LittleEndian.PutUint64(buf[8:16], h.DataSize)
+	copy(buf[16:], h.DataHash[:])
+	buf[24] = byte(h.ChunkType)
+	binary.LittleEndian.PutUint64(buf[25:33], h.NumRecords) // overwrite buf[32] below
```
creachadair: Subtle. Can you please comment what's going on here (i.e., using the 8th byte as scratch for a 56-bit value)?

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+
+	r.decodedSize += size
+	r.numRecords++
+	return nil
+}
+
+// encode returns the binary-encoding of the Riegeli record chunk.
+func (r *recordChunk) encode() []byte {
+	// TODO(schroederc): compression
+	buf := bytes.NewBuffer(make([]byte, 0, 1+binary.MaxVarintLen64+len(r.CompressedSizes)+len(r.CompressedValues)))
```
creachadair: Again maybe for the future: Consider passing in a buffer here to avoid reallocating (the writer doesn't support concurrent writers anyway)

kythe/go/util/riegeli/writer.go
```diff
@@ -0,0 +1,202 @@
...
+
+// A blockWriter interleaves blockHeaders inside chunks of data.  Each
+// blockHeader interrupts a single chunk, providing both its relative starting
+// and ending positions.
+type blockWriter struct {
+	w   io.Writer
+	pos int
+}
+
+// Write implements the io.Writer interface.  The given []byte must be a single
```
creachadair: If this is really meant to be a record write, I think we should consider using a name other than `io.Writer` which doesn't imply any framing. Someone passing this into, say `fmt.Fprint`, would be surprised to discover multiple records if the implementation happens to issue two writes.

Maybe `WriteRecord`?

kythe/go/util/riegeli/reader.go
schroederc: Maybe once I implement the RecordsMetadata and transposed chunks.  For now, it should happen about once per file for the file signature.

kythe/go/util/riegeli/writer.go
schroederc: ✓

kythe/go/util/riegeli/writer.go
schroederc: ✓

kythe/go/util/riegeli/writer.go
schroederc: ✓

kythe/go/util/riegeli/writer.go
schroederc: ✓

kythe/go/util/riegeli/writer.go
schroederc: ✓

kythe/go/util/riegeli/writer.go
schroederc: ✓

kythe/go/util/riegeli/reader.go
schroederc: It's without the block boundaries, not the chunk headers.  The block headers are unnecessary except for verifying their hash and seeking to a particular position.  I added the hash check and the seeking will happen in the future.

kythe/go/util/riegeli/reader.go
schroederc: ✓

kythe/go/util/riegeli/block_test.go
schroederc: ✓

schroederc:
The missing features of compression, transposed blocks, and seeking make it undesirable to upstream at this time, but maybe once the library is fleshed out we can ask.

PR #173, opened 2018-06-19
Add copy output step to the end of kythe maven extraction.

danielmoy-google:
Yet another affordance for docker-in-docker described in #156, which we
probably should get rid of long term.  For the short/medium term though,
this makes calling the eventual docker image a lot nicer, since we can
specify output to arbitrary top-level directories.

### DIFF
--- a/kythe/go/extractors/config/extractor.go
+++ b/kythe/go/extractors/config/extractor.go
@@ -56,15 +56,23 @@ type Repo struct {
 	// default config.
 	ConfigPath string
 
-	// An optional path that dictates where temporary repo copies should go.
-	// TODO(#156): this should be removed as soon as we refactor code to no
-	// longer use docker-in-docker.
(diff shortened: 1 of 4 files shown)


kythe/go/extractors/config/extractor.go
```diff
@@ -60,50 +60,76 @@ type Repo struct {
 	// TODO(#156): this should be removed as soon as we refactor code to no
 	// longer use docker-in-docker.
 	TempRepoDir string
+
+	// An optional path that dicates where temporary output should go before
```
creachadair: "dictates" sp. 

But I'd simplify:  `// An optional temporary directory path to write output before copying to OutputPath.`

Maybe say something about why this helps with d-in-d?

creachadair: Also: What happens if it's not set.

PR #176, opened 2018-06-23
Inherit language from file vname when creating new anchor.

nbeloglazov:
This helps to create anchors from other languages. For example JavaScript indexer can create "imputes" edges pointing from .java files.

Tested:
  bazel test //kythe/javatests/...


creachadair:
Ah, not only that, but this was supposed to have been the behaviour all along.

I should probably document that better.

nbeloglazov:
I didn't update [implicit anchors](https://github.com/google/kythe/blob/master/kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java#L155) because it was breaking [jvm edges](https://github.com/google/kythe/blob/master/kythe/javatests/com/google/devtools/kythe/analyzers/jvm/testdata/jvm_edges.kythe_verifier.txt#L11).

PR #178, opened 2018-06-25
entrystream: add support for Riegeli files

schroederc:


kythe/go/platform/tools/entrystream/entrystream.go
```diff
@@ -85,6 +92,24 @@ func main() {
...
+				rec, err := r.Next()
+				if err == io.EOF {
+					return nil
+				} else if err != nil {
+					return err
+				}
+				var e spb.Entry
+				if err := proto.Unmarshal(rec, &e); err != nil {
+					return err
+				} else if err := emit(&e); err != nil {
```
salguarnieri: You could get rid of the else since you are returning in the preceding if

kythe/go/platform/tools/entrystream/entrystream.go
```diff
@@ -55,8 +59,11 @@ type entrySet struct {
 }
 
 var (
-	readJSON        = flag.Bool("read_json", false, "Assume stdin is a stream of JSON entries instead of protobufs")
-	writeJSON       = flag.Bool("write_json", false, "Print JSON stream as output")
+	readJSON     = flag.Bool("read_json", false, "Assume stdin is a stream of JSON entries instead of protobufs")
+	writeJSON    = flag.Bool("write_json", false, "Print JSON stream as output")
```
salguarnieri: Can you either add input validation to make sure conflicting options aren't set (readJSON && readRiegeli) or switch to an enum since there are 3 options now?

schroederc: Switched to enum

kythe/go/platform/tools/entrystream/entrystream.go
schroederc: I don't think the style guide says anything about this case.  Most of our codebase joins the lines with an `else` like this, though.

PR #2794, opened 2018-06-27
Riegeli: add support for decoding transposed chunks

schroederc:
The transposed chunk decoder is used for both RecordsMetadata chunks and
regular transposed record chunks.  Transposed chunks decompose records
into their component key-value entries and store alike entries (by type
and field) together to achieve better compression.  Transposed chunks
are decoded by interpreting a state machine encoded alongside the
compressed record data buffers.

Current testing involves reading golden data until an encoder is
implemented.


kythe/go/util/riegeli/golden_test.go
```diff
@@ -19,24 +19,60 @@ package riegeli_test
...
 )
 
+func BenchmarkGoldenTestData(b *testing.B) {
+	for _, variant := range goldenRiegeliFileVariants {
+		file := strings.Join([]string{goldenRiegeliFilePrefix, variant, "riegeli"}, ".")
+		b.Run(variant, func(b *testing.B) { benchGoldenData(b, file) })
+	}
+}
+
+func benchGoldenData(b *testing.B, goldenRiegeliFile string) {
```
creachadair: Can you please document what this benchmark is benchmarking, at least in broad terms?

kythe/go/util/riegeli/reader.go
```diff
@@ -22,36 +22,73 @@ import (
...
+			// ignore chunks with no records; even for unknown chunk types
 			continue
 		}
-		r.recordReader, err = newRecordChunkReader(c)
-		if err != nil {
-			return fmt.Errorf("bad record chunk: %v", err)
+
+		switch c.Header.ChunkType {
+		case fileSignatureChunkType:
+			// TODO (schroederc): verify
```
creachadair: What does ignoring this do? Will the results be valid if we skip the signatures?

kythe/go/util/riegeli/golden_test.go
schroederc: ✓

kythe/go/util/riegeli/reader.go
schroederc: Ignoring it means we could be reading a corrupt file, but skipping it won't be a problem for a valid file.

creachadair: It might be worth logging that. Although, if you're going to implement this soon anyway, it's not crucial.

schroederc: Better to just fix it.

kythe/go/util/riegeli/reader.go
```diff
@@ -94,6 +96,15 @@ func (r *Reader) ensureRecordReader() error {
 	return nil
 }
 
+func verifySignature(c *chunk) error {
```
creachadair: I might just inline this, since it's shorter without. Your call, though.

PR #2796, opened 2018-06-28
viewindex: add --file flag to print a single file's contents

schroederc:


kythe/go/platform/tools/viewindex/viewindex.go
```diff
@@ -70,6 +72,17 @@ func main() {
...
+	} else if *printFile != "" {
+		for _, f := range idx.Files {
+			if f.Info.GetDigest() == *printFile {
+				if _, err := os.Stdout.Write(f.Content); err != nil {
+					log.Fatal(err)
+				}
+				return
+			}
+		}
+		fmt.Fprintf(os.Stderr, "File digest %q not found\n", *printFile)
```
creachadair: The rest of the code uses log.Fatal for this (which, modulo timestamps, is basically equivalent).

schroederc: Fix'd

PR #2805, opened 2018-07-02
Riegeli: properly attempt to read byte until an error

schroederc:


schroederc:
https://godoc.org/io#Reader:

> Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) == 0. Callers should treat a return of 0 and nil as indicating that nothing happened; in particular it does not indicate EOF.

kythe/go/util/riegeli/compression.go
```diff
@@ -56,8 +56,13 @@ type byteReadCloser struct{ io.ReadCloser }
 // ReadByte implements the io.ByteReader interface.
 func (b byteReadCloser) ReadByte() (byte, error) {
 	var buf [1]byte
-	_, err := b.Read(buf[:])
-	return buf[0], err
+	for {
```
creachadair: This seems dangerous. What circumstances lead to the need for a retry? Can that happen more than once?

I'm concerned that if we hit the pathological state, this could turn into a CPU spinner. I wonder if we could get the same benefit from just "trying again" and failing if it fails the second time.

schroederc: I converted to using `io.ReadFull`.  It does the exact same thing, but it looks less awful.

kythe/go/util/riegeli/compression.go
```diff
@@ -56,13 +56,8 @@ type byteReadCloser struct{ io.ReadCloser }
...
 func (b byteReadCloser) ReadByte() (byte, error) {
 	var buf [1]byte
-	for {
-		n, err := b.Read(buf[:])
-		if n == 0 && err == nil {
-			continue
-		}
-		return buf[0], err
-	}
+	_, err := io.ReadFull(b.ReadCloser, buf[:])
```
creachadair: Maybe:

    // A read can return no data without error, and Riegeli does this.
    // ReadFull to ensure we keep trying till we get the byte we want.

PR #2812, opened 2018-07-09
viewindex: add support for .kzip files

schroederc:


kythe/go/platform/tools/viewindex/viewindex.go
```diff
@@ -61,6 +66,61 @@ func main() {
 	}
 
 	path := flag.Arg(0)
+	if !strings.HasSuffix(path, ".kzip") {
+		// Assume deprecated .kindex, by default
```
danielmoy-google: does it actually have a .kindex suffix usually?  I thought so, but maybe not.

kythe/go/platform/tools/viewindex/viewindex.go
```diff
@@ -61,6 +66,61 @@ func main() {
...
+			idx.Files = append(idx.Files, &apb.FileData{Content: data, Info: f.Info})
+		}
+
+		return en.Encode(idx)
+	}); err != nil {
+		log.Fatalf("Error scanning kzip %q: %v", path, err)
+	}
+}
+
+func viewKIndex(path string) {
```
danielmoy-google: the idea of not having viewZip / viewKIndex is that the latter is going to go away sometime?  If so that sounds ok, otherwise maybe have separate functions

kythe/go/platform/tools/viewindex/viewindex.go
schroederc: It does, but the previous behavior took any extension.   I'm leaving that (minus ".kzip") until .kindex is removed.

kythe/go/platform/tools/viewindex/viewindex.go
schroederc: Yeah, .kindex is going to be removed.

PR #2814, opened 2018-07-10
Migrate from java_common.create_provider to new constructor JavaInfo() [1]

schroederc:
java_common.create_provider will be deprecated soon.

[1] https://docs.bazel.build/versions/master/skylark/lib/JavaInfo.html#JavaInfo


shahms:
Does `JavaInfo` work all the way back to Bazel 0.13 or does that need updating as well?

schroederc:
Yeah, it works for me with 0.13.0.

PR #2815, opened 2018-07-10
properly handle utf16 surrogate pairs

evmar:
The 'utf8' module maps nodejs UTF-16 offsets <-> Kythe UTF-8 offsets.
Because of surrogate pairs, a single code point in UTF-8 may be multiple
UTF-16 code units.

The fix also includes a unit test for the module that includes some
surrogate pairs.  To run this unit test we bring in a dependency on
Jasmine and adjust the test-running machinery to run both the unit
test and the end-to-end test.

This code was mostly written by Bowen Ni <bowen1897@gmail.com>; I just
added the Jasmine glue for running as part of open source Kythe.

### DIFF
--- a/kythe/typescript/utf8.ts
+++ b/kythe/typescript/utf8.ts
@@ -28,24 +28,36 @@
 class Scanner {
   constructor(private buf: Buffer, public ofs = 0) {}
 
-  /** Scans forward one codepoint. */
-  scan(): void {
-    let byte = this.buf[this.ofs++];
+  /** Scans forward one codepoint and returns the delta in UTF-16 offset. */
+  scan(): number {
+    const byte = this.buf[this.ofs++];
+
+    // This code interprets the bytes following the bit patterns found in
+    //   https://en.wikipedia.org/wiki/UTF-8#Description
+    // to scan forward by one code point.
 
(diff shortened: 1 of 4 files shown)


kythe/typescript/utf8.ts
```diff
@@ -28,24 +28,32 @@
...
+  scan(): number {
+    const byte = this.buf[this.ofs++];
 
     if ((byte & 0b10000000) === 0) {
       // Common case: ASCII.
-      return;
+      return 1;
     }
 
-    if ((byte & 0b11100000) == 0b11000000) {
```
zrlk: Please use named constants instead of binary literals.

evmar: These come directly from the UTF-8 definition.  See the table at
  https://en.wikipedia.org/wiki/UTF-8#Description

So I'm not quite sure how to name them except by their value.  Any suggestions?

zrlk: That's a valid objection. Maybe leave a comment describing where they came from (with the link you'd posted)?

evmar:
Oops, I rebased and it nuked your review.  (On the projects I work on doing this doesn't drop the review status.)  Sorry!  Can you reapprove?

PR #2816, opened 2018-07-10
Support extracting/analyzing Go packages with vendor/ dirs

schroederc:


kythe/go/indexer/indexer.go
```diff
@@ -944,6 +952,9 @@ func vnameToImport(v *spb.VName, goRoot string) string {
 		return strings.TrimSuffix(tail, filepath.Ext(tail))
 	}
 	trimmed := strings.TrimSuffix(v.Path, filepath.Ext(v.Path))
+	if path := trimVendorPrefix(trimmed); path != trimmed {
```
danielmoy-google: Might be worth a comment up here, but maybe not.

PR #2818, opened 2018-07-11
Go indexer: add --continue flag to log errors without halting analysis

schroederc:


kythe/go/indexer/cmd/go_indexer/go_indexer.go
```diff
@@ -104,9 +105,19 @@ func main() {
 		docURL = u
 	}
 
+	visit := indexGo
+	if *contOnErr {
```
danielmoy-google: I don't know go style much.  What's the tradeoff to doing this vs having indexGo return a function?

(putting the logic inside indexGo instead of out here, and then calling visitPath(ctx, path, indexGo(*contOnErr)))

schroederc: It's simpler to just handle the logic inside of indexGo.

kythe/go/indexer/cmd/go_indexer/go_indexer.go
```diff
@@ -143,14 +133,18 @@ func checkMetadata(ri *apb.CompilationUnit_FileInput, f indexer.Fetcher) (*index
...
-// indexGo is a visitFunct that invokes the Kythe Go indexer on unit.
+// indexGo is a visitFunc that invokes the Kythe Go indexer on unit.
 func indexGo(ctx context.Context, unit *apb.CompilationUnit, f indexer.Fetcher) error {
 	pi, err := indexer.Resolve(unit, f, &indexer.ResolveOptions{
 		Info:       indexer.XRefTypeInfo(),
 		CheckRules: checkMetadata,
 	})
 	if err != nil {
-		return err
+		if !*contOnErr {
```
danielmoy-google: so this only continues on errors for indexer.Resolve - if pi.Emit issues an error it still returns error.

If that's intentional then this is ok, otherwise what I meant by earlier comment was to wrap this whole logic with the same if statement you had up top in previous commit, but keep it inside indexGo.

I'm unclear whether it'd be worth still dying on emit errors, so maybe that's what you actually want anyways.

schroederc: Good point.  I made that change too quickly.

PR #2821, opened 2018-07-11
Sample web UI: keep track of current xrefs target state

schroederc:


robinp:
Could add a bit of background in the description, the change subject is quite opaque for the external observer.

PR #2833, opened 2018-07-13
don't warn on module imports that fail to resolve

evmar:
As discussed in the test case, valid type-checked programs can be in
this state, so we shouldn't spam the logs about it.


evmar:
Do I need to do anything to get this merged?

PR #2836, opened 2018-07-16
Update LLVM/Clang.

zrlk:


schroederc:
As mentioned offline, the buildbot checks will run `./tools/modules/update.sh` to account for any LLVM version changes.

PR #2839, opened 2018-07-16
Add go.mod file based on current WORKSPACE

schroederc:
This file defines kythe.io/ as the root of a Go 1.11 module requiring
the given imports.  When using a `go` tool that includes support for
modules (currently the tool must be built from HEAD), Kythe's Go code
will now be built and tested with the same dependency versions as
`bazel test` uses.

Note: the `go.mod` file must currently be kept in sync with our
WORKSPACE file manually


schroederc:
No, there's on an ongoing issue: https://github.com/google/kythe/issues/2834

Issue #2834, opened 2018-07-13
Bazel `-c opt` C++ tests fail


schroederc:
Various failures occur from 12 tests when testing with `bazel test -c opt //...`.

Example 1:

```
INFO: From Testing //kythe/cxx/common:path_utils_test:
==================== Test output for //kythe/cxx/common:path_utils_test:
[libprotobuf ERROR external/com_google_protobuf/src/google/protobuf/descriptor_database.cc:58] File already exists in database:
 google/protobuf/any.proto
[libprotobuf FATAL external/com_google_protobuf/src/google/protobuf/descriptor.cc:1394] CHECK failed: generated_database_->Add(
encoded_file_descriptor, size):
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: generated_database_->Add(encoded_file_descriptor, size):
```

Example 2:

```
FAIL: //kythe/cxx/extractor/testdata:extract_verify_std_string_test (see /usr/local/google/home/schroederc/.cache/bazel/_bazel_
schroederc/c48582765972a15180970afb1bdc98a2/execroot/io_kythe/bazel-out/k8-opt/testlogs/kythe/cxx/extractor/testdata/extract_ve
rify_std_string_test/test.log)
INFO: From Testing //kythe/cxx/extractor/testdata:extract_verify_std_string_test:
==================== Test output for //kythe/cxx/extractor/testdata:extract_verify_std_string_test:
Two /kythe/code facts about a node differed in value:
  App(vname, (vsprintf#n#builtin, "", "", "", c++))
...
```

zrlk:
Proto's upset because we're evidently linking in two copies of any.proto's generated code.

In the second issue, we're seeing different definitions for some builtin functions. This happens with -c opt because Bazel passes `-D_FORTIFY_SOURCE=1`, whereas in the default configuration it does not.

zrlk:
For path_utils_test, the linker invocation isn't appreciably different between opt and fastbuild (modules_checked_empty jumps around, but that shouldn't matter).

We're passing `-lexternal_Scom_Ugoogle_Uprotobuf_Slibany_Uproto`, though, which could be relevant. Maybe something has changed with protobuf?

zrlk:
`-c opt` adds the following to the (static) descriptor pool (breaking on `google::protobuf::DescriptorPool::InternalAddGeneratedFile`):
```
[...]

PR #2841, opened 2018-07-17
Update protobuf to 3.6.0.

zrlk:
This addresses #2834. The actually important change
is adding :_internal_wkt_protos_genrule to a [proto_lang_toolchain](https://docs.bazel.build/versions/master/be/protocolbuffer.html#proto_lang_toolchain.blacklisted_protos)
rule's blacklist. It is not clear how things were expected to work without it.


WORKSPACE
```diff
@@ -256,34 +256,18 @@ http_archive(
...
-http_archive(
-    name = "com_google_protobuf_cc",
-    strip_prefix = "protobuf-106ffc04be1abf3ff3399f54ccf149815b287dd9",
-    urls = ["https://github.com/google/protobuf/archive/106ffc04be1abf3ff3399f54ccf149815b287dd9.zip"],
-)
-
-# This is required by the proto_library implementation for its
-# :java_toolchain rule.
-http_archive(
-    name = "com_google_protobuf_java",
```
shahms: Are these other aliases no longer required?

zrlk: Evidently not, as building //... works. I don't know what they were pointing to and I can't find documentation saying that they're necessary.

PR #2850, opened 2018-07-18
Update third_party/bazel extra_actions_base.proto

schroederc:


third_party/bazel/src/main/protobuf/extra_actions_base.proto
```diff
@@ -20,6 +20,8 @@ package blaze;
 
 option java_multiple_files = true;
 option java_package = "com.google.devtools.build.lib.actions.extra";
+// option cc_api_version = 2;
```
salguarnieri: do these need to be here?

Issue #2795, opened 2018-06-27
Should support extractrepo with gradle in addition to maven


danielmoy-google:
Currently the extractrepo binary only works with maven.  This is a tracking issue to make that work with gradle as well.

Have a test repo set up at [here](https://github.com/danielmoy-google/gradle-kythe-extract-test) that we can test extraction on.

danielmoy-google:
Open things:

* Need an equivalent of `mvn-extract.sh` for gradle
  * This would necessitate figuring out what gradle's analog of `forceJavacCompilerUse`, `maven.compiler.fork`, and `maven.compiler.executable` are.
* Likely need an equivalent of `MvnPomPreprocessor.java` to handle modifying `build.gradle` in place.  Unless the above subpoint is somehow easier to accomplish in gradle, in which case maybe this step can be skipped.
* extractrepo & dependencies likely need massive refactoring to not tunnelvision on maven-only.

danielmoy-google:
Ah back looking at this, some update:

* have extraction working with a local repo.
  * `maven.compiler.fork` and `.executable` are done inside the build.gradle file
* Not seeing a readily available library for modifying gradle's dsl for build.gradle
  * [jetbrains](https://github.com/JetBrains/android/tree/master/android/src/com/android/tools/idea/gradle/parser) has one, but it doesn't appear to work for arbitrary keys
  * [one of the core devs says there's no support for this as of a year ago](https://discuss.gradle.org/t/loading-updating-and-saving-a-build-gradle-file-using-java/22932/2)

So that's... less than ideal.  With Maven we (by we I mean craig) got around this because they just use xml, and there's a bunch of xml libraries.

For the immediate short term we can do an unspeakable hacky job there.  It's unclear whether we'd want to build up a whole parser library for build.gradle and maintain it on our own.

danielmoy-google:
Tangential concern/design point here, on where detection should happen.  Right now I'm leaning towards having `extractrepo` & co just take the maven/gradle/?? as an input bit.  Less convoluted logic in there would be better.  Given that the long term plan is to have a separate system figuring out language, dependencies, and whatnot, it would be a better fit in there anyways.

PR #2859, opened 2018-07-24
Add logic to append kythe javac-wrapper to gradle.build files.

danielmoy-google:
Initial work towards #2795.

### DIFF
--- a/kythe/go/extractors/config/wrapper/README.md
+++ b/kythe/go/extractors/config/wrapper/README.md
@@ -1,0 +1,59 @@
+# Kythe Extractor Wrapping
+
+This package contains resources for preprocessing a repository's build
+configuration before running Kythe Extraction via `extractrepo.go`.
+
+## Gradle
+
+`gradle_build_modifier.go` takes an input `gradle.build` file and appends the
+bits necessary for replacing javac calls with Kythe's `javac-wrapper.sh`.
+
+```
+allprojects {
+  gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+      options.fork = true
+      options.forkOptions.executable = '/opt/kythe/extractors/javac-wrapper.sh'
+    }
+  }
+}
+```
+
+If the input file already contains reference to
+`options.forkOptions.executable`, then `gradle_build_modifier.go` does nothing.
+
+### Future work
+
+The current implementation uses some simple file parsing, without actually
+understanding the structure.  If that becomes necessary in the future, it might
+be better to use the existing java libraries for `org.codehaus.groovy.ast` to
+properly parse the build.gradle file and have more precise picture.  In
+particular `org.codehaus.groovy.ast.CodeVisitorSupport` might be sufficient.
+
+## Maven
+
+Maven is handled in java by
+`com.google.devtools.kythe.platform.tools.MvnPomPreprocessor`.  It utilizes an
+xml library to parse and modify the mvn `pom.xml` config file in a similar way
+as described above for gradle.
+
+### Future work
+
+In theory if we can find a nice xml library for golang that supports reflecting
+into specific elements and modifying without knowing the whole file structure,
+then we could do away with the java binary for maven preprocessing.
+
+## CMake
+
+Coming soon™.
+
+## Bazel
+
+Actually we have no custom work here.  The logic was as follows:
+
+1. We already have a substantial pipeline using bazel extractor, so it should
+   "just work"
+
+2. Not many external projects use bazel, so we're targeting
+   maven/gradle/cmake/etc first.
+
--- a/kythe/go/extractors/config/wrapper/README.md
+++ b/kythe/go/extractors/config/wrapper/README.md
@@ -24,15 +24,15 @@ If the input file already contains reference to
 
 ### Future work
 
-The current implementation uses some simple file parsing, without actually
+The current implementation uses simple string-based matching, without actually
 understanding the structure.  If that becomes necessary in the future, it might
-be better to use the existing java libraries for `org.codehaus.groovy.ast` to
+be better to use the existing Java libraries for `org.codehaus.groovy.ast` to
 properly parse the build.gradle file and have more precise picture.  In
 particular `org.codehaus.groovy.ast.CodeVisitorSupport` might be sufficient.
 
 ## Maven
 
-Maven is handled in java by
+Maven is handled in Java by
 `com.google.devtools.kythe.platform.tools.MvnPomPreprocessor`.  It utilizes an
 xml library to parse and modify the mvn `pom.xml` config file in a similar way
 as described above for gradle.
@@ -41,19 +41,16 @@ as described above for gradle.
 
 In theory if we can find a nice xml library for golang that supports reflecting
 into specific elements and modifying without knowing the whole file structure,
-then we could do away with the java binary for maven preprocessing.
+then we could do away with the Java binary for maven preprocessing.
 
 ## CMake
 
-Coming soon™.
+Coming soon™.  https://github.com/google/kythe/issues/2861
 
 ## Bazel
 
-Actually we have no custom work here.  The logic was as follows:
-
-1. We already have a substantial pipeline using bazel extractor, so it should
-   "just work"
-
-2. Not many external projects use bazel, so we're targeting
-   maven/gradle/cmake/etc first.
+Actually we have no custom work here.  We extract compilation records from Bazel
+using the extra action mechanism.  The extractrepo tool therefore doesn't handle
+Bazel directly, but repositories using Bazel for languages we already support
+should work without extra effort.
 
--- a/kythe/go/extractors/config/wrapper/gradle_build_modifier.go
+++ b/kythe/go/extractors/config/wrapper/gradle_build_modifier.go
@@ -20,11 +20,11 @@ allprojects {
 `
 
 // This matches a line which sets the javac to use Kythe's javac-wrapper.sh
-var kytheMatcher = regexp.MustCompile(`^\s*options\.forkOptions\.executable\ =\ '/opt/kythe/extractors/javac-wrapper.sh'\s*$`)
+var kytheMatcher = regexp.MustCompile(`\s+options\.forkOptions\.executable\ =\ '/opt/kythe/extractors/javac-wrapper.sh'`)
 
 // This matches any line which sets a new javac executable, useful for detecting
 // edge cases which already modify javac.
-var javacMatcher = regexp.MustCompile(`^\s*options\.forkOptions\.executable\ =.*$`)
+var javacMatcher = regexp.MustCompile(`\s+options\.forkOptions\.executable\ =`)
 
 // PreProcessGradleBuild takes a gradle.build file and either verifies that it
 // already has the bits necessary to run kythe's javac wrapper, or adds that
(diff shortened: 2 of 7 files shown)


kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"regexp"
+)
+
+// These are the lines necessary for gradle build to use a different javac.
+var kytheJavacWrapper = `
```
danielmoy-google: I was ambivalent about whether to plop this into a datafile or just inline it here.

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+const testDataDir = "testdata"
+
+var multipleNewLines = regexp.MustCompile("\n{2,}")
+
+func filesEqual(got, want []byte) (bool, string) {
```
danielmoy-google: This is now the second or third or so file where we have this copied in.  Wonder if we should make a test util for it.  #2860 to track

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func TestPreprocess(t *testing.T) {
+	testcases := []struct {
+		inputFile          string
+		expectedOutputFile string
+	}{
+		{"modified-gradle.build", "modified-gradle.build"},
+		{"plain-gradle.build", "modified-gradle.build"},
+	}
```
danielmoy-google: I mean I guess I could add a case for the one that's supposed to fail and verify it forwards the error?

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if k {
+		// Already has the kythe javac-wrapper.
+		return nil
+	}
+	return appendKytheWrapper(gradleBuildFile)
+}
+
+func hasKytheWrapper(gradleBuildFile string) (bool, error) {
+	f, err := os.Open(gradleBuildFile)
```
creachadair: I think this'd be simpler if you slurp up the whole file:

```
bits, err := ioutil.ReadFile(path)
if err != nil {
  return false, err
}
if kytheMatcher.Match(bits) {
  return true, nil
}
if javacMatcher.Match(bits) {
  return false, fmt.Errorf(…)
}
return false, nil
```

Granted you'll get more than you need, but these files shouldn't be so big that reading the whole thing matters. (If they are, we're going to have a much bigger problem when we try to actually build from it :)

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
creachadair: Although it's unexported, please document this one, as it's being clever about whitespace. :)

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
```diff
@@ -0,0 +1,123 @@
...
+	if diff != "" {
+		return false, diff
+	}
+
+	return true, ""
+}
+
+func TestHasKythe(t *testing.T) {
+	testcases := []struct {
+		f string
```
creachadair: I suggest descriptive names instead of comments:

```
[]struct {
  fileName   string
  hasWrapper bool
  wantError  bool
}{
```

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
```diff
@@ -0,0 +1,123 @@
...
+		// Whether kythe javac wrapper should be present.
+		k bool
+		// Whether we expect an error.
+		e bool
+	}{
+		{"modified-gradle.build", true, false},
```
creachadair:
```
{fileName: "modified-gradle.build", hasWrapper: true},
{fileName: "plain-gradle.build"},
{fileName: "other-gradle.build", wantError: true},
```

kythe/go/extractors/config/wrapper/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Future work
+
+In theory if we can find a nice xml library for golang that supports reflecting
+into specific elements and modifying without knowing the whole file structure,
+then we could do away with the java binary for maven preprocessing.
+
+## CMake
+
+Coming soon™.
```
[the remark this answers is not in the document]
danielmoy-google: there was not, but now there is

kythe/go/extractors/config/wrapper/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Bazel
+
+Actually we have no custom work here.  The logic was as follows:
```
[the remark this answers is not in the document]
danielmoy-google: Ah right good point, especially for the context of repo preprocessing modification, we shouldn't need any

(*knock on wood*)

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -0,0 +1,79 @@
+package wrapper
```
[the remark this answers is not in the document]
danielmoy-google: Yea I fully expect to refactor it like this once it gets bigger.  For now I'm intentionally keeping it like this though.

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
danielmoy-google: Mm yea this is busted, the regex no longer match, I wonder if it's due to the newlines:

https://github.com/google/kythe/pull/2859/commits/d7825ad8d182fb292dc45205afdd5de1cd2d3ba9#diff-f5927cf9c3c71d011a630eb46cdbd0af

danielmoy-google: Ok getting rid of the ^$ checking fixes as expected:

https://github.com/google/kythe/pull/2859/commits/a62f2d4c6000b64b768653d5f401fe6d0d7c2b26

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -20,11 +20,11 @@ allprojects {
 `
 
 // This matches a line which sets the javac to use Kythe's javac-wrapper.sh
-var kytheMatcher = regexp.MustCompile(`^\s*options\.forkOptions\.executable\ =\ '/opt/kythe/extractors/javac-wrapper.sh'\s*$`)
+var kytheMatcher = regexp.MustCompile(`\s+options\.forkOptions\.executable\ =\ '/opt/kythe/extractors/javac-wrapper.sh'`)
```
creachadair: _Must_ there be spaces before `options`, or is this simply trying to ensure a word break? 

If it's the latter: Would `\b` suffice?

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -20,11 +20,11 @@ allprojects {
...
 
 // This matches any line which sets a new javac executable, useful for detecting
 // edge cases which already modify javac.
-var javacMatcher = regexp.MustCompile(`^\s*options\.forkOptions\.executable\ =.*$`)
+var javacMatcher = regexp.MustCompile(`\s+options\.forkOptions\.executable\ =`)
```
creachadair: 23, ibid.

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -45,25 +45,16 @@ func PreProcessGradleBuild(gradleBuildFile string) error {
...
-		if javacMatcher.MatchString(scanner.Text()) {
-			return false, fmt.Errorf("Found existing non-kythe javac override for file %s, which we can't handle yet.", gradleBuildFile)
-		}
+	if kytheMatcher.Match(bits) {
+		return true, nil
 	}
-	if err := scanner.Err(); err != nil {
-		return false, fmt.Errorf("Reading file %s: %v", gradleBuildFile, err)
+	if javacMatcher.Match(bits) {
+		return false, fmt.Errorf("Found existing non-kythe javac override for file %s, which we can't handle yet.", gradleBuildFile)
```
creachadair: nit: s/Found/found/ https://github.com/golang/go/wiki/CodeReviewComments#initialisms

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
```diff
@@ -16,6 +16,10 @@ const testDataDir = "testdata"
 
 var multipleNewLines = regexp.MustCompile("\n{2,}")
 
+// filesEqual compares two strings by collapsing irrelevant whitespace.
```
creachadair: It collapses newlines, what about spaces/tabs? 
(I don't know if newlines are allowed inside string literals, which I guess could be an issue for 0x20)

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
danielmoy-google: I guess if we want to be precise about it, actually I should stuff a \n out front there, because we really want this to be a separate new statement with no preceding // comment markers or anything.

Then \s+ to match arbitrary whitespace

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
danielmoy-google: ack

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
danielmoy-google: So it only collapses whitespace at the beginning/end of lines, which should be unnecessary.  I guess if there's a multiline string it could be a problem if the language being compared supports that (gradle and xml I think do not, so so far it hasn't come up)

kythe/go/extractors/config/wrapper/gradle_build_modifier.go
```diff
@@ -20,11 +20,11 @@ allprojects {
 `
 
 // This matches a line which sets the javac to use Kythe's javac-wrapper.sh
-var kytheMatcher = regexp.MustCompile(`\s+options\.forkOptions\.executable\ =\ '/opt/kythe/extractors/javac-wrapper.sh'`)
+var kytheMatcher = regexp.MustCompile(`\n\s+options\.forkOptions\.executable\ =\ '/opt/kythe/extractors/javac-wrapper.sh'`)
```
creachadair: I think you can get the same effect via `(?m)^\s*options\.forkOptions\.Executable\ =…`

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
creachadair: Well, XML supports `<![CDATA[…]>` sections, which can span multiple lines, but I would be slightly surprised to find a lot of those in build configs.

PR #2862, opened 2018-07-25
RDF: Encode /kythe/code facts in Base64.

### LINKED ISSUE #2855: MarkedSources messages do not survive encoding to RDF triples
[Original report](https://groups.google.com/d/msg/kythe/yK5fC2bP0Uo/gk-FUVOjBQAJ).

When a stream of entries is encoded to RDF triples, it appears the fact values for `/kythe/code` facts may not be encoded correctly. I suspect the issue is that we are reading out Unicode runes from a string value, and I'm not convinced the recoding preserves the binary value properly.

This may also fall out of interactions with Java String encoding, but I haven't yet tried to get a minimal repro.

creachadair:
Fixes #2855. Per https://www.w3.org/TR/2014/REC-n-triples-20140225, the syntax
of N-Triples is expressed over UTF-8 encoded Unicode code points.

We are escaping code sequences correctly, but not binary values.  The
/kythe/code fact carries a wire-format protobuf message, which is not UTF-8 and
must therefore be escaped. This commit checks specifically for that fact when
encoding triples, and Base64 encodes the value. This matches the JSON encoding
for proto3 messages.

### DIFF
--- a/kythe/go/storage/tools/triples/triples.go
+++ b/kythe/go/storage/tools/triples/triples.go
@@ -27,6 +27,7 @@ package main
 
 import (
 	"context"
+	"encoding/base64"
 	"flag"
 	"fmt"
 	"io"
@@ -41,6 +42,7 @@ import (
 	"kythe.io/kythe/go/util/flagutil"
 	"kythe.io/kythe/go/util/kytheuri"
 	"kythe.io/kythe/go/util/schema/edges"
+	"kythe.io/kythe/go/util/schema/facts"
 
 	spb "kythe.io/kythe/proto/storage_go_proto"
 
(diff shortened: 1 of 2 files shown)


kythe/go/storage/tools/triples/triples.go
```diff
@@ -156,6 +158,9 @@ func toTriple(entry *spb.Entry) (*rdf.Triple, error) {
 	if graphstore.IsEdge(entry) {
 		t.Predicate = entry.EdgeKind
 		t.Object = kytheuri.FromVName(entry.Target).String()
+	} else if entry.FactName == facts.Code {
```
jrtom: Test?

creachadair: There isn't one, and it's not worth the trouble to add one for this.

PR #2863, opened 2018-07-25
Add Kzip IndexWriter implementation; update protobuf

shahms:
Implement and test a C++ KzipWriter

### DIFF
--- a/kythe/cxx/common/kzip_writer.h
+++ b/kythe/cxx/common/kzip_writer.h
@@ -25,25 +25,40 @@
 
 namespace kythe {
 
+/// \brief Kzip implementation of IndexWriter.
+/// see https://www.kythe.io/docs/kythe-kzip.html for format description.
 class KzipWriter : public IndexWriterInterface {
  public:
+  /// \brief Constructs a Kzip IndexWriter which will create and write to
+  /// \param path Path to the file to create. Must not currently exist.
   static StatusOr<IndexWriter> Create(absl::string_view path);
+  /// \brief Constructs an IndexWriter from the libzip source pointer.
+  /// \param source zip_source_t to use as backing store.
+  /// See https://libzip.org/documentation/zip_source.html for ownership.
+  /// \param flags Flags to use when opening `source`.
   static StatusOr<IndexWriter> FromSource(zip_source_t* source,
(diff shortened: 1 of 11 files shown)


kythe/cxx/common/kzip_writer.h
```diff
@@ -0,0 +1,51 @@
...
+  StatusOr<std::string> WriteUnit(
+      const kythe::proto::IndexedCompilation& unit) override;
+  StatusOr<std::string> WriteFile(absl::string_view content) override;
+  Status Close() override;
+
+ private:
+  explicit KzipWriter(zip_t* archive);
+
+  zip_t* archive_;  // Owned, but must be manually deleted via `Close`.
+  std::vector<std::string>
```
creachadair: This might be a little nicer with the comment above. I get what clang-format is doing here but it seems silly. YMMV, though.

kythe/cxx/common/kzip_writer.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <zip.h>
+#include <vector>
+
+#include "kythe/cxx/common/index_writer.h"
+#include "kythe/cxx/common/status_or.h"
+
+namespace kythe {
+
+class KzipWriter : public IndexWriterInterface {
```
creachadair: Please document this—in particular the requirement that the caller arrange to call Close and check its result before giving up on the object.

kythe/cxx/common/kzip_writer_test.cc
```diff
@@ -0,0 +1,127 @@
...
+
+template <typename T> WithStatusFn<T> WithStatus(Status* status, T function) {
+  return WithStatusFn<T>{status, std::move(function)};
+}
+
+TEST(KzipReaderTest, RecapitulatesSimpleKzip) {
+  StatusOr<IndexReader> reader = KzipReader::Open(TestFile("stringset.kzip"));
+  ASSERT_TRUE(reader.ok()) << reader.status();
+
+  StatusOr<IndexWriter> writer =
```
creachadair: I think a comment here outlining (roughly) what this is actually writing into the test file would probably help. There's rather a lot of nesting to decode otherwise.

shahms: I refactored this to hopefully make it clearer.

PR #2870, opened 2018-07-25
Don't hard-code the workspace name in kzip tests

shahms:
This should help when importing.


danielmoy-google:
Hmm buildbot fails?

PR #2871, opened 2018-07-25
Move the correct formation of golang test data reading into testutil.

danielmoy-google:
Refactor existing usages to use that testutil.

Refactor broken existing usages to use that testutil.

### DIFF
--- a/kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
+++ b/kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
@@ -43,7 +43,7 @@ func TestHasKythe(t *testing.T) {
 
 	for _, tcase := range testcases {
 		// This should just ignore the file and do nothing.
-		hasWrapper, err := hasKytheWrapper(getPath(tcase.fileName))
+		hasWrapper, err := hasKytheWrapper(getPath(t, tcase.fileName))
 		if err != nil && !tcase.wantError {
 			t.Fatalf("Failed to open test gradle file: %v", err)
 		} else if err == nil && tcase.wantError {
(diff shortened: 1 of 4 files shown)


kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
```diff
@@ -98,16 +98,16 @@ func TestPreprocess(t *testing.T) {
...
 			t.Errorf("Expected input file %s to be %s, but got diff %s", tcase.inputFile, tcase.expectedOutputFile, diff)
 		}
 	}
 }
 
 // getPath returns a resolved filepath name for a file living in testdata/ directory.
-func getPath(f string) string {
-	return os.ExpandEnv(filepath.Join(testDataDir, f))
+func getPath(t *testing.T, f string) string {
+	return testutil.TestFilePath(t, filepath.Join(testDataDir, f))
```
creachadair: If you're going to pass in the `*testing.T`, please also add `t.Helper()` at the top.
(It ensures logs from the function get attributed to the right location)

kythe/go/test/testutil/testutil.go
```diff
@@ -171,6 +164,14 @@ func TrimmedEqual(got, want []byte) (bool, string) {
 	return diff == "", diff
 }
 
+func caller(up int) (file string, line int) {
```
creachadair: Maybe for a future commit: I think we can probably get rid of this if we use `t.Helper()` judiciously here too.

kythe/go/extractors/config/wrapper/gradle_build_modifier_test.go
danielmoy-google: done

PR #2879, opened 2018-07-27
Factor out common test-source-root routine

shahms:
This should further facilitate more reliable imports.

### DIFF
--- a/kythe/cxx/common/testutil.cc
+++ b/kythe/cxx/common/testutil.cc
@@ -26,6 +26,8 @@
 namespace kythe {
 namespace {
 
+// This must match the name from the workspace(name={name})
+// rule in the root WORKSPACE file.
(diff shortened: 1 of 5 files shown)


kythe/cxx/common/testutil.cc
```diff
@@ -0,0 +1,43 @@
...
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/string_view.h"
+#include "absl/strings/strip.h"
+#include "glog/logging.h"
+
+namespace kythe {
+namespace {
+
+constexpr char kDefaultWorkspace[] = "io_kythe";
```
creachadair: I assume this has to match the `workspace(name=X)` definition in the root `WORKSPACE` file? (If that's correct, would you mind adding a comment to that effect?)

shahms: Done.

PR #2886, opened 2018-07-31
Stop using some deprecated proto facilities

shahms:
This stops using deprecated proto facilities (`ByteSize()` and `SetTotalBytesLimit(int, int)`) as well as migrating the `CompilationUnit::FileInput.context` field from `deprecated` to removed


shahms:
I have a change restoring and emitting the `context` field from the extractor (as well as using `details`) for a more gradual rollout, but am stuck behind the bazel 0.16 change.

shahms:
PTAL

Issue #2899, opened 2018-08-03
ref/call anchors inside lambdas in macro invocations misattributed to macro itself


shahms:
```
#define CHECK(x) (x)
//- @Function defines/binding FunctionDecl
void Function();
void wrapper() {
  CHECK([]{
    //- @Function ref FunctionDecl
    //- @"Function()" ref/call FunctionDecl <-- fails
    Function();
  });
}
```

PR #2900, opened 2018-08-03
Add minimal reproduction for #2899

shahms:
Adds a disabled test case for mis-anchored ref/call edges nested in macros nested in lambdas. Regular `ref` is accurate, but `ref/call` points to the start of the macro invocation.

### DIFF
--- a/kythe/cxx/indexer/cxx/testdata/basic/macros_lambda_call.cc
+++ b/kythe/cxx/indexer/cxx/testdata/basic/macros_lambda_call.cc
@@ -1,0 +1,12 @@
+#define CHECK(x) (x)
+
+//- @Call defines/binding CallDecl
+void Call();
+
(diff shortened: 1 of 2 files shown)


kythe/cxx/indexer/cxx/testdata/basic/macros_lambda_call.cc
```diff
@@ -0,0 +1,12 @@
+#define CHECK(x) (x)
+
+//- @Call defines/binding CallDecl
```
creachadair: Can you add some non-verifier comments to this file, documenting what it's checking for?

shahms: Shoot, missed this. I can do that in a follow up.

PR #2907, opened 2018-08-08
doc comments are good

danielmoy-google:
I really should make a pre-commit hook for running lint

### DIFF
--- a/kythe/go/extractors/config/runextractor/backup/backup.go
+++ b/kythe/go/extractors/config/runextractor/backup/backup.go
@@ -36,10 +36,13 @@ import (
 	"path/filepath"
 )
 
+// A File is a simple holder for a backup of some file that supports Recovery
+// to restore the original version.
 type File struct {
 	orig, tmp string // file paths
 }
 
+// New backs a given file up to an adjacent temporary file.
 func New(orig string) (*File, error) {
 	f, err := os.Open(orig)
 	if err != nil {


kythe/go/extractors/config/runextractor/backup/backup.go
```diff
@@ -36,10 +36,13 @@ import (
 	"path/filepath"
 )
 
+// A File is a simple holder for a backup of some file that supports Recovery
```
creachadair: I don't think "simple" adds much here. Suggestion:

```
// A File records the locations of an original file and a backup copy of that file.
// The Restore method replaces the contents of the original with the backup,
// overwriting any changes that were made since the backup was created.
```

or something like that.

kythe/go/extractors/config/runextractor/backup/backup.go
```diff
@@ -36,10 +36,13 @@ import (
...
+// to restore the original version.
 type File struct {
 	orig, tmp string // file paths
 }
 
+// New backs a given file up to an adjacent temporary file.
```
creachadair: I'd suggest more detail here:

```
// New creates a backup copy of the specified file, located in the same directory.
// The caller should ensure the Release method is called when the backup is no
// longer needed, to clean up.
```

or words to that effect.

kythe/go/extractors/config/runextractor/backup/backup.go
danielmoy-google: done

kythe/go/extractors/config/runextractor/backup/backup.go
danielmoy-google: done

PR #2908, opened 2018-08-08
Add list of required env vars.

danielmoy-google:
Also populate an empty cmake extractor shim and remove some of the
java-specific verbiage in prep for #2861.


kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/google/subcommands"
+)
+
+type cmakeCommand struct {
+	cmdutil.Info
+
+	buildFile       string
+	javacWrapper    string
+	pomPreProcessor string
```
shahms: none of these fields are relevant for CMake

PR #2910, opened 2018-08-09
Use the new Bazel C++ toolchain API

mhlopko:
This PR makes Kythe forward compatible with Bazel 0.16 which introduced
a new Skylark API to the C++ toolchain
(https://docs.google.com/document/u/1/d/1g91BWJITcYw_X-VxsDC0VgUn5E9g0kRBGoBSpoO41gA/edit).
Backwards compatibility is maintained.

While at it, I:

* removed unused crosstool fields
* since I cannot write Skylark code without buildifier formatting it if my life
  dependent on it, I also reformatted cc_indexer_test.bzl file.

I'll gladly split this work into multiple PRs if that's preferred.

### DIFF
--- a/tools/build_rules/verifier_test/cc_indexer_test.bzl
+++ b/tools/build_rules/verifier_test/cc_indexer_test.bzl
@@ -60,7 +60,7 @@ _INDEXER_FLAGS = {
     "experimental_drop_cpp_fwd_decl_docs": False,
 }
 
-def _compiler_options(cpp, copts, includes):
+def _compiler_options(ctx, cpp, copts, includes):
     """Returns the list of compiler flags from the C++ toolchain."""
 
     # Bazel is missing these attributes until 0.16.0,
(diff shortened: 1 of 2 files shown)


mhlopko:
Still trying to run tests locally. Not giving up :)

mhlopko:
Ok, ran tests locally, output identical with master. Which makes me feel like this is not tested too well? In all other migrations there were corner cases that needed some attention. Thanks for taking a look!

creachadair:
We've had some buildbot issues because of other Bazel-induced breakage, but I'll tag it and see if it works.

creachadair:
> Which makes me feel like this is not tested too well? 

I'm not sure which "this" you mean. It's not as if there's meaningful test support in Skylark apart from whether the rest of the repository actually builds and invokes its tests. There should be fairly complete unit test coverage throughout the workspace, though: `bazel test //kythe/...` should exercise the things we depend upon directly.

shahms:
Also, the local toolchain options are generally only used on macOS so there's a non-zero chance this breaks there.  If so, I'll take a look.

mhlopko:
Thank you all for a quick review and merge! Let me know if you need help with debugging.

Issue #2864, opened 2018-07-25
Update existing uses of .kindex and indexpack to use .kzip instead


creachadair:
Various command-line tools (chiefly indexers and tests) uses .kindex files and indexpack directories to ship compilation records around. Update these to use the .kzip format instead, and remove the .kindex and indexpack support once that is complete.

PR #2911, opened 2018-08-09
Add package extutil, to bridge .kzip/.kindex output.

creachadair:
In support of #2864, add a library to make it easier for extractors that
currently emit .kindex files to switch between that and .kzip format.  This
will allow an extractor to emit either format until we are finished with the
.kindex migration, at which point this can be collapsed and removed.


creachadair:
I will fix the typo in the commit log before merging. >.<

creachadair:
> Tests?

Not worth it. The code it calls is already tested, and this is just temporary plumbing.

Issue #2905, opened 2018-08-06
Do maven preprocessing in golang, not a java jar


danielmoy-google:
For now the java xml library we're using works pretty well.

However it does require an additional os.exec - it'd be nice to do this preprocessing in golang native, since that's what we appear to be settled on for most of the execution.

danielmoy-google:
https://github.com/beevik/etree maybe?

creachadair:
https://github.com/subchen/go-xmldom is another option too.

PR #2917, opened 2018-08-09
Port of MvnPomPreprocessor to golang

danielmoy-google:


danielmoy-google:
Addresses #2905

kythe/go/extractors/config/runextractor/README.md
```diff
@@ -32,6 +32,7 @@ Because `runextractor` is invoked in a manner not conducive to cleanly passing
 commandline flags, there is non-trivial setup done with environment variables.
 When calling `runextractor`, here are the relevant environment variables:
 
+* **KYTHE_CORPUS**: The corpus set for extracted files.
```
creachadair: Can you please also mention which of these are "allowed" vs. "required" for things to work?

And maybe give some examples or other context?

creachadair: elsewhere we've mostly used the phrase "corpus label" as the term of art for this—the corpus is the group of files, the corpus label is its identifier.

kythe/go/extractors/config/runextractor/README.md
```diff
@@ -32,6 +32,7 @@ Because `runextractor` is invoked in a manner not conducive to cleanly passing
...
 * **KYTHE_ROOT_DIRECTORY**: The root path for file input to be extracted.
```
creachadair: Is this a relative path? (Or may it be either?)

kythe/go/extractors/config/runextractor/README.md
```diff
@@ -32,6 +32,7 @@ Because `runextractor` is invoked in a manner not conducive to cleanly passing
...
 * **KYTHE_OUTPUT_DIRECTORY**: The path for storing output.
```
creachadair: Include `REAL_JAVAC` and `JAVAC_EXTRACTOR_JAR` too?

kythe/go/extractors/config/runextractor/gradlecmd/gradlecmd.go
```diff
@@ -52,6 +53,11 @@ func (g *gradleCommand) SetFlags(fs *flag.FlagSet) {
 }
 
 func (g gradleCommand) verifyFlags() error {
+	for _, key := range []string{"KYTHE_CORPUS", "KYTHE_ROOT_DIRECTORY", "KYTHE_OUTPUT_DIRECTORY", "JAVAC_EXTRACTOR_JAR", "REAL_JAVAC"} {
```
creachadair: Consider maybe defining constants for these someplace. It doesn't have to be in this PR, though.

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/beevik/etree"
+)
+
+// PreProcessPomXML takes a pom.xml file and either verifies that it already has
+// the bits necessary to specify a separate compiler on commandline, or adds
+// functionality by dropping in a maven-compiler-plugin to the build.
+//
+// Note this potentially modifies the input file, so make a copy beforehand if
+// you need to keep the original.
+func PreProcessPomXML(pomXMLFile string) error {
```
creachadair: Can this be unexported?

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
```diff
@@ -0,0 +1,88 @@
...
+	"os"
+
+	"github.com/beevik/etree"
+)
+
+// PreProcessPomXML takes a pom.xml file and either verifies that it already has
+// the bits necessary to specify a separate compiler on commandline, or adds
+// functionality by dropping in a maven-compiler-plugin to the build.
+//
+// Note this potentially modifies the input file, so make a copy beforehand if
```
creachadair: s/potentially modifies/overwrites/

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
```diff
@@ -0,0 +1,88 @@
...
+func hasCompilerPlugin(doc *etree.Document) bool {
+	for _, p := range doc.FindElements("//project/build/plugins/plugin/artifactId") {
+		if p.Text() == "maven-compiler-plugin" {
+			return true
+		}
+	}
+	return false
+}
+
+func appendCompilerPlugin(doc *etree.Document) error {
```
creachadair: Please add a comment describing roughly what this does, something like

```
// appendCompilerPlugin modifies a doc containing the root element of a pom.xml
// configuration, by injecting a custom Maven compiler <plugin> element into the 
// plugin list for the top-level project.
```

or whatever is actually true. Sometimes a schematic example can help, YMMV.

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
creachadair: Relatedly: Please document clearly that the file may be modified even if the function reports an error, so that the caller is responsible for cleanup either way.

creachadair: I worry this might be somewhat brittle to use. It might be worthwhile to have this function create a backup and return it to the caller, so they can choose to clean up and/or restore.

```
// AddPluginToPOM injects …
// A backup is made of the original file before any changes are made. The caller
// is responsible for releasing or restoring the backup.
func AddPluginToPOM(pomFile string) (*backup.Backup, error) {
    bkup, err := backup.New(pomFile)
    if err != nil {
        return nil, err
    }
    // do real work
    return bkup, nil
}
```

(To make this nice, we'd probably want to make the backup methods no-op on a nil receiver)

Again not necessarily something for this change, though.

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	if hasCompilerPlugin(doc) {
+		return nil
+	}
+	if err := appendCompilerPlugin(doc); err != nil {
+		return err
+	}
+	f, err := os.OpenFile(pomXMLFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return fmt.Errorf("opening file %s for append: %v", pomXMLFile, err)
```
creachadair: the errors from package `os` already include the path, this can be simpler:

    return fmt.Errorf("opening output: %v", err)

creachadair: Not actually using `O_APPEND` though?

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	for _, tcase := range testcases {
+		// Copy the file into a temp file.
+		tf, err := ioutil.TempFile("", tcase.inputFile)
+		if err != nil {
+			t.Fatalf("creating temp file: %v", err)
+		}
+		tfName := tf.Name()
+		defer os.Remove(tfName)
```
creachadair: The defers don't happen till the function returns. Which is fine, given there are only three, but you could also use a subtest for this if you like: https://blog.golang.org/subtests

```
t.Run(tcase.inputFile, func(t *testing.T) {
   …the code you have now…
})
```

creachadair: That also has the nice side benefit that a fatal in the test body doesn't prevent the rest of the cases in the loop from running (whereas, at present, that terminates the whole test).

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+}
+
+// getPath returns a resolved filepath name for a file living in testdata/ directory.
+func getPath(t *testing.T, f string) string {
+	return testutil.TestFilePath(t, filepath.Join(testDataDir, f))
+}
+
+func readBytes(t *testing.T, f string) []byte {
```
creachadair: The Go analogue of the "OrDie" convention is "must", e.g., `mustReadBytes`

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier_test.go
```diff
@@ -0,0 +1,89 @@
...
+	b, err := ioutil.ReadFile(f)
```
creachadair: `t.Helper()`

kythe/go/extractors/config/runextractor/gradlecmd/gradlecmd.go
danielmoy-google: done

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
danielmoy-google: ah yea I copy/pasted this from a gradle where I was appending.  here I'm not, because the xml structure isn't amenable to it.

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
danielmoy-google: Hmm I'd kinda leave it as-is for now and keep backup done one level up.  That way if we either need multiple modifications we can do it in one spot where we really care, and if we end up not needing back up at all it removes pretty cleanly too.

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
danielmoy-google: oh yea I guess I moved this into the same package so it can.  Originally I had it outside in its own place.

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
danielmoy-google: I wonder if doc also includes path.  Guess we can clean that up later if it's spammy

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier.go
danielmoy-google: Oo I wonder if pointing to the test files is a good sufficient example

danielmoy-google: Also including a comment to make clear the difference between maven & gradle, since it's a bit funky to do the preprocess for maven and then also still have to pass compiler flags

kythe/go/extractors/config/runextractor/README.md
creachadair: I was trying to ask whether those two environment variables should also be mentioned in this section. But if this is only the strictly-required ones, maybe not.

kythe/go/extractors/config/runextractor/README.md
```diff
@@ -30,13 +30,14 @@ Use:
...
 commandline flags, there is non-trivial setup done with environment variables.
-When calling `runextractor`, here are the relevant environment variables:
+When calling `runextractor`, here are the required environment variables:
 
-* **KYTHE_CORPUS**: The corpus set for extracted files.
-* **KYTHE_ROOT_DIRECTORY**: The root path for file input to be extracted.
-* **KYTHE_OUTPUT_DIRECTORY**: The path for storing output.
+* **KYTHE_ROOT_DIRECTORY**: The absolute path for file input to be extracted.
+* **KYTHE_OUTPUT_DIRECTORY**: The absolute path for storing output.
+* **KYTHE_CORPUS**: The corpus label for extracted files.  If not set it
```
creachadair: If it has a default, it's not really required, right?

kythe/go/extractors/config/runextractor/README.md
```diff
@@ -30,13 +30,14 @@ Use:
 
 Because `runextractor` is invoked in a manner not conducive to cleanly passing
 commandline flags, there is non-trivial setup done with environment variables.
-When calling `runextractor`, here are the relevant environment variables:
+When calling `runextractor`, here are the required environment variables:
 
-* **KYTHE_CORPUS**: The corpus set for extracted files.
-* **KYTHE_ROOT_DIRECTORY**: The root path for file input to be extracted.
-* **KYTHE_OUTPUT_DIRECTORY**: The path for storing output.
+* **KYTHE_ROOT_DIRECTORY**: The absolute path for file input to be extracted.
```
creachadair: Absolute paths remove one kind of uncertainty, but replace it with another: Instead of not knowing what the paths are supposed to be relative to, the absolute path makes it unclear which part of the path belongs to the build tool and which is happenstance (e.g., I run in a directory named `/tmp/build/1ca039beefc0d3000/test/kythe/go/platform/foo/testdata/kythe/go/platform/input.go`, how do I know where the root of the repo is so I can generate paths?)

Of course, that doesn't mean we _can't_ use absolute paths, but it might wind up making more sense to stipulate that the current working directory when the tool starts up affects the interpretation of these values. (We might not have enough data to really define that yet, I'm just thinking ahead).

kythe/go/extractors/config/runextractor/constants/constants.go
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package constants defines relevant constants common to multiple extractors.
+package constants
+
+import ()
```
creachadair: dd

creachadair: Does your editor add these? I've seen this a couple times now, and I'm not sure where they come from.

kythe/go/extractors/config/runextractor/constants/constants.go
```diff
@@ -0,0 +1,33 @@
...
+
+func requiredEnvVars() []string {
```
creachadair: This probably does not need to be a function. I think it could more simply be:

```
var (
	requiredEnv     = []string{"KYTHE_CORPUS", "KYTHE_ROOT_DIRECTORY", "KYTHE_OUTPUT_DIRECTORY"}
	requiredJavaEnv = append(requiredEnv, "JAVAC_EXTRACTOR_JAR", "REAL_JAVAC")
)
```

(exported or not doesn't matter, the same thing works either way—static initialization order is well-defined)

kythe/go/extractors/config/runextractor/constants/constants.go
danielmoy-google: Yea auto drop in by the editor.

kythe/go/extractors/config/runextractor/README.md
danielmoy-google: Yea this was an explanation of the underlying behavior of the java and other extractors, guess it isn't necessary to put here because we force caller to set it.

kythe/go/extractors/config/runextractor/README.md
danielmoy-google: Yea and there's another wrench in here which is docker-in-docker does funky things with paths set (the path might be an absolute path on a different image, or the root (or both!).

For now, absolute path, if only until we get rid of docker in docker.

But I will put a note here for the dind removal bit to reevaluate what to do here.

kythe/go/extractors/config/runextractor/constants/constants.go
danielmoy-google: done

kythe/go/extractors/config/runextractor/mavencmd/pom_xml_modifier_test.go
danielmoy-google: oh that's slick

PR #2918, opened 2018-08-10
bazel_go_extractor: Update for current bazel_go rules. 

### LINKED ISSUE #2912: Bazel Go extractor fails trying to load a directory path
Repro:
```
$ bazel build kythe/go/util/schema:schema --experimental_action_listener=//kythe/go/extractors/cmd/bazel:extract_kindex_go --experimental_extra_action_top_level_only
```

Expected: A compilation record.
Observed: 
```
ERROR:$W/kythe/kythe/go/util/schema/BUILD:6:1: Executing extra_action //kythe/go/extractors/cmd/bazel:extra_action on //kythe/go/util/schema:schema failed (Exit 1)
2018/08/09 10:27:29 Extracting XA for "//kythe/go/util/schema:schema" with 11 inputs
2018/08/09 10:27:29 Found 10 required inputs, 0 source files
2018/08/09 10:27:29 Extraction failed: reading input files failed: read bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/linux_amd64_stripped/stdlib~/pkg: is a directory
Target //kythe/go/util/schema:schema up-to-date:
  bazel-bin/kythe/go/util/schema/linux_amd64_stripped/schema~/kythe.io/kythe/go/util/schema.a
```

Briefly, the Go rule is advertising the directory containing the standard library, and the extractor is trying to read it. This code hasn't been properly updated since we switched to https://github.com/bazelbuild/rules_go, so it's almost certainly a bug in the extractor.

creachadair:
This program was not updated since we switched away from the legacy Go rules,
and was still expecting the build to generate a symlink forest and other such
nonsense that no longer applies.

Replace that with a much simpler implementation. The action's command line now
includes the file locations directly, so we do not need to guess and remap path
prefixes. Most of the old code was able to be simply deleted.

Fixes #2912.

### DIFF
--- a/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
+++ b/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
@@ -91,6 +91,7 @@ func main() {
 		CheckAction: ext.checkAction,
 		CheckInput:  ext.checkInput,
 		CheckEnv:    ext.checkEnv,
+		IsSource:    ext.isSource,
 		FixUnit:     ext.fixup,
 	}
 	ai, err := bazel.SpawnAction(info)
@@ -148,6 +149,8 @@ func (e *extractor) checkInput(path string) (string, bool) {
 	}
 	return path, false
 }
+
+func (*extractor) isSource(name string) bool { return filepath.Ext(name) == ".go" }
 
 func (*extractor) checkEnv(name, _ string) bool { return name != "PATH" }
 
--- a/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
+++ b/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
```diff
@@ -32,16 +31,12 @@ import (
...
-	"bitbucket.org/creachadair/stringset"
-
 	apb "kythe.io/kythe/proto/analysis_go_proto"
 	gopb "kythe.io/kythe/proto/go_go_proto"
-	spb "kythe.io/kythe/proto/storage_go_proto"
 )
 
 var (
-	corpus = flag.String("corpus", "kythe", "The corpus label to assign (required)")
+	corpus = flag.String("corpus", "kythe.io", "The corpus label to assign (required)")
```
schroederc: The other extractors use "kythe" as a default.  Can we keep them consistent?

kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
```diff
@@ -109,352 +106,145 @@ func main() {
...
 	for name, value := range info.Environment {
 		switch name {
 		case "GOOS":
 			e.goos = value
 		case "GOARCH":
 			e.goarch = value
+		case "GOROOT":
+			e.goroot = value
+		case "CGO_ENABLED":
+			e.cgoEnabled = value == "1"
```
schroederc: Does it have to be "1" or can it be any non-empty value?

kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
```diff
@@ -109,352 +106,145 @@ func main() {
...
 }
 
-func (*extractor) checkEnv(name, _ string) bool {
-	switch name {
-	case "PATH", "GOOS", "GOARCH":
-		return false
+func (e *extractor) checkInput(path string) (string, bool) {
+	switch filepath.Ext(path) {
+	case ".go", ".a":
+		return path, true // keep source files, archives
```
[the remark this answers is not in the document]
creachadair: Right now the indexer can't do anything with them, so no—but if we get to the point where we can handle cgo in the future, we might want those files.

kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
creachadair: I'm not sure. `"1"` is the value the rules use now—and I assume that `"0"` and `""` could probably be taken as false. I didn't actually look at the implementation, though.

kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
creachadair: ✔︎

Issue #2874, opened 2018-07-26
C++ compilation of rule '//kythe/cxx/common:k\ zip_writer_test' failed


kamahen:
```
kythe/cxx/common/kzip_writer_test.cc:89:5: error: return type 'kythe::Status' must match previous return\
 type 'const kythe::Status' when lambda expression has unspecified explicit return type
    return OkStatus();
    ^
kythe/cxx/common/kzip_writer_test.cc:112:5: error: return type 'kythe::Status' must match previous retur\
n type 'const kythe::Status' when lambda expression has unspecified explicit return type
    return OkStatus();
    ^
```

Fixed by:   
```
--- a/kythe/cxx/common/status.h
+++ b/kythe/cxx/common/status.h
@@ -87,7 +87,7 @@ class ABSL_MUST_USE_RESULT Status final {
 };

 /// \brief Returns a non-error status.
-inline Status OkStatus() { return {}; }
+inline const Status OkStatus() { return {}; }

 /// \brief Return a Cancelled error with the given message.
 Status CancelledError(absl::string_view message);
```

creachadair:
Thanks, this is #2873. Can you tell me a bit more about your setup? I haven't been able to repro it locally yet.

creachadair:
In particular, can you tell us what `clang --version` you are using?

zrlk:
I can repro with 3.5.2. This may be a bug in clang that's since been fixed.

creachadair:
#2877 should hopefully give you a working build again.

shahms:
This is a compiler bug which we should fix by requiring a compiler without the bug.

shahms:
Yes, at least 3.6. 3.5 is ancient. 3.6 is also pretty old, but is still a better choice.  3.8 and 4.0 should both be broadly available for distros as well.

creachadair:
> Should I install clang-3.6?

I don't know if this particular issue is fixed there. I'm using 4.0.1, and would probably advise using at least 4. To your point, though, we should probably update the documentation to stipulate a higher version.

shahms:
3.8 will work, but Ubuntu has a variety of clang's available including 4.0 (but it's the `clang-4.0` package, rather than just `clang`).  You'll likely also have to set the `CLANG` environment variable to the path to the clang binary so the Bazel repository rules will pick the right version.

kamahen:
3.8 didn't work (on regular Ubuntu 16.04). I'll install clang-4.0 later
[...]

PR #2919, opened 2018-08-10
Increase minimum required version of clang to 3.6.

creachadair:
This is intended to address #2874.


creachadair:
I think we concluded that 3.6 should be sufficient. I know 3.8 works.

PR #2921, opened 2018-08-10
Add libcurl external repository

schroederc:
third_party/curl.BUILD was borrowed from https://github.com/tensorflow/tensorflow/blob/bb890ea5fa4c8a2100a295e93c5bf37b3c88b55a/third_party/curl.BUILD


WORKSPACE
```diff
@@ -38,6 +38,17 @@ bind(
     actual = "@com_google_code_gson_gson//jar",
 )
 
+new_http_archive(
+    name = "curl",
```
shahms: `se_haxx_curl` ?

third_party/curl.BUILD
```diff
@@ -0,0 +1,679 @@
...
+        "#endif",
+        "",
+        "#endif  // EXTERNAL_CURL_INCLUDE_CURL_CONFIG_H_",
+        "EOF",
+    ]),
+)
+
+config_setting(
+    name = "ios",
+    values = {"crosstool_top": "//tools/osx/crosstool:crosstool"},
```
shahms: This isn't a valid target

third_party/curl.BUILD
```diff
@@ -0,0 +1,679 @@
...
+)
+
+config_setting(
+    name = "windows",
+    values = {"cpu": "x64_windows_msvc"},
+)
+
+config_setting(
+    name = "android",
+    values = {"crosstool_top": "//external:android/crosstool"},
```
shahms: I don't think this is either

WORKSPACE
schroederc: ✓

third_party/curl.BUILD
schroederc: ✓

third_party/curl.BUILD
schroederc: ✓

PR #2923, opened 2018-08-13
Update LLVM and change uses of deprecated Loc functions

shahms:


tools/modules/versions.sh
```diff
@@ -20,12 +20,12 @@
 # be set to the new (matched) strings.
 
 # llvm
-MIN_LLVM_SHA="d8c8d742377d60b3cdacd4b7131115f5ac1e2bc2"
-MIN_LLVM_REV="337145"
+MIN_LLVM_SHA="36f54002c931a026f490f9fb074c11d91e3487a2"
+MIN_LLVM_REV="339407"
 # clang
 # NOTE: when updating Clang, make sure to adjust the path in
```
salguarnieri: Does this note apply?

Issue #2930, opened 2018-08-14
Kythe Columnar Serving Format


schroederc:
Tracking issue for implementing the columnar serving format proposed in https://github.com/google/kythe/pull/2909.

PR #2931, opened 2018-08-14
Add columnar FileDecorations protobuf message definitions

schroederc:
These are foundational definitions for #2930.

A large part of this change is the move of the Node message to the schema package so that it can be used for file decoration target nodes.


kythe/go/serving/pipeline/nodes/nodes.go
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-// Package nodes provides Beam transformations over *ppb.Nodes.
+// Package nodes provides Beam transformations over *scpb.Nodes.
```
creachadair: To the reader in http://godoc.org/, the `scpb` won't denote anything; I suggest "…over `kythe.proto.schema.Node` messages." or words to that effect.

kythe/proto/schema.proto
```diff
@@ -165,3 +167,56 @@ enum EdgeKind {
...
+  }
+}
+
+// A Fact represents a single Kythe graph node fact.
+message Fact {
+  kythe.proto.VName source = 1;
+
+  oneof name {
+    kythe.proto.schema.FactName kythe_name = 2;
+    string generic_name = 3;
```
creachadair: Please document these.

kythe/proto/schema.proto
```diff
@@ -165,3 +167,56 @@ enum EdgeKind {
...
+  bytes value = 4;
+}
+
+// An Edge represents a single Kythe graph edge.
+message Edge {
+  kythe.proto.VName source = 1;
+  kythe.proto.VName target = 2;
+
+  oneof kind {
+    kythe.proto.schema.EdgeKind kythe_kind = 3;
```
creachadair: These too, please.

kythe/go/serving/pipeline/nodes/nodes.go
schroederc: The package name does show up in the documentation: https://godoc.org/kythe.io/kythe/go/serving/pipeline/nodes#Kind.  I wanted to be consistent with the function types.

kythe/proto/schema.proto
schroederc: ✓

kythe/proto/schema.proto
schroederc: ✓

kythe/go/serving/pipeline/nodes/nodes.go
creachadair: That makes sense. SGTM.

PR #2932, opened 2018-08-14
Add orderedcode utility library for Kythe types

schroederc:
These encodings will be used as part of #2930.


kythe/go/util/keys/keys.go
```diff
@@ -0,0 +1,80 @@
...
+//  - int32/int
+//
+// More detail at: https://godoc.org/github.com/google/orderedcode#Append
+func Append(key []byte, items ...interface{}) ([]byte, error) {
+	expanded := make([]interface{}, 0, len(items))
+	for _, item := range items {
+		switch x := item.(type) {
+		case *spb.VName:
+			expanded = append(expanded, x.Corpus, x.Language, x.Path, x.Root, x.Signature)
+		case int32:
```
danielmoy-google: worth a comment about why forcing to int64?  I dunno.  Maybe I'm missing context.

schroederc: ✓

PR #2950, opened 2018-08-20
Unbreak verifier on macOsX

### LINKED ISSUE #2943: Verifier build broken on mac
#2853 may break the verifier build on the mac because it will pick up the bison in /usr/bin instead of the user's preferred path to bison. On my mac (and probably others) the bison version in /usr/bin is pretty old and will not work for our build. It is not uncommon for people to install updated versions of tools in somewhere other than /usr/bin and editing $PATH so bash finds the updated version before the one in /usr/bin.

zrlk:
Support ancient Bison again. Fixes #2943.

Please check that this doesn't also upset Linux (I have limited access to that environment just this moment)

### DIFF
--- a/kythe/cxx/verifier/assertions.yy
+++ b/kythe/cxx/verifier/assertions.yy
@@ -1,8 +1,9 @@
 // Started from the calc++ example code as part of the Bison-3.0 distribution.
+// NOTE: This file must remain compatible with Bison 2.3.
 %skeleton "lalr1.cc"
 %defines
-%define parser_class_name {AssertionParserImpl}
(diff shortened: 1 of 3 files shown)


kythe/cxx/verifier/assertions.yy
```diff
@@ -1,8 +1,9 @@
 // Started from the calc++ example code as part of the Bison-3.0 distribution.
+// NOTE: This file must remain compatible with Bison 2.3.
```
salguarnieri: Is it worth adding an issue to add bison as an external repo?

PR #2952, opened 2018-08-22
Remove deprecated use of built-in http/git rules

edbaunton:
The built-in http and git rules are being deprecated in favour of the
starlark versions so load them from their respective starlark modules.

Tested using `--incompatible_remove_native_git_repository` and
`--incompatible_remove_native_http_archive` flags.

See https://groups.google.com/forum/#!topic/bazel-discuss/dO2MHQLwJF0


salguarnieri:
I think buildbot might only run automatically for core kythe team PRs

edbaunton:
Thanks!

PR #2954, opened 2018-08-24
Use our shasum tool instead of md5 for releases

salguarnieri:
md5 is not in the normal mac PATH, so it is easier if we just use a signature
generated from our own shasum tool.

This is necessary now since we are setting
experimental_strict_action_env in our bazel.rc.


salguarnieri:
Sorry, missed that you already LGTM'd. I just added a comment to the build file.

PR #2958, opened 2018-08-25
Use wc -c instead of du -b

salguarnieri:
du -b doesn't exist on macOS. We can use wc -c instead to get the character count.

### DIFF
--- a/kythe/go/platform/tools/kindex_file_generator/BUILD
+++ b/kythe/go/platform/tools/kindex_file_generator/BUILD
@@ -26,7 +26,7 @@ shell_tool_test(
       --output "$$kindex_output" \\
       --source "$$input"
 
-    size="$$(du -b "$$kindex_output" | cut -f1)"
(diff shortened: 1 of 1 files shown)


kythe/go/platform/tools/kindex_file_generator/BUILD
```diff
@@ -26,7 +26,7 @@ shell_tool_test(
       --output "$$kindex_output" \\
       --source "$$input"
 
-    size="$$(du -b "$$kindex_output" | cut -f1)"
+    size="$$(wc -c < "$$kindex_output")"
     if [[ "$$size" -eq 0 ]] ; then
```
creachadair: I think this could also be more simply

    if [[ -s "$$kindex_output" ]] ; then

creachadair: Er, sorry, that's `! -s "$$kindex_output"`

Issue #148, opened 2018-06-11
Implement kzip reader and writer support in Java


creachadair:
As part of migrating away from .kindex and index pack formats to .kzip, we will need a way for the Java extractor to generate .kzip output, and for the Java indexer to read it (to keep the verifier tests working).

PR #2959, opened 2018-08-25
Add Kzip to java extractors

salguarnieri:
Add the option to output a kzip file instead of a kindex file to the
various java extractors. In most cases this involves adding another
command line argument that specifies which format should be output.

Part of #148


kythe/java/com/google/devtools/kythe/extractors/java/bazel/JavaExtractor.java
```diff
@@ -49,17 +50,30 @@
...
+  private static final String KINDEX_OUTPUT_FORMAT = "kindex";
 
   public static void main(String[] args) throws IOException, ExtractionException {
-    if (args.length != 3) {
-      System.err.println("Usage: java_extractor extra-action-file output-file vname-config");
+    JsonUtil.usingTypeRegistry(JavaCompilationUnitExtractor.JSON_TYPE_REGISTRY);
+
+    if (args.length != 4) {
+      System.err.println(
+          "Usage: java_extractor extra-action-file output-file vname-config output-format");
```
schroederc: Why not infer `output-format` from `output-file`'s extension?

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
```diff
@@ -127,6 +116,28 @@ public void process(String[] args) {
     }
   }
 
+  private static void outputIndexInfo(CompilationDescription indexInfo) throws IOException {
+    String outputFile = System.getenv("KYTHE_OUTPUT_FILE");
```
schroederc: Same as above.  We wouldn't need separate env vars if the extractor inferred the output format from the file extension.  I think makes more sense since the `.kindex` file is being deprecated anyway.

kythe/java/com/google/devtools/kythe/extractors/shared/IndexInfoUtils.java
```diff
@@ -66,8 +80,38 @@ public static CompilationDescription readIndexInfoFromStream(InputStream inputSt
     return new CompilationDescription(compilationUnit, fileContents);
   }
 
-  public static void writeIndexInfoToStream(CompilationDescription description, OutputStream stream)
-      throws IOException {
+  public static Collection<CompilationDescription> readKZip(String path)
```
schroederc: This is unused.  Meant to be in a later commit?

kythe/java/com/google/devtools/kythe/extractors/java/bazel/JavaExtractor.java
salguarnieri: same as below

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
salguarnieri: Works for me. I was worried that somewhere we might have some kindex files without an extension.

kythe/java/com/google/devtools/kythe/extractors/shared/IndexInfoUtils.java
salguarnieri: yeah, its for the next change.

kythe/java/com/google/devtools/kythe/extractors/java/bazel/JavaExtractor.java
```diff
@@ -162,7 +152,7 @@ public static void main(String[] args) throws IOException, ExtractionException {
                 javacOpts,
                 jInfo.getOutputjar());
 
-    if (outputFormat.equals(KZIP_OUTPUT_FORMAT)) {
+    if (outputPath.endsWith(IndexInfoUtils.KZIP_FILE_EXT)) {
```
schroederc: Can we flip this (i.e. if not .kzip, use .kindex) to ensure the legacy behavior of always using `.kindex` mostly remains until we delete support for it?

salguarnieri: Isn't that what this says, only with the positive case coming first?

schroederc: Carry on

Issue #2955, opened 2018-08-24
asciidoc not in PATH on macOS


salguarnieri:
For example, you cannot build `//kythe/docs:modeling-libraries` because bazel cannot find asciidoc since it only looks in /bin and /usr/bin and brew installs asciidoc in /usr/local/bin/asciidoc.

This prevents you from running `bazel build ...`

bzz:
Same happens with `bazel test //kythe/extractors/cmake:compilation_database` and GNU parallel installed from brew in `/usr/local/bin/parallel`.

shahms:
We should probably define and use Bazel toolchains for the system tools we require...

PR #2961, opened 2018-08-26
Add brew bin dir to path in asciidoc rule

bzz:
Very crude attempt of bypass #2955 - adds standard brew bin path to PATH of the asciidoc rule.


shahms:
This shouldn't be necessary if `/usr/local/bin` is already in the PATH (which it should be if that's where the build tools reside).  We shouldn't be hacking PATH manually here.

salguarnieri:
#2970 should have fixed the issue so this isn't necessary.

bzz:
Makes perfect sense, thank you for prompt fix and the feedback!

shahms:
Thanks for the bug report!

PR #2962, opened 2018-08-26
Remove ProtoTypeAdapter for Java JSON code

salguarnieri:
I think this is unused.

If it is used in the future, I think we would need to change it because
currently it uses java naming style (camel case) for field names. I think we
would need to set the naming scheme with something like:
ProtoTypeAdapter.newBuidler().setFieldNameSerializationFormat(CaseFormat.LOWER_UNDERSCORE, CaseFormat.LOWER_UNDERSCORE).build().


salguarnieri:
PTAL, fixed build

PR #2964, opened 2018-08-27
Add Go 1.11 to Buildbot

schroederc:


kythe/release/appengine/buildbot/master/master.cfg.template
```diff
@@ -118,6 +115,15 @@ goKytheSteps.addStep(steps.ShellCommand(command=["go", "get", "-t", "kythe.io/ky
...
                                         env={"GOPATH": util.Property('builddir')}))
 
+goModuleKytheSteps = util.BuildFactory()
+goModuleKytheSteps.addStep(steps.GitHub(repourl=util.Property('repository', 'git://github.com/google/kythe.git'),
+                                        mode='incremental'))
+goModuleKytheSteps.addStep(steps.ShellCommand(command=["go", "test", "./kythe/..."],
+                                              haltOnFailure=True))
+# Check that the Kythe module files are up-to-date
+goModuleKytheSteps.addStep(steps.ShellCommand(command="! { git diff --color=always -- go.{mod,sum} | grep .; }",
+                                              warnOnFailure=True))
```
salguarnieri: Why don't we want to halt if the files are not up to date?

schroederc: I'm letting this be a soft requirement for a while since there's not a great solution for keeping the file synchronized with our `WORKSPACE` file.  It's not the end of the world if we don't specify a dep.  Go will just pick up the latest, as is the current `GOPATH` default.

Issue #2801, opened 2018-06-30
RFC: A uniform analyzer-driver interface


creachadair:
# RFC: A uniform analyzer-driver interface

## Background

Kythe defines a loosely-coupled interface between build tools and analyzers. There are broadly two components of interest: _Extractors_ that consume source repositories and emit compilation records, and _analyzers_ that consume compilation records and emit analysis results (such as Kythe graph entries).

A **compilation record** in Kythe consists of a [kythe.proto.CompilationUnit][kcu], together with the contents of all the files referenced as required inputs to that unit. Presently there are three file-based storage formats for compilation records ([indexpack][], [kindex][], and [kzip][]), and analyzer tools must explicitly support any of these they want to consume.

## Problem Statement

There are two problems we'd like to address in this space:

1. There are too many file formats, and it is a lot of work to add support for them in a new language. 

2. There is no way to implement an analyzer without supporting at least one of these file formats, and thus each analyzer must—unnecessarily, we argue—subsume some important implementation details of the analysis infrastructure.

A solution to Problem (1) is already underway, and is the subject of a separate milestone, [Consolidate compilation record storage formats](https://github.com/google/kythe/milestone/9). The rest of this issue is concerned with Problem (2), which is not yet resolved.

We would like analyzers to have a simple and uniform interface to the rest of Kythe. Moreover, we would like that interface to depend as little as possible on concerns that are not relevant to the analyzer, such as how the compilation records are stored, how files are stored, cached, access-controlled, and fetched, and how the results of analysis are captured, stored, or propagated.

## Proposal

We propose that the interface between an analyzer and the rest of the Kythe infrastructure should be implemented as a text-based client-server protocol between the _analyzer_ (which is the client) and a _driver_ (which is the server). In summary, the interaction between analyzer and driver will work as follows:

1. The analyzer requests work from the driver, and the driver replies with a `CompilationUnit` message to be processed.

2. The analyzer requests any files it requires from among those listed in the `CompilationUnit`, and the driver replies with the contents of those files.

3. The analyzer sends analysis outputs to the driver, and the driver disposes of them as appropriate.

The choice to use a text-based protocol is intentional. While all the Kythe data formats can be described by [protocol buffers][protobuf], we have found in practice that the quality of protobuf support varies widely among languages, and that in many cases the overhead of depending on the protobuf compiler and adding a code-generation step is a deterrent to contribution.

[...]

PR #2969, opened 2018-08-29
Add driver.proto defining messages for the analyzer protocol.

creachadair:
This file defines protobuf versions of the message types described in #2966,
for use in the JSON-RPC exchange between analyzers and their drivers.  This is
a step toward a working implementation of #2801.


kythe/proto/driver.proto
```diff
@@ -0,0 +1,92 @@
...
+
+  // The file path and/or digest being requested. At least one of these fields
+  // must be non-empty.
+  string path = 2;
+  string digest = 3;
+}
+
+// A reply from the driver delivering the contents of a file.
+message FileReply {
+  string path = 2;
```
schroederc: Start at field number 1.

creachadair: ✔︎

PR #2970, opened 2018-08-29
Add toolchain for tools needed for asciidoc rules

salguarnieri:
Our doc targets use asciidoc which isn't in the bazel approved binary location on macOS. Create a toolchain that gathers all the tools needed for doc generation.


kythe/docs/asciidoc.bzl
```diff
@@ -1,26 +1,75 @@
...
-        'export OUTDIR="$$PWD/$(@D)"',
-        'export LOGFILE="$$(mktemp -t \"XXXXXXasciidoc\")"',
-        'export PATH',
-        'trap "rm -f \"$${LOGFILE}\"" EXIT ERR INT',
-        "asciidoc %s -o $(@) $(location %s) 2> \"$${LOGFILE}\"" % (' '.join(args), src),
-        'cat $${LOGFILE}',
-        '! grep -q -e "filter non-zero exit code" -e "no output from filter" "$${LOGFILE}"']))
+def _asciidoc_impl(ctx):
+    asciidoc = ctx.toolchains["//tools/build_rules/external_tools:external_tools_toolchain_type"].asciidoc
+    args = ["--backend", "html", "--no-header-footer"]
```
shahms: Use `ctx.actions.args()` rather than a list? https://docs.bazel.build/versions/master/skylark/lib/Args.html

kythe/docs/asciidoc.bzl
```diff
@@ -1,26 +1,75 @@
...
+        outputs = [ctx.outputs.out, ctx.outputs.logfile],
+        env = {
+            "PATH": tool_path,
+        },
+        command = "\n".join([
+            '"%s" %s 2> "%s"' % (
+                asciidoc,
+                " ".join(args),
+                ctx.outputs.logfile.path,
+            ),
```
shahms: Consider quoting these with `@bazel_skylib//:lib.zl` `shell.quote`

kythe/docs/asciidoc.bzl
```diff
@@ -1,26 +1,75 @@
...
+        ),
+        "example_script": attr.label(
+            doc = "Script to pass to asciidoc as --attribute=example_script=VALUE.",
+            allow_single_file = True,
+        ),
+    },
+    doc = "Generate asciidoc",
+    outputs = {
+        "out": "%{name}.html",
+        "logfile": "%{name}.logfile",
```
shahms: Why declare this as an explicit output?  This shouldn't be required (although, admittedly, output file handling in Bazel rules is complicated).  In particular https://docs.bazel.build/versions/master/skylark/rules.html#files (use `ctx.actions.declare_file` )

tools/build_rules/external_tools/external_tools_configure.bzl
```diff
@@ -0,0 +1,73 @@
...
+
+    repository_ctx.file("BUILD", _BUILD_TEMPLATE.format(
+        asciidoc = asciidoc,
+        path = repository_ctx.path(""),
+    ))
+
+def symlink_command(repository_ctx, command):
+    binary = repository_ctx.which(command)
+    if binary == None:
+        fail("Unable to find '%s' executavle on path." % command)
```
shahms: executable

tools/build_rules/external_tools/external_tools_toolchain.bzl
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+external_tools_toolchain = rule(
+    implementation = _external_tools_toolchain_impl,
+    attrs = {
+        "asciidoc": attr.string(),
+        "dot": attr.string(),
+        "python": attr.string(),
+        "cat": attr.string(),
+        "path": attr.string(),
```
shahms: If you're providing `path` the others shouldn't be necessary, no?

tools/build_rules/external_tools/external_tools_configure.bzl
```diff
@@ -0,0 +1,73 @@
...
+        "tee",
+        "rm",
+        "cut",
+        "sha1sum",
+    ]
+    for tool in tools:
+        symlink_command(repository_ctx, tool)
+
+    repository_ctx.file("BUILD", _BUILD_TEMPLATE.format(
+        asciidoc = asciidoc,
```
shahms: should probably provide the other tools explicitly listed in rule attrs?

kythe/docs/asciidoc.bzl
salguarnieri: I think Args can't easily be turned back into a string list and passing it in as the arguments attribute means I'd have to rewrite the command to use $1, $2, etc. syntax to access the arguments. I didn't like how that looked so I kept it like this.

I could have just been holding it wrong though.

kythe/docs/asciidoc.bzl
salguarnieri: done

tools/build_rules/external_tools/external_tools_configure.bzl
salguarnieri: fixed by removing them from the attrs.

tools/build_rules/external_tools/external_tools_configure.bzl
salguarnieri: done

tools/build_rules/external_tools/external_tools_toolchain.bzl
salguarnieri: Yeah. I left asciidoc because we call it explicitly, but removed the others.

kythe/docs/asciidoc.bzl
salguarnieri: Done

PR #2971, opened 2018-08-29
Don't use parallel

salguarnieri:
The easiest way to get this test to work on macOS is to not use parallel since parallel isn't in the PATH that bazel uses.


shahms:
That script has largely (but not entirely) been subsumed by https://github.com/google/kythe/tree/master/kythe/go/extractors/config/runextractor/cmakecmd anyway.

PR #2972, opened 2018-08-30
Add ResolvedDiagnostic/CorpusPath proto definitions

schroederc:


kythe/proto/common.proto
```diff
@@ -89,6 +89,19 @@ message Diagnostic {
   string context_url = 4;
 }
 
+// A Diagnostic with its associated file.
+message ResolvedDiagnostic {
```
creachadair: I'm not sure what "resolved" here means. Can you talk a bit more about what resolution means in the comment?

schroederc: 👍

PR #2977, opened 2018-09-02
Record @deprecated javadoc tags in the Java indexer

cushon:
Record the contents of `@deprecated` javadoc tags, and attach them to the
javadoc documentation node under `/kythe/java/deprecated`.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
@@ -18,10 +18,14 @@ package com.google.devtools.kythe.analyzers.java;
 
 import static com.google.devtools.kythe.analyzers.java.KytheTreeScanner.DocKind.JAVADOC;
 
+import com.google.common.base.Splitter;
 import com.google.common.flogger.FluentLogger;
 import com.google.devtools.kythe.analyzers.base.EntrySet;
 import com.google.devtools.kythe.proto.Storage.VName;
+import com.sun.source.doctree.DeprecatedTree;
 import com.sun.source.doctree.ReferenceTree;
+import com.sun.source.tree.CompilationUnitTree;
+import com.sun.source.util.DocSourcePositions;
 import com.sun.source.util.DocTreePath;
 import com.sun.source.util.DocTreePathScanner;
 import com.sun.source.util.DocTrees;
@@ -31,14 +35,19 @@ import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.tree.DCTree.DCDocComment;
 import com.sun.tools.javac.tree.DCTree.DCReference;
 import com.sun.tools.javac.util.Context;
+import com.sun.tools.javac.util.Position;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 public class KytheDocTreeScanner extends DocTreePathScanner<Void, DCDocComment> {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
   private final KytheTreeScanner treeScanner;
   private final List<MiniAnchor<Symbol>> miniAnchors;
+  private Optional<String> deprecated;
   private final DocTrees trees;
 
   public KytheDocTreeScanner(KytheTreeScanner treeScanner, Context context) {
@@ -55,7 +64,9 @@ public class KytheDocTreeScanner extends DocTreePathScanner<Void, DCDocComment> 
     }
 
     miniAnchors.clear();
+    deprecated = Optional.empty();
     scan(new DocTreePath(treePath, doc), doc);
+
 
     String bracketed =
         MiniAnchor.bracket(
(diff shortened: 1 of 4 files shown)


creachadair:
We've talked about a couple of different ways of representing modifiers like this. Deprecation probably should not be Java-specific, and probably shouldn't require documentation support to work. In any case, we ought to update the schema before (or at least at the same time as) we modify any indexers.

creachadair:
I added #2978 for this topic.

cushon:
I switched to the `tag/deprecated` proposal from #2978.

> shouldn't require documentation support to work

I'm not sure what this means. You mention in the bug that `@Deprecated` annotatoins are indexed separately and don't need treatment here. What does that leave besides the `@deprecated` javadoc tag?

creachadair:
> You mention in the bug that @Deprecated annotatoins are indexed separately and don't need treatment here. What does that leave besides the @deprecated javadoc tag?

Your comment above says "Record the contents of @deprecated javadoc tags, and attach them to the
javadoc documentation node under /kythe/java/deprecated." This assumes there will always be a documentation node, which is not the case. My point is that the deprecated tag should go on the semantic node—regardless whether that's because of a `@Deprecated` source annotation, or a `@deprecated` JavaDoc annotation.

jrtom:
When can there be a Javadoc tag but no documentation node?  That is, are there cases where we might be parsing Javadoc comments but not creating a documentation node?

(I agree that we should probably be attaching the Kythe "deprecated" tag to the semantic node anyway, but I'm puzzled at the premise.)

cushon:
Thanks for the explanation. (I was thinking about the implementation not the schema--we need support for processing javadoc comments to make this work for Java, but attaching to the semantic node sounds good.)

creachadair:
> When can there be a Javadoc tag but no documentation node? 

There can't—my point wasn't about the presence or absence of the JavaDoc annotation, but to ensure that the resulting tag goes to the semantic node, regardless how it got there.

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Deprecation.java
```diff
@@ -0,0 +1,25 @@
+package pkg;
+
+// - @+9Deprecation defines/binding DeprecationClass
```
zrlk: The verifier looks for exactly `//-`, so none of these assertions are being checked.

kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
```diff
@@ -98,4 +114,36 @@ public Void visitReference(ReferenceTree tree, DCDocComment doc) {
 
     return null;
   }
+
+  @Override
+  public Void visitDeprecated(DeprecatedTree node, DCDocComment doc) {
+    if (node.getBody().isEmpty()) {
+      // deprecated tag is empty
```
jaysachs: Wouldn't the method/class still want to have the deprecated fact emitted even if the tag was empty?

kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
```diff
@@ -98,4 +114,36 @@ public Void visitReference(ReferenceTree tree, DCDocComment doc) {
...
+    int end = (int) positions.getEndPosition(unit, doc, node);
+    if (end == Position.NOPOS) {
+      // deprecated tag is empty
+      return null;
+    }
+    CharSequence source;
+    try {
+      source = unit.getSourceFile().getCharContent(/* ignoreEncodingErrors= */ true);
+    } catch (IOException e) {
+      return null;
```
jaysachs: Related question here.

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Deprecation.java
cushon: Thanks, I think this got clobbered by the formatter. (The style guide technically requires a space after `//` in line comments, so it inserts them.)

kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
cushon: The `@Deprecated` annotation is the source of truth for whether the API is deprecated, the javadoc is mostly useful for documentation, and the documentation isn't useful if it's empty. So I'm not sure tools would actually want to surface an empty `@deprecated` tag, but there's no reason not to record it just in case and let it get filtered out later.

I updated the implementation to save empty strings here and added a test.

kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
cushon: I'm not saving empty strings for the other cases where the tag is empty. The case where `unit.getSourceFile().getCharContent` throws seems distinct.

It should be impossible, so I could also rethrow as a `UncheckedIOException` if you prefer.

PR #2982, opened 2018-09-04
 Define a deprecated tag and use it for C++

zrlk:
In partial service to 2978


kythe/docs/schema/schema.txt
```diff
@@ -1450,6 +1453,19 @@ Brief description::
...
 
+
+[[tagdeprecated]]
+tag/deprecated
+~~~~~~~~~~~~~~
+
+Brief description::
+  If this node should no longer be used, it should be marked with
+  `tag/deprecated`. If this fact has a nonempty value, that value should be
+  set to a human-readable reason why the node was deprecated, and/or what should
```
creachadair: Perhaps: "…to a UTF-8 encoded human-readable string describing why…"

(just to make it clear that this is text)

PR #2987, opened 2018-09-06
Implement the encoding for columnar FileDecorations

schroederc:
First part of serving support for #2930.

### DIFF
--- a/kythe/go/serving/xrefs/columnar/columnar_encoding.go
+++ b/kythe/go/serving/xrefs/columnar/columnar_encoding.go
@@ -101,6 +101,7 @@ func encodeDecorText(prefix []byte, file *spb.VName, t *xspb.FileDecorations_Tex
 	if err != nil {
 		return nil, err
 	}
+	// Encode the subset of the value proto not encoded in the key.
 	val, err := proto.Marshal(&xspb.FileDecorations_Text{Text: t.Text})
 	if err != nil {
 		return nil, err
@@ -115,7 +116,6 @@ func encodeDecorTarget(prefix []byte, file *spb.VName, t *xspb.FileDecorations_T
 	if err != nil {
 		return nil, err
 	}
-
 	val, err := proto.Marshal(&xspb.FileDecorations_Target{})
 	if err != nil {
 		return nil, err
@@ -151,6 +151,7 @@ func encodeDecorTargetDefinition(prefix []byte, file *spb.VName, td *xspb.FileDe
 	if err != nil {
 		return nil, err
 	}
+	// Encode the subset of the value proto not encoded in the key.
 	val, err := proto.Marshal(&xspb.FileDecorations_TargetDefinition{Definition: td.Definition})
 	if err != nil {
 		return nil, err
@@ -178,6 +179,7 @@ func encodeDecorOverride(prefix []byte, file *spb.VName, o *xspb.FileDecorations
 	if err != nil {
 		return nil, err
 	}
+	// Encode the subset of the value proto not encoded in the key.
 	val, err := proto.Marshal(&xspb.FileDecorations_Override{MarkedSource: o.MarkedSource})
 	if err != nil {
 		return nil, err
--- a/kythe/go/serving/xrefs/columnar/columnar_encoding.go
+++ b/kythe/go/serving/xrefs/columnar/columnar_encoding.go
@@ -1,0 +1,366 @@
+/*
+ * Copyright 2018 The Kythe Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


kythe/go/serving/xrefs/columnar/columnar_encoding.go
```diff
@@ -0,0 +1,366 @@
...
+	xspb "kythe.io/kythe/proto/xref_serving_go_proto"
+)
+
+// ColumnarDecorationsKeyPrefix is the common key prefix for all Kythe columnar
+// FileDecoration key-value entries.
+var ColumnarDecorationsKeyPrefix, _ = keys.Append(nil, "fd")
+
+func init() {
+	// Restrict the capacity of the key prefix to ensure appending to it creates a new array.
+	ColumnarDecorationsKeyPrefix = ColumnarDecorationsKeyPrefix[:len(ColumnarDecorationsKeyPrefix):len(ColumnarDecorationsKeyPrefix)]
```
salguarnieri: what does slice[:N:N] do?

kythe/go/serving/xrefs/columnar/columnar_encoding.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	return &KV{key, val}, nil
+}
+
+func encodeDecorText(prefix []byte, file *spb.VName, t *xspb.FileDecorations_Text) (*KV, error) {
+	key, err := keys.Append(prefix, file, columnarDecorationsTextGroup, t.StartOffset, t.EndOffset)
+	if err != nil {
+		return nil, err
+	}
+	val, err := proto.Marshal(&xspb.FileDecorations_Text{Text: t.Text})
```
salguarnieri: Can you add a comment why you are creating a new proto instead of using t?

kythe/go/serving/xrefs/columnar/columnar_encoding.go
```diff
@@ -0,0 +1,366 @@
...
+
+func encodeDecorTarget(prefix []byte, file *spb.VName, t *xspb.FileDecorations_Target) (*KV, error) {
+	key, err := keys.Append(prefix, file,
+		columnarDecorationsTargetGroup, t.StartOffset, t.EndOffset,
+		t.GetGenericKind(), int32(t.GetKytheKind()), t.Target)
+	if err != nil {
+		return nil, err
+	}
+
+	val, err := proto.Marshal(&xspb.FileDecorations_Target{})
```
salguarnieri: Why don't we need to marshal any content?

kythe/go/serving/xrefs/columnar/columnar_encoding.go
```diff
@@ -0,0 +1,366 @@
...
+		return nil, err
+	}
+	val, err := proto.Marshal(&xspb.FileDecorations_Override{MarkedSource: o.MarkedSource})
+	if err != nil {
+		return nil, err
+	}
+	return &KV{key, val}, nil
+}
+func encodeDecorDiagnostic(prefix []byte, file *spb.VName, d *xspb.FileDecorations_Diagnostic) (*KV, error) {
+	val, err := proto.Marshal(d)
```
salguarnieri: Can you move this down a few lines to match the pattern of the other encode* functions?

kythe/proto/xref_serving.proto
```diff
@@ -108,7 +123,8 @@ message FileDecorations {
   //
   // Columnar key: 60-override
   message Override {
-    kythe.proto.common.MarkedSource marked_source = 1;
+    kythe.proto.VName override = 1;
+    kythe.proto.common.MarkedSource marked_source = 2;
```
salguarnieri: Is it safe to reorder these fields?

kythe/proto/xref_serving.proto
```diff
@@ -55,7 +69,7 @@ message FileDecorations {
   // Columnar key: 10-start-end-kind-target
   message Target {
     // Starting offset of this file decoration.
-    int32 start_offest = 1;
+    int32 start_offset = 1;
```
salguarnieri: My diff view is saying something changed on this line, but I can't tell what it is.

schroederc: `offset` was misspelled.

kythe/proto/xref_serving.proto
schroederc: Yes, no one uses this format yet.

kythe/go/serving/xrefs/columnar/columnar_encoding.go
schroederc: The value is marshaled first because the key needs its hash.

kythe/go/serving/xrefs/columnar/columnar_encoding.go
schroederc: It's all encoded in the key.

kythe/go/serving/xrefs/columnar/columnar_encoding.go
schroederc: It ensures that the capacity of the `[]byte` is the same as its length.  This means that any `append` operation will copy the underlying bytes instead of using any leftover capacity.  If we didn't do this, any concurrent use of the prefix would be a race.

kythe/go/serving/xrefs/columnar/columnar_encoding.go
schroederc: ✓

PR #2988, opened 2018-09-07
Move common schema functions into schema package

schroederc:


kythe/go/util/schema/schema.go
```diff
@@ -96,3 +97,35 @@ func (n *Node) ToEntries() []*spb.Entry {
...
+// GetSubkind returns the string representation of the node's subkind.
+func GetSubkind(n *scpb.Node) string {
+	if k := n.GetGenericSubkind(); k != "" {
+		return k
+	}
+	return SubkindString(n.GetKytheSubkind())
+}
+
+// GetFactName returns the string representation of the fact's name.
+func GetFactName(f *scpb.Fact) string {
```
salguarnieri: Why the switch to Get* for the function names?

schroederc: The unprefixed names are already used.  :-/

PR #3013, opened 2018-09-14
Make the asciidoc rule emit outputs to a zip file.

creachadair:
Because asciidoc may generate an unknown number of output files (as the result of running filters), 
fix the rule to capture these in a ZIP file instead.  Also add some documentation so that future readers 
can see what's going on. This addresses an issue where recent versions of Bazel, which have become
more aggressive about cleaning up the sandbox after an action executes, would lose the support files
from the asciidoc filters.

Also update the website documentation builder to account for the above.

### DIFF
--- a/kythe/docs/asciidoc.bzl
+++ b/kythe/docs/asciidoc.bzl
@@ -1,7 +1,13 @@
 load("@bazel_skylib//:lib.bzl", "shell")
 
+_toolchain_type = "//tools/build_rules/external_tools:external_tools_toolchain_type"
+
 def _asciidoc_impl(ctx):
-    asciidoc = ctx.toolchains["//tools/build_rules/external_tools:external_tools_toolchain_type"].asciidoc
+    # Locate the asciidoc binary from the toolchain and construct its args.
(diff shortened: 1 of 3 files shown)


kythe/docs/asciidoc.bzl
```diff
@@ -11,38 +17,51 @@ def _asciidoc_impl(ctx):
...
+                  ctx.files.data),
         outputs = [ctx.outputs.out, logfile],
-        env = {
-            "PATH": tool_path,
-        },
+        env = {"PATH": tool_path},  # so we can locate the binaries asciidoc needs
         command = "\n".join([
-            '%s %s 2> %s' % (
+            # Create the temporary staging directory.
+            "mkdir out",
```
shahms: use https://docs.bazel.build/versions/master/skylark/lib/actions.html#declare_directory and provide the path to that declared directory here rather than hardcoding `out`?

creachadair: I did that originally, but there wasn't any point to it—the directory is a nonce, and doesn't outlive the rule, so declaring it doesn't add any value (and complects the code quite a bit)

Issue #1213, opened 2018-06-25
Write an informal survey/overview for the Kythe schema [phab:T133]


kythebot:
_Created by **fromberger** at 2016-05-24 17:17:47_: 

The current schema document is primarily a reference manual for the naming conventions, facts, and edges defined by the Kythe graph. For authors of new indexers, it would be helpful to have an informal survey/orientation document for the schema. Some topics I think would be useful to gather together include:

* Naming conventions (tickets and VNames, how to ensure uniqueness using examples like files, anchors, built-in types).
* How Kythe maps a graph of nodes and edges to entry tuples (node = facts, edge = nodes + label).
* A high-level tour of some of the common graph structures: File info, definition/usage cross-reference, call-graph, maybe a simple type structure.

It's an open question whether this should be part of the existing schema, or a separate document. The reference component of the schema should remain the source of truth, but an informal survey would I think help people figure out which pieces of the schema they need to care about—and why.

kythe:
_**fromberger** wrote at 2016-05-24 17:42:19_:

Luke's blog posts from last year are probably a good source of input here, too: [[ http://www.kythe.io/2015/07/13/flow/ | Indexing JavaScript with Flow ]] and [[ http://www.kythe.io/2015/08/05/flow-2/ | Part 2]] of the same. Some of the info in those posts may be slightly stale, but I think the essentials are still true.

creachadair:
I can't close this myself, but since #3015 is merged I think it's complete.

PR #3015, opened 2018-09-15
Add a Kythe schema overview doc.

creachadair:
This document serves as a high-level outline of a select handful of graph
structures from the Kythe schema, intended to help the developers of Kythe
indexers get oriented. It does not cover every detail, but provides some basic
terminology and examples drawn from the Kythe repository.


creachadair:
I think that the buildbot is missing some parts of what asciidoc needs.

creachadair:
Addresses #1213.

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+
+Kythe Schema Overview
+=====================
+:Revision: 1.0
+:toc2:
+:toclevels: 3
+:priority: 999
+
+This document is intended as a high-level overview of the
+http://kythe.io/docs/schema/[Kythe graph schema], and explains some of the
```
schroederc: %s/http/https/g

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+
+From the same example above, we can think of `Revisions` as the definition site
+of a semantic node (`function`) belonging to the type `Reader`. It has an
+anchor in `kcd.go` that defines where it is in the file. There is also another
+anchor in
+https://github.com/creachadair/kythe/blob/master/kythe/go/platform/kcd/memdb/memdb.go[`memdb.go`]
+where a `Revisions` function is implemented for the (concrete) `DB` type in
+that file.
+
+A **fact** is a named string value associated with a node. The
```
schroederc: Fact values are technically just `bytes` (not `string` with its implication of encoding).  This is only really important for `/kythe/code` and `/kythe/text`, but it should be taken into account.

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+http://kythe.io/docs/schema/#_common_node_facts[Kythe schema], defines the
+names and expected formats for various facts; for example the `doc/uri` fact is
+commonly used to attach a reference to some externally-located documentation to
+a node in the graph.
+
+An **edge** is a directed, labelled relationship between two nodes. In the above
+example, we would paint an edge from the anchor at `Revisions` to the semantic
```
schroederc: Instead of "to the ... node for its ... node", maybe: "to its semantic `function` node".

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+node for its `function` node, with the `defines/binding` label. This indicates
+that the anchor is a "definition" for the function (specifically, one that binds
+it to an identifier).
+
+NOTE: In the diagrams below, not all the possible anchors, semantic nodes,
+facts, or edges are shown, just a few examples. We will continue to refer back
```
schroederc: s/, just a few examples//

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+to the same `kcd.go` file where possible.
+
+== [[Examples]] Edge and Node Examples
+
+Now we will explain some of the basic edges and nodes that you'll see in a kythe
```
schroederc: %s/kythe/Kythe/g

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+The first mention of `matchRevision` (commented as "definition") records an
+anchor with a `defines/binding` edge to the semantic node for that variable. The
+references (commented "reference 1" and "reference 2") have anchors with `ref`
+edges to the same semantic node.
+
+There may in general be more than one definition site for a given semantic node,
+and in some cases there may be no definition at all. When a unique definition
+does exist we call it the **target definition** of the node.
+
+An example with no definition sites is an implicit constructor in Java: If you
```
schroederc: s/If/if/

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+
+=== [[FeatureFunc]] Functions and Parameters
+
+Functions are represented by semantic nodes that have additional edges to
+indicate their relationship to the parameters of the function. These edges are
+named `param.0`, `param.1`, etc., where the ordinal reflects the order of
+declaration.
+
+As illustrated in this example, the parameters in turn may have their own
+declaration sites, following the conventions of the jump-to-definition example:
```
schroederc: We should explicitly note that declarations also use `defines` edges before this point.

kythe/docs/schema-overview.txt
```diff
@@ -0,0 +1,421 @@
...
+
+`kcd.go`:
+
+[source,Go]
+----
+type Reader interface {
+  Revisions(_ context.Context, filter *RevisionsFilter, f func(Revision) error) error
+----
+
+NOTE: In Go, the blank (`_`) parameter name is used when the syntax requires the
```
schroederc: We should just name the parameter and remove the note since it is not relevant.

kythe/docs/schema-overview.txt
creachadair: ✔︎

kythe/docs/schema-overview.txt
creachadair: ✔︎

kythe/docs/schema-overview.txt
creachadair: Agreed. I also cleaned up the edge-label description similarly.

kythe/docs/schema-overview.txt
creachadair: ✔︎

kythe/docs/schema-overview.txt
creachadair: ✔︎

kythe/docs/schema-overview.txt
creachadair: ✔︎

kythe/docs/schema-overview.txt
creachadair: I'm not sure I follow why it matters here—the example that follows uses only `defines/binding`.

kythe/docs/schema-overview.txt
creachadair: Sure, though it's no longer a snippet from the original source then.

But that probably doesn't matter too much, so done.

kythe/docs/schema-overview.txt
schroederc: Because the document talks multiple times of declarations but only implies the connection to `defines` edges.

creachadair:
Per your suggestion offline, I also tagged this rule `manual` for now, till we can update the buildbot.

kythe/docs/schema-overview.txt
creachadair: Ah, I see what you mean. Yes, that makes sense. PTAL.

PR #3026, opened 2018-09-19
Add a 'nobuildbot' label for skipping buildbot.

danielmoy-google:
Presumably if you know for sure build isn't relevant (say the commit
doesn't affect buildable files), or if buildbot itself is hosed, give
ourselves an easy out.

(I didn't know if we already had some way to do this, if so we can
just kill this commit)


danielmoy-google:
(From irc, schroederc notes that the required nature of buildbot check gating merge will prevent the original idea I had from actually working - but might still be useful for things in development)

PR #3032, opened 2018-09-20
Add Bazel RBE builder in Buildbot

schroederc:


tools/bazel.rc
```diff
@@ -1,5 +1,8 @@
 # TODO(#2897): move to %workspace%/.bazelrc
 
+# Remote Build Execution requires a strong hash function, such as SHA256.
+startup --host_jvm_args=-Dbazel.DigestFunction=SHA256
```
schroederc: Turns out you can't put a startup option under a config.

creachadair: Oh. I guess I should have realized that, since presumably the configs aren't read till after startup. ☹️

PR #3036, opened 2018-09-21
Fix asciidoc targets in RBE environment

schroederc:


kythe/docs/asciidoc.bzl
```diff
@@ -36,7 +32,6 @@ def _asciidoc_impl(ctx):
                   ([ctx.file.example_script] if ctx.file.example_script else []) +
                   ctx.files.data),
         outputs = [ctx.outputs.out, logfile],
-        env = {"PATH": tool_path},  # so we can locate the binaries asciidoc needs
```
schroederc: This leads the action to not being able to find standard tools like `mkdir`.

shahms: This change will break macOS (which is why this was added in the first place).

schroederc: Does that work for macOS?

shahms: We'll see; I can take a look this afternoon to make sure.

shahms: This works.

PR #3056, opened 2018-09-25
Add a data dep on extractor runtime dep docker image.

danielmoy-google:
I foresee problems importing this, but worth a shot because it'll avoid bugs during dev of the form "but I fixed this and it's still broken, why?".

### DIFF
--- a/kythe/go/extractors/config/BUILD
+++ b/kythe/go/extractors/config/BUILD
@@ -8,6 +8,9 @@ go_library(
         "config_parser.go",
         "extractor.go",
     ],
+    data = [
+        "//kythe/java/com/google/devtools/kythe/extractors/java/artifacts",
+    ],
     deps = [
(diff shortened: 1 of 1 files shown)


kythe/go/extractors/config/BUILD
```diff
@@ -8,6 +8,9 @@ go_library(
         "config_parser.go",
         "extractor.go",
     ],
+    data = [
+        "//kythe/java/com/google/devtools/kythe/extractors/java/artifacts",
```
schroederc: Buildbot doesn't currently support `docker` so this rule will need to be tagged `"manual"`.

danielmoy-google: Hmm.  So then we have to figure out if we want to make the whole transitive stack up for this manual or not.

Probably not.

danielmoy-google: I'll just close this for now then.

PR #3058, opened 2018-09-25
Excise remaining references to old maven scripts and preprocessor.

danielmoy-google:


danielmoy-google:
Oh yea this should fix #2905

Issue #3069, opened 2018-09-26
Java kzip writer incorrectly string-quotes JSON output


creachadair:
When writing a unit file, the Java kzip writer renders the unit as a quoted string containing the JSON blob, rather than the JSON blob itself, e.g.,

    "{\"foo\":3}"

rather than

    {"foo":3}

This causes the reader to fail upon re-entry because a string is not an object, which is what the wrapper is intended to be.

PR #3072, opened 2018-09-26
Make maven/gradle extractors default to using kzip

### LINKED ISSUE #3059: Maven/gradle extraction should port to .kzip
Still writing .kindex for now.

danielmoy-google:
Also remove an unused method in java extractor.

Fixes #3059

### DIFF
--- a/kythe/go/extractors/config/base/gradle/gradle.go
+++ b/kythe/go/extractors/config/base/gradle/gradle.go
@@ -71,6 +71,10 @@ const DefaultConfig = `{
         {
           "name": "JAVAC_EXTRACTOR_JAR",
           "value": "/opt/kythe/extractors/javac_extractor.jar"
+        },
+        {
+          "name": "KYTHE_OUTPUT_FILE",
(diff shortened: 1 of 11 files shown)


creachadair:
#3069 may wind up preventing this from working until it is fixed.

danielmoy-google:
> #3069 may wind up preventing this from working until it is fixed.

Noted, though we might be fine with that since there's a big enough stack of work to be done on just extracting before any indexing is done.

Also this just affects the default, we can always use the config json to just specify a different value for KYTHE_OUTPUT_FILE, and it'd be set accordingly.

creachadair:
> Also this just affects the default, we can always use the config json to just specify a different value for KYTHE_OUTPUT_FILE, and it'd be set accordingly.

Yeah, so maybe for now we should include the setting, but use `.kindex` for  suffix, just so the fallback continues to work?

creachadair:
Per our offline discussion: A fix for #3069 is in progress, and nothing important depends on the default right now, so we can go ahead.

PR #3074, opened 2018-09-26
Fix KzipWriter to include mandatory directory entries and update test…

### LINKED ISSUE #3054: C++ kzip writer does not generate a root directory
Per http://www.kythe.io/docs/kythe-kzip.html, a kzip file is intended to have a directory as its first entry. The C++ writer emits the correct subdirectory structure, but does not include the root directory itself.

shahms:
…s to check/allow for this

Fixes #3054

### DIFF
--- a/kythe/cxx/common/kzip_writer.cc
+++ b/kythe/cxx/common/kzip_writer.cc
@@ -101,9 +101,8 @@ StatusOr<std::string> KzipWriter::WriteUnit(
     auto status = InitializeArchive(archive_);
     if (!status.ok()) {
       return status;
-    } else {
-      initialized_ = true;
     }
+    initialized_ = true;
   }
   if (auto json = WriteMessageAsJsonToString(unit)) {
     contents_.push_back(std::move(*json));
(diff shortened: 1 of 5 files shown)


kythe/cxx/common/kzip_writer.cc
```diff
@@ -84,6 +97,14 @@ KzipWriter::~KzipWriter() {
 
 StatusOr<std::string> KzipWriter::WriteUnit(
     const kythe::proto::IndexedCompilation& unit) {
+  if (!initialized_) {
+    auto status = InitializeArchive(archive_);
+    if (!status.ok()) {
+      return status;
+    } else {
```
schroederc: Remove `else` since the first branch escapes the method?  That's go style.  I'm not sure if it is C++.

kythe/cxx/common/kzip_writer.cc
```diff
@@ -97,6 +118,14 @@ StatusOr<std::string> KzipWriter::WriteUnit(
 }
 
 StatusOr<std::string> KzipWriter::WriteFile(absl::string_view content) {
+  if (!initialized_) {
+    auto status = InitializeArchive(archive_);
+    if (!status.ok()) {
+      return status;
+    } else {
+      initialized_ = true;
```
schroederc: Same as above.

shahms:
The C++ style guide makes no nevermind over that particular nit ;-)

PR #3076, opened 2018-09-26
Write out JSON correctly for Java kzips

### LINKED ISSUE #3069: Java kzip writer incorrectly string-quotes JSON output

salguarnieri:
Previously we were getting the JSON string from proto and asking gson to
turn that into an json element. This was wrong because it was taking a
json object and treating it as a single json string.

Now we get the json string, parse it Json java object, and then pass it
back to gson for it to use. This likely is inefficient because we are
then just asking gson to turn it back into a String, but this lets us
keep the current code structure with minimal changes.

Fixes #3069


salguarnieri:
I added a todo, ptal

PR #3084, opened 2018-09-28
Implement StandardJavaFileManager for wrapper class

schroederc:
This interface implementation is required by javac for compilations specifying a `--release` version.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/UsageAsInputReportingFileManager.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/UsageAsInputReportingFileManager.java
@@ -17,6 +17,8 @@
 package com.google.devtools.kythe.extractors.java;
 
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/UsageAsInputReportingFileManager.java
```diff
@@ -109,6 +112,38 @@ public boolean isSameFile(FileObject a, FileObject b) {
     return super.isSameFile(unwrap(a), unwrap(b));
   }
 
+  @Override
+  public Iterable<? extends JavaFileObject> getJavaFileObjects(String... names) {
+    return getJavaFileForSources(Lists.newArrayList(names));
```
danielmoy-google: I think Lists.newArrayList is weakly-ish deprecated.  Whether or not we actually care about that at this point, I dunno.

If that list doesn't need to be modified later, prefer ImmutableList.of(names) instead.

If it does need to be modified later, then this is fine as-is (or use Arrays.asList if you want to get ahead of any impending deprecation).


ditto below on 134

schroederc:
PTAL

PR #3091, opened 2018-10-01
Decouple verifier from LLVM dependency by splitting //kythe/cxx/commo…

shahms:
…n:lib some

This is the short-path to LLVM independence by re-implementing CleanPath inside kythe_uri.cc.  Ideally, we wouldn't depend on LLVM for path_utils and the header-search stuff would live elsewhere.

For now, this allows verifier to be built without LLVM.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -6,7 +6,7 @@ load("//:version.bzl", "check_version")
 
 # Check that the user has a version between our minimum supported version of
 # Bazel and our maximum supported version of Bazel.
-check_version("0.16", "0.17")
+check_version("0.16", "0.18")
 
 load("//tools/cpp:clang_configure.bzl", "clang_configure")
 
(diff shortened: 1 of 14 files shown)


shahms:
I don't know what happened to those BUILD files, that was not what they looked like when I ran `git commit`

shahms:
PTAL, seem to have resolved the mysterious BUILD reformatting (although there does seem to be a commit-hook-like behavior I'm not sure of the origin for), e.g.

$ git commit -m "Fix funky formatting on BUILD files"
[llvm-is-yuge 5577708a] Fix funky formatting on BUILD files
 4 files changed, 740 insertions(+), 841 deletions(-)
 rewrite kythe/cxx/common/BUILD (99%)
 rewrite kythe/cxx/doc/BUILD (98%)
 rewrite kythe/cxx/tools/fyi/BUILD (95%)
 rewrite kythe/cxx/verifier/BUILD (98%)

PR #3092, opened 2018-10-01
Update setup_release.sh to build release archive

schroederc:


schroederc:
`//kythe/release:release_test` builds the release archive as a dependency and then spot checks it.

salguarnieri:
> `//kythe/release:release_test` builds the release archive as a dependency and then spot checks it.

Gotcha, can you add a comment saying you depend on that side effect?

Issue #3075, opened 2018-09-26
maven/gradle extraction needs to support jdk of version newer than 8


danielmoy-google:
# Some things will break

## xbootclasspath/p

https://github.com/google/kythe/blob/master/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh#L65

There's some random discussions you can find online about this, including [the official java9 notes](https://www.oracle.com/technetwork/java/javase/9-relnote-issues-3704069.html) which suggest --patch-module is the replacement, and [stack overflow](https://stackoverflow.com/questions/48148607/how-to-add-a-jar-to-the-boot-classpath-in-java-9) which say to just use xbootclasspath/a.

Cody suggests also jvm flag `--add-opens=jdk.compiler/com.sun.tools.javac.api=java.compiler`

## base docker images need to not ask for 8

https://github.com/google/kythe/blob/master/kythe/go/extractors/config/base/mvn/mvn.go#L34

## tools/bazel.rc would need to not be pinning java8:

https://github.com/google/kythe/blob/master/tools/bazel.rc#L13

## will depend on javac8wrapper being not java8

https://github.com/google/kythe/blob/master/kythe/java/com/google/devtools/kythe/extractors/java/standalone/Javac8Wrapper.java

# other random thoughts

Also see bazel issue for jdk 11: https://github.com/bazelbuild/bazel/issues/6245

Maybe just jump straight to jdk 11.  Skip 9 & 10 completely.

cushon:
If you need to override the javac version, the replacement flags for `-Xbootclasspath/p` are `--patch-module=java.compiler=java_compiler.jar --patch-module=jdk.compiler=jdk_compiler.jar`. Here's an example in Bazel:

https://github.com/bazelbuild/bazel/blob/610e8ff2baebe996eee8dfea1c13be554c9464ff/tools/jdk/default_java_toolchain.bzl#L36

You cold also try using the javac version in the host JDK instead of overriding it.

schroederc:
We already use those options in Bazel: https://github.com/google/kythe/blob/master/kythe/java/com/google/devtools/kythe/extractors/java/bazel/BUILD#L26.  This issue is for catching up the Maven extraction.

schroederc:
You can piggy-back off of this branch to ensure that the Maven extraction works with 9: https://github.com/schroederc/kythe/tree/java9

cushon:
[...]

PR #3093, opened 2018-10-01
Migrate javac_extractor to use ambient langtools

schroederc:
Part of #3075.

This removes our hard-coded usage of `-Xbootclasspath/p` and allows the `javac-wrapper.sh` to be version-agnostic.  When running on JRE 8, the `-Xbootclasspath/p` flag can be passed using the `KYTHE_JAVA_RUNTIME_OPTIONS` environment variable.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
@@ -20,7 +20,9 @@
 # KYTHE_OUTPUT_DIRECTORY set to understand where the root of the compiled source
 # repository is and where to put the resulting .kindex files, respectively.
 #
-# This script assumes a usable java binary is on $PATH.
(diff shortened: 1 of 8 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
```diff
@@ -40,6 +40,9 @@
 
 export TMPDIR="${TMPDIR:-/tmp}"
 
+if [[ -z "$JAVABIN" ]]; then
+  readonly JAVABIN="$(which java)"
```
danielmoy-google: huh wonder if we need REAL_JAVAC still then.

maybe fine to leave in

schroederc: I changed it to use `JAVA_HOME` instead.

The reason `javac` uses a different var because this wrapper be be added to `PATH` to replace `javac` silently.

PR #3099, opened 2018-10-02
First pass at fixing a build error witing kzip.

danielmoy-google:
Right now duplicates crash the pipeline, but should probably be safely
ignored.


danielmoy-google:
I am unconvinced that this is the right approach, and also I'm not quite sure yet how to test it.

PR #3100, opened 2018-10-02
Don't crash if we try to write the same file to a kzip more than once

salguarnieri:
It looks like we may see the same file multiple times during extraction. If we try to add the file to the zip again, the zip library will throw a generic zip exception. We can avoid this situation by just not appending a file if we know we have already added it to the zip.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
+++ b/kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
@@ -97,7 +97,7 @@ public final class KZipWriter implements KZip.Writer {
       output.write(data);
       output.closeEntry();
     } else {
-      logger.atWarning().log("Warning: Already wrote " + path + " to kzip.");
+      logger.atWarning().log("Warning: Already wrote %s to kzip.", path);
     }
   }
 
--- a/kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
+++ b/kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
@@ -15,6 +15,7 @@
  */
 package com.google.devtools.kythe.platform.kzip;
 
+import com.google.common.flogger.FluentLogger;
 import com.google.devtools.kythe.proto.Analysis;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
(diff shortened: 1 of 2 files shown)


kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
```diff
@@ -81,10 +88,14 @@ public String writeFile(byte[] data) throws IOException {
...
-    output.putNextEntry(entry);
-    output.write(data);
-    output.closeEntry();
+    if (pathsWritten.add(path)) {
+      ZipEntry entry = new ZipEntry(path);
+      output.putNextEntry(entry);
+      output.write(data);
+      output.closeEntry();
+    } else {
+      System.err.println("Warning: Already wrote " + path + " to kzip.");
```
danielmoy-google: maybe flogger if cody pulled that in (I forget)

kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
```diff
@@ -31,6 +33,9 @@
   private final ZipOutputStream output;
   private final Gson gson;
 
+  // We may
```
danielmoy-google: maybe fill out the comment

// We store our own set of paths written, analogous to ZipOutputWriter.names, so that we avoid
// raising exceptions by writing duplicates.

kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
```diff
@@ -94,7 +96,7 @@ private void appendZip(byte[] data, String path) throws IOException {
       output.write(data);
       output.closeEntry();
     } else {
-      System.err.println("Warning: Already wrote " + path + " to kzip.");
+      logger.atWarning().log("Warning: Already wrote " + path + " to kzip.");
```
danielmoy-google: nit: use the built in log formatters

"Warning: already wrote %s to kzip.", path);

In this case it doesn't actually matter, but the habit is worth forming because that one time you write a logger.atInfo().log("String appended with " + someExpensiveToMarshalObject); in a performance-critical loop, you'll have a bad time :P

kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
salguarnieri: Done

kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
salguarnieri: done

kythe/java/com/google/devtools/kythe/platform/kzip/KZipWriter.java
salguarnieri: done

PR #3106, opened 2018-10-03
Begin migrating external repository definitions to importable WORKSPACE function

schroederc:
This allows other Bazel projects to more easily depend on Kythe without
needing to copy our entire WORKSPACE file.

Only repositories needed to build targets exposed by #3105 have been
migrated.


schroederc:
Buildbot is complaining about the switch to the non-native `maven_jar` repository rule.  It requires `mvn` installed to work properly.

external.bzl
```diff
@@ -0,0 +1,120 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
```
danielmoy-google: maybe worth a comment in here explaining what should go here vs directly in WORKSPACE

PR #3109, opened 2018-10-04
Make kythe Entry XOR nature more explicit.

danielmoy-google:


kythe/proto/storage.proto
```diff
@@ -91,7 +91,7 @@ message VNameMask {
   bool language = 5;
 }
 
-// An Entry associates a fact with a graph object (node or edge).  This is the
+// An Entry associates a fact with a graph object (node XOR edge).  This is the
```
schroederc: We should probably clean this up further.  Originally, all Entry messages were considered "facts", but now we tend to refer to an Entry as either a fact XOR an edge and a group of them sharing the same source VName as a node.

creachadair: This is a reasonable choice, but we intentionally left room in the data model for edges to carry facts. We chose not to use that (and collapsed the ordinal case into the edge label to avoid the one place where we used to rely on it). But it's not intrinsic in the model.

schroederc: We've been pretty consistent on not wanting facts on edges.  We're probably at the point we can just declare it as truth.

creachadair: Sure, but I think if we want to be explicit about that, we should document more than simply "or" becoming "xor". Probably the storage model doc could use some colour too.

danielmoy-google: new wording

danielmoy-google: https://github.com/google/kythe/pull/3109/commits/c783b50d9e8a8ca7f0b238e1c8dda5b6251823f0

?

danielmoy-google:
We ok with this wording?

kythe/docs/kythe-storage.txt
```diff
@@ -346,24 +346,34 @@ specific rules that apply to their construction and relationships.
...
  * *Source.* The <<TermTicket,ticket>> of the source node (must not be empty).
- * *Kind.* An edge label (may be empty). The schema defines which labels are
-   meaningful.
- * *Target.* The <<TermTicket,ticket>> of the target node (may be empty).
- * *Fact.* A fact label (must not be empty). The schema defines which fact labels
-   are meaningful.
- * *Value.* The string value of the fact (may be empty).
+ * *EdgeKind.* An edge label (only used for edge). The schema defines which
+   labels are meaningful.
+ * *Target.* The <<TermTicket,ticket>> of the target node (ionly used for edge).
```
creachadair: typo: ionly

kythe/docs/kythe-storage.txt
```diff
@@ -346,24 +346,34 @@ specific rules that apply to their construction and relationships.
...
 where
 
  * *Source.* The <<TermTicket,ticket>> of the source node (must not be empty).
- * *Kind.* An edge label (may be empty). The schema defines which labels are
-   meaningful.
- * *Target.* The <<TermTicket,ticket>> of the target node (may be empty).
- * *Fact.* A fact label (must not be empty). The schema defines which fact labels
-   are meaningful.
- * *Value.* The string value of the fact (may be empty).
+ * *EdgeKind.* An edge label (only used for edge). The schema defines which
```
creachadair: Please don't lose the bit about which fields may be empty.

We've talked about getting rid of the requirement that the fact label (currently) must be nonempty, but we shouldn't do that as part of this change IMO.

danielmoy-google:
Unclear what I just did with git.  Probably pushed the wrong button.

danielmoy-google:
> Unclear what I just did with git. Probably pushed the wrong button.

Ah I should not have updated this branch in the middle, oh well.  Should fix itself on squash.

kythe/docs/kythe-storage.txt
danielmoy-google: done

kythe/docs/kythe-storage.txt
danielmoy-google: done

shahms:
Daniel, is this PR still relevant?

schroederc:
Closing extremely stale PR.

PR #3119, opened 2018-10-08
Use kzip format in cc_verifier_test rule

schroederc:


kythe/cxx/common/indexing/frontend.cc
```diff
@@ -135,6 +136,34 @@ void DecodeIndexFile(const std::string& path,
   close(fd);
 }
 
+/// \brief Reads data from a .kzip file into memory.
+/// \param path The path from which the file should be read.
+/// \param virtual_files A vector to be filled with FileData.
+/// \param unit A `CompilationUnit` to be decoded from the .kzip.
```
shahms: This doesn't deal with the fact that .kzip files can contain more than one CompilationUnit, is that intentional?

schroederc: Yeah, it's the same way in java indexer.

shahms: I suspect that is going to cause confusing problems in the future.

PR #3125, opened 2018-10-09
Add Javac9Wrapper to support --release flag

schroederc:


kythe/java/com/google/devtools/kythe/extractors/java/standalone/Javac9Wrapper.java
```diff
@@ -0,0 +1,138 @@
...
+    // Use javac's argument parser to get the list of source files
+    Context context = new Context();
+
+    JavacFileManager fileManager = new JavacFileManager(context, true, null);
+    Arguments args = Arguments.instance(context);
+    args.init("kythe_javac", arguments);
+    fileManager.handleOptions(args.getDeferredFileManagerOptions());
+    Options options = Options.instance(context);
+
+    List<String> sources = new ArrayList<>();
```
danielmoy-google: If it doesn't need to be mutable, maybe

List<String> sources = args.getFileObjects().stream().map(fo -> fo.getName()).collect(toImmutableList());

Or not, I'm fine this way too.

Issue #3120, opened 2018-10-08
Support analyzing kzip files with multiple compilation units


schroederc:
- [X] Support in [go_indexer.go](https://github.com/google/kythe/blob/6d1bfe86d8350d3e9bf402542fa43551bbd88bb1/kythe/go/indexer/cmd/go_indexer/go_indexer.go#L173)
- [x] Support in [JavaIndexer.java](https://github.com/google/kythe/blob/f32f399c943c6d6be65648918c54a45b4812b0ac/kythe/java/com/google/devtools/kythe/analyzers/java/JavaIndexer.java#L107)
- [x] Support in [ClassFileIndexer.java](https://github.com/google/kythe/blob/f32f399c943c6d6be65648918c54a45b4812b0ac/kythe/java/com/google/devtools/kythe/analyzers/jvm/ClassFileIndexer.java#L101)
- [x] Support in C++ [frontend.cc](https://github.com/google/kythe/blob/c69b7c50650f7b51f037193f33ed706fe1d4aeee/kythe/cxx/common/indexing/frontend.cc#L139)

justbuchanan:
Currently when using index pack files, usage looks like `cxx_indexer --index_pack=path/to/file hash [more hashes...]`

Do we want also want the ability to select specific CU hashes when working with kzips, or should we always read all CUs when given a kzip file? From what I can tell, the java and go indexers just read all contents of kzips.

schroederc:
I think analyzing all compilations in a `.kzip` is a good default.

justbuchanan:
C++ support added in #3232.

PR #3135, opened 2018-10-10
Java indexer: allow for analysis of multiple compilations

schroederc:
Part of #3120

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/java/JavaIndexer.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/java/JavaIndexer.java
@@ -49,6 +49,7 @@ import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.ServiceLoader;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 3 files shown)


kythe/java/com/google/devtools/kythe/analyzers/java/JavaIndexer.java
```diff
@@ -83,53 +89,74 @@ public static void main(String[] args) throws AnalysisException, IOException {
...
-        // java_indexer kindex-file
-        try {
-          desc = IndexInfoUtils.readKindexInfoFromFile(compilationPath);
-        } catch (EOFException e) {
-          if (config.getIgnoreEmptyKIndex()) {
-            return;
+        } else {
+          // java_indexer kindex-file
+          if (!loggedDeprecationWarning) {
+            logger.atWarning().log(
```
shahms: `logger.atWarning().atMostEvery(1, DAYS)` ? I wish there was a clearer way to say `logger.atWarning().once()`, but given the expected duration of a single extraction, using DAYS seems fine and will avoid an extraneous logging local.  But up to you if you think it's worth it.

shahms: Particularly since you're never actually setting the local bool ;-)

schroederc: There is a Google-internal method that would work for `once()`, but...

PR #3136, opened 2018-10-10
Prefer String.getBytes(Charset)

danielmoy-google:


danielmoy-google:
Only one I could find

PR #3159, opened 2018-10-17
build: satisfy Bazel 0.18.0 `--all_incompatible_changes`

schroederc:


schroederc:
Note: our full build does not work with `--all_incompatible_changes` because of a couple of dependencies

third_party/libzip.BUILD
```diff
@@ -42,73 +44,104 @@ _CMAKE_VARIABLES = {
...
 
+_VARS = {
+    "LIBZIP_TYPES_INCLUDE": "#include <stdint.h>",
+    "PACKAGE_VERSION": "1.5.1",
+    "PACKAGE_VERSION_MAJOR": "1",
+    "PACKAGE_VERSION_MINOR": "5",
+    "PACKAGE_VERSION_MICRO": "1",
+}
+
+_VARS.update(dict([
```
shahms: We can't add dictionaries any more? Ugh. Starlark is getting worse.

schroederc:
The `user_compile_flags` change is not compatible with 0.16.0 so I left it alone for now.  I was thinking about updating our min once 0.18.1. is out.

Issue #3150, opened 2018-10-16
Support kythe extraction on Cloud Build


danielmoy-google:
We want some way to build in "the cloud" instead of just locally, presumably pushing results to gcs for later indexing or static analysis.

There will be a variety of child issues on this.

danielmoy-google:
#3219 has gen code

PR #3160, opened 2018-10-17
Moves build preprocessing libraries into a binary

danielmoy-google:
Will need to call this binary in a dockerfile up on cloud build to satisfy #3150.


kythe/go/extractors/config/preprocessor/README.md
```diff
@@ -0,0 +1,73 @@
...
+    tasks.withType(JavaCompile) {
+      options.fork = true
+      options.forkOptions.executable = '/opt/kythe/extractors/javac-wrapper.sh'
+    }
+  }
+}
+```
+
+If the input file already contains reference to
+`options.forkOptions.executable`, then `gradle_build_modifier.go` does nothing.
```
schroederc: Can we not extract from those builds?

kythe/go/extractors/config/preprocessor/README.md
```diff
@@ -0,0 +1,73 @@
...
+  gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+      options.fork = true
+      options.forkOptions.executable = '/opt/kythe/extractors/javac-wrapper.sh'
+    }
+  }
+}
+```
+
+If the input file already contains reference to
```
schroederc: s/reference/a reference/

kythe/go/extractors/config/preprocessor/README.md
```diff
@@ -0,0 +1,73 @@
...
+The current implementation uses simple string-based matching, without actually
+understanding the structure.  If that becomes necessary in the future, it might
+be better to use the existing Java libraries for `org.codehaus.groovy.ast` to
+properly parse the build.gradle file and have more precise picture.  In
+particular `org.codehaus.groovy.ast.CodeVisitorSupport` might be sufficient.
+Ideally we find such a library in golang.
+
+### Maven
+
+Maven's build config is handled by `pom_xml_modifier.go`.  It utilizes the etree
```
schroederc: I don't think it matter which xml library we use for the purposes of this documentation.

kythe/go/extractors/config/preprocessor/README.md
```diff
@@ -0,0 +1,73 @@
...
+xml library to parse and modify the mvn `pom.xml` config file in a similar way
+as described above for gradle.  One notable difference between gradle and maven
+here is that gradle actually embeds the refrence to the javac wrapper directly
```
schroederc: s/refrence/reference/

kythe/go/extractors/config/preprocessor/README.md
```diff
@@ -0,0 +1,73 @@
...
+into the build file, while the modifications to the maven pom xml file merely
```
schroederc: s/into/in/

kythe/go/extractors/config/preprocessor/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+<build>
+  <plugins>
+    <plugin>
+      <groupId>org.apache.maven.plugins</groupId>
+      <artifactId>maven-compiler-plugin</artifactId>
+      <version>3.7.0</version>
+      <configuration>
+        <source>1.8</source>
+        <target>1.8</target>
```
schroederc: What about other Java versions?

kythe/go/extractors/config/preprocessor/README.md
danielmoy-google: That is a good question.

I have been banking on the theory that

1. yes we will be able to extract from those builds, and
2. we can do this by just overriding the executable, and then pointing to the overridden executable in javac-wrapper.sh

Then we'd just need some small modification in here in the mutator, and also a way to propagate the overridden bit back out and stuff it into REAL_JAVAC or whatever environment variable (I forget off the top of my head).

Of course, I have no proof that any of this would work.

In any event, filing an issue:

#3162

kythe/go/extractors/config/preprocessor/README.md
danielmoy-google: done

kythe/go/extractors/config/preprocessor/README.md
danielmoy-google: sure sg

kythe/go/extractors/config/preprocessor/README.md
danielmoy-google: done

kythe/go/extractors/config/preprocessor/README.md
danielmoy-google: done

kythe/go/extractors/config/preprocessor/README.md
danielmoy-google: TBD still #3075

In particular my first efforts towards getting this to work have failed.

kythe/go/extractors/config/preprocessor/modifier/BUILD
```diff
@@ -0,0 +1,34 @@
...
+    srcs = [
+        "build_gradle_modifier.go",
+        "pom_xml_modifier.go",
+    ],
+    deps = [
+        "@com_github_beevik_etree//:go_default_library",
+    ],
+)
+
+go_test(
```
schroederc: The `go` tool considers all `_test.go` files as part of the same package so the build will break since the same top-level names are used in this test and the one below.

danielmoy-google: ahh damn.  prefer separate packages for both of these then?

schroederc: I'd just make sure the variable names don't clash in the tests.

danielmoy-google: This feels like a way I will shoot myself in the foot again in the future though, so I'm gonna separate them out now.  (Well, again, since that's the way they were before too)

schroederc: You can prevent the foot shooting by combining the Bazel `go_test` targets into one.

danielmoy-google: If we get rid of runextractor binary I may do this then.

For now having separate libraries means we don't have a maven subcommand depending on a bunch of gradle stuff and v. versa.

If the future is combined (which I'm leaning more towards now), I can come back and fix this.

danielmoy-google:
(I shot myself in the foot by updating this branch too early)

danielmoy-google:
Ok my lack of will to maintain the local running stack on  top of the cloud build stack will quickly outweigh my interest in keeping these libs separate.

Though I do wonder if we'll have a way of running cloudbuild.yaml stuff locally.

PR #3163, opened 2018-10-18
fix for upcoming bazel change

shahms:
Specifically addresses https://github.com/bazelbuild/bazel/issues/6383


schroederc:
As I said in #3159, this requires a Bazel min version bump which I was planning on doing after 0.18.1 is out.

PR #3166, opened 2018-10-18
fix: recover file VNames using file content; reorder singleton checking

### LINKED ISSUE #3157: Range type needs to disambiguate wrt files
The verifier range primitive needs a file ID attached; otherwise, if you're really unlucky, anchors could alias across files.

zrlk:
The verifier was initially used to process a single file's subgraph.
As such, it only used begin and end offsets to represent anchor ranges.
Now that it supports multiple files, it ought to disambiguate beween
ranges in different files. Unfortunately, if the verifier is used in
the normal manner (without --use_file_nodes), there is no direct way to
recover all of the VName fields necessary to build anchor goals, since
the filename passed to the verifier may have no relation to what an
indexer produces. This change works around this problem by searching for
file nodes with the same *content* as the verifier input files and using
their VNames to construct anchor goals.

The problem is exacerbated if you are a heavy user of goal groups,
as backtracking might otherwise undo the wrong anchor being selected
for a variable.

BREAKING CHANGE: Particularly if your tests rely on files that your
indexer does not emit as file nodes (e.g., you supply the verifier
with an out-of-band text file with goals), this change will cause your
tests to fail. For behavior that's similar to what the verifier was
doing before, assert the --nofile_vnames flag.

BREAKING CHANGE: If you were using --use_file_nodes and the (default)
--check_for_singletons, you may discover new singletons in your code,
as previous to this change the singleton check was performed *before*
file nodes were read from the input graph.

Fixes #3157.


kythe/cxx/verifier/assertion_ast.h
```diff
@@ -200,18 +200,28 @@ class AstNode : public ArenaObject {
 /// \brief A range specification that can unify with one or more ranges.
 class Range : public AstNode {
  public:
-  Range(const yy::location& location, size_t begin, size_t end)
+  Range(const yy::location& location, size_t begin, size_t end, Symbol source,
```
schroederc: `s/Symbol source/Symbol path/`

kythe/cxx/verifier/testdata/regex_expected_error.txt
```diff
@@ -1,2 +1,2 @@
 Could not verify all goals. The furthest we reached was:
-  regex_input.txt:2:6-2:19 _Aaa bbb _Ccc
+  App(vname, ("", "", "", "", "")):2:6-2:19 _Aaa bbb _Ccc
```
shahms: This makes these error messages substantially less informative.

kythe/cxx/verifier/verifier.h
```diff
@@ -41,16 +41,19 @@ class Verifier {
 
   /// \brief Loads an in-memory source file.
   /// \param vname The AstNode of the vname for the file.
-  /// \param text The symbol for the text to load
+  /// \param corpus The corpus to use for anchors
```
schroederc: This change doesn't match the decl.

kythe/cxx/verifier/verifier.cc
```diff
@@ -375,7 +380,9 @@ class Solver {
       }
     } else if (Range* sr = s->AsRange()) {
       if (Range* tr = t->AsRange()) {
-        if (sr->begin() == tr->begin() && sr->end() == tr->end()) {
+        if (sr->begin() == tr->begin() && sr->end() == tr->end() &&
+            sr->path() == tr->path() && sr->root() == tr->root() &&
+            sr->corpus() == tr->corpus()) {
```
shahms: There are enough components being compared here that this should likely be a separate function or an intrinsic part of Range

kythe/cxx/verifier/verifier.cc
```diff
@@ -681,24 +688,68 @@ bool Verifier::LoadInlineProtoFile(const std::string& file_data) {
...
+  int fd = ::open(filename.c_str(), 0);
+  if (fd < 0) {
+    LOG(ERROR) << "Can't open " << filename;
     return false;
   }
-  return true;
+  struct stat fd_stat;
+  if (::fstat(fd, &fd_stat) < 0) {
+    LOG(ERROR) << "Can't stat " << filename;
+    return false;
```
shahms: fd is leaked here

kythe/cxx/verifier/verifier.cc
```diff
@@ -681,24 +688,68 @@ bool Verifier::LoadInlineProtoFile(const std::string& file_data) {
...
+  }
+  std::string content;
+  content.resize(fd_stat.st_size);
+  if (::read(fd, const_cast<char*>(content.data()), fd_stat.st_size) !=
+      fd_stat.st_size) {
+    LOG(ERROR) << "Can't read " << filename;
+    return false;
```
shahms: and here

kythe/cxx/verifier/verifier.cc
```diff
@@ -681,24 +688,68 @@ bool Verifier::LoadInlineProtoFile(const std::string& file_data) {
...
+  }
+  ::close(fd);
```
shahms: this should likely use kythe/cxx/indexer/cxx/scope_guard.h (or that file should be moved to common, there's nothing indexer specific about it).

kythe/cxx/verifier/verifier.h
shahms: Description of the parameter is changed, but not the parameter name itself.

kythe/cxx/verifier/verifier.h
```diff
@@ -376,6 +382,12 @@ class Verifier {
 
   /// Identifier for MarkedSource false values.
   AstNode* marked_source_false_id_;
+
+  /// Maps from file content to (verified) VName.
+  std::unordered_map<Symbol, AstNode*> content_to_vname_;
```
shahms: absl::flat_hash_map is a thing now and should be preferred (although we might have to update the absl dependency).

schroederc:
`//kythe/release:release_test` needs to be fixed as well.

PR #3169, opened 2018-10-18
Add a build-preprocessor docker image for config/preprocessor.

danielmoy-google:
Expose at gcr.io/kythe-public/build-preprocessor for use in cloud build
for #3150.

### DIFF
--- a/kythe/go/extractors/config/preprocessor/Dockerfile
+++ b/kythe/go/extractors/config/preprocessor/Dockerfile
@@ -1,0 +1,3 @@
+FROM scratch
+ADD kythe/go/extractors/config/preprocessor/preprocessor preprocessor
+ENTRYPOINT ["/preprocessor"]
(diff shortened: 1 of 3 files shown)


kythe/go/extractors/config/preprocessor/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM scratch
```
danielmoy-google: this is definitely up for changing later, for now just want to get something up there that I can test with

In particular I'm not so keen on this approach because it'll probably rely on mounting volumes to pass input/output file back and forth.  I'm still a bit fuzzy on how cloud build handles that.

danielmoy-google: actually, it turns out this just works as-is in cloud build.

I was wrong, custom volume mapping logic will not be necessary here.

PR #3174, opened 2018-10-19
style: add standard scripts for git hooks

schroederc:
Initially, the `pre-commit` hook runs the legacy arcanist unit/lint
commands (see #3173) and the `commit-msg` hook runs `commitlint` to
check for the Conventional Commit format.

See http://go/gh/marionebl/commitlint#getting-started for the new
`commitlint` tool dependency.


tools/arc/linter.sh
```diff
@@ -47,6 +50,20 @@ case $name in
       shellcheck -f json "$file" | \
         jq -r '.[] | "shellcheck::" + (if .level == "info" then "advice" else .level end) + ":" + (.line | tostring) + " " + .message'
     fi ;;
+  *.java)
+    if command -v google-java-format &>/dev/null; then
+      google-java-format -n "$file" | sed 's/^/google-java-format::error:1 /'
+    fi ;;
+  *.go)
+    if command -v gofmt &>/dev/null; then
```
salguarnieri: If they don't have the formatter installed, can we output a warning message instead of silently skipping the formatting step?

schroederc: We can do that per-file but not once per `arc lint` invocation.

PR #3184, opened 2018-10-22
fix: Adding error message to git hook when commitlint not installed.

danielmoy-google:
Unclear if we'd want this somewhere else.

### DIFF
--- a/tools/git/commit-msg.sh
+++ b/tools/git/commit-msg.sh
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 # Check commit message: https://www.conventionalcommits.org/en/v1.0.0-beta.2/
-if ! which commitlint; then
+if ! which commitlint &>/dev/null; then
   echo "FAILURE: commitlint not installed"
   exit 1
 fi
-command -v commitlint &>/dev/null && commitlint --edit "$1"
+commitlint --edit "$1"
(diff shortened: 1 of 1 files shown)


tools/git/commit-msg.sh
```diff
@@ -15,4 +15,8 @@
 # limitations under the License.
 
 # Check commit message: https://www.conventionalcommits.org/en/v1.0.0-beta.2/
+if ! which commitlint; then
+  echo "FAILURE: commitlint not installed"
+  exit 1
+fi
```
schroederc: There's now no need for the `command -v commitlint &>/dev/null` below.

tools/git/commit-msg.sh
```diff
@@ -15,4 +15,8 @@
 # limitations under the License.
 
 # Check commit message: https://www.conventionalcommits.org/en/v1.0.0-beta.2/
+if ! which commitlint; then
```
schroederc: Pipe stdout/stderr to /dev/null: `! which commitlint &>/dev/null`

tools/git/commit-msg.sh
```diff
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 # Check commit message: https://www.conventionalcommits.org/en/v1.0.0-beta.2/
-if ! which commitlint; then
+if ! which commitlint &>/dev/null; then
   echo "FAILURE: commitlint not installed"
```
schroederc: Sorry for the late comment, but this should probably go to stderr: `>&2`

danielmoy-google: done

PR #3190, opened 2018-10-26
feat: add columnar encoding for edges api

schroederc:
Part of #2930


kythe/go/serving/graph/columnar/columnar_encoding.go
```diff
@@ -0,0 +1,177 @@
...
+	val, err := proto.Marshal(idx)
+	if err != nil {
+		return nil, err
+	}
+	return &KV{key, val}, nil
+}
+
+func encodeEdge(prefix []byte, src *spb.VName, e *gspb.Edges_Edge) (*KV, error) {
+	key, err := keys.Append(prefix, src, columnarEdgesEdgeGroup,
+		e.GetGenericKind(), int32(e.GetKytheKind()), e.Ordinal, e.Reverse, e.Target)
```
salguarnieri: What is the type of e.GetKytheKind() that we have to cast it to int32? If it is int64 why don't we just store the int64?

kythe/go/serving/pipeline/beam.go
```diff
@@ -596,12 +642,24 @@ func nodeToEdges(n *scpb.Node, emit func(*spb.VName, *scpb.Edge)) {
...
 // reverseEdge emits the reverse of each *scpb.Edge, embedding the associated TargetNode.
 func reverseEdge(src *spb.VName, nodeStream func(**scpb.Node) bool, edgeStream func(**scpb.Edge) bool, emit func(*spb.VName, *scpb.Edge)) {
 	var node *scpb.Node
 	if !nodeStream(&node) {
-		node = &scpb.Node{Source: src}
+		node = &scpb.Node{}
+	} else {
+		node = nodeWithoutEdges(node)
 	}
+	node.Source = src
```
salguarnieri: why do we always set node.Source now?

kythe/go/serving/graph/columnar.go
```diff
@@ -0,0 +1,255 @@
...
+// Edges implements part of the graph.Service interface.
+func (c *ColumnarTable) Edges(ctx context.Context, req *gpb.EdgesRequest) (*gpb.EdgesReply, error) {
+	// TODO(schroederc): implement edge paging
+	reply := &gpb.EdgesReply{
+		EdgeSets: make(map[string]*gpb.EdgeSet, len(req.Ticket)),
+		Nodes:    make(map[string]*cpb.NodeInfo),
+
+		// TODO(schroederc): TotalEdgesByKind: make(map[string]int64),
+	}
+	patterns := xrefs.ConvertFilters(req.Filter)
```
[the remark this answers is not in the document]
schroederc: The default is different between the methods.  Empty filter in Nodes => all facts.  Empty filter in Edges => no facts.

kythe/go/serving/graph/columnar/columnar_encoding.go
schroederc: It's an enum: https://github.com/google/kythe/blob/fe5b532f75f08e1847483b566b6302da1a7a4ab0/kythe/proto/schema_go_proto/schema.pb.go#L262.  It must be cast for the `keys` package .

kythe/go/serving/pipeline/beam.go
schroederc: Nodes passed through this will not have their `Source` set since it was pulled out in the input `KV`'s key.

PR #3195, opened 2018-10-29
build: update Apache Beam to commit cdd3458

schroederc:


schroederc:
PTAL; I missed the go.mod change.

PR #3201, opened 2018-10-30
feat: limit disksort memory by custom Size method

schroederc:


kythe/go/util/disksort/disksort.go
```diff
@@ -160,7 +177,11 @@ func (m *mergeSorter) Add(i interface{}) error {
 	}
 
 	m.buffer = append(m.buffer, i)
-	if len(m.buffer) >= m.opts.MaxInMemory {
+	if sizer, ok := i.(sizer); ok {
```
danielmoy-google: Just out of curiosity, when would this not trigger?  (And would the append be hidden from bufferSize in that case?)

schroederc: It would not trigger if the element does not implement the `Size() int` method.  It's size would not be counted in `m.bufferSize`.  Generally, this means **no** elements will be counted by size since the sorter assumes a homogeneous collection of elements.

danielmoy-google: Right I meant in practice, is that going to happen (would we want to log that case?).  Or is it never supposed to really happen (just ignore it)?

schroederc: It's never going to happen in a local Beam pipeline and it's always going to happen in the legacy pipeline (where we don't care).  There's no need to log it.

danielmoy-google: ok awesome :+1:

PR #3202, opened 2018-10-30
refactor: share common proto comparison logic

schroederc:


schroederc:
The other two don't compare protos.

danielmoy-google:
> The other two don't compare protos.

ah, indeed

PR #3203, opened 2018-10-31
retain a reference to zip source on validation error

shahms:
The reference counting semantics of libzip object are such that it is the caller's responsibility to free constructor objects on error.  Retain a reference to the source in `KzipReader::FromSource` when validation fails so callers can do this reliably.


shahms:
PTAL (I added tests).

PR #3206, opened 2018-10-31
feat: add proto for configuring kythe extraction on a repo

danielmoy-google:
I am not heavily sold on the exact contents of this proto.  I'll discuss it with Andrey later this week or early next week too, since probably we'll end up sharing a single config at some point.

But, getting something in lets me write code based off of it.


kythe/proto/BUILD
```diff
@@ -19,6 +19,7 @@ filegroup(
         "graph.proto",
         "identifier.proto",
         "java.proto",
+        "repo.proto",
```
schroederc: Don't make this public yet.

kythe/proto/repo.proto
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// A single target in this repo to extract.
+message BuildTarget {
+  BuildSystem build_system = 1;
+  // The root of this extraction target, relative to the repo's root.
+  // The build config (e.g. pom.xml) should live in this directory.
+  string root = 2;
+
+  JavaOptions java_options = 3;
```
schroederc: We probably want the options per language to be highly extensible so maybe this should be a repeated Any field.

kythe/proto/BUILD
danielmoy-google: :+1:

kythe/proto/repo.proto
danielmoy-google: offline: current rough plan is to go the other way.  Make things very coarse here, but then allow a given repo to fully specify a kytheextract.yaml or something if they need extremely custom behavior.

Anyways, some changes here.

kythe/proto/repo.proto
```diff
@@ -0,0 +1,52 @@
...
+enum BuildSystem {
+  UNKNOWN = 0;
+
+  MAVEN = 1;
+  GRADLE = 2;
+  BAZEL = 3;
+}
+
+message JavaOptions {
+  enum Version {
```
jaysachs: Making this an enum is interesting. I understand the motivation for it, and will strive to be optimistic that it sticks this way for all the versions we need to identify.

kythe/proto/repo.proto
```diff
@@ -0,0 +1,52 @@
...
+
+package kythe.proto.repo;
+
+// Represents configuration data on how to extract a repo.
+message Config {
+  repeated ExtractionHints extraction = 1;
+}
+
+// Settings specific to extracting a single root and build system in this repo.
+message ExtractionHints {
```
jaysachs: Maybe singular, since it applies to a single root?

kythe/proto/repo.proto
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package kythe.proto.repo;
+
+// Represents configuration data on how to extract a repo.
+message Config {
+  repeated ExtractionHints extraction = 1;
```
jaysachs: Standard style is for repeated fields to be plural.

danielmoy-google: Ah yea, had this singular before and never remembered to update the name.  Thanks :)

kythe/proto/repo.proto
danielmoy-google: :+1:

kythe/proto/repo.proto
danielmoy-google: Yup, and we can always change it later.

kythe/proto/repo.proto
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package kythe.proto.repo;
+
+// Represents configuration data on how to extract a repo.
+message Config {
+  repeated ExtractionHint extractions = 1;
```
astroilov: Is this repeated because there could be multiple ways to extract the same source, because there are multiple ExtractionHints needed for one extraction for the same source, or because there could be multiple source roots in the same repo?

schroederc:
As a concrete example, can you update the `%workspace%/.kythe` file we currently have to be this expected format?

kythe/proto/repo.proto
danielmoy-google: Sort of in between?  In particular, I want some flexibility on how to handle one repo that has either multiple source roots (e.g. separate maven sub parts), or even multiple source roots with different builders (I've seen maven+gradle, but I also imagine backend+frontend will be a common theme in the future).

On the other hand, I could be easily convinced to make this not repeated for the short term (in fact the code I have ready and waiting explicitly only supports size=1 right now).

danielmoy-google:
> As a concrete example, can you update the `%workspace%/.kythe` file we currently have to be this expected format?

Yes, once I have bazel working on cloudbuild and have successfully extracted kythe on it, I'd definitely want to do that.  Opened #3210 for that.

PR #3207, opened 2018-10-31
refactor: make beamio.EncodeKeyValues a reusable Beam transform

schroederc:


kythe/go/serving/pipeline/beamio/beamio.go
```diff
@@ -0,0 +1,71 @@
...
+	for _, table := range tables {
+		encoded := beam.ParDo(s, &encodeKeyValue{beam.EncodedCoder{table.Coder()}}, table)
+		encodings = append(encodings, encoded)
+	}
+	return beam.Flatten(s, encodings...)
+}
+
+type encodeKeyValue struct{ Coder beam.EncodedCoder }
+
+func (e *encodeKeyValue) ProcessElement(key beam.T, val beam.U) (KeyValue, error) {
```
salguarnieri: Add doc comment

schroederc: I'm going to push back on this.  Exported methods should _generally_ be documented but when it comes to Beam, this would cause of lot of redundant comments.  This method **is required** to be exported by the Beam framework, but its receiver is not exported because it is an implementation detail (like an unexported function).  This ParDo is already commented through the exported `EncodeKeyValues` PTransform above.

salguarnieri: sgtm

PR #3209, opened 2018-10-31
fix: modernize Bazel end-to-end example scripts

### LINKED ISSUE #3089: Documentation on how to set up Kythe in a Bazel repo is out of date
https://github.com/google/kythe/blob/master/kythe/examples/bazel/setup-bazel-repo.sh appears to not reflect the current state of the art WRT setting up Kythe in a Bazel repo. Is there any more recent documentation that should be preferred?

schroederc:
Extract Bazel using Kythe as an external repository instead from a binary release.

Fixes #3089


schroederc:
Probably not worth it right now.  It's not the most stable example.

PR #3219, opened 2018-11-05
feat: adding generator for kythe maven/gradle based on config proto

danielmoy-google:
Reading config proto in json format, then translating to cloudbuild objects, and serializing out to yaml.


kythe/go/extractors/constants/constants.go
```diff
@@ -36,19 +36,43 @@ var (
...
+	KytheBuildPreprocessorImage = "gcr.io/kythe-public/build-preprocessor"
+
 	// DefaultJavacLocation points to a common location for a javac binary.
 	// The binary will usually be symlinked here.
 	DefaultJavacLocation = "/usr/bin/javac"
 
+	// DefaultExtractorsDir is the canonical directory for extractors that any
+	// kythe docker image will use.  This can be useful if you need to load that
+	// volume to get access to kythe extractors.
+	DefaultExtractorsDir = "/opt/kythe/extractors"
```
jaysachs: What is the organization principle for the ordering in this file?

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// Constants ephemeral to a single kythe cloudbuild run.
+const (
+	outputDirectory = "/workspace/out"
+	codeDirectory   = "/workspace/code"
+	javaVolumeName  = "kythe_extractors"
+)
+
+// KytheToYAML takes an input file, parses it as a JSON defined
```
jaysachs: Primary doc is for the caller. The intermediate JSON step is just an implementation detail -- comment that inside the function.

kythe/go/extractors/gcp/config/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var (
+	input  = flag.String("input", "", "The input config proto to read.")
+	output = flag.String("output", "cloudbuild.yaml", "The output yaml file to write.")
+)
+
+func main() {
+	flag.Parse()
+
```
schroederc: Check flags are non-empty.

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+	initializeArtifacts(&build)
+	build.Artifacts.Objects.Location = fmt.Sprintf("gs://%s/", outputGsBucket)
+	build.Artifacts.Objects.Paths = append(build.Artifacts.Objects.Paths, path.Join(outputDirectory, outputFilePattern))
+	build.Steps = append(build.Steps, commonSteps()...)
+
+	if len(conf.Extractions) == 0 {
+		return build, fmt.Errorf("config has no extraction specified")
+	} else if len(conf.Extractions) > 1 {
+		return build, fmt.Errorf("we don't yet support multiple extraction in a single repo")
+	}
```
jaysachs: Maybe make these checks up front?

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+
+	hints := conf.Extractions[0]
+	bs := hints.BuildSystem
```
jaysachs: And unfortunate abbreviation.

jaysachs: Is it even worth capturing?

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, fmt.Errorf("marshalling cloudbuild.Build to JSON: %v", err)
+	}
+	return yaml.JSONToYAML(json)
+}
+
+// KytheToBuild takes a kythe.proto.extraction.RepoConfig and generates the
+// necessary cloudbuild.Build with BuildSteps for running kythe extraction.
+func KytheToBuild(conf rpb.Config) (cloudbuild.Build, error) {
+	build := cloudbuild.Build{}
+	initializeArtifacts(&build)
```
jaysachs: I'm not go-savvy, but why create it just to initialize rather than having a function return a pointer to a new one?

jaysachs: (already initialized)

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"fmt"
+	"os"
+	"path"
+
+	rpb "kythe.io/kythe/proto/repo_go_proto"
+
+	"github.com/ghodss/yaml"
+	"github.com/golang/protobuf/jsonpb"
+	cloudbuild "google.golang.org/api/cloudbuild/v1"
```
schroederc: Don't rename import; it's already named `cloudbuild`.

kythe/go/extractors/gcp/config/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,45 @@
...
+	yamlData, err := config.KytheToYAML(*input)
+	if err != nil {
+		log.Fatalf("failure converting %s to %s: %v", *input, *output, err)
+	}
+
+	if err := ioutil.WriteFile(*output, yamlData, 0644); err != nil {
```
jaysachs: Does this take empty or "-" as meaning stdout? and similarly for the call to KytheToYAML?

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Constants ephemeral to a single kythe cloudbuild run.
+const (
+	outputDirectory = "/workspace/out"
+	codeDirectory   = "/workspace/code"
+	javaVolumeName  = "kythe_extractors"
+)
+
+// KytheToYAML takes an input file, parses it as a JSON defined
+// kythe.proto.extraction.RepoConfig, and converts it to YAML.
```
schroederc: s/converts it to/returns it as marshaled/

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, fmt.Errorf("converting cloudbuild.Build: %v", err)
+	}
+	json, err := build.MarshalJSON()
+	if err != nil {
+		return nil, fmt.Errorf("marshalling cloudbuild.Build to JSON: %v", err)
+	}
+	return yaml.JSONToYAML(json)
+}
+
+// KytheToBuild takes a kythe.proto.extraction.RepoConfig and generates the
```
schroederc: s/generates/returns/

kythe/go/extractors/gcp/config/converter.go
schroederc: Trivial function used once.  Just inline:

```
build := cloudbuild.Build{
  Artifacts: &cloudbuild.Artifacts{
    Object: &cloudbuild.ArtifactObjects{
      Location: fmt.Sprintf("gs://%s/", outputGsBucket),
      // etc.
    },
  },
}
```

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+	return build, nil
+}
+
+func readConfigFile(input string) (rpb.Config, error) {
+	var conf rpb.Config
+	file, err := os.Open(input)
+	if err != nil {
+		return conf, fmt.Errorf("opening input file %s: %v", input, err)
+	}
+	if err := jsonpb.Unmarshal(file, &conf); err != nil {
```
schroederc: `defer file.Close()`

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+	hints := conf.Extractions[0]
+	bs := hints.BuildSystem
+
+	// TODO(danielmoy): when we support bazel, we probably want to pull out the
+	// switch cases here into something a bit easier to read.
+	// I imagine something like a proper interface that each build system can
+	// override, which would provide things like artifactPaths(), steps().
+	switch bs {
+	case rpb.BuildSystem_MAVEN:
+		build.Steps = append(build.Steps, javaArtifactsStep())
```
schroederc: You can join the appends together:

```
build.Steps = append(build.Steps,
  javaArtifactsStep(),
  mavenStep(hints),
)
```

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+		build.Steps = append(build.Steps, mavenStep(hints))
+		build.Artifacts.Objects.Paths = append(build.Artifacts.Objects.Paths, path.Join(outputDirectory, "javac-extractor.err"))
+	case rpb.BuildSystem_GRADLE:
+		//build.Steps = append(build.Steps, GradleSteps())
+		build.Artifacts.Objects.Paths = append(build.Artifacts.Objects.Paths, path.Join(outputDirectory, "javac-extractor.err"))
+	default:
+		return build, fmt.Errorf("unsupported build system %s", bs)
+	}
+
+	return build, nil
```
schroederc: Return a pointer to the protocol buffer.

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -0,0 +1,119 @@
...
+func readConfigFile(input string) (rpb.Config, error) {
+	var conf rpb.Config
+	file, err := os.Open(input)
+	if err != nil {
+		return conf, fmt.Errorf("opening input file %s: %v", input, err)
+	}
+	if err := jsonpb.Unmarshal(file, &conf); err != nil {
+		return conf, fmt.Errorf("parsing json file %s: %v", input, err)
+	}
+	return conf, nil
```
schroederc: Return pointers to protocol buffers.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/DataDog/zstd v0.0.0-20180321192404-aebefd9fcb99
 	github.com/apache/beam v0.0.0-20181024230410-cdd3458380a6
 	github.com/beevik/etree v0.0.0-20180609112216-9d7e8feddccb
+	github.com/ghodss/yaml v1.0.0
```
schroederc: This should be locked down to the commit used by Bazel.

kythe/go/extractors/gcp/config/converter.go
creachadair: +1

kythe/go/extractors/gcp/config/yamlgen/yamlgen.go
creachadair: Re: writing to stdout. There's no magic pathname, but you could write to /dev/stdout probably.

schroederc: `/dev/stdout` does work, but "-" has to be handled specially.  Some of this is handled in our vfs package: https://github.com/kythe/kythe/blob/master/kythe/go/platform/vfs/vfs.go#L133

go.mod
danielmoy-google: :+1:

go.mod
```diff
@@ -33,4 +34,5 @@ require (
 	google.golang.org/api v0.0.0-20180404000327-3097bf831ede
 	google.golang.org/genproto v0.0.0-20180716172848-2731d4fa720b // indirect
 	google.golang.org/grpc v0.0.0-20180510165436-d07538b1475e
+	gopkg.in/yaml.v2 v2.2.1-20180328165020-5420a8b6744d // indirect
```
schroederc: This is invalid.  This script should work:

```
git checkout master -- go.mod
rm -f go.sum
go mod edit --require=github.com/ghodss/yaml@0ca9ea5df5451ffdf184b4428c902747c2c11cd7 --require=gopkg.in/yaml.v2@5420a8b6744d3b0345ab293f6fcba19c978f1183
go test ./kythe/...
```

It'll leave the yaml version floating, but that's probably fine for now.

danielmoy-google: That's almost exactly what I did in 43ca'

https://github.com/kythe/kythe/pull/3219/commits/43ca290300c4a073dfb080f37c85de54d453f45f#diff-37aff102a57d3d7b797f152915a6dc16

I am starting to suspect I have another transitive dep missing, because the go test ./kythe/... step clobbers out the commit hash from go.mod

danielmoy-google: I couldn't find such a missing dep yesterday, and the only other idea I had was maybe my branch started too far out of date.

danielmoy-google:
My current line of thinking is that this is actually not a problem




Because I picked out a commit from the actual release branch that has a proper version tag thing (semvar?), it doesn't need the hash in there.  I think what's happening is that go test ./kythe/... is resolving the commit hash to the bona fide version, and then dropping the commit hash.

For example, if I just run `go mod edit .blahblah..` without `go test./kythe/...`, I do see a commit hash there.  But after `go test` runs, the commit hash is gone and replaced with a version number.

The rest of our deps are using pseudo versions: https://tip.golang.org/cmd/go/#hdr-Pseudo_versions

which is why they appear different.

I dunno

My understanding of this is still shaky at best.

creachadair:
Do you have `GO111MODULE` set in your environment? That can in some cases affect how the tools treat modules.

go.sum
```diff
@@ -63,5 +65,7 @@ google.golang.org/genproto v0.0.0-20180716172848-2731d4fa720b/go.mod h1:JiN7NxoA
 google.golang.org/grpc v0.0.0-20180510165436-d07538b1475e h1:X6P5OjvrKe7dZIhFpqU7P6JWZVwnSvT8bB0iRauPo3I=
 google.golang.org/grpc v0.0.0-20180510165436-d07538b1475e/go.mod h1:yo6s7OP7yaDglbqo1J04qKzAhqBH6lvTonzMVmEdcZw=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v2 v2.2.1-0.20180326061214-4fc5987536ef h1:KngkwkexysX8F+cPblZfFFu3eAF/xsJJej9Jt93h3qI=
+gopkg.in/yaml.v2 v2.2.1-0.20180326061214-4fc5987536ef/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.1 h1:mUhvW9EsL+naU5Q3cakzfE91YhliOondGd6ZrsDBHQE=
```
schroederc: Remove the old lines: `rm -f go.sum; go test ./kythe/...`

danielmoy-google:
> Do you have `GO111MODULE` set in your environment? That can in some cases affect how the tools treat modules.

Ah no, this is the first I've ever heard of GO111MODULE env var.

go.sum
danielmoy-google: done, thanks

danielmoy-google:
> It's probably worthwhile to read through https://github.com/golang/go/wiki/Modules since we're now using modules. There are still some rough edges. Given that you _don't_ have it set, I wouldn't expect any problems from the default behavour, but I admit I'm not completely familiar with all the corners yet either.

Right I think the issue we are running in to here is that go modules that import a specific release version with a semvar tag have go mod edit / go test drop the commit hash from the go.mod file.

If there's some syntax to have it always use a [pseudo version](https://tip.golang.org/cmd/go/#hdr-Pseudo_versions) maybe that'd make everyone happy?

PR #3226, opened 2018-11-06
feat: Support adding Clang USRs to the graph

zrlk:
Clang tools use USRs ("Unified Symbol Resolution") as ticket-equivalents
(but they're insufficiently specific for Kythe). This PR adds a new flag,
--experimental_usr_byte_size, which (if given a nonzero value) instructs
the indexer to take that many bytes off the SHA1 hash of USRs and
add them as clang/usr nodes with /clang/usr edges pointing at the
C++ nodes to which they refer:

    //- @foo defines/binding Foo
    //- FooUsr /clang/usr Foo
    //- FooUsr=vname(_,"","","","usr").node/kind clang/usr
    void foo() { }

Similar to the way we deal with JVM names, the corpus, path,
and root fields of a usr vname are cleared. Clients are permitted
to write their own USR tickets. The USR value itself is encoded
in capital hex (to match Clang's own internal USR stringification,
modulo the configurable size of the SHA1 prefix).

USRs are added only for NamedDecls that:
  * are FunctionDecls
  * are not under implicit template instantiations
  * are not in a DeclContext inside a function body
  * can actually be assigned USRs from Clang

Future PRs may allow various other NamedDecl subtypes to be granted
USRs, but it is unlikely that we will support assigning USRs to
decls that are not at TU/namespace scope.

### DIFF
--- a/kythe/cxx/common/indexing/KytheGraphRecorder.cc
+++ b/kythe/cxx/common/indexing/KytheGraphRecorder.cc
@@ -31,8 +31,7 @@ static const std::string* const kNodeKindSpellings[] = {
     new std::string("interface"),  new std::string("package"),
     new std::string("tsigma"),     new std::string("doc"),
     new std::string("tbuiltin"),   new std::string("meta"),
-    new std::string("diagnostic"),
-};
+    new std::string("diagnostic"), new std::string("clang/usr")};
 
 static const std::string* kEdgeKindSpellings[] = {
     new std::string("/kythe/edge/defines"),
@@ -77,6 +76,7 @@ static const std::string* kEdgeKindSpellings[] = {
     new std::string("/kythe/edge/ref/init/implicit"),
     new std::string("/kythe/edge/imputes"),
     new std::string("/kythe/edge/tagged"),
+    new std::string("/clang/usr"),
 };
 
 bool of_spelling(absl::string_view str, EdgeKindID* edge_id) {
--- a/kythe/cxx/common/indexing/KytheGraphRecorder.h
+++ b/kythe/cxx/common/indexing/KytheGraphRecorder.h
@@ -45,7 +45,8 @@ enum class NodeKindID {
   kDoc,
   kTBuiltin,
   kMeta,
-  kDiagnostic
+  kDiagnostic,
+  kUsr
 };
 
 /// \brief Known properties of nodes. See the schema for details.
@@ -115,7 +116,8 @@ enum class EdgeKindID {
   kRefInit,
   kRefInitImplicit,
   kImputes,
-  kTagged
+  kTagged,
+  kUsr
 };
 
 /// \brief Returns the Kythe spelling of `node_kind_id`
--- a/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
+++ b/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
@@ -43,6 +43,7 @@
 #include "clang/AST/TemplateName.h"
 #include "clang/AST/Type.h"
 #include "clang/AST/TypeLoc.h"
+#include "clang/Index/USRGeneration.h"
 #include "clang/Lex/Lexer.h"
 #include "clang/Sema/Lookup.h"
 #include "gflags/gflags.h"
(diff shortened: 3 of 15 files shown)


zrlk:
Note that the bot needs to rebuild Clang, or be ordered to do a `cmake --build . -- "-jNN" clangIndex` in the llvm build directory. We could do this with a version bump.

schroederc:
There's no way to tell Buildbot to rebuild LLVM due to a configuration change yet.  Updating the version with the config change would work, though.

kythe/cxx/common/indexing/KytheGraphRecorder.cc
```diff
@@ -31,8 +31,7 @@ static const std::string* const kNodeKindSpellings[] = {
     new std::string("interface"),  new std::string("package"),
     new std::string("tsigma"),     new std::string("doc"),
     new std::string("tbuiltin"),   new std::string("meta"),
-    new std::string("diagnostic"),
-};
+    new std::string("diagnostic"), new std::string("clang/usr")};
```
salguarnieri: Is the } on the same line what clang format wanted?

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -3259,6 +3262,16 @@ bool IndexerASTVisitor::AddNameToStream(llvm::raw_string_ostream& Ostream,
   return true;
 }
 
+void IndexerASTVisitor::AssignUSR(const GraphObserver::NodeId& TargetNode,
```
salguarnieri: Can you add docs (here or in the .h) that specify when we do and when we don't assign usr? I'm pretty sure you already have the text in the PR description.

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
```diff
@@ -870,6 +872,20 @@ void KytheGraphObserver::recordFunctionNode(
   }
 }
 
+void KytheGraphObserver::assignUsr(const NodeId& node, llvm::StringRef usr,
+                                   int byte_size) {
+  usr = usr.take_front(byte_size);
```
salguarnieri: What happens if byte_size is larger than the usr?

kythe/cxx/common/indexing/KytheGraphRecorder.cc
zrlk: Yes.

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
zrlk: As many bytes as are available are returned.

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
zrlk: Done

salguarnieri:
I'll have to do another approval once you update the branch and make the buildbot happy

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
```diff
@@ -870,6 +872,20 @@ void KytheGraphObserver::recordFunctionNode(
...
+  auto hash = llvm::SHA1::hash(llvm::arrayRefFromStringRef(usr));
```
hokein: No sure it is intended.
USR length is not fixed, and could be very long for functions with long type arguments.

`SHA1` value is always 20 bytes, truncating the USR doesn't help to reduce the length of the tickets, I think a more reasonable way is to truncate the hash value, which is also what clangd index currently does.

kythe/cxx/common/indexing/KytheGraphRecorder.h
```diff
@@ -45,7 +45,8 @@ enum class NodeKindID {
   kDoc,
   kTBuiltin,
   kMeta,
-  kDiagnostic
+  kDiagnostic,
+  kUsr
```
shahms: minor: I think `kClangUsr` makes it clear that this is in a distinct namespace, particular given the potential confusion around the 'Usr' abbreviation.

kythe/cxx/common/indexing/KytheGraphRecorder.h
```diff
@@ -115,7 +116,8 @@ enum class EdgeKindID {
   kRefInit,
   kRefInitImplicit,
   kImputes,
-  kTagged
+  kTagged,
+  kUsr
```
shahms: ditto

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
shahms: minor: IMO, the `byte_size` argument is redundant here; this is kind of the purpose of StringRef/string_view classes and I'd rather have the caller make this adjustment if necessary.

PR #3231, opened 2018-11-07
test: move pre-commit hook to pre-push

schroederc:


salguarnieri:
pre-push fits in better with the general PR flow where you can keep you own local workflow and then fix everything up before you make it public via a PR.

danielmoy-google:
> pre-push fits in better with the general PR flow where you can keep you own local workflow and then fix everything up before you make it public via a PR.

Ah that's a good point.

PR #3232, opened 2018-11-08
feat(cxx_indexer): read all compilations from a kzip file (#3120)

justbuchanan:
See #3120.


kythe/cxx/common/indexing/frontend.cc
```diff
@@ -137,18 +137,21 @@ void DecodeIndexFile(const std::string& path,
...
+/// \param silent The silent flag is copied to each of the jobs created from the
+/// kzip file.
+void DecodeKZipFile(const std::string& path, std::vector<IndexerJob>* jobs,
+                    bool silent) {
   StatusOr<IndexReader> reader = kythe::KzipReader::Open(path);
   CHECK(reader) << "Couldn't open kzip from " << path;
   bool compilation_read = false;
   auto status = reader->Scan([&](absl::string_view digest) {
-    CHECK(!compilation_read) << "Found more than 1 compilation in " << path;
+    jobs->emplace_back();
```
schroederc: Is it easy/possible for the indexer to only have a single compilation unit in memory at a time?  If not, this should probably be a `TODO`.

justbuchanan: Should definitely be possible, but probably would require a fair amount of refactoring. Right now the `IndexerFrontend` builds a list of "jobs" from the specified files/units, then the list is later consumed by the indexers.

The other languages just call `visit()` for each compilation unit found, which seems like a good way to avoid storing all units at once during the process.

PR #3003, opened 2018-09-12
Add test for compilation database extraction

shahms:

PR #3236, opened 2018-11-08
fix(docs): fixed a few broken links

justbuchanan:
There's also a link in the "Build System Integration" section of kythe/web/site/contributing.md that points to some cmake files that seem to have moved. I'm not sure where that code lives now.

### DIFF
--- a/kythe/docs/exploring-kythe-sample-web-ui.txt
+++ b/kythe/docs/exploring-kythe-sample-web-ui.txt
@@ -107,6 +107,6 @@ http://www.kythe.io/docs/schema/#param[/kythe/edge/param]:: declared parameters
 
 These examples are just a taste of the data exposed in our sample web UI.  With
 other specialized interfaces such as the
-https://kythe.io/repo/kythe/go/serving/tools/kythe.go[Kythe command-line],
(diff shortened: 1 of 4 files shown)


danielmoy-google:
> I'm not sure where that code lives now.

Ah I think it's still in flux.  There's some bits in kythe/go/extractors/config/runextractor/cmake now.  But *that* is also in flux and might end up in kythe/go/extractors/gcp at some point soon (maybe? if we can get it working?).

shahms might have another idea where to point there.

In any event, it'll be on us to update that section of the docs once the gcp extractor stuff is settled more.

danielmoy-google:
> > I'm not sure where that code lives now.
> 
> Ah I think it's still in flux. There's some bits in kythe/go/extractors/config/runextractor/cmake now. But _that_ is also in flux and might end up in kythe/go/extractors/gcp at some point soon (maybe? if we can get it working?).
> 
> shahms might have another idea where to point there.
> 
> In any event, it'll be on us to update that section of the docs once the gcp extractor stuff is settled more.

Was removed in #3003 if you're curious

justbuchanan:
Thanks for the pointers. I'll leave that one alone for now and we can come back and fix it once that gets figured out.

Issue #945, opened 2018-06-25
Remove LLVM dependencies from "common" C++ code [phab:T91]


kythebot:
_Created by **mazurak** at 2016-01-21 20:42:17_: 

There are a bunch of little dependencies on LLVM within the kythe/cxx/common (e.g., StringRefs), even though nothing says that a language whose indexer is written in C++ necessarily cares about LLVM at all. It would be nice if this dependency could be broken.

kythe:
_**schroederc** wrote at 2017-05-31 23:09:43_:

I believe this is fixed, right?

kythe:
_**zarko** wrote at 2017-05-31 23:14:50_:

It's not. The biggest blocker is the lack of `std::string_view`. We also use some of LLVM's path-manipulation APIs.

schroederc:
Rough list of targets from which we want to prune LLVM:

```
//kythe/cxx/common:file_vname_generator_test
//kythe/cxx/common:file_vname_generator_testlib
//kythe/cxx/common/indexing:caching_output
//kythe/cxx/common:index_pack
//kythe/cxx/common:index_pack_test
//kythe/cxx/common:index_pack_testlib
//kythe/cxx/common:lib (cxx_details probably needs to be pulled out)
//kythe/cxx/common:net_client
//kythe/cxx/common:net_client_test
//kythe/cxx/common:net_client_testlib
//kythe/cxx/common:path_utils
//kythe/cxx/common:path_utils_test
//kythe/cxx/common:path_utils_testlib
//kythe/cxx/common/testdata:net_client_test_runner
//kythe/cxx/doc:doc
//kythe/cxx/doc:doclib
//kythe/cxx/doc:html_renderer
//kythe/cxx/doc:html_renderer_test
//kythe/cxx/doc:html_renderer_testlib
```

These targets on the ones with direct dependencies on LLVM:

```
//kythe/cxx/common/indexing:caching_output
//kythe/cxx/common:index_pack_testlib
//kythe/cxx/common:lib
//kythe/cxx/common:path_utils_testlib
```

justbuchanan:
I'm going to work on removing llvm from cxx/extractor:index_pack since it's a dependency of the in-progress proto indexer.

zrlk:
Index packs have been deprecated; it would be better to remove index pack support from the proto indexer (and add kzip support).

justbuchanan:
Sounds good, I'll do that instead.

justbuchanan:
[...]

PR #3237, opened 2018-11-08
chore: prise away some LLVM deps for #945

zrlk:
This PR drops some of the easy deps.


kythe/cxx/common/BUILD
```diff
@@ -43,7 +43,6 @@ cc_library(
     deprecation = "Please use :kzip_{reader,writer}",
     visibility = [
```
schroederc: Remove visibility attribute.

PR #3241, opened 2018-11-09
Revert "build: remove unused build targets (#3240)"

schroederc:
This reverts commit 4e52a30a6479df904fe8e8dfafac67953a721e00.

The dependency is only used by license because we use the project
through an external repo (com_google_common_html_types).

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -26,6 +26,7 @@ filegroup(
         "//third_party/re2:LICENSE",
         "//third_party/re2j:LICENSE",
         "//third_party/riegeli:LICENSE",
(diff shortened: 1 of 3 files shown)


third_party/BUILD
```diff
@@ -26,6 +26,7 @@ filegroup(
         "//third_party/re2:LICENSE",
         "//third_party/re2j:LICENSE",
         "//third_party/riegeli:LICENSE",
+        "//third_party/safe_html_types:LICENSE",
```
salguarnieri: Can you add a comment saying why we need this so we don't try to delete it again?

schroederc: PTAL

PR #3242, opened 2018-11-09
chore: remove more LLVM from common (#945)

zrlk:


kythe/cxx/common/indexing/KytheCachingOutput.cc
```diff
@@ -76,8 +75,7 @@ bool MemcachedHashCache::SawHash(const Hash& hash) {
 }
 
 std::string FileOutputStream::Stats::ToString() const {
-  std::string out;
-  llvm::raw_string_ostream ostream(out);
+  std::stringstream ostream;
```
shahms: This is fine, but you might also consider:
```
 return absl::StrCat(...);
```

PR #3243, opened 2018-11-09
chore: remove the remaining llvm references from cxx/common (#945)

zrlk:


kythe/cxx/common/kythe_metadata_file.cc
```diff
@@ -210,9 +212,9 @@ bool KytheMetadataSupport::LoadMetaElement(const rapidjson::Value& value,
 #undef JSON_SAFE_LOAD_STRING
 
 std::unique_ptr<MetadataFile> KytheMetadataSupport::LoadFromJSON(
-    llvm::StringRef json) {
+    absl::string_view json) {
   rapidjson::Document document;
-  document.Parse(json.str().c_str());
+  document.Parse(std::string(json).c_str());
```
shahms: Does this accept (const char*, size) as well? Or just null-terminated string?

shahms: https://github.com/Tencent/rapidjson/blob/master/include/rapidjson/document.h#L2342

It does! we can save this copy and conversion to string.

zrlk: rapidjson has its own custom collection of input streams we could enjoy if we choose.

zrlk: Oh, they must build their docs without `RAPIDJSON_HAS_STDSTRING` set

shahms: That I can't say, but there is a `Parse(const char*, size_t)` overload, which should be directly compatible with `json.data(), json.size()`

PR #3244, opened 2018-11-09
chore: rename header guards (#945)

zrlk:
We need a stronger tag than `chore` for changes like this

### DIFF
--- a/kythe/cxx/extractor/language.h
+++ b/kythe/cxx/extractor/language.h
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#ifndef KYTHE_CXX_EXTRACTOR_KYTHE_LANGUAGE_H_
-#define KYTHE_CXX_EXTRACTOR_KYTHE_LANGUAGE_H_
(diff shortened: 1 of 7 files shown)


danielmoy-google:
> We need a stronger tag than `chore` for changes like this

`migration:` or `refactor:` or something

kythe/cxx/extractor/language.h
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#ifndef KYTHE_CXX_COMMON_KYTHE_LANGUAGE_H_
-#define KYTHE_CXX_COMMON_KYTHE_LANGUAGE_H_
+#ifndef KYTHE_CXX_EXTRACTOR_KYTHE_LANGUAGE_H_
```
salguarnieri: No second KYTHE

zrlk: I'd prefer `toil:`

schroederc:
`refactor` is an acceptable type.

PR #3248, opened 2018-11-12
typescript: fix some lint issues

evmar:
These are flagged by the Google linter, which we don't enforce for
the Kythe project, but it seems harmless to occasionally fix these on a
case by case basis.

- Use 'const' whereever possible.
- Fix a missing semicolon by using the preferred function style.


evmar:
Sorry, missed one of the files in my first patch.

evmar:
(I don't have commit access, so please merge for me.)

Issue #3233, opened 2018-11-08
Refactor C++ frontend to avoid storing multiple compilation units in memory at a time


justbuchanan:
See discussion on #3232.

justbuchanan:
Closed by https://github.com/kythe/kythe/pull/3251.

PR #3251, opened 2018-11-13
refactor(cxx_indexer): enumerate compilations as they're read

justbuchanan:
This way there is only one compilation in memory at a time (#3233), which is more memory-efficient than building a list of all compilations and processing them all afterwards.

Previously the `IndexerContext` constructor did all the work and saved a list of jobs. Now the constructor just does some basic setup and the `EnumerateCompilations()` method handles reading files and passing compilations to the indexer callback.

### DIFF
--- a/kythe/cxx/indexer/cxx/frontend.cc
+++ b/kythe/cxx/indexer/cxx/frontend.cc
@@ -317,7 +317,6 @@ void IndexerContext::LoadDataFromUnpackedFile(
     const std::string& default_filename,
     const CompilationVisitCallback& visit) {
   IndexerJob job;
-  job.allow_filesystem_access = true;
   int read_fd = STDIN_FILENO;
   std::string source_file_name = default_filename;
   llvm::SmallString<1024> cwd;
@@ -421,6 +420,8 @@ IndexerContext::IndexerContext(const std::vector<std::string>& args,
       ignore_unimplemented_(FLAGS_ignore_unimplemented) {
   args_.erase(std::remove(args_.begin(), args_.end(), std::string()),
               args_.end());
+  // Only allow filesystem access for unpacked inputs
+  unpacked_inputs_ = !HasIndexArguments();
 
   InitializeClaimClient();
   OpenOutputStreams();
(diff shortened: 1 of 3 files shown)


kythe/cxx/indexer/cxx/frontend.h
```diff
@@ -56,14 +59,17 @@ class IndexerContext {
                  const std::string& default_filename);
   ~IndexerContext();
 
+  /// \brief A callback used when enumerating compilations.
+  typedef std::function<void(IndexerJob&)> CompilationVisitCallback;
```
shahms: Prefer `using` over `typedef` when defining new aliases.

kythe/cxx/indexer/cxx/frontend.h
```diff
@@ -56,14 +59,17 @@ class IndexerContext {
...
+  /// \brief Reads compilations from the file(s) specified in the constructor
+  /// and calls the callback for each one.
+  /// \param visit A callback to call for each compilation unit.
+  void EnumerateCompilations(CompilationVisitCallback visit);
```
shahms: minor: `const CompilationVisitCallback&`

shahms: (and below).

kythe/cxx/indexer/cxx/frontend.cc
```diff
@@ -165,7 +165,7 @@ void DecodeIndexFile(const std::string& path,
 /// \param silent The silent flag is copied to each of the jobs created from the
 /// kzip file.
 void DecodeKZipFile(const std::string& path, bool silent,
-                    IndexerContext::CompilationVisitCallback visit) {
```
shahms: `const&` ?

zrlk:
LGTM, but how would you manage to invoke the indexer where you have some jobs that can touch the local filesystem and some that can't?

justbuchanan:
Good point, filesystem access shouldn't vary between jobs. I moved that bit back to the context object.

PR #3252, opened 2018-11-13
fix: forgot test files from earlier commit, adding

danielmoy-google:


kythe/go/extractors/gcp/config/converter_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"kythe.io/kythe/go/test/testutil"
+)
+
+const testDataDir = "testdata"
+
+func TestFoo(t *testing.T) {
```
salguarnieri: Can you pick a more descriptive test name?

kythe/go/extractors/gcp/config/converter_test.go
```diff
@@ -0,0 +1,73 @@
...
+	// Do nothing intentionally
```
salguarnieri: What does this comment mean? It looks like you are testing some logic below.

danielmoy-google: ah is from earlier commit

kythe/go/extractors/gcp/config/converter_test.go
danielmoy-google: done

kythe/go/extractors/gcp/config/converter_test.go
```diff
@@ -0,0 +1,72 @@
...
+				t.Fatalf("reading config %s: %v", tcase.basename, err)
+			}
+			expected, err := tcase.getExpectedYAML(t)
+			if err != nil {
+				t.Fatalf("reading expected testdata %s: %v", tcase.basename, err)
+			}
+
+			eq, diff := testutil.TrimmedEqual(actual, expected)
+			if !eq {
+				t.Errorf("Expected config %s to be equal, but got diff %s", tcase.basename, diff)
```
salguarnieri: Is it possible to combine these three errorfs into one?

danielmoy-google: for now I'd rather leave separate because it makes fixing the test later easier.

but, filing issue for the real problem - the diff doesn't print in such a way to cleanly isolate the diff (#3253)

PR #3254, opened 2018-11-14
fix(cxx frontend): set CompilationUnit.source_file for unpacked input

justbuchanan:
When using the indexer like: `indexer -i file.cc`, sets the `source_file` field to `["file.cc"]`.

The C++ indexer seems to work fine without this field set, but other indexers do require it. Setting this field could allow for other indexers implemented in C++ to reuse `IndexerContext`. Why does C++ not require this field? Does it just read thorough all of the files in the list passed alongside the `CompilationUnit`?

### DIFF
--- a/kythe/cxx/indexer/cxx/frontend.cc
+++ b/kythe/cxx/indexer/cxx/frontend.cc
@@ -348,6 +348,7 @@ void IndexerContext::LoadDataFromUnpackedFile(
   file_data.mutable_info()->set_path(source_file_name);
   file_data.set_content(source_data.str());
   job.virtual_files.push_back(std::move(file_data));
+  job.unit.add_source_file(source_file_name);
   for (const auto& arg : args_) {
     job.unit.add_argument(arg);
   }


shahms:
ISTR that the "source file" inferred from the .cc file specified on the command line rather than being pulled from the field in the CU.

kythe/cxx/indexer/cxx/frontend.cc
```diff
@@ -348,6 +348,7 @@ void IndexerContext::LoadDataFromUnpackedFile(
   file_data.mutable_info()->set_path(source_file_name);
   file_data.set_content(source_data.str());
   job.virtual_files.push_back(std::move(file_data));
+  job.unit.add_source_file(source_file_name);
```
shahms: Ultimately, it's the job of the thing constructing the compilation unit to set the source file.  CUs coming from the extractor should have this field set appropriately, so unconditionally adding it here is most likely to result in two identical source files being present.

creachadair: That's true, but I think in this case the CLI indexer is simulating the extractor as well as indexing.

PR #3260, opened 2018-11-15
use the first token in the file as the module link

evmar:
See discussion in the commit message.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -102,12 +102,6 @@ class Vistor {
    */
   rootDirs: string[];
 
-  /**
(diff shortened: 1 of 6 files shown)


evmar:
This feels wrong to me, so your advice on how to fix it is welcomed.

zrlk:
[ref/file](https://kythe.io/docs/schema/#reffile) is probably the right thing for this (though the example in the schema doc is mismatched somehow)

evmar:
@zrlk do you know if ref/file will do the right thing in the Google-internal xref tool?

evmar:
I convinced myself that this is actually the best we can do.

PR #3261, opened 2018-11-16
fix example for expands/transitive

evmar:
The example was after a different section.


evmar:
Need someone to rebase/merge

PR #3263, opened 2018-11-18
Use rules_docker to build http_server image

achew22:
This is in addition to the existing docker rules and shouldn't interfere
with the normal release process and provides a toehold for moving more
of that logic into rules_docker.

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -7,6 +7,14 @@ load("@io_kythe//:setup.bzl", "maybe")
 load("@io_bazel_rules_go//go:def.bzl", "go_register_toolchains", "go_rules_dependencies")
 load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
 load("@build_bazel_rules_nodejs//:package.bzl", "rules_nodejs_dependencies")
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    _container_repositories = "repositories",
+)
+load(
+    "@io_bazel_rules_docker//go:image.bzl",
+    _go_image_repos = "repositories",
+)
 
 def _rule_dependencies():
     gazelle_dependencies()
@@ -609,6 +617,11 @@ def _bindings():
         actual = "@net_zlib//:zlib",
     )
 
+def _docker_dependencies():
+    _container_repositories()
+
+    _go_image_repos()
+
 def kythe_dependencies():
     """Defines external repositories for Kythe dependencies.
 
@@ -617,6 +630,7 @@ def kythe_dependencies():
     _cc_dependencies()
     _go_dependencies()
     _java_dependencies()
+    _docker_dependencies()
 
     # proto_library, cc_proto_library, and java_proto_library rules implicitly
     # depend on @com_google_protobuf for protoc and proto runtimes.
--- a/kythe/go/serving/tools/http_server/BUILD
+++ b/kythe/go/serving/tools/http_server/BUILD
@@ -1,3 +1,4 @@
+load("@io_bazel_rules_docker//go:image.bzl", "go_image")
 load("//tools:build_rules/shims.bzl", "go_binary")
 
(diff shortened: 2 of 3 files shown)


external.bzl
```diff
@@ -617,6 +630,7 @@ def kythe_dependencies():
     _cc_dependencies()
     _go_dependencies()
     _java_dependencies()
+    _docker_dependencies()
```
schroederc: The tests are failing because our `bazel_skylib` version is incompatible with whatever the docker rules declare.  You can just move this in the after `_rule_dependencies()` below.

kythe/go/serving/tools/http_server/BUILD
```diff
@@ -20,3 +21,25 @@ go_binary(
         "@org_golang_x_net//http2:go_default_library",
     ],
 )
+
+go_image(
```
schroederc: What's the context/reasoning/need for this target?

kythe/go/serving/tools/http_server/BUILD
```diff
@@ -20,3 +21,25 @@ go_binary(
...
+    name = "http_server_image",
+    srcs = ["http_server.go"],
+    goarch = "amd64",
+    goos = "linux",
+    static = "on",
+    deps = [
+        "//kythe/go/services/filetree",
```
schroederc: Is it possible to not completely mirror the `go_binary` attributes?

schroederc:
Since this is your first contribution, you should add yourself to the end of the `CONTRIBUTORS` file.

kythe/go/serving/tools/http_server/BUILD
achew22: [`go_image`](https://github.com/bazelbuild/rules_docker#go_image) is a `rules_docker` [macro](https://github.com/bazelbuild/rules_docker/blob/master/go/image.bzl#L65) that does two things.

1. It creates a `rules_go` `go_binary` named `name + ".binary"`.
1. Creates a docker image using that binary.

If you `blaze run` this target on a machine with Docker installed it will create a new tagged docker image named `bazel/kythe/go/serving/tools/http_server/http_server_image:http_server_image` (name:tag). This docker image is constructed entirely hermetically in Bazel without invoking the Docker tools which means you don't need `dind` in order to build an image. By default the `go_image` rule uses the `grc.io/distroless/base` as the base for the binaries. Interestingly the `gcr.io/distroless/base` image is itself created in Bazel this way.

In what way would you like me to expand in the PR to make this more clear? Would you prefer this as a change to the Kythe documentation, to my commit message, or inline in the `BUILD` file itself?

kythe/go/serving/tools/http_server/BUILD
schroederc: Based on https://github.com/bazelbuild/rules_docker/blob/master/go/image.bzl#L65, it looks like you can just specify a `binary` attribute.

kythe/go/serving/tools/http_server/BUILD
schroederc: I would like for more of the motivation for this target to be added to the description.  While I understand that this creates a Docker image, it's not clear at all why we would want this image since the Kythe team does not need it.

achew22: How does the Kythe team serve their UI if not through the sample web UI? Is there a more fleshed out example that can be used? Is the expectation that a user of Kythe will fork off the sample UI and write their own server?

schroederc: I don't follow.  A Docker image is not necessary to run the server.  The lack of a Docker image solely for this binary does not imply an expectation to fork the server code.

shahms:
Cleaning up old PRs.  Feel free to re-open and address the previous comments if this is still relevant.

PR #3265, opened 2018-11-19
build: fix incompatibility with Bazel 0.19.2

schroederc:


tools/git/commit-msg.sh
```diff
@@ -15,4 +15,5 @@
 # limitations under the License.
 
 # Check commit message: https://www.conventionalcommits.org/en/v1.0.0-beta.2/
-bazel run @io_kythe//tools/git:commitlint -- --edit "$1"
```
jaysachs: bazel run was removed, or is just broken?

shahms: ditto, it looks like bazel run should still work?

shahms: This mostly matters because the binaries which result from `bazel build` don't necessarily have predictable paths (see the changes to the go_rules, for example) so bazel run is the only reliable way of actually executing those binaries AFAIK.  It's dumb, but ...

PR #3275, opened 2018-11-20
fix: make workaround source compatible with JDK 11

schroederc:


schroederc:
Secondary reviewer: @cushon

cushon:
LGTM, thanks

Issue #3210, opened 2018-11-01
Bootstrap/dogfood extract kythe+bazel on cloudbuild


danielmoy-google:
kythe/go/extractors/gcp right now is just maven+gradle.  Needs bazel support, and then as a first customer we should support ourselves by extracting github.com/google/kythe using it.

Cody also points out that the %workspace%/.kythe file should be updated to have these contents (I believe it was originally intended to use the more fine-grained `ExtractionConfiguration`).

danielmoy-google:
Cody fixed this

PR #3288, opened 2018-12-04
feat: add (not yet working) poc bazel extraction on gcp

danielmoy-google:
Also include the setup-bazel-repo.sh script in a docker image
gcr.io/kythe-public/kythe-bazel-extractor-artifacts.

Includes extraction command for java and c++, but lacks logic for exporting the resulting kzips anywhere.  The built in cloud build artifacts doesn't support any sort of glob/wildcard, so we'll have to do that ourselves in a follow-up step later.

### DIFF
--- a/kythe/examples/bazel/BUILD
+++ b/kythe/examples/bazel/BUILD
@@ -1,3 +1,6 @@
 package(default_visibility = ["//visibility:public"])
 
-exports_files(["kythe_config.json"])
+exports_files([
+    "kythe_config.json",
+    "setup-bazel-repo.sh",
+])
--- a/kythe/go/extractors/gcp/README.md
+++ b/kythe/go/extractors/gcp/README.md
@@ -32,7 +32,7 @@ authorize it, associate it with a valid project id, create a test gs bucket.
 ## Maven Proof of Concept
 
 To extract a maven repository using Kythe on Cloud Build, use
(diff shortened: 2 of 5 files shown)


creachadair:
Can I prevail upon you to make the commit log more descriptive? 

Specifically: It'd be helpful to see a sentence or two about what is extracted (which languages, which build targets, etc.) and if this is part of an issue could you please add a link to it?

danielmoy-google:
Yea this is working towards #3210.

danielmoy-google:
Ah, do we have opinions on exactly how to handle bazel on gcp w.r.t. binary deps?  The way I see it, we have two realistic options:

* make our own cloud builder (just a docker image) that wraps [gcr.io/cloud-builders/bazel](https://cloud.google.com/cloud-build/docs/cloud-builders), and has clang and any other deps we currently need for kythe (sadly this also includes asciidoc...)
* make an image that just contains those necessary binaries and sideload it using volumes

The latter is the way I've been doing most things for maven/gradle so far.  I could see the appeal of the former way, though I'm not super comfortable with inheriting images like that.

shahms:
I think the sideload-by-volumes approach is more flexible as it allows both inheritance and ... not.

danielmoy-google:
Ok abandoning this in favor of a slightly different approach, which will be coming in pieces in separate pulls.

PR #3289, opened 2018-12-05
fix: add -fsanitize= to the unsupported args list

asmundak:
If clang invocation has -fsanitize=cfi argument, running cxx_extractor on it results in
error: no such file or directory: '/kythe_builtins/share/cfi_blacklist.txt'
(the other allowed values for the -fsanitize= will result in a different file to be missing).


shahms:
I *think* this should be fine (we don't generally want to run the indexer with any sanitizer enabled anyway), Luke?

zrlk:
This looks OK--while you *can* test at compile time whether a sanitizer's running (eg `__has_feature(address_sanitizer)`), I'm not sure that this is actually useful to support.

PR #3290, opened 2018-12-05
fix(gotool_extractor): when no global corpus is given, use package's corpus for each file

schroederc:


kythe/go/extractors/golang/golang.go
```diff
@@ -435,11 +435,20 @@ func (p *Package) addFiles(cu *apb.CompilationUnit, root, base string, names []s
...
 		trimmed := strings.TrimPrefix(path, root+"/")
+		vn := &spb.VName{
+			Corpus: p.ext.Corpus,
+			Path:   trimmed,
+		}
+		if vn.Corpus == "" {
+			vn.Corpus = p.VName.Corpus
+			components := strings.SplitN(vn.Path, string(filepath.Separator), 2)
+			vn.Path = strings.TrimPrefix(components[1], vn.Corpus+"/")
+			if components[0] != "src" {
```
salguarnieri: Is this a known pattern for go?

schroederc: Yes, it's the structure of a GOPATH.

creachadair: Unfortunately it's not quite that simple, because different repository hosts use different structures. kythe/kythe/go/extractors/govname has the logic that the Go tool uses—or at least that it used when I wrote that (it may need updated).

creachadair: (specifically: Knowing which part is the actual package name from the file path alone is tricky)

schroederc: For custom repository hosts, the corpus is the top-level domain (e.g. "k8s.io).  I plan to expand the config to better support these cases.

creachadair: Could you add a comment to that effect? Now that you describe it it makes sense, but it wasn't obvious that it does the right thing from the code alone.

creachadair:
> Meta note, we gonna be generally ok with using the repo URL as the corpus for most external things?

It's not clear there's something obviously better, at least to start with. Longer term the main confounding issue is that repositories sometimes have alternate vanity URLs (e.g., `gopkg.in`) that redirect to the real one. Canonicalizing those—if we care to—will either require a database or some other identification like a root-tip hash pair.

But starting with the base URL makes as much sense as anything else. After all, that's how anyone who wants to use the code will find it.

PR #3293, opened 2018-12-06
feat: support reading kzip files in kindex_tool

zrlk:
This tool is used for some C++ extractor tests.


zrlk:
@shahms: what's the right way to ensure the protos kzip_reader needs get linked in?

shahms:
@zrlk: You need to use a symbol from the proto in question. Internally, I have a library target that exports a function that does nothing other that:
```
namespace {
template <typename T>
void Use() {
  T t;
}
}
void UseDemProtos() {
 Use<kythe::proto::BuildDetails>(); // etc.
}
```

It's not ideal, but better than `alwayslink` (which doesn't work for protos anyway).

kythe/cxx/tools/kindex_tool_main.cc
```diff
@@ -83,6 +86,59 @@ void CanonicalizeHash(std::map<google::protobuf::string, size_t>* hashes,
       google::protobuf::string("hash" + std::to_string(inserted.first->second));
 }
 
+void DumpCompilationUnit(const std::string& path,
+                         kythe::proto::CompilationUnit* unit) {
+  using namespace google::protobuf::io;
```
shahms: This doesn't have the semantics you want it to and is pretty error-prone (all of these symbols will be found as if they were at global scope in this function).  Please using-declare the entities you actually want.

kythe/cxx/tools/kindex_tool_main.cc
```diff
@@ -83,6 +86,59 @@ void CanonicalizeHash(std::map<google::protobuf::string, size_t>* hashes,
...
+  std::map<google::protobuf::string, size_t> hash_table;
```
shahms: Do you actually want this to be ordered? If no, use absl::flat_hash_map

kythe/cxx/tools/kindex_tool_main.cc
```diff
@@ -83,6 +86,59 @@ void CanonicalizeHash(std::map<google::protobuf::string, size_t>* hashes,
...
+
+void DumpFileData(const std::string& path,
+                  const kythe::proto::FileData& content) {
+  using namespace google::protobuf::io;
+  CHECK(content.has_info() && !content.info().digest().empty());
+  std::string out_path = path + "_" + content.info().digest();
+  int out_fd =
+      open(out_path.c_str(), O_WRONLY | O_CREAT | O_TRUNC, S_IREAD | S_IWRITE);
+  CHECK_GE(out_fd, 0) << "Couldn't open " << out_path << " for writing.";
+  FileOutputStream file_output_stream(out_fd);
```
shahms: I think you need to tell the FileOutputStream to close the FD when it's done? (Same above)

zrlk: It's closed on line 139. Above, on 124.

shahms: Ah, fair enough. (I was confusing the Close() semantics with SetCloseOnDelete())

PR #3294, opened 2018-12-06
fix(java_indexer): ensure static import refs are static

schroederc:


dimo414:
Does this cover a private static field too? e.g.

```java
// MyClass.java
private static String foo = ...;
public static void foo() { ... }
```

`import static ....MyClass.foo` should not be considered a reference to the `foo` field.

schroederc:
No, this just handles `static`.  I can make a separate change for an accessibility check.

PR #3295, opened 2018-12-07
fix(sh_test): use "set -e" instead of adding "-e" to shebang

justbuchanan:
Bazel ignores the shebang for sh_test targets, so the "-e" was previously being ignored.

//kythe/go/serving/tools/testdata:kwazthis_test is now failing and I'm not sure why. Any ideas? It looks like the `jq` commands at the end aren't passed any json input.

### DIFF
--- a/kythe/cxx/common/testdata/net_client_test_runner.sh
+++ b/kythe/cxx/common/testdata/net_client_test_runner.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
 # Copyright 2015 The Kythe Authors. All rights reserved.
 #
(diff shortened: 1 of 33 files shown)


danielmoy-google:
> //kythe/go/serving/tools/testdata:kwazthis_test is now failing and I'm not sure why. Any ideas? It looks like the `jq` commands at the end aren't passed any json input.

jq commands are passed $JSON as specified up above on line 23.

But, I don't see immediately how it's failing.

justbuchanan:
It looks like the Generics.java file has changed since the byte offset in the kwazthis request was initially set. Also, both a defines and a defines/binding are returned - was one of those added since this test was written?

PR #3297, opened 2018-12-07
refactor: remove deprecated kindex from cxx extractor

justbuchanan:
KZip has replaced the index pack and kindex formats.

Previously the extractor only wrote kzip files if KYTHE_OUTPUT_FILE was
specified in the environment and ended with '.kzip'. This remains
unchanged, but now the extractor also writes kzip files if
KYTHE_OUTPUT_DIRECTORY is specified - this used to result in .kindex
files being written instead. When given a directory, the extractor names
the .kzip files based on a hash of the compilation unit just as it did
with kindex.

Extractor tests have been updated to use kzip instead of kindex.

Part of #2867

Issue #3264, opened 2018-11-19
Usage of KZip extraction for multi-module Maven projects


c-goetz:
Hi all,

we have a multi-module Maven project and I wanted to change the extraction so that it uses KZip instead of KIndex. The problem I'm experiencing is that the KZip file gets overwritten for each Maven module.
Is this behavior intended? The documentation for KZip implies that it's possible to store multiple compilation units in one KZip.

Thanks for your help (and Kythe in general)
Carlo

danielmoy-google:
Hi Carlo,

Yes indeed that is unsupported right now, but we would like to support it.

You are correct that you can put multiple CU in a single kzip (it's just a zip archive really).

Roughly speaking we're probably looking at modifying [AbstractJavacWrapper](https://github.com/kythe/kythe/blob/master/kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java#L123) to do one of two things differently:

1. output multiple files instead of to a single file (perhaps using KYTHE_OUTPUT_DIRECTORY like the old kindex
1. read in the file first to see if there's something in there, and add more results to it.

There's no right answer there about which way is better to do it.  It is worth noting that the first option there is closer to how the bazel indexer already does things.  Also there might be some advantage to having a unique filename there - I don't suppose we could make a filename based on the hash of the cu like bazel does?  That'd be sweet.

If we do that first option, we would need to build a utility that combines multiple kzip - but we already need such a library for bazel extraction on google cloud build anyways, so that won't be extra work.

c-goetz:
I would prefer the first variant. We make use of Mavens parallel builds. Writing concurrently to the same zip-archive does not sound like fun.

Would the following changes to the extractor be ok?
- check if `KYTHE_KZIP` is set to non-empty value (like `KYTHE_INDEX_PACK`)
- if so write kzips to `KYTHE_OUTPUT_FILE` folder

For merging there is zipmerge from libzip (https://libzip.org/documentation/zipmerge.html).

Are there any pointers to setting up a development environment (preferably IntelliJ) or is it straight forward (never worked with bazel)?

shahms:
zipmerge should "just work" depending on how it deals with conflicting files.  All of the extractors currently use the same root directory (called "root") although this isn't required by the spec (so could conceivably result in invalid .kzip files).

creachadair:
[...]

PR #3298, opened 2018-12-07
refactor: write kzip always for JavaStandalone extractor

danielmoy-google:
Follow-up to #3297, getting rid of kindex and indexpack support.

Relevantly, this will mean that setting KYTHE_OUTPUT_DIRECTORY now
writes potentially multiple kzips, instead of kindex.  We can bundle
those together later with zipmerge if we want.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
@@ -52,10 +52,11 @@ import java.util.List;
  * <p>KYTHE_ROOT_DIRECTORY: required root path for file inputs; the {@link FileData} paths stored in
  * the {@link CompilationUnit} will be made to be relative to this directory
  *
- * <p>KYTHE_OUTPUT_FILE: if set to a non-empty value, write the resulting .kindex file to this path
+ * <p>KYTHE_OUTPUT_FILE: if set to a non-empty value, write the resulting .kzip file to this path
  * instead of using KYTHE_OUTPUT_DIRECTORY
  *
- * <p>KYTHE_OUTPUT_DIRECTORY: required directory path to store the resulting .kindex file
+ * <p>KYTHE_OUTPUT_DIRECTORY: directory path to store the resulting .kzip file, if KYTHE_OUTPUT_FILE
+ * is not set
  *
  * <p>KYTHE_INDEX_PACK: if set to a non-empty value, interpret KYTHE_OUTPUT_DIRECTORY as the root of
  * an indexpack instead of a collection of .kindex files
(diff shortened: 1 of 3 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
```diff
@@ -122,18 +121,25 @@ private static void outputIndexInfo(CompilationDescription indexInfo) throws IOE
       if (outputFile.endsWith(IndexInfoUtils.KZIP_FILE_EXT)) {
         IndexInfoUtils.writeKzipToFile(indexInfo, outputFile);
       } else {
-        IndexInfoUtils.writeKindexToFile(indexInfo, outputFile);
+        System.err.printf("Unsupported output file: %s%n", outputFile);
```
salguarnieri: %n -> \n ?

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
```diff
@@ -122,18 +121,25 @@ private static void outputIndexInfo(CompilationDescription indexInfo) throws IOE
...
-    new Archive(outputDir).writeDescription(indexInfo);
+    // Just rely on the underlying compilation unit's signature to get the filename, if we're not
+    // writing to a single kzip file.
+    String name =
+        indexInfo
+            .getCompilationUnit()
+            .getVName()
+            .getSignature()
+            .trim()
+            .replaceAll("^/+|/+$", "")
```
salguarnieri: Do these replaces create a risk of name collision?

danielmoy-google:
Ah this will probably be required for #3264 if I'm remembering correctly.

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
danielmoy-google: I'm unsure here.  The rest of this file is already using %n, so I'd like to keep it that way for now.  If we want to move it all to \n, rather do that as a separate refactor step, and all at once (and maybe other files as well?).

In any event, %n is more platform-agnostic, I think?  (Windows?)

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
danielmoy-google: In theory maybe?  In practice, this is what we already used before.  So, might be ok.

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
danielmoy-google: Yup ok, from the java docs, %n is the platform-agnostic one

https://docs.oracle.com/javase/tutorial/java/data/numberformat.html

salguarnieri: TIL

kythe/java/com/google/devtools/kythe/extractors/java/standalone/AbstractJavacWrapper.java
salguarnieri: sgtm

danielmoy-google:
Ok should have a fix for the failing test - kzip writing needs to populate parent directory too, just like kindex.

We just hadn't run into this problem yet, because all our kzip writing has been done from more precise wrapping logic that creates the output directory ahead of time.

PR #3303, opened 2018-12-10
feat(build_details): add a build_config field to BuildDetails proto

shahms:
This enables indexers to include relevant facts on anchor nodes, once the rest of the plumbing is in place.  I will update the schema in this PR as well (forthcoming).


schroederc:
Can you add this fact name to http://go/gh/kythe/kythe/blob/master/kythe/data/schema_index.textproto#L92 as well?

creachadair:
I understand the nature of the change, but is there any documentation (e.g., an RFC or a design doc) that motivates it?

danielmoy-google:
Internally, yes.

I suppose once it's solidified more maybe shahms will push it external?

shahms:
There's nothing particularly confidential about it, but I would have to double check.  The change is also being made to accommodate the needs of external corpora anyway (Chromium, etc.).

Mostly we don't have an established procedure for doing design docs externally.  I should probably look at how the Bazel team does so.

creachadair:
> There's nothing particularly confidential about it, but I would have to double check. The change is also being made to accommodate the needs of external corpora anyway (Chromium, etc.).

Makes sense. For design changes that aren't specifically intended for internal use, I think it would be best to start in the open-source repo. We haven't always been perfect about that, but it was one of our original goals to try and insulate Kythe from the inevitable flood Google-specific special cases by keeping the documentation as public as possible.

> Mostly we don't have an established procedure for doing design docs externally. I should probably look at how the Bazel team does so.

We created `kythe/docs/rfc` as a repository for design RFCs, and that seems like a reasonable place to put them. I think @schroederc did that by fiat, though, so I imagine improvements to that idea are probably welcome. :)

shahms:
While that would be ideal, these changes often start in ad-hoc hallway discussions and get a bit wider audience and discussion in a weekly meeting where impacts on google-internal indexing are mentioned and often tied up in the design.  Those can't be opened up for a variety of reasons.

However, it would certainly be good to engage with the community as early as feasible.

creachadair:
> While that would be ideal, these changes often start in ad-hoc hallway discussions and get a bit wider audience and discussion in a weekly meeting where impacts on google-internal indexing are mentioned and often tied up in the design. Those can't be opened up for a variety of reasons.

Of course, and that makes sense as a starting point. One of the reasons we open-sourced Kythe when we did was to help force ourselves to motivate these design questions in a more general way. In the original design of Grok, everything was a completely-pragmatic compromise with the internal infrastructure, and that made it nearly impossible to share anything about Grok—even the data model—with anyone outside Google. 

Naturally, a lot of the design issues still reflect the need to maintain that internal service. But I think it's valuable to view Kythe, rather than Google3, as the source of truth. Although there's nothing wrong with hallway discussions and internal use cases, I think it's useful to consider those a starting point rather than the _primum movens_ for a design. That's a principle we set out when we started, and I'd argue it's worth keeping.

PR #3305, opened 2018-12-10
fix(java_indexer): ensure static import refs are accessible

schroederc:


schroederc:
`¯\_(ツ)_/¯` At that point we'd be testing javac rather than the indexer.

creachadair:
What was the motivation for this change? Is the issue that the model for static glob imports has wrong results if we paint edges to the inaccessible members? (That would make sense, but even if that's wrong it'd be helpful to include that kind of context in the commit logs)

schroederc:
You're spot on.  This follows https://github.com/kythe/kythe/pull/3294.

Issue #3307, opened 2018-12-10
FR: Consider supporting shallow clones of LLVM and Clang


creachadair:
The LLVM and Clang repositories are very large, both in size and in number of objects. The `tools/modules/update.sh` script currently clones and checks out out the whole repository for each. Both projects have a long history and a lot of code, and everyone who wants to build/develop against Kythe winds up having to download that complete history.

The obvious solution would be to use shallow clones (e.g., `git clone --depth=1 …`), but that does not quite work because we pin (via `versions.sh`) to specific commit hashes. In _principle_ one could fetch just a single commit, but the Git hosting sites (including GitHub) do not permit you to request an arbitrary hash, only "known" commits. As far as I can tell, "known" means commits which are tagged, or are branch heads. Unfortunately, it looks like LLVM doesn't use tags or branches.

We could still make this better for our contributors, though, if  if we are willing to maintain a fork of the mirror and attach tags to our fork. Then we could get a shallow clone of the fork by doing something like:

    git clone --depth 1 --branch v1.2.3 --single-branch $FORKREPO $TARGET
    (cd $TARGET && git checkout -b v1.2.3)

I did this as an experiment, and compare:

```
# github.com/llvm-mirror/llvm
$ du -sh third_party/llvm/llvm
2.3G	third_party/llvm/llvm

# github.com/creachadair/llvm
$ du -sh test/llvm
426M	test/llvm
```

It adds a couple steps to updating the commit hash: Rebasing the fork, and adding/updating the tag. But that happens infrequently enough that it might be worthwhile (and I think it could be scripted).

creachadair:
> We don't have to use `git clone` we can just fetch an http archive of the particular revision, like we do for our other dependencies in WORKSPACE, there's no need to fork.

Does that work for arbitrary commits? That would certainly be easier, but I couldn't see a way to get the exact version we want without a tag or a branch.

shahms:
$ wget https://github.com/llvm-mirror/llvm/archive/e744b686ed7b30e1c6a172ede09f012de33aadac.zip
$ du -hs llvm-e744b686ed7b30e1c6a172ede09f012de33aadac/
386M	llvm-e744b686ed7b30e1c6a172ede09f012de33aadac/

shahms:
there's a missing unzip step there, but yes ;-)

creachadair:
Nevermind, I see @danielmoy-google is already working on it.

creachadair:
[...]

PR #3308, opened 2018-12-10
fix: change tools/modules/update.sh to use an archive instead of clone

danielmoy-google:
One way of fixing #3307.

This will make the initial download of llvm and clang faster, by not
requiring a clone of the entire ~3GB repo.  It also save on the ongoing
disk used.  However, it does make updates marginally slower because you
have to re-download a new archive.  Given that updates are vastly
dominated by the cost of building llvm&clang, this is probably ok.

Also this has the side effect of making images for Google Cloud Build +
bazel significantly smaller, which will be nice for #3210.

### DIFF
--- a/kythe/web/site/getting-started.md
+++ b/kythe/web/site/getting-started.md
@@ -44,6 +44,7 @@ Kythe relies on the following external dependencies:
 * libssl-dev
 * bison-3.0.4 (2.3 is also acceptable)
 * flex-2.5
+* wget
 * [docker](https://www.docker.com/) (for release images `//kythe/release/...` and `//buildtools/docker`)
 * [leiningen](http://leiningen.org/) (used to build `kythe/web/ui`)
 * [ninja](https://ninja-build.org/) (optional; improves LLVM build speed)
--- a/kythe/web/site/getting-started.md
+++ b/kythe/web/site/getting-started.md
@@ -61,7 +61,8 @@ apt-get install \
     asciidoc asciidoctor source-highlight graphviz \
     gcc libssl-dev uuid-dev libncurses-dev libcurl4-openssl-dev flex clang-3.5 bison \
     openjdk-8-jdk \
-    parallel
+    parallel \
+    wget
 
 # https://golang.org/dl/ for Golang installation
 # https://docs.docker.com/installation/debian/#debian-jessie-80-64-bit for Docker installation
--- a/tools/modules/update.sh
+++ b/tools/modules/update.sh
@@ -22,21 +22,24 @@
 # repositories are at the correct version and will configure and build them.
 
 wget_copy_archive() {
+  # The dependency to download (llvm, clang).
   local target="$1"
-  local dir="$2"
+  # The directory to download the dependency into.
+  local dir="${2:?missing directory}"
+  # The specific version of the dependency to use.
   local sha="$3"
 
   if [[ ! -f "$dir/$sha.sentinel" ]]; then
     wget "https://github.com/llvm-mirror/$target/archive/$sha.zip"
+    trap 'rm "$sha.zip"' EXIT ERR INT
     rm -rf "$dir"
     mkdir -p "$dir"
     local tmpdir=$(mktemp -d)
+    trap 'rm -rf "$tmpdir"' EXIT ERR INT
     unzip "$sha.zip" -d "$tmpdir/"
     # This relies on the behavior of github to always produce a zip archive with
     # a subdirectory named "repo-###sha###" that contains the repo inside it.
     mv -T "$tmpdir/$target-$sha" "$dir"
-    rmdir "$tmpdir"
-    rm "$sha.zip"
     # Leave an empty file so that we know what version we have checked out.
     touch "$dir/$sha.sentinel"
   fi
--- a/tools/modules/update.sh
+++ b/tools/modules/update.sh
@@ -21,8 +21,27 @@
 # If the argument `--build_only` is passed, the script will assume that the
 # repositories are at the correct version and will configure and build them.
 
+ZIPFILES=()
+TMPDIRS=()
+
+trap_cleanup() {
+  ZIPFILES+=("$1")
+  TMPDIRS+=("${2?:missing directory}")
+  trap wget_cleanup EXIT
+}
+
+wget_cleanup() {
+  cd "$ROOT"
+  for file in "${ZIPFILES[@]}"; do
+    rm "$file"
+  done
+  for dir in "${TMPDIRS[@]}"; do
+    rm -rf "$dir"
+  done
+}
+
 wget_copy_archive() {
-  # The dependency to download (llvm, clang).
+  # The dependency to download from llvm-mirror (llvm, clang).
   local target="$1"
   # The directory to download the dependency into.
   local dir="${2:?missing directory}"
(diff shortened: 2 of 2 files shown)


tools/modules/update.sh
```diff
@@ -16,28 +16,30 @@
...
 #
-# If the single argument `--git_only` is passed, the repositories will be
+# If the single argument `--fetch_only` is passed, the repositories will be
 # updated to the pinned versions without configuring or building them.
 # If the argument `--build_only` is passed, the script will assume that the
 # repositories are at the correct version and will configure and build them.
 
-git_maybe_clone() {
-  local repo="$1"
+wget_copy_archive() {
```
creachadair: You'll need to make `wget` a requirement on the installation instructions, too.

tools/modules/update.sh
```diff
@@ -16,28 +16,30 @@
...
+  local target="$1"
```
creachadair: Please document this in some wise—"target" is very general.

tools/modules/update.sh
```diff
@@ -16,28 +16,30 @@
...
   local dir="$2"
```
creachadair: Given that you are about to `rm -rf` on this, please verify it is sane, e.g., `local dir="${2:?missing directory}"` or words to that effect.

tools/modules/update.sh
```diff
@@ -16,28 +16,30 @@
...
+  local sha="$3"
 
-git_checkout_sha() {
-  local repo="$1"
-  local sha="$2"
-  cd "$repo"
-  if [[ "$sha" != "$(git rev-parse HEAD)" ]]; then
-    git fetch origin master
-    git checkout -f "$sha"
+  if [[ ! -f "$dir/$sha.sentinel" ]]; then
```
creachadair: Another approach that might be safer would be to always unpack into a directory named for the SHA, and just update a symlink pointing to that directory, e.g., `llvm -> a0c310991b…30c0`

creachadair: (though that would mean these directories would accumulate over time, so maybe that's not great either—though clobbering a whole directory structure from a script always gives me pause)

tools/modules/update.sh
```diff
@@ -16,28 +16,30 @@
...
+    wget "https://github.com/llvm-mirror/$target/archive/$sha.zip"
+    rm -rf "$dir"
+    mkdir -p "$dir"
+    local tmpdir=$(mktemp -d)
+    unzip "$sha.zip" -d "$tmpdir/"
+    # This relies on the behavior of github to always produce a zip archive with
+    # a subdirectory named "repo-###sha###" that contains the repo inside it.
+    mv -T "$tmpdir/$target-$sha" "$dir"
```
schroederc: `-T` is not a valid option on the OSX `mv` tool.

tools/modules/update.sh
```diff
@@ -16,28 +16,30 @@
...
-  local sha="$2"
-  cd "$repo"
-  if [[ "$sha" != "$(git rev-parse HEAD)" ]]; then
-    git fetch origin master
-    git checkout -f "$sha"
+  if [[ ! -f "$dir/$sha.sentinel" ]]; then
+    wget "https://github.com/llvm-mirror/$target/archive/$sha.zip"
+    rm -rf "$dir"
+    mkdir -p "$dir"
+    local tmpdir=$(mktemp -d)
```
schroederc: Please add a `trap` to cleanup the temporary directory: `trap 'rm "$tmpdir"' EXIT ERR INT`

tools/modules/update.sh
danielmoy-google: :+1:

tools/modules/update.sh
danielmoy-google: Yup, my two exact thoughts.  Shahms also independently had the same idea.

I agree that clobbering the directory structure from script doesn't sit well with me, but gradually accumulating 500MBs of stuff every dep doesn't sit well with me either.

I'm not convinced which way is better yet, this is just the one I settled on so far.

tools/modules/update.sh
danielmoy-google: :+1:

Also above for the .zip file while we're at it.

tools/modules/update.sh
danielmoy-google: Ah shit, what's the behavior in OSX mv to rename a directory then?

I'm not seeing one on https://www.unix.com/man-page/osx/1/mv/

:(

tools/modules/update.sh
danielmoy-google: Added to kythe/web/site

kythe/web/site/getting-started.md
```diff
@@ -61,7 +61,8 @@ apt-get install \
     asciidoc asciidoctor source-highlight graphviz \
     gcc libssl-dev uuid-dev libncurses-dev libcurl4-openssl-dev flex clang-3.5 bison \
     openjdk-8-jdk \
-    parallel
+    parallel \
+    wget
```
creachadair: Above too, please (under "external dependencies")

tools/modules/update.sh
```diff
@@ -22,21 +22,24 @@
 # repositories are at the correct version and will configure and build them.
 
 wget_copy_archive() {
+  # The dependency to download (llvm, clang).
```
creachadair: e.g., "This must be a repository hosted by the LLVM mirror" or words to that effect.

tools/modules/update.sh
creachadair: `mv olddir newdir/newname`

tools/modules/update.sh
```diff
@@ -22,21 +22,24 @@
...
+  # The specific version of the dependency to use.
   local sha="$3"
 
   if [[ ! -f "$dir/$sha.sentinel" ]]; then
     wget "https://github.com/llvm-mirror/$target/archive/$sha.zip"
+    trap 'rm "$sha.zip"' EXIT ERR INT
     rm -rf "$dir"
     mkdir -p "$dir"
     local tmpdir=$(mktemp -d)
+    trap 'rm -rf "$tmpdir"' EXIT ERR INT
```
creachadair: I believe `EXIT` subsumes the others.

tools/modules/update.sh
danielmoy-google: actually, to make this even more robust I am going to cd back to the $ROOT, since some parts later on will cd away.  So depending on where it dies, it'll be safer to always cd back to known $ROOT and rm from there.

(Also I just learned the hard way that purposefully interrupting tools/modules/update.sh in the middle and leaving clang/llvm in a bad state makes the presubmit very unhappy)

tools/modules/update.sh
danielmoy-google: ah k yea because I'm removing the directory this should work.  I was leaving it in place with a previous revision (before settling on removing old versions).

kythe/web/site/getting-started.md
danielmoy-google: :+1:

tools/modules/update.sh
schroederc: I believe it depends on the shell implementation, so I always just play it safe.

danielmoy-google: Sure I am fine having ERR INT in there as well

tools/modules/update.sh
danielmoy-google: :+1:

tools/modules/update.sh
```diff
@@ -21,25 +21,42 @@
 # If the argument `--build_only` is passed, the script will assume that the
 # repositories are at the correct version and will configure and build them.
 
+ZIPFILES=()
+TMPDIRS=()
+
+trap_cleanup() {
+  ZIPFILES+=("$1")
+  TMPDIRS+=("${2?:missing directory}")
+  trap wget_cleanup EXIT ERR INT
```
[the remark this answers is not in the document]
danielmoy-google: Yea, but if I'm fitting my mental model of trap correctly, it behaves like setting an interrupt handler, so setting it twice to the same thing doesn't hurt?

I'm not opposed to pulling the trap outside above the initial wget calls or something though, if people would rather it not be in here.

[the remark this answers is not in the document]
creachadair: I think it's harmless to re-set it. As far as I can see this is only called once anyway, but it just stood out to me. I don't think it's dangerous.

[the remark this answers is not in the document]
schroederc: There's another call to `trap` below that changes the handler so we need to join them or call the first manually before the later handler replacement.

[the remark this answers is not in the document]
danielmoy-google: Ah I missed that.  I'm in favor of moving everything into one trap cleanup handler.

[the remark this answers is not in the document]
danielmoy-google: wait no.  The other trap explicitly doesn't occur on clean EXIT?

hmm.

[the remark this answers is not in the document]
danielmoy-google: Ok, how about we just clear the trap used for fetching and leave things in a clean state for the rest of the script.

danielmoy-google:
Ah merge conflicts

danielmoy-google:
Any thoughts/concerns with this version?

PR #3309, opened 2018-12-10
docs(rfc): add an RFC for incorporating build description into the graph

shahms:
@creachadair Some background for the earlier change (and upcoming ones).


kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+## Background {#background}
+
+Currently, the Kythe graph largely assumes that it is built for a single configuration, at a consistent revision for a particular corpus.  Historically, this has worked well.  But as time marches onward and the diversity of platforms built from a single corpus has grown it has left a widening gap in the information available to users with only partial cross references for some files (due to conditional compilation) or other files entirely omitted due to build-time selection.  In order to work around this lack, some third-party corpora (most notably Chromium) have begun to use the "root" VName field to encode this information.  Similarly, the extant "prodkernel" corpus has need of this, but currently makes do without. We reasonably anticipate that other soon-to-be-corpora such as Android and Tensorflow would encounter similar limitations.
+
+
+## Overview {#overview}
+
+Rather than introduce any new VName fields, nodes or edges build configuration can be adequately represented via a signature-relevant fact on anchor nodes and some ancillary clarifications to the guidelines for root and corpus.  This is predicated on:
```
creachadair: The Background above talks about build configuration, but I don't see anything that discusses the specific problem that causes for the graph at the node level. In other words, it'd be helpful to answer the question "why is _this_ solution preferred?" in addition to "what are we planning to do?"

Chromium is using the root in that way because they have to serve all their xrefs from a UI that doesn't know how to distinguish "variants" like platform strings, talking to a service that only serves a single corpus. But that's an historical accident. The question I see here is—even if the code browser _did_ know how to do that, why does that mean we have to add data to the _graph_ for that information? 

Using C++ again (running with the Chromium example), each of the build configurations has to be built and indexed separately anyway, since they use different flags, include-paths, etc. Assuming we fixed the problem that the UI doesn't have a knob for the platform—could this problem be as effectively solved by storing the graph for (say) the "android" variant separately from the graph for (say) the "linux" variant?

In other words: What do we get out of stuffing this information into the graph at the node level, that we could not get by having multiple graphs?  The list below mentions filtering—could that not be done by selecting from a set of graphs rather than examining each node individually? (more on this below)

(I assume there is a reason for this, but it isn't spelled out anywhere that I can see)

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+1.  Non-generated source files exist independent of build configuration.
+1.  Nodes must already incorporate relevant information into their signature.  This includes any elements related to build-configuration which materially alters the represented entity. This is currently already done by non-anchor (aka "semantic") nodes, but build configuration has not been considered relevant for anchors.
+1.  Certain languages will necessarily emit different anchors under different build configurations (C/C++ are particularly subject to this).
```
creachadair: I don't get this example. C++ emits different anchors because the compiler compiles the headers differently, and/or stores output in different directories. Is there some information we are currently adding to anchors _without_ the addition of new build-config tags, that we would be able to get rid of by including it?

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,64 @@
...
+1.  Corpus represents a stable name to a particular logical body of code.  Different corpus names may map to different parts of the same underlying repository.  That mapping resides outside the Kythe graph[^1].  In particular, this can be used to accommodate source code which resides in a particularly repository, but is logically distinct from the primary corpus (perhaps due to using a different build system).  For example, within Piper the primary corpus is "google3" but the source code for several languages' standard library resides in third_party.  Presently, Java and C++ use distinct corpora for these references, but Go uses a separate root.  Go will be changed to use the "gostdlib" corpus for this.
+
+
+1.  A new fact "build/config" will be allowed on anchor nodes.  This should be a user-meaningful string representation of all relevant build characteristics (think "android" not "-DHAVE_GLIB").
```
creachadair: Allowed _only_ on anchor nodes?

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Detailed Design {#detailed-design}
+
+
+### Graph {#graph}
+
+The Kythe graph will be augmented such that:
+
+
+1.  File nodes with a non-empty root VName field are explicitly considered to have been generated by the build.  If the contents of generated files differ with the build configuration, the root must include a unique string per build configuration.  This follows from the general requirements that unique graph entities be given unique VNames.  This allows clients to trivially know if they should request the source for a particular file from the underlying repository or from Kythe.
```
creachadair: Here I think you've touched on an interesting idea: The distinction between "parts of the name related to output" (e.g., code generation) vs "parts of the name related to input" (e.g., configuration settings). Does that separation buy us something?

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+1.  A new fact "build/config" will be allowed on anchor nodes.  This should be a user-meaningful string representation of all relevant build characteristics (think "android" not "-DHAVE_GLIB").  
+1.  As we expect distinct anchors at the same position, this fact must be incorporated into the anchor signature by the indexer when present.
+
+
+### Extraction and Indexing {#extraction-and-indexing}
+
+As build configuration is a high-level description of a build, but must be incorporated into anchors by the indexer, it must be provided or derived at build time and incorporated into each resulting CompilationUnit.  This will be done by adding a string-value "build_config" field to the[ kythe.proto.BuildDetails](https://cs.corp.google.com/piper///depot/google3/third_party/kythe/kythe/proto/buildinfo.proto) message.  Indexers must incorporate this field into emitted anchors if the language in question makes a meaningful distinction between build configurations.  This is primarily relevant to C/C++ family languages, but is up to the individual language indexers to make this determination.  When in doubt, it is recommended that these behavior be flag controlled.
```
creachadair: It's probably worth mentioning somewhere that build configurations—like corpus labels—are assumed to have some kind of stability to them over time. This is kind of implied by the examples of platform, but it would bear saying it explicitly.

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,64 @@
...
+1.  File nodes with a non-empty root VName field are explicitly considered to have been generated by the build.  If the contents of generated files differ with the build configuration, the root must include a unique string per build configuration.  This follows from the general requirements that unique graph entities be given unique VNames.  This allows clients to trivially know if they should request the source for a particular file from the underlying repository or from Kythe.
+1.  Corpus represents a stable name to a particular logical body of code.  Different corpus names may map to different parts of the same underlying repository.  That mapping resides outside the Kythe graph[^1].  In particular, this can be used to accommodate source code which resides in a particularly repository, but is logically distinct from the primary corpus (perhaps due to using a different build system).  For example, within Piper the primary corpus is "google3" but the source code for several languages' standard library resides in third_party.  Presently, Java and C++ use distinct corpora for these references, but Go uses a separate root.  Go will be changed to use the "gostdlib" corpus for this.
```
zrlk: nit: a particular repository (was particularly)

kythe/docs/rfc/2967.md
shahms: The answer to your first question is the first bullet point "Non-generated source files exist independent of build configuration."  If the data is in a different graph for each build configuration every xrefs request has to be split across each build configuration.  And then stitched together either at the client or in a proxy.  It's inefficient. This also applies to graph storage generally; using distinct graphs means we have N copies of essentially the same data.

C++ is the poster child because a single file is likely to have different anchors (possible to the same underlying semantic entities) depending on the build configuration.  Keeping them in the same graph allows us to provide references to each distinct entity, regardless of platform (e.g. header w/ multiple implementations).  Splitting the graph does not allow for this.

shahms: Figuring out all of the xrefs for a "particular" semantic entity will also be lossy as that entity will potentially have different tickets in each graph and the only key that they are the "same" being the fact that the anchors happen to overlap.

creachadair: > If the data is in a different graph for each build configuration every xrefs request has to be split across each build configuration.

Is there any expectation that xrefs will be shared _across_ build configurations? In C++, for example, that will ~never happen because of how tickets are generated.

kythe/docs/rfc/2967.md
creachadair: My question was to relevancy: This bullet point offers itself as partial explanation for why a new fact is required for anchors. But I don't see the connection between "certain languages will need to emit different anchors under different build configurations" when the example language—C++—already _does_ emit different anchors (with or without the distinction). So my question was that if it's _not_ about distinguishability, then I don't know why this point supports the design.

kythe/docs/rfc/2967.md
creachadair: I think those points would be helpful to include in the motivation for the design.

kythe/docs/rfc/2967.md
shahms: It emits some different anchors and some "identical" anchors.  There is no way of distinguishing the two at present and (importantly) no way for a user to do so.  It is about distinguishability, but that was not the question (which was about "things we can remove" due to this change).

kythe/docs/rfc/2967.md
shahms: Perhaps worth ruminating on, but won't impact this particular design, IMO.

kythe/docs/rfc/2967.md
creachadair: Go only uses the root this way inside Google, because of how Blaze is organized. It doesn't use the root in the "native" Go implementation. 

cf. https://github.com/kythe/kythe/blob/c453a319b76da635c7915c8376d0121118c2d8bd/kythe/go/extractors/govname/govname.go#L34

kythe/docs/rfc/2967.md
shahms: For the moment, yes.  It's possible it might be useful on file nodes if those files are generated and that generation depends on the build configuration.  I can add a comment to that effect.

kythe/docs/rfc/2967.md
shahms: yes, this is probably more relevant internally but is included for completeness and reference

kythe/docs/rfc/2967.md
shahms: Done.

shahms:
PTAL

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Post-processing
+
+The post processor will collect the build configurations facts and emit them for
+use by the StatusService. (TODO)
+
+### Serving
+
+The available build configurations will be made available via StatusService.
+(TODO)
```
creachadair: Is there an issue this could point to?

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,123 @@
...
+message. Indexers must incorporate this field into emitted anchors if the
+language in question makes a meaningful distinction between build
+configurations. This is primarily relevant to C/C++ family languages, but is up
+to the individual language indexers to make this determination. When in doubt,
+it is recommended that these behavior be flag controlled.
+
+### Post-processing
+
+The post processor will collect the build configurations facts and emit them for
+use by the StatusService. (TODO)
```
creachadair: Is there an issue this could point to?

kythe/docs/rfc/2967.md
```diff
@@ -0,0 +1,123 @@
...
+1.  As we expect distinct anchors at the same position, this fact must be
+    incorporated into the anchor signature by the indexer when present.
+
+### Extraction and Indexing
+
+As build configuration is a high-level description of a build, but must be
+incorporated into anchors by the indexer, it must be provided or derived at
+build time and incorporated into each resulting CompilationUnit. This will be
+done by adding a string-value "build_config" field to
+the[ kythe.proto.BuildDetails](https://cs.corp.google.com/piper///depot/google3/third_party/kythe/kythe/proto/buildinfo.proto)
```
creachadair: You'll need to update the link—this is an internal one.

kythe/docs/rfc/2967.md
shahms: Done (#3322)

PR #3310, opened 2018-12-11
fix: Cleanup typos in objc test

salguarnieri:
A few verifier variables were mis-typed in this file, which meant it
wasn't really testing what we wanted it to test. I fixed the names and
set this target to disallow singletons so this won't happen again.


creachadair:
Ha! My reviews are now officially useless. :)

salguarnieri:
I think checking for singletons is the default for the binary, but for cxx indexer tests, we explicitly set the flag to false by default. https://github.com/kythe/kythe/blob/master/tools/build_rules/verifier_test/cc_indexer_test.bzl#L48

danielmoy-google:
> Ha! My reviews are now officially useless. :)

well, approvals at least

Hopefully your reviews are still very useful :P

PR #3317, opened 2018-12-11
feat(java_indexer): emit implicit anchors for default constructors

schroederc:
Emit implicit `defines` anchors for default constructors and emit
implicit `ref/call` edges for implicit `super` constructor calls (even
if the callsites are in an implicit constructor).


schroederc:
This is ready for review, but I will hold back on merging it until I'm back from vacation next week.

PR #3318, opened 2018-12-12
feat: add docker image with a kythe install for use in google cloud build

danielmoy-google:
Creates `gcr.io/kythe-public/kythe-bazel-extractor-artifacts`.

For now basing this off of launcher.gcr.io/google/ubuntu16_04 and exposing relevant pieces via /opt/kythe/ with the expectation that we can sideload necessary pieces in subsequent build steps:

* /opt/kythe/extractors contains `setup-bazel-repo.sh`, a script for modifying WORKSPACE so that a bazel repo can extract using kythe extractors
* /opt/kythe/repo/kythe contains an installed kythe repo (./tools/modules/update.sh already run, so `bazel build some:target`should work).

Another step towards #3210.

### DIFF
--- a/kythe/go/extractors/gcp/bazel/Dockerfile
+++ b/kythe/go/extractors/gcp/bazel/Dockerfile
@@ -35,26 +35,12 @@ FROM launcher.gcr.io/google/ubuntu16_04
 # TODO(danielmoy): for modifying other bazel repos, use the setup-bazel-repo script.
 ADD kythe/examples/bazel/setup-bazel-repo.sh /opt/kythe/extractors/setup-bazel-repo.sh
 
-# These are instructions for installing bazel from apt-get, but it doesn't work
-# because one of the deps 'openjdk-8-jdk' cannot be found.
-#ENV KYTHE_PRE_DEPS \
-#    curl ca-certificates
-#
-#RUN apt-get update && \
-#    apt-get upgrade -y && \
-#    apt-get install -y --no-install-recommends $KYTHE_PRE_DEPS && \
-#    apt-get clean
-#
-# RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
-# RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
-
 ENV KYTHE_DEPS \
     asciidoc asciidoctor source-highlight graphviz git \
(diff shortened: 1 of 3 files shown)


kythe/go/extractors/gcp/bazel/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+#     - name: 'gcr.io/kythe-public/kythe-bazel-extractor-artifacts'
+#       volumes:
+#         - name: 'kythe_extractors'
+#           path: '/opt/kythe/extractors/'
+#         - name: 'kythe_repo'
+#           path: '/opt/kythe/repo/kythe/'
+
+FROM launcher.gcr.io/google/ubuntu16_04
+
+# TODO(danielmoy): for modifying other bazel repos, use the setup-bazel-repo script.
```
salguarnieri: I'm a little confused about the todo. Do you need to change the line below, or are you just noting that you need to use this file in the future?

danielmoy-google: Noting that I'll want to use this file in the future.  Right now, I haven't tested that part, so the TODO is mostly just to remind myself where to look next.

kythe/go/extractors/gcp/bazel/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+ADD kythe/examples/bazel/setup-bazel-repo.sh /opt/kythe/extractors/setup-bazel-repo.sh
+
+# These are instructions for installing bazel from apt-get, but it doesn't work
+# because one of the deps 'openjdk-8-jdk' cannot be found.
+#ENV KYTHE_PRE_DEPS \
```
danielmoy-google: Well actually, now

(1) I did get this working, but

(2) the more I think about it, the less I want to isntall apt-get install bazel anyways, so just gonna go with the wget route to a specific pinned version.

salguarnieri: yeah, wget and pinning sounds like a safer way to get bazel, especially since kythe may be sensitive to the exact bazel version you pull down.

justbuchanan: Keep in mind you can also ask apt for a specific version of a package with something like `sudo apt install bazel=0.20.0`.

danielmoy-google: Hmm that's true.

Other random thought - the wget/script install instructions are listed first on bazel.build's installation instructions, if that has any effect.

PR #3323, opened 2018-12-14
fix shell lint warnings in tools/modules/update.sh

shahms:
shellcheck lint errors keep preventing me from pushing anything, so fix them :-)


shahms:
No worries!

PR #3330, opened 2018-12-17
fix(java_common): allow analyzers to throw InterruptedExceptions

schroederc:
By convention/best-practices, `InterruptedException` should not be swallowed or turned into other exceptions without resetting the interrupt status.  Adding the `throws` clause allows subclasses of `AbstractCompilationAnalyzer` to follow this practice when handling the futures returned from a `FileDataProvider`.


kythe/java/com/google/devtools/kythe/analyzers/base/AbstractCompilationAnalyzer.java
```diff
@@ -65,6 +66,8 @@ public void analyzeRequest(AnalysisRequest req, FactEmitter emitter) throws Anal
           "Analyzing compilation: {%s}",
           TextFormat.shortDebugString(req.getCompilation().getVName()));
       analyzeCompilation(req.getCompilation(), Optional.ofNullable(revision), fileData, emitter);
+    } catch (InterruptedException e) {
```
jaysachs: Thread.currentThread().interrupt();

creachadair:
Can you say more what you mean by "allow", and why it matters?

kythe/java/com/google/devtools/kythe/analyzers/base/AbstractCompilationAnalyzer.java
schroederc: Based on the last paragraph on this page, this should not reset the interrupt bit since, by convention, the interrupt bit should be cleared when throwing an `InterruptedException`: https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html

schroederc:
> Can you say more what you mean by "allow", and why it matters?

I added more to the description.

Issue #2881, opened 2018-07-30
Support Go 1.11 modules


schroederc:
The new Go modules system debuting in 1.11 requires changes to all developer tools. 
 There's a supported migration path using the https://godoc.org/golang.org/x/tools/go/packages package.

schroederc:
Note: Kythe itself should be able to build as a module after https://github.com/google/kythe/pull/2839

schroederc:
The https://godoc.org/golang.org/x/tools/go/packages package has a nice interface, but it lacks the details necessary for a full extraction.  However, that package is really just a wrapper around calling `go list` (or `$GOPACKAGESDRIVER`).  I have a prototype of the `gotool` extractor that calls out to the `go` tool in the same way for extraction, and I believe that's the best path forward to handle all of Go's corner-cases (including the new module support).

PR #3338, opened 2018-12-21
feat(gotool_extractor): use Go tool to extract package data

schroederc:
As a consequence, the `gotool` extractor now supports extracting Go 1.11 modules (#2881).

### DIFF
--- a/kythe/go/extractors/golang/golang.go
+++ b/kythe/go/extractors/golang/golang.go
@@ -196,10 +196,12 @@ func (e *Extractor) dirToImport(dir string) (string, error) {
 	return dir, nil
 }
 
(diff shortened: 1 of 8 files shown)


schroederc:
@creachadair

kythe/go/extractors/golang/golang.go
```diff
@@ -200,21 +200,38 @@ func (e *Extractor) dirToImport(dir string) (string, error) {
 // If the package has already been located, its existing package is returned.
```
danielmoy-google: does the comment up here need updating at all?

schroederc: I added a note about multiple packages being resolved.

Issue #3306, opened 2018-12-10
RFC: A(nother) uniform analyzer-driver interface


jaysachs:
# RFC: A(nother) uniform analyzer-driver interface

## Evolution

This proposal is an evolution of [RFC: A uniform analyzer-driver interface](#2801), and shamelessly incorporates large portions verbatim. The primary differences are:

1. Analyzers retrieve file contents from the filesystem and emit output to the filesystem; we essentially push the burden of virtualizing the filesystem from the analyzers (of which there are many, in many languages) to the drivers (which we expect there to be far fewer, in a small subset of languages).

2. We assume the the reduction of standardization on a single representation for compilation units as kzip to be accepted and implementation in-progress.

## Background

Kythe defines a loosely-coupled interface between build tools and analyzers. There are broadly two components of interest: _Extractors_ that consume source repositories and emit compilation records, and _analyzers_ that consume compilation records and emit analysis results (such as Kythe graph entries).

A **compilation record** in Kythe consists of a [kythe.proto.CompilationUnit][kcu], together with the contents of all the files referenced as required inputs to that unit. Work is underway to standardize on one format ([kzip][]), though two legacy formats ([indexpack][] and [kindex][]) still have some usage.

## Goals

There are two goals we'd like to address in this space:

1. To present analyzer authors with the most intuitive interface for obtaining artifacts for analysis (ultimately sourced from compilation records). Analyzers are often based on or derived from compilers, which typicially interact with file system APIs for obtaining sources.

2. To insulate analyzer authors from (changes in) the storage format of compilation records. Currently there is no way to implement an analyzer without supporting at least one of the existing file formats, and thus each analyzer must—unnecessarily, we argue—subsume some important implementation details of the analysis infrastructure.

We would like analyzers to have a simple and uniform interface to the rest of Kythe. Moreover, we would like that interface to depend as little as possible on concerns that are not relevant to the analyzer, such as how the compilation records are stored, how files are stored, cached, access-controlled, and fetched, and how the results of analysis are captured, stored, or propagated.

## Proposal

We propose that the interface between an analyzer and the rest of the Kythe infrastructure should be implemented as a text-based client-server protocol between the _analyzer_ (which is the client) and a _driver_ (which is the server). In summary, the interaction between analyzer and driver will work as follows:

1. The analyzer requests work from the driver, and the driver replies with a `AnalysisWorkUnit` message to be processed.

[...]

PR #3339, opened 2018-12-21
An alternate RFC for common analyzer/driver interface.

jaysachs:
This document serves to satisfy issue #3306, and is an evolution of [this RFC](https://github.com/kythe/kythe/commits/master/kythe/docs/rfc/2966.md). It defines the
basic terminology, channel framing, message formats, and methods for a driver
and analyzer to communicate uniformly via pipes or other reliable byte streams for the "control plane". File contents and analyzer output are satisfied via the analyzer's filesystem, presumably virtualized to some degree by the driver. It furthermore attempts to apply to less kythe-specific applications.


kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+
+The basic types referenced above are:
+
+-   `<analysis-id>`: an integer ID assigned by the server (JSON: number).
+-   `<bytes>`: a base64-encoded byte string (JSON: string).
+-   `<diagnostic>`: a `kythe.proto.common.Diagnostic` message (JSON: object).
+-   `<digest>`: a lower-case hex-coded SHA-256 digest (JSON: string).
+-   `<file-path>`: a file path (JSON: string).
+-   `<protocol>`: a protocol version label (JSON: string).
+-   `<analysis-type>`: an identifier for a specific type or types of analyses.
```
schroederc: How are type**S** of analyses specified?  CSV?

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+    
+    The reply includes the "root" file on which to begin analysis, as well as
+    the output file to which to emit output. The driver (and driver environment)
+    MUST make contents available for each file referenced in the compilation
+    record at the file system path specified in the compilation record.
+
+    Example:
+
+        » {"jsonrpc":"2.0","id":2,"method":"analyze","params":{"type":"kythe/indexer/go"}}
+        « {"jsonrpc":"2.0","id":2","result":{"id":101,"unit":{...}}}
```
schroederc: Update example response.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+-- Protocol handshake.
+
+» {"jsonrpc":"2.0", "id":1, "method":"init", "params":{"protocol":"kythe1","output-encoding":"protobuf"}}
+« {"jsonrpc":"2.0", "id":1, "result":{"protocol":"kythe1"}}
+
+-- Request two analysis tasks. The analyzer may process as many analysis
+-- tasks concurrently as it wishes.
+
+» {"jsonrpc":"2.0", "id":2, "method":"analysis", "params":{"type":"kythe/indexer/go"}}
+« {"jsonrpc":"2.0", "id":2, "result":{"id":101, "unit":{...1...}}}
```
schroederc: Update example response.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+» {"jsonrpc":"2.0", "id":3, "method":"analysis", "params":{"type":"kythe/indexer/go"}}
+« {"jsonrpc":"2.0", "id":3, "result:"{"id":384, "unit":{...2...}}}
```
schroederc: Update example response.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+
+    Request a fresh compilation record from the environment, blocking until one
+    is available. The environment assigns a unique integer analysis ID to each
+    request, distinct from any other ID currently assigned to that analyzer.
+    Once an analysis ID has been assigned, it is designated as **pending** until
+    the analyzer sends a `done` notification (see below) for that ID. A driver
+    SHOULD NOT issue the same compilation record again while there is a pending
+    analysis ID for that record.
+    
+    The reply includes the "root" file on which to begin analysis, as well as
```
schroederc: Elaborate on what is in this "root" file.  I'm assuming it depends on both the analyzer type and the output encoding but more detail is needed.

Add an explicit example for Kythe analyses.  Where is the CompilationUnit?  What structure is assumed for the required inputs?  By digest?  By path?  Both?

Also, at least note what "other" analyzer may do as well.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+    the output file to which to emit output. The driver (and driver environment)
+    MUST make contents available for each file referenced in the compilation
```
schroederc: Add note that contents need not be literally copied.  FUSE, hardlinks, or symlinks (etc.) may be used.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+    Request a fresh compilation record from the environment, blocking until one
+    is available. The environment assigns a unique integer analysis ID to each
+    request, distinct from any other ID currently assigned to that analyzer.
+    Once an analysis ID has been assigned, it is designated as **pending** until
+    the analyzer sends a `done` notification (see below) for that ID. A driver
+    SHOULD NOT issue the same compilation record again while there is a pending
+    analysis ID for that record.
+    
+    The reply includes the "root" file on which to begin analysis, as well as
+    the output file to which to emit output. The driver (and driver environment)
```
schroederc: Elaborate on the format of the output file and how it depends on the analysis request.

Add an explicit example for Kythe analyses.  Also, at least note what "other" analyzer may do as well.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+    {
+      "type": <analysis-type>
+    }
+    ```
+
+-   `AnalyzeReply`
+
+    ```
+    {
+      "id": <analysis-id>,
```
schroederc: Can the compilation unit remain as part of the `AnalyzeReply`?  Or must it be part of the FS?

kythe/docs/rfc/3339.md
creachadair: It should probably be a list if we want multiple values, that will make it compatible with a protobuf `repeated` field.

kythe/docs/rfc/3339.md
jaysachs: Changed so that the non-file part of the CU is in the AnalyzeReply.

kythe/docs/rfc/3339.md
jaysachs: Reworked the AnalyzeReply and added the "vname" API which should take care of this.

kythe/docs/rfc/3339.md
jaysachs: Let me know what you think here.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,308 @@
...
+### Messages
+
+The following message structures are used by the protocol:
+
+-   `InitRequest`
+
+    ```
+    {
+      "protocol": <protocol>
+      "output-encoding": <output-encoding>
```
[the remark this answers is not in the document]
jaysachs: Done.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+    Once an analysis ID has been assigned, it is designated as **pending** until
+    the analyzer sends a `done` notification (see below) for that ID. A driver
+    SHOULD NOT issue the same compilation record again while there is a pending
+    analysis ID for that record.
+    
+    The reply includes the input file(s) on which to begin analysis,
+    as well as the output file to which to emit analysis output. It
+    reflects some of the kythe.proto.CompilationUnit structure. The
+    driver (and driver environment) MUST make contents available for
+    each file referenced in the compilation record at the file system
```
creachadair: This is actually a little tricky. The compilation unit records _physical_ paths (in the `info` field), as they were spelled by the build system at the moment when the data were captured. Those paths often assume the existence of a file tree with symlinks pointing outside the repository root.

To actually set that up you'd need to solve a set of constraints: Your virtual (or actual) filesystem root has to be high enough that all the up-pointers have something to traverse, and that down-steps are correctly populated within that hierarchy. 

If the analyzer already virtualizes file access, then this is easy: You just map the path it requests directly to the data in the compilation record. But allowing the analyzer to access the filesystem through the system API means you have to actually create those structures (links, directories, etc.) and get their access permissions semi-right.

This isn't a dealbreaker, but it is an aspect of this design that requires much more detail that the old one didn't need.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+    path specified in the compilation record.
+
+    The output file should contain the content appropriate to the
+    analysis-type, encoded as specified in the `InitRequest`. For
+    Kythe index analyses, this would be `kythe.common.Entry`
+    objects. Other analyziers might emit a "findings" structure.
```
creachadair:
```suggestion
     objects. Other analyzers might emit a "findings" structure.
```

kythe/docs/rfc/3339.md
jaysachs: I want to make sure I understand your point here. If you're "simply" saying that the driver may need to virtualize an arbitrary portion of the filesystem, I agree. That's effectively what the in-analyzer file system virtualization is doing.

What I (naively) imagined is that we'd set up the virtualization mapping precisely for every path mentioned in the compilation record. I believe(d) that this is independent of symlinks, or permissions -- that the driver virtualization gets first crack at any path. 

If that's *not* the case, i.e. the driver is further downstream, then I agree there are some more subtleties here.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+-   `<analysis-id>`: an integer ID assigned by the server (JSON: number).
+-   `<bytes>`: a base64-encoded byte string (JSON: string).
+-   `<diagnostic>`: a `kythe.proto.common.Diagnostic` message (JSON: object).
+-   `<digest>`: a lower-case hex-coded SHA-256 digest (JSON: string).
+-   `<file-path>`: a file path (JSON: string).
+-   `<file-path-list>`: a list of `file-path` (JSON: list).
+-   `<vname>`: a `kythe.proto.VName` message (JSON: object).
+-   `<protocol>`: a protocol version label (JSON: string).
+-   `<analysis-type-list>`: a list of an identifier for a specific type or types of analyses.
+-   `<output-encoding>`: an identifier indicating how the output will be encoded, must be one of `json` or `protobuf` (JSON: string).
```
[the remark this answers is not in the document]
jaysachs: Would you consider it sufficient constraint that the underlying POSIX file descriptor be a pipe? Overly-constraining?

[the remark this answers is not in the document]
creachadair: You can stipulate that the output is append-only—no seek, no read. That would get you most of the way there. But it's also important to fix a format for the output, too.

kythe/docs/rfc/3339.md
jaysachs: I've spent a lot of time thinking about this, and I don't see the issue. But I'm new here, so clearly there's plenty of room for ignorance and misunderstanding ...

The compilation record stores, along with the byte content, the actual path (possibly relativized, and maybe bastardized for e.g. jar files) requested from the host OS, i.e. before any symlink/hardlink/overlay/blah is traversed/resolved. It's precisely that path that the virtualization layer will intercept requests for. Why is there anything to worry about here?

jaysachs: Ah. What you're pointing out is if the idea is to leave the "original compiler code" essentially intact, then the analyzer may do all sorts of operations other than open(name, "r") on the files it actually requires. And, the logic follows that we attempt to circumvent these other operations, then we're still doing some amount of intra-analyzer virtualization.

I wonder how much this actually occurs in practice, though.

creachadair: > I wonder how much this actually occurs in practice, though.

I don't know. But I have seen some compilers do some odd things with the filesystem in order to obtain locking, memory-map files, and so on. Hopefully most of that stuff is ancillary to reading input files and can be ignored—but I do think this spec will need to spend some more time laying out the rules for what the virtualization promises (and thus what the analyzer can rely on). When we were going to make the analyzer figure out how to inject itself into the I/O mechanism of the underlying compiler explicitly, these things didn't need much discussion.

shahms: Implementing any and all of the weird POSIX or non-POSIX-but-damn-the-torpedoes corner cases expected by compilers-used-to-implement-analyzers is unlikely to be feasible.  In the real world, I expect most compilers will fall into one of three categories:

1) Compilers which do weird corner-casey-crap but have a Filesystem abstraction which can be implemented by the analyzer to not do so (or use a custom protocol directly as would be required by RFC2966).
2) Compilers which do straightforward create/write open/read operations (but have no filesystem abstraction)
3) Lost causes

Both this RFC and RFC2966 can support 1 with roughly the same amount of work, but this or something similar is required for 2. Given this I'd be very tempted to piggyback on the operations supported by something like 9P2000 if only because there are a broad set of implementations on a lot of platforms and it has better specified (and more limited) semantics.  The "file operations" specified by RFC2966 are unlikely to be sufficient for pretending to be a filesystem.  That said, I suspect including an RPC for emitting analysis output rather than necessitating the using the filesystem for output (rather than only input) will be simpler and sufficient.

The driver will also have some limited ability to rewrite the paths it sends to the analyzer.  It *should* also have a pretty good idea of what files the analyzer may try to open as it shouldn't be opening anything not in the CU.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+-- Protocol handshake.
+
+» {"jsonrpc":"2.0", "id":1, "method":"init", "params":{"protocol":"kythe1","output-encoding":"protobuf"}}
+« {"jsonrpc":"2.0", "id":1, "result":{"protocol":"kythe1"}}
+
+-- Request two analysis tasks. The analyzer may process as many analysis
+-- tasks concurrently as it wishes.
+
+» {"jsonrpc":"2.0", "id":2, "method":"analysis", "params":{"types":["kythe/indexer/go"]}}
+« {"jsonrpc":"2.0", "id":2, "result":{"id":101, "input":"path/to/file1.go"}}
```
shahms: s/input/inputs/ (it's specified as a list)

this also doesn't include the output file path, which seems like it would be required to do anything meaningful?

kythe/docs/rfc/3339.md
jaysachs: I'll admit that the main reason I went with filesystem for output rather than RPC is pushback-on-multiple-encodings. I don't have a strong reason for pushing the filesystem as long as the output encoding can be negotiated (and efficiently supports proto delimited, in addition to JSON), except it keeps the rpc protocol simple and pure text.

creachadair: It is certainly true that using a for _output_ is a separate concern from using a virtual filesystem for _input_. I think the latter is the better-motivated case, and doesn't strictly need to be tied to the former.

creachadair: Either way though, my main point here was that the spec should probably say more about what the VFS will provide.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+    server.
+
+    The server MUST support concurrent JSON-RPC requests from the analyzer. The
+    server MUST support batch JSON-RPC requests from the analyzer.
+
+-   Under certain circumstances, the analyzer process **must exit**. In a case
+    where the analyzer process must exit, it MUST terminate within a grace
+    period. If it does not, the operating environment SHOULD forcibly terminate
+    the process (and may do so at any time after the grace period ends). The
+    exit code of an analyzer process is ignored. The grace period is defined by
```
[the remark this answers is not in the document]
jaysachs: I interpreted this paragraph as applying to the "certain circumstances", and those circumstances are when told by the analyzer to exit.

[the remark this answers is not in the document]
shahms: That's fair.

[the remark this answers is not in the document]
creachadair: That was the intent, but I agree I should probably have been more explicit.

kythe/docs/rfc/3339.md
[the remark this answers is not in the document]
jaysachs: So my inclination here is to use "standard" delimited proto output, and streaming concatenated JSON. Because it's a file sink, and write-only, I see no need for sync points or checksums. Does anyone have a strong opinions for/against?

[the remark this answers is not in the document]
creachadair: Go handles unframed JSON just fine, but other languages' JSON libraries do not. I recall that TypeScript explicitly had a problem with this, and we wound up requiring each message on a line (which is easy to do since there is no internal vertical whitespace in JSON tokens, so you need only remove optional whitespace).

On that basis I'd recommend using some framing—and if you are going to use varint-prefixed records for proto, you might as well for JSON too IMO.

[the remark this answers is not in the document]
jaysachs: This only really matters for the driver, so I think unframed is fine. But I went with newline-delimited since that seemed most tool-friendly. It's different than delimited proto, which is length-delimited, but the tool/library support is already there for delimited proto, so I'm fine with the discrepancy.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+encoded using the canonical
+[JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json)
+for the named message type.
+
+#### analysis-type
+
+Analysis types are described by the following grammar:
+
+```grammatical-framework
+analysis-type         = analysis-type-segment 0*{"/" analysis-type-segment}
```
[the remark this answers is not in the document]
jaysachs: It was, and a good suggestion to unify them.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,348 @@
...
+-   A **corrupt frame** is one in which the length tag does not have the correct
+    form. A frame is corrupt if the newline terminator is missing, the string
+    preceding the newline terminator contains non-digits or is empty, or the
+    stream ends before the specified number of bytes is delivered.
+
+    If the analyzer receives a corrupt frame from the driver, it must exit. If
+    the driver receives a corrupt frame from the analyzer, it must close the
+    input stream (in which case the analyzer therefore must exit).
+
+-   When an analyzer process terminates, the driver MUST consider any pending
```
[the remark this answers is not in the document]
jaysachs: I added a note about this exact use case.

kythe/docs/rfc/3339.md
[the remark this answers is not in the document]
creachadair: No, it matters for the analyzer too, which will need to be able to receive responses in the control channel shared with the driver.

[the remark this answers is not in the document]
creachadair: (responded to the wrong comment, replaced)

I think if you're going to choose a framing you should use the same regardless of format. There's no sensible reason to multiply code paths in this case.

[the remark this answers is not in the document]
jaysachs: I don't see the multiple code paths. Any given analyzer will emit exactly one of {json, protobuf}; nothing is shared between those two encodings.

Protobuf libraries already easily support reading and writing delimited, so that takes care of both producer (analyzer) and consumers (driver and tools). 

For JSON, I suspect there's more support for newline-delimited universally. Certainly you can take such a file and break apart the records on the command line.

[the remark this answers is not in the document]
creachadair: The driver will have to have multiple code paths for reading records.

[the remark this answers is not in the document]
creachadair: C++ has some built-in code for handling varint-prefixed records, but Go doesn't. So they're separate concerns: Reading the next record, vs. decoding it. Only one of them needs to switch, the other is a generic data-handling routine.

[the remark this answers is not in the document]
jaysachs: [Buffer.DecodeMessage](https://github.com/golang/protobuf/blob/master/proto/decode.go#L368)

I don't imagine a driver being written in other than C++/Go/Java.

Just to be clear, I don't feel super strongly about this.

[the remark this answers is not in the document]
jaysachs: I can't find a standard for length-encoded JSON. Just some implementations, which run the gamut of emitting the length as uint32BE, uint32LE, or even text. If you can point me to a specification, I'll adopt that.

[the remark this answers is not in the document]
creachadair: I don't see any reason not to use the same encoding as for protobuf messages: varint length + message. If support for command-line tools is important, the fact that you're allowing multiple different encodings means you'll want a translation tool anyway, so we can just provide that in the Kythe repo and not worry about it. That makes the encoding rule for output simpler: A stream of varint-prefixed records, containing either JSON or protobuf wire format. You can use the filename to distinguish which is which, perhaps.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+    analyzers in a sandbox.
+
+2.  To make it easier to run analyzers as standalone tools, and reduce or
+    eliminate the need to implement separate wrapper tools for testing vs.
+    production.
+
+3.  To reduce friction for analyzers written in languages that do not have good
+    support for efficient manipulation of binary data, or which require FFI to
+    get good performance.
+
```
jaysachs: So something I just realized about this proposal: it essentially prohibits an analyzer from performing concurrent analyses. The core of the reason is that the file system virtualization is not sensitive to which analysis is executing. It's definitely possible to change the virtualized file data in between sequential analyses, but concurrent analyses would see each others' input files. In the case where a generated file is generated from two different build configs and has different content, we'd have a correctness issue. Even outside that case, if the virtualization synthesizes directory listings, that could also cause problems. I suspect there are other potential issues here.

Is it worth while giving up on concurrent analyses?

creachadair: > Is it worth while giving up on concurrent analyses?

IMO yes. The existing driver only offers one analysis at a time to the analyzer, partly because in the past when we tried to do more we discovered many of the compiler libraries are not safe for concurrency (e.g., using static data and caches without locks). When I wrote #2966 I had in mind to allow the _analyzer_ to decide, but I don't actually expect that feature to be used much in practice, and I think it could be sacrificed.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+    hardlinks, symlinks, or other may be used.
+
+    Example:
+
+        » {"jsonrpc":"2.0","id":2,"method":"analyze","params":{"types":["kythe/indexer/go"]}}
+        « {"jsonrpc":"2.0","id":2","result":{"input":"path/to/main.go", "output":"path/to/output.json"}}
+
+-   `vname(VNameRequest)`
+
+    Requests the complete VName to emit for the specified path and signature.
```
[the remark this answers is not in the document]
jaysachs: Yes, it relieves the analyzer from having to thread through those two fields. More significantly, though, it opens up the possibility to handle the case when semantic entities "belong" to some other corpus and/or root: one could imagine wanting to associate files of dependencies with some "well known" corpus name, or that build-time generated files should have some specific root. In these cases, some component needs to "understand" the mapping of path to corpus/root, and it makes sense to insulate the myriad indexers from how to determine that mapping.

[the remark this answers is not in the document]
creachadair: The analyzer was already freed of the need to reason about vname mapping by the compilation record. As I understand it, the real change here is providing that information _dynamically_ during the analysis instead of ahead-of-time. That's actually an interesting change, and I'm not entirely clear on the motivation for it.

We talked about various ways to solve this issue in the original implementation, and settled on making the build system capture that knowledge in the CU (that's what the vname fields are for, in the CU and its required inputs). I'm not sure what the argument for requesting it on the fly is, but it's a viable approach. 

That said: We should probably discuss that question.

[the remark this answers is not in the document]
jaysachs: Right, that was too early in the AM for a reply. The per-file VName still would be present in the CU proto. The real reason for this API is that with the files being read from the file system, the only remaining motivation to expose the raw compilation unit mapping (FileInput) to the analyzer would be to supply the VName overrides. Thus the introduction of this API to encapsulate things. While it does permit dynamic non-CU-supplied VName construction, that wasn't its primary intention.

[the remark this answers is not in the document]
creachadair: Ah! Yes, that makes sense.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+    environment does not reply.
+
+    Example:
+
+        » {"jsonrpc":"2.0","method":"log","params":{"message":{...}}}
+
+-   `done(LogRequest)`
+
+    A notification from the analyzer to the environment, signifying that the
+    analyzer has completed work on the specified analysis ID. The environment
```
clconway: Where is the analysis ID specified? It looks like the LogRequest.message field is free-form?

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+    The reply includes the input file(s) on which to begin analysis, as well as
+    the output file to which to emit analysis output. It reflects some of the
+    kythe.proto.CompilationUnit structure. The driver (and driver environment)
+    MUST make contents available for each file referenced in the compilation
+    record at the file system path specified in the compilation record.
+
+    The output file should contain the content appropriate to the
+    analysis-type, encoded as specified in the `InitRequest`. For Kythe index
+    analyses, this would be `kythe.common.Entry` objects. Other analyzers might
+    emit a "findings" structure.
```
clconway: This freedom is going to add complexity. Maybe the canonical output should be a gRPC Any proto? Or a finding proto with an Any field that can be used to encode Entries, etc.?

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+    analyzer has completed work on the specified analysis ID. The environment
+    does not reply. After receiving this notification, the driver MUST discard
+    any further requests for the analysis ID (and return an error, if
+    appropriate) unless and until it issues a new analysis reusing that ID.
+
+    Example:
+
+        » {"jsonrpc":"2.0","method":"done","params":{"message":"success"}}
+
+### Errors
```
clconway: How does the analyzer communicate that it had a unrecoverable internal error?

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+-   For each file path occurring in the compilation record, the driver MUST 
+    support `open()`, `read()` and `lseek()` (or equivalent) operations. The
+    contents backing such operations MUST be precisely the contents specified
+    in the compilation record. Other file operations MAY be supported.
+
+-   For each directory implied by the file paths in the compilation record, the
+    driver MAY support a subset of read-only directory operations, e.g. in order
+    to provide directory listings via `readdir(3)` or equivalent.
+
+-   For all other paths, the driver MAY fall back to the underlying host
```
clconway: I may be misunderstanding: does this mean to imply that paths in the compilation record MUST NOT fall back to the underlying host? Or is it just saying that access to the filesystem is not necessarily limited to compilation record paths?

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,389 @@
...
+## Output
+
+The output file MUST be treated as append-only by the analyzer, with no seek,
+read or other operations permitted. Data MUST written to the output file based
+on the encoding negotiated in the Init request/response.
+
+-   For "json" encoding, the analyzer MUST write the output as
+    [varint](https://developers.google.com/protocol-buffers/docs/encoding#varints) length-prefixed records.
+
+-   For "protobuf" encoding, the analyzer MUST write the output as [delimited
```
clconway: This is under-specified. What is the format of the message size delimiter?

kythe/docs/rfc/3339.md
jaysachs: Where do you see the complexity? I'd expect the driver would know a priori what type of data to expect.

kythe/docs/rfc/3339.md
jaysachs: The current proposal is that it terminates with a non-zero exit code. Do you see cases where an analyzer would want to abort the current analysis (e.g. due to unrecoverable malformed input) but continue to be able to request analyses, without restarting? Is that common enough that it warrants additional complexity?

kythe/docs/rfc/3339.md
jaysachs: The latter. Does the "MAY" not convey that sense sufficiently? I'm open to other wording.

kythe/docs/rfc/3339.md
jaysachs: Whatever [standard proto delimited methods](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/MessageLite#writeDelimitedTo-java.io.OutputStream-) do. Is there a canonical place that's documented?

kythe/docs/rfc/3339.md
shahms: Maybe if we can get more people piling on https://github.com/protocolbuffers/protobuf/issues/5226 we can get a better resolution there (I'm not optimistic, but still...)

kythe/docs/rfc/3339.md
clconway: I more had in mind that the analyzer could return a stack trace or other rich description of the error condition, which could be cleaner than scraping stderr.

shahms: While some analyzers may be able to return structured data for internal errors quite a few of them will segfault and (if lucky) print a stack trace on stderr, so if we want to capture that we'll have to scrape stderr regardless.

creachadair: There are different ways an analyzer can fail. If it fails under control, it can report an error in-channel, which along with the `"done"` message. If it fails out-of-control, however, the program will simply exit. The protocol needs to account for both.

Even when an analyzer _tries_ to control failures, it can't always. Even in a managed runtime the OS may kill a process for exceeding resource limits or poor use of FFI. Moreover some compiler toolchains do not manage memory well (GHC, Python, Dart, and SBCL for some examples we've encountered in the past) and will pile up state and garbage if you re-use the analyzer without an exit in between.

So at some point you have to deal with the fact that the program either exits uncontrollably, or exits because it is simpler to exit and restart than to rewrite the compiler. I think the protocol should deal with that as simply as possible—at which point reading the exit code is probably the best we can hope for. Scraping stderr is probably not useful for an in-band message, though it makes sense to capture stderr for logging purposes.

kythe/docs/rfc/3339.md
clconway: I am perhaps being overly literal and getting myself confused. Since falling back to the underlying host is not mentioned above, one might infer it is not allowed. It may help to replace "For all other paths..." with "For paths that are available via the host's local filesystem..."

kythe/docs/rfc/3339.md
clconway: There doesn't seem to be a canonical reference for this and there doesn't appear to be an equivalent of writeDelimitedTo() in every language (even though it's the canonical representation of an embedded message).

jaysachs: I specified varint, without any canonical reference.

kythe/docs/rfc/3339.md
```diff
@@ -0,0 +1,391 @@
...
+    analysis requests to have failed, and thereafter the environment MAY retry
+    analysis of the corresponding compilation records.
+
+-   If the analyzer voluntarily (i.e. not at behest of the driver) exits
+    succesfully (e.g. exit code of 0) with no outstanding analyses, the driver
+    SHOULD cause a new instance of the analyzer to start. This is motivated by
+    analyzers for which the best, or only, mechanism for clearing state is to
+    restart.
+
+### Messages
```
shahms: (aside: it would be nice to specify the protocol using proto, even if the implementation is JSON-RPC)

kythe/docs/rfc/3339.md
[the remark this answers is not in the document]
shahms: It also turns out that it's incredibly common to get this mapping wrong, particularly for generated files and (presumably) files from a different corpus.  Indexer authors routinely emit references using the raw "path" disregarding the CU/required_input VName, often up to hard-coding various values for "corpus" as well.  Abstracting this behind an interface should substantially help with this problem.

That said, I think the generic `vname()` interface may still be a bit too low-level.  There are a variety of other node kinds which have structural, factual or other VName requirements (tapp comes to mind), so it might be nice to express those requirements via an RPC/proto interface in some fashion.

But file/anchor VNames are the low-hanging fruit here.

kythe/docs/rfc/3339.md
creachadair: Yeah, that's what kythe/proto/driver.proto was meant to represent, but I think it may be out-of-date with the current proposal.

shahms:
Jay, is this still relevant?

jaysachs:
Yes it is, though very low priority.

creachadair:
> Yes it is, though very low priority.

Priority notwithstanding, I'd be interested in helping some version of this happen. I'd be happy to serve as a sounding-board for ideas if that would help.

PR #3369, opened 2019-01-15
update llvm to r350964 

shahms:
This also switches getSourceText to using the upstream static member function in Lexer, originally to workaround the bug in r350891 but now just to accomplish the same goal in cleaner way.

As `getThisType()` no longer accepts an ASTContext parameter, remove the extra Kythe plumbing passing it around.

### DIFF
--- a/kythe/cxx/indexer/cxx/ImputedConstructorSupport.cc
+++ b/kythe/cxx/indexer/cxx/ImputedConstructorSupport.cc
@@ -33,9 +33,8 @@ namespace {
 
 class TargetTypeVisitor : public clang::RecursiveASTVisitor<TargetTypeVisitor> {
  public:
-  explicit TargetTypeVisitor(clang::QualType target_type,
-                             clang::ASTContext* context)
-      : context_(context), target_type_(std::move(target_type)) {}
+  explicit TargetTypeVisitor(clang::QualType target_type)
+      : target_type_(std::move(target_type)) {}
 
   const clang::CXXConstructorDecl* FindConstructor(clang::Stmt* stmt) {
     Enqueue(stmt);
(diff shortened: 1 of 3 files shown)


shahms:
> Could you please update the commit description before you merge to mention briefly what's changing? (Obviously the update itself isn't a problem, but the fact that it requires surgery probably bears recording)

I'm not sure what you're saying here.  It doesn't require "surgery" any more than any upstream LLVM integration typically requires changes.

creachadair:
> I'm not sure what you're saying here. It doesn't require "surgery" any more than any upstream LLVM integration typically requires changes.

Some of our functions are changing signatures—that presumably isn't forced upon us by the LLVM change, so I assume there's some secondary reason for it (e.g., the API no longer requires that argument, for example).

shahms:
> > I'm not sure what you're saying here. It doesn't require "surgery" any more than any upstream LLVM integration typically requires changes.
> 
> Some of our functions are changing signatures—that presumably isn't forced upon us by the LLVM change, so I assume there's some secondary reason for it (e.g., the API no longer requires that argument, for example).

The Clang API changed to no longer accept that argument.  The only reason for the argument in the Kythe functions was to pass it to the Clang API. While we can certainly continue to pass an unused argument through the other API layers it seems ... silly?

creachadair:
> The Clang API changed to no longer accept that argument. The only reason for the argument in the Kythe functions was to pass it to the Clang API. While we can certainly continue to pass an unused argument through the other API layers it seems ... silly?

No doubt! I just meant could you please describe that in the description? E.g., LLVM removed the context argument from blah so we no longer need to pass it. Or words to that effect.

shahms:
update to r351210, PTAL

PR #3370, opened 2019-01-15
refactor: move more path utils to llvm-free common library

justbuchanan:
* Removed duplicate `CleanPath()`
* Moved `MakeCleanAbsolutePath()`, `RelativizePath()` to common path_utils library
* Added `IsAbsolutePath()`, `GetCurrentDirectory()` to replace our usage of the corresponding clang functions
* Removed cxx_extractor/path_utils_test.cc because its contents are now in common/path_utils.cc

I'm going to need `RelativizePath()` in the proto extractor to support the `KYTHE_ROOT` config option, which is why I'm making these changes.

The release targets currently fail because the proto indexer has its own implementation of `GetCurrentDirectory()`. I'll remove it from there and make it depend on the shared utils here.


kythe/cxx/common/path_utils.cc
```diff
@@ -71,4 +76,73 @@ std::string CleanPath(absl::string_view input) {
...
+    } else if (errno == ERANGE) {
+      len += len;
+      buffer = absl::make_unique<char[]>(len);
+    } else {
+      return false;
+    }
+  }
+}
+
+std::string MakeCleanAbsolutePath(const std::string& in_path) {
```
shahms: absl::string_view parameter?

kythe/cxx/common/path_utils.cc
```diff
@@ -71,4 +76,73 @@ std::string CleanPath(absl::string_view input) {
...
+      LOG(ERROR) << "Unable to get current working directory";
+      return "";
+    }
+    abs_path = JoinPath(dir, in_path);
+  }
+  return CleanPath(abs_path);
+}
+
+std::pair<absl::string_view, absl::string_view> SplitPath(
+    absl::string_view path) {
```
shahms: Is this conforming to an existing function? If not, prefer to use a named struct with meaningful members rather than `std::pair`.  `.first` and `.second` are pretty hard to read at callsites.

kythe/cxx/common/path_utils.cc
justbuchanan: :+1:

kythe/cxx/common/path_utils.cc
justbuchanan: No particular interface, I just borrowed this from google3. I added a `PathParts` struct to name the parts of the return value.

PR #3371, opened 2019-01-15
build: fix buildbot scripts to work on darwin

schroederc:


kythe/release/appengine/buildbot/cache-llvm.sh
```diff
@@ -32,6 +32,15 @@
 
 CACHE="$HOME/.cache/kythe-llvm"
 
+copy() {
+  if which gcp &>/dev/null; then
```
creachadair: I assume this is because macOS `cp` doesn't support `-l`?

I think you're going to have to add `coreutils` to the installation list, because it isn't by default even if Homebrew is present.

PR #3380, opened 2019-01-17
fix(gotool_extractor): mark dep as missing if it cannot be built

schroederc:


kythe/go/extractors/golang/golang.go
```diff
@@ -514,7 +514,7 @@ func (p *Package) addDeps(cu *apb.CompilationUnit, importPaths []string, localPa
 	for _, ip := range importPaths {
 		if ip == "unsafe" {
 			// package unsafe is intrinsic; nothing to do
-		} else if dep, err := p.ext.addPackage(ip, localPath); err != nil {
+		} else if dep, err := p.ext.addPackage(ip, localPath); err != nil || dep.PkgObj == "" {
```
creachadair: I would expect a package that doesn't build to report an error from addPackage (specifically, from the call to Import). In what circumstances does it report success but not have a package?

schroederc: Dependencies are not necessarily added here anymore.  The standard Locate method now also adds packages for dependencies and keeps around failed builds.  They're not reported as successful in Locate, but addPackage just returns whatever is in the cache as a success.

creachadair: Ah, I see. I guess we knew the Locate method was going to change semantics, but would you mind adding a comment to that effect nearby?

PR #3384, opened 2019-01-18
build: add java_proto_verifier_test Starlark rule

schroederc:
java_proto_verifier_test extracts/analyzes/verifies protocol buffer
sources, their generated Java sources, and Java sources depending on the
generated sources as test of the cross-language references between Java
and proto.

A single test is implemented as a proof-of-concept for the Starlark
rule.  More tests will follow.

### DIFF
--- a/tools/build_rules/verifier_test/java_verifier_test.bzl
+++ b/tools/build_rules/verifier_test/java_verifier_test.bzl
@@ -224,11 +224,20 @@ def _generate_java_proto_impl(ctx):
     # Generate the Java protocol buffer sources into a directory.
     # Note: out contains .meta files with annotations for cross-language xrefs.
     out = ctx.actions.declare_directory(ctx.label.name + "_gen")
-    ctx.actions.run(
+    protoc = ctx.executable._protoc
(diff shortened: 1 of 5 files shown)


tools/build_rules/verifier_test/java_verifier_test.bzl
```diff
@@ -224,11 +224,20 @@ def _generate_java_proto_impl(ctx):
...
+    ctx.actions.run_shell(
         outputs = [out],
         inputs = ctx.files.srcs,
-        executable = ctx.executable._protoc,
-        arguments = ["--java_out=annotate_code:" + out.path] + [src.path for src in ctx.files.srcs],
+        tools = [protoc],
+        command = "\n".join([
+            "#/bin/bash",
+            "set -e",
+            "mkdir -p " + out.path,
```
shahms: can't you just run this as a separate command rather than writing a whole tiny shell script for this?

shahms: That's a good point, maybe include a comment referencing the upstream issue around directories?

PR #3390, opened 2019-01-22
fix: fix a typo in a comment

danielmoy-google:
I accidentally closed pull/3379 instead of merging it, this is the same.


danielmoy-google:
Ok time to hit the *green* button, not the button next to the green button

PR #3391, opened 2019-01-22
build: allow non-tests to depend on java_proto_verifier_test kzips

schroederc:


schroederc:
For the internal fingerprinting.

PR #3399, opened 2019-01-24
fix: use static_cast to avoid sign-compare warning

justbuchanan:


justbuchanan:
The warning might already be disabled in this repo actually - I've only noticed it showing up when I use this from lang-proto. Do you know where the config is to suppress warnings?

schroederc:
Using `bazel test --client_env=CC=clang` or `CC=clang bazel test //...` should force Bazel to use clang locally.  As a note, Buildbot (and `--config=remote`) already tests using clang.

PR #3400, opened 2019-01-24
fix(gotool_extractor): fix file paths with --canonicalize_package_corpus

schroederc:


kythe/go/extractors/govname/govname.go
```diff
@@ -152,3 +151,57 @@ func ForStandardLibrary(importPath string) *spb.VName {
...
+type repoRootCacheNode struct {
+	root *vcs.RepoRoot
+
+	children map[string]*repoRootCacheNode
+}
+
+// add puts the given *vcs.RepoRoot into the prefix tree for the given path
+// components.
+func (n *repoRootCacheNode) add(components []string, r *vcs.RepoRoot) {
+	if len(components) == 0 {
```
danielmoy-google: so to be clear, a second call to .add() with another empty components would reset root, and that's expected?

schroederc: Yup.

PR #3403, opened 2019-01-25
fix: adding example config for guava-android, distinct from guava

danielmoy-google:


danielmoy-google:
Added some bits in readme too.

PR #3408, opened 2019-01-28
fix(go): use GoPackageInfo to resolve input import paths

schroederc:
This allows a required input's VName to differ from the import path used to refer to it within a CompilationUnit (e.g. with `--canonicalize_package_corpus`).


kythe/proto/go.proto
```diff
@@ -36,3 +36,9 @@ message GoDetails {
   // Whether cgo is enabled for this compilation.
   bool cgo_enabled = 7;
 }
+
+// GoPackageInfo provides details about a Go package.  This may be in relation
+// to a CompilationUnit as a whole or specific required input.
+message GoPackageInfo {
```
creachadair: Is there some reason this isn't included as part of the `GoDetails`? Since the tags, CGo, and other environment bits there are also per-package, it's not obvious that this shouldn't be there too.

schroederc: This message can also be attached to a required input.

schroederc: Unless you mean to nest this message as a field within `GoDetails`.  In which case, that would also be fine.

creachadair: I meant more the first. So it's not in the `GoDetails` because in the latter case you don't know the compilation details, apart from the package name?

schroederc: It's more about the context of the message. `GoPackageInfo` may annotate either the compilation unit as a whole or a required input.  A `GoDetails` is only attached to the compilation unit.

creachadair: Right, that's what I mean: In both cases you're referring to a package, but in one case you know the compilation details for that package (in its own CU) and in the other you have only its complete form (as a dependency via required_input).

Anyway, that makes sense.

Issue #3405, opened 2019-01-25
preprocessor needs to be smarter about avoiding unnecessary modfiications


danielmoy-google:
Stemming from #3404, it preprocessor should at least look at `pluginManagement` before making modifications.

PR #3409, opened 2019-01-28
fix: support pluginManagement in mvn.pom preprocessor

danielmoy-google:
Support both ways:

1. Parent pom.xml files which specify `maven-compiler-plugin` inside of
   `pluginManagement/plugins/plugin` instead of `plugins/plugin` should
   be correctly handled, and verified.
2. Child pom.xml files which specify a bare `maven-compiler-plugin` that
   references a parent pom.xml specification in `pluginManagement`
   should also be skipped.

Caveat: In the latter case there, we don't yet have support for
verifying that the parent pom.xml file even exists, let alone has the
correct reference for the child.

Addresses a bit of #3405 and #3404.


kythe/go/extractors/config/preprocessor/modifier/modifier_test.go
```diff
@@ -42,6 +43,7 @@ func TestBuildGradleHasKythe(t *testing.T) {
 		{fileName: "other-build.gradle", javacWrapper: "/tmp/javac-wrapper.sh", wantError: true},
 		// Look at other-build.gradle but use the correct javac wrapper.
 		{fileName: "other-build.gradle", javacWrapper: "/different/javac-wrapper.sh", hasWrapper: true},
+		{fileName: "other-build.gradle", javacWrapper: "/different/javac-wrapper.sh", hasWrapper: true},
```
danielmoy-google: I forget what I was doing here, but this pull should be maven only, so axing this

kythe/go/extractors/config/preprocessor/modifier/pom_xml_modifier.go
```diff
@@ -50,10 +56,128 @@ func PreProcessPomXML(pomXMLFile string) error {
 	return f.Close()
 }
 
-func hasCompilerPlugin(doc *etree.Document) bool {
-	for _, p := range doc.FindElements("//project/build/plugins/plugin/artifactId") {
-		if p.Text() == "maven-compiler-plugin" {
+// Returns whether there is a maven-compiler-plugin specified already.
```
schroederc: `// hasCompilerPlugin returns whether...`

kythe/go/extractors/config/preprocessor/modifier/pom_xml_modifier.go
```diff
@@ -50,10 +56,128 @@ func PreProcessPomXML(pomXMLFile string) error {
...
+			// sibling elements.
+			return true, wellformedCompilerPlugin(p)
+		}
+	}
+
+	// If we didn't find anything, we can proceed with modification.
+	return false, nil
+}
+
+// Check to see if the plugin fits the minimum supported versions we've
```
schroederc: `// wellformedCompilerPlugin checks...`

kythe/go/extractors/config/preprocessor/modifier/pom_xml_modifier.go
```diff
@@ -50,10 +56,128 @@ func PreProcessPomXML(pomXMLFile string) error {
...
+	if v, err := javaVersion(k.Text()); err != nil || v.earlierThan(versiontuple{[]int{1, 7}}) {
+		// TODO(#3075): Should test an upper bound here too.
+		return fmt.Errorf("unsupported %s version: %s, need 1.7+", kind, k.Text())
+	}
+	return nil
+}
+
+// Some container that supports semvar (3.7.0) and java versions (1.6) for
+// comparison purposes only.
+type versiontuple struct {
```
schroederc: This can probably just be `type versionTuple []int`.

danielmoy-google: ah yea.  Originally I had this only for semvar and it made sense with the struct, but after refactor to handle both semvar and java version in the same thing, the struct makes very little sense.

Thanks.

kythe/go/extractors/config/preprocessor/modifier/pom_xml_modifier.go
danielmoy-google: :+1:

kythe/go/extractors/config/preprocessor/modifier/pom_xml_modifier.go
danielmoy-google: :+1:

PR #3413, opened 2019-01-30
feat(bazel_extractor): infer corpus from source files

schroederc:
This allows a caller to push all corpus configuration logic to the VName rules file.

### DIFF
--- a/kythe/go/extractors/bazel/extractor.go
+++ b/kythe/go/extractors/bazel/extractor.go
@@ -198,6 +198,10 @@ func (c *Config) extract(ctx context.Context, info *ActionInfo, file fileReader)
 		return nil, err
 	}
 
+	if c.Corpus == "" {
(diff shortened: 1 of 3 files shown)


kythe/go/extractors/bazel/extractor.go
```diff
@@ -275,6 +279,31 @@ func (c *Config) fetchInputs(ctx context.Context, paths []string, file func(int,
...
+			continue
+		}
+
+		vname, ok := c.Rules.Apply(path)
+		if ok && vname.Corpus != "" {
+			sourceCorpora.Add(vname.Corpus)
+		}
+	}
+
+	corpora := sourceCorpora.Elements()
```
creachadair: Could more simply be:

```go
if sourceCorpora.Len() != 1 {
  …
}
corpus, _ := sourceCorpora.Pop()
```

if you like.

PR #3414, opened 2019-01-30
build: split vnames config by language with configurable corpus

schroederc:


creachadair:
As a request for future changes, could you please include some more context in the descriptions for some of these changes? The force of the change is clear enough, but the rationale isn't at all.

PR #3422, opened 2019-01-31
Move LLVM dependency to an external repository

shahms:
The generated_cmake_targets.bzl file was generated by a separate tool(s) which I'm working on getting in shape to push to a separate repository, but for now this should get the LLVM/Clang build working as an external repo, obviating and removing the `tools/modules/` scripts.

I'm currently running the tests on Darwin, but so far things look relatively good.


tools/build_rules/llvm/cmakedefines/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["cmakedefines.go"],
+    importpath = "github.com/shahms/cmaketobzl/tools/cmakedefines",
```
jaysachs: Is that the canonical location?

shahms: That was where it was during development, but the current location is canonical for now (although fairly academic as nothing should reference it by that name).

tools/build_rules/llvm/cmakedefines/cmakedefines.go
```diff
@@ -0,0 +1,203 @@
...
+	case bool:
+		*cv = cmakeValue(strings.ToUpper(fmt.Sprint(v)))
+	case float64:
+		*cv = cmakeValue(fmt.Sprint(v))
+	case string:
+		*cv = cmakeValue(v)
+	case nil:
+		*cv = ""
+	default:
+		return fmt.Errorf("expected bool, string or number, found: %v", data)
```
schroederc: You can use `%T` to print the type.

shahms: Done.

PR #3429, opened 2019-02-01
build: update to tagged versions of stringset and shell

creachadair:
Rather than pinning to specific random commit hashes, point to the release tags
for these libraries. This required a small change to the shim, which did not
handle the tag/commit combination quite correctly.

More libraries could probably productively be pinned this way, and it is
generally safer to do so since that's how Go modules will do it too. For now,
though, I'm just doing the ones I know the versions for.


creachadair:
> Please update the `go.mod` module file as well.

What is the best way to do that? Last time I tried doing a `go build` and `go mod tidy` it made a lot of unrelated changes.

creachadair:
> > Please update the `go.mod` module file as well.
> 
> What is the best way to do that? Last time I tried doing a `go build` and `go mod tidy` it made a lot of unrelated changes.

Nevermind, I was able to update those packages individually without breaking anything.

schroederc:
```
git checkout go.mod
rm -f go.sum
go mod edit --require=bitbucket.org/creachadair/stringset@v0.0.3 --require=bitbucket.org/creachadair/shell@v0.0.4
go test ./kythe/...
```

PR #3430, opened 2019-02-01
fix: add an ordering-agnostic yaml comparator in testutil

danielmoy-google:
For testdata at least, this means we can compare proper example yaml
that are checked in and ordered according to proto order.  This
unfortunately doesn't fix the underlying ordering problem we have
reported up in googleapis/google-api-go-client/issues/302, but sidesteps
it for increased test coverage at least.


danielmoy-google:
why is cla stalled :(

danielmoy-google:
> why is cla stalled :(

ask and ye shall receive apparently

PR #3433, opened 2019-02-04
build: update to a tagged version of the go-cmp module

creachadair:


shahms:
That go test failure seems unrelated, but I'd appreciate it if you could take a quick look to verify

creachadair:
I am pretty sure that failure is not related. It works locally:

```
$ go test -race -cpu=1,4 ./kythe/go/extractors/govname
ok  	kythe.io/kythe/go/extractors/govname	3.323s
```

shahms:
Thanks, I'll bug @schroederc to look into it.

schroederc:
The go test failure stems from this upstream change: https://go.googlesource.com/tools/+/0a99049195aff55f007fc4dfd48e3ec2b4d5f602

creachadair:
Ah, that makes sense. In that case I'll just update the test case not to depend on that.

PR #3438, opened 2019-02-05
build: use default cc_configure

schroederc:


tools/build_rules/verifier_test/cc_indexer_test.bzl
```diff
@@ -728,16 +723,14 @@ def cc_extractor_test(
         tags = [],
         restricted_to = ["//buildenv:all"]):
     """C++ verifier test on an extracted source file."""
-    args = ["-std=" + std, "-c"]
```
shahms: This needs to remain; we have tests which require a specific version of C++

PR #3442, opened 2019-02-05
fix: remove an obsolete test case

creachadair:
The go/vcs package removed support for go.googlesource.com as a well-known VCS.
Thus, this test is no longer valid even though it was previously correct.


schroederc:
Can you update `go.mod` to use the updated `golang.org/x/net/tools` version?  Maybe the `external.bzl` version as well if it doesn't conflict with the patch file.

PR #3449, opened 2019-02-06
feat(api): add DecorationsRequest filter for build config

schroederc:


schroederc:
R: @hildrum

kythe/go/serving/xrefs/xrefs.go
```diff
@@ -292,6 +293,11 @@ func (t *Table) Decorations(ctx context.Context, req *xpb.DecorationsRequest) (*
 		bindings := stringset.New()
 
 		for _, d := range decor.Decoration {
+			// Filter decorations by requested build configs.
+			if len(buildConfigs) != 0 && !buildConfigs.Contains(d.Anchor.BuildConfiguration) {
+				continue
+			}
+
```
jaysachs: This filter decorations. Isn't it possible that asking for xrefs for a ticket might also need filtering? (i.e. if the xref is in some other build config)?

schroederc: Filtering for xrefs will be done in a separate change.

hildrum:
LGTM, though you don't appear to have a test where the request has an empty build config and there decorations returned for all build configs.

kythe/proto/xref.proto
```diff
@@ -179,6 +184,9 @@ message DecorationsReply {
     // The reference's span within the source file.
     common.Span span = 5;
 
+    // The build config of the file containing this reference.
+    string build_config = 6;
```
creachadair: Couldn't more than one configuration apply?

schroederc: Each anchor has a single build configuration.

schroederc:
> LGTM, though you don't appear to have a test where the request has an empty build config and there decorations returned for all build configs.

The existing tests cover the empty filter case.

PR #3458, opened 2019-02-07
release(dev): add stricter checking of commit messages

schroederc:
Add 'deprecation' as a commit type.

Warn when a commit does not specify a scope or it is not a standard scope.

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -16,5 +16,57 @@
 
 module.exports = {
   // https://www.conventionalcommits.org/en/v1.0.0-beta.2/
-  extends: ['@commitlint/config-conventional']
+  extends: ['@commitlint/config-conventional'],
+  rules: {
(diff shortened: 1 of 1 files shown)


.commitlintrc.js
```diff
@@ -16,5 +16,57 @@
 
 module.exports = {
   // https://www.conventionalcommits.org/en/v1.0.0-beta.2/
-  extends: ['@commitlint/config-conventional']
+  extends: ['@commitlint/config-conventional'],
+  rules: {
+    'scope-case': [2, 'always', 'snake-case'],
+    'scope-empty': [1, 'never'],
+    'scope-enum': [
```
danielmoy-google: Maybe an example comment on how scopes are used

// fix(extraction): or whatever

schroederc: That documentation is at the link above.

PR #3462, opened 2019-02-08
fix: fix #3459 by removing call to emitComment

### LINKED ISSUE #3459: Java indexer should not associate inline comments on annotations with the following method as a documents
For example, right now this comment shows up as a /kythe/edge/documents on someMethod:

```
@SomeAnnotation // comment about the annotation, incorrectly documents someMethod
void someMethod() {
  ...
}
```

Probably, an inline comment like that should not be attached to the annotation.

In fact I think if you put a non-javadoc comment above the annotation, it doesn't get attached to the method (I could be wrong there):

```
// This doesn't document someMethod
@SomeAnnotation
void someMethod() {
  ...
}
```

That one is more ambiguous to me, but we probably should try to fix the first case there.

danielmoy-google:
I am not clear what would have been relying on the documents edges emitted by this before, so I'm really uncomfortable with this fix.

I also suspect maybe we just don't have ample test coverage or something.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
@@ -477,9 +477,6 @@ public class KytheTreeScanner extends JCTreeScanner<JavaNode, TreeContext> {
     }
 
     if (bindingAnchor != null) {
-      if (!documented) {
-        emitComment(methodDef, methodNode);
-      }
       if (absNode != null) {
         emitAnchor(bindingAnchor, EdgeKind.DEFINES_BINDING, absNode.getVName(), getScope(ctx));
         Span span = filePositions.findIdentifier(methodDef.name, methodDef.getPreferredPosition());
--- a/kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/AnnotationComments.java
+++ b/kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/AnnotationComments.java
@@ -4,10 +4,7 @@ package pkg;
 public class AnnotationComments {
   //- @+3fooString defines/binding FooString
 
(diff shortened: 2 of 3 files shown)


danielmoy-google:
> This seems like it would just indiscriminately remove line comments from all methods. Please add tests to Comments.java if that is not the intention.
> 
> Does this problem generalize to other comments on annotations (e.g. classes)?

Probably right on both counts.  This is likely not the correct fix, and also we should do classes as well.

danielmoy-google:
> > This seems like it would just indiscriminately remove line comments from all methods. Please add tests to Comments.java if that is not the intention.
> > Does this problem generalize to other comments on annotations (e.g. classes)?
> 
> Probably right on both counts. This is likely not the correct fix, and also we should do classes as well.

Indeed, see #3463

schroederc:
`emitCommentsOnLine` already "claims" comments per definition line.  I wonder if claiming the line comments (without emitting any nodes) for the annotation line would be sufficient to mitigate this issue.  This would also claim these comments for the annotation rather than the method, but I'm not sure we care:

```
@Deprecated method() { prinln(); } // i am a comment
```

danielmoy-google:
> `emitCommentsOnLine` already "claims" comments per definition line. I wonder if claiming the line comments (without emitting any nodes) for the annotation line would be sufficient to mitigate this issue. This would also claim these comments for the annotation rather than the method, but I'm not sure we care:
> 
> ```
> @Deprecated method() { prinln(); } // i am a comment
> ```

Yea or something similar:

```
@Inject private injectedVariable; // with a comment
```

I don't think we have any of the inline method-with-annotatino-and-comment happening internally, but I'd bet that someone's doing crazy injection with comments.  Probably not too much, and I agree we probably don't care too much.

danielmoy-google:
Abandoning, have the relevant half (not class annotation) fixed in #3489

Not sure on proper fix for the class annotations in there, but this at least handles the direct bug report.

PR #3463, opened 2019-02-08
test: add tests for line comments on methods

danielmoy-google:
Shows why #3462 is a terrible, terrible idea.

Also add test for class annotation docs as well.

### DIFF
--- a/kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/AnnotationComments.java
+++ b/kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/AnnotationComments.java
@@ -1,13 +1,14 @@
 package pkg;
 
-// Tests for comments that should be ignored.
+//- @+3AnnotationComments defines/binding AnnotationComments
+
+@Deprecated // TODO(#3459): This should not annotate the class, but does.
 public class AnnotationComments {
+  //- !{ _DeprecatedDoc documents AnnotationComments }
+
   //- @+3fooString defines/binding FooString
 
   @SuppressWarnings("unchecked") // TODO(#3459): This should not documents fooString, but does.
   public String fooString() { return ""; }
-  // These tests are all busted:
-  //- !{ UncheckedDoc2 documents FooString }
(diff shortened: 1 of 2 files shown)


kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Comments.java
```diff
@@ -126,4 +126,12 @@
   //- OverrideDoc.node/kind doc
   //- OverrideDoc param.0 _Override
   //- OverrideDoc.text "This documents {@link #compareTo()} over {@link [Override]}. "
+
+  //- @+3fooFunction defines/binding FooFunction
+
+  // This documents FooFunction
```
salguarnieri: Is there a similar test for a class?

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/AnnotationComments.java
```diff
@@ -1,13 +1,14 @@
 package pkg;
 
-// Tests for comments that should be ignored.
+//- @+3AnnotationComments defines/binding AnnotationComments
+
+@Deprecated // TODO(#3459): This should not annotate the class, but does.
 public class AnnotationComments {
+  //- !{ _DeprecatedDoc documents AnnotationComments }
```
salguarnieri: Why use _DeprecatedDoc instead of _? We want to say this class has no documentation, not just that the deprecated comment isn't documentation (though that means the same thing in this case).

danielmoy-google: sure sounds good

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Comments.java
danielmoy-google: yea there's already a test for that up on line 71 I think.

PR #3480, opened 2019-02-11
chore(ts_indexer): Use more standard syntax for lambda

salguarnieri:


salguarnieri:
@evmar can you verify this change?

evmar:
LGTM

zrlk:
I'm confused as to why we need to eta-expand this, but OK

evmar:
In general,
`f(g.p)` is not equivalent to `f(x => g.p(x))`
because 
1) if `g.p` depends on the value of `this`, those two provide different values for it (depends on what `f` does in the former, is bound to `g` in the latter)
2) if `f` passes multiple arguments to the given function, the latter swallows all but the first

I'm not sure why it was necessary here though.

PR #3481, opened 2019-02-12
fix two warnings

evmar:
See separate commits.  These both print warnings when running TS tests.


salguarnieri:
Please update the branch and then we can merge it.

danielmoy-google:
One of these days I'll remember to merge this after updating the branch

PR #3482, opened 2019-02-12
fix: adding support to gcp extraction generation for subdirs

danielmoy-google:
Also supports multiple extraction targets for a single corpus if the
top-level pom.xml file does not pull in all of the subdirectories.

This commit has a somewhat undesireable intermediate state here where we
haven't settled on where CORPUS is specified.  Right now we support both
${_CORPUS} and having it codified in the repo config file.  I am leaning
towards not supporting ${_CORPUS} and having a config generator that
adds it in if necessary, but that's not addressed in this commit.


kythe/go/extractors/gcp/config/common.go
```diff
@@ -84,8 +100,10 @@ func zipMergeStep() *cloudbuild.BuildStep {
 			fmt.Sprintf(
 				"%s merge --output %s %s/*.kzip",
 				constants.DefaultKzipToolLocation,
-				path.Join(outputDirectory, outputFilePattern),
+				path.Join(outputDirectory, outputFileName(corpus)),
 				outputDirectory),
 		},
+		// We explicitly don't include a WaitFor here because this should
```
schroederc: Shouldn't it then wait for the penultimate steps?

danielmoy-google: AH yes, exactly.  I was confused on the same thing while re-writing this code.

Turns out that not having the `waitFor` *is* waiting for everything (BECAUSE THAT'S NOT CONFUSING)

Default behavior is to wait for everything, so leaving this blank is intentional.  I added this comment because I was confusing myself.

But now that you point it out, I'll go add even more info this comment

danielmoy-google: There's a megamind meme in here somewhere...

Issue #3483, opened 2019-02-12
Static claim tool uses unit vname, which doesn't seem overly specific.


robinp:
Not sure what the intention is. The `static_claim` tool uses `unit.v_name` as the claimant, but the (non-Bazel)-extracted cxx kzip's seem to have just `{language:c++, corpus:some_corpus}` as the unit vname. So static claim generation doesn't work.

Should the unit vname be more specific? If so, how can it be controlled? Thank you!

shahms:
VName's should generally be unique, this includes compilation units (although they aren't considered in the schema proper).  In practice, the information-light CompilationUnit vname's haven't proved much of a problem as the facts which actually make it into the graph tend to derive the full vname from the compilation unit and source files.

That said, the compilation units should have properly unique vnames or use the sha256 of the normalized compilation unit as the claim key.

creachadair:
The goals may have changed, but the original idea of the `v_name` field in the CompilationUnit was not to _identify_ the unit. CompilationUnits are not first-class entities in the graph schema (currently), though of course they could be promoted to that status if necessary.

Instead, the `v_name` in the CompilationUnit was meant to be the basis for the VNames of objects _analyzed_ within the unit. In other words, it tells you what language, corpus, root, etc. you should use when you are looking at some object in the AST of the package/module/etc. you are analyzing. That information isn't intrinsic in the language, so it must be provided from outside.

We chose to do this proactively, based on the observation that the build system already has to know this information (e.g., what repo it's looking at, what compiler to invoke). But it is also technically possible to map it on after the fact. 

In any case: The CompilationUnit `v_name` wasn't intended to be an identifier, so this is currently working as intended. It is, however, possible to change our intention. :)

creachadair:
> the basis for the VNames

Also note that this labelling ("basis") is predicated on the view that the VName is the projection of a notional "vector" of attributes that each object has. We haven't really chased that model very deeply, so arguably it's not accurate anymore.

shahms:
Ah, fair enough.  The field should probably call that out explicitly:
```
  // The base VName for the compilation and any generated VNames from its
  // analysis. Generally, the `language` component designates the language of
  // the compilation's sources.
```
Reading it carefully, it's implied but not entirely clear that's the case.  We should also probably document the "canonical" naming for CU's within a kzip file, right now each of the implementations do something different.

creachadair:
https://github.com/kythe/kythe/pull/3484

creachadair:
[...]

PR #3484, opened 2019-02-12
docs: clarify the comment on the CompilationUnit.v_name field

creachadair:
Updates #3483, where it arose that the existing description is not sufficiently
detailed to make the intention clear. Specifically, point out that the v_name
of a compilation record is not an identifier. To the extent that compilation
records have identifiers, it is the unit digest of the complete record.


shahms:
Thanks!

PR #3486, opened 2019-02-12
build: use github.com/google/uuid for Go UUID generation

creachadair:
Recent updates of the github.com/pborman/uuid library have become thin wrappers
around the Google version, and in fact they're both maintained at least in part
by the same person. Instead of updating the wrapper, which would require us to
also depend on the underlying library, just use the Google one directly.

This required a small change to the call site for uuid.New(), since the type
has changed slightly. However, otherwise this is a fairly straightforward update.

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -300,10 +300,10 @@ def _go_dependencies():
 
     maybe(
         go_repository,
-        name = "com_github_pborman_uuid",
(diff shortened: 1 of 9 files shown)


creachadair:
Assuming the CI is happy after the rebase, is there anything further you'd like me to do before this is merged?

shahms:
Was just waiting for buildbot, I think.

PR #3490, opened 2019-02-13
Make pipeline more tolerant to some errors (affects write_tables tool).

robinp:
There were two places that bailed out on error. I observed these happening when using the `beam` pipeline or the older `graphstore`-based one. The added flag can turn bailing out to just logging the warning but continuing processing.

### DIFF
--- a/kythe/go/serving/pipeline/pipeline.go
+++ b/kythe/go/serving/pipeline/pipeline.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"log"
 	"sort"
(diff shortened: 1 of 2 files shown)


kythe/go/serving/pipeline/pipeline.go
```diff
@@ -51,6 +52,10 @@ import (
 	spb "kythe.io/kythe/proto/storage_go_proto"
 )
 
+var (
+	lenientErrors = flag.Bool("lenient_errors", false, "Don't die on errors, rather warn and skip processing.")
+)
+
```
shahms: There's already a Options struct below; it seems like it would be better to extend that than add flags to a library package which doesn't otherwise use them.

robinp: Can do for `pipeline.go`. Should `beam.go` still introduce the flag? It is not apparent how to plumb the options to a beam-registered function (idea welcome).

shahms: I do not know; let's rope in Cody :-)

schroederc: The standard Go idiom is to declare a public variable in a library for the flag:

```
// LenientErrors controls whether the pipeline dies on critical errors......
var LenientErrors bool
```

and then attach a flag to that in the binary:

```
func init() {
  flag.StringVar(pipeline.LenientErrors, ...)
}
```

but... let's just make the pipeline lenient to errors at all times and get rid of the flag.  Seems like the better default.

robinp: Who could argue not having to write more code. Thank you, done.

PR #3491, opened 2019-02-13
Make kythe/cxx/common/path_utils.cc relativize against root dir.

robinp:
This is needed if `KYTHE_ROOT_DIRECTORY=/` is passed to the cxx
extractor.

Also check that relativization only happens if the relativization base
is a proper path prefix of the target, and not just a stringy prefix.

### DIFF
--- a/kythe/cxx/common/path_utils.cc
+++ b/kythe/cxx/common/path_utils.cc
@@ -132,14 +132,27 @@ absl::string_view Dirname(absl::string_view path) {
   return SplitPath(path).dir;
 }
 
+namespace {
+bool IsProperPathPrefix(absl::string_view s, absl::string_view prefix) {
+  return absl::StartsWith(s, prefix) &&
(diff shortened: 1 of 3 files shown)


kythe/cxx/common/path_utils.cc
```diff
@@ -132,14 +132,27 @@ absl::string_view Dirname(absl::string_view path) {
   return SplitPath(path).dir;
 }
 
+namespace {
+bool IsProperPathPrefix(absl::string_view s, absl::string_view prefix) {
+  return absl::StartsWith(s, prefix) &&
+         (s.size() == prefix.size() || s[prefix.size()] == '/');
+}
+}  // namespace
+
```
shahms: Put this in the extant anonymous namespace rather than introducing another to put it here.

robinp: Done.

PR #3497, opened 2019-02-14
Beam: support reading input files from a directory.

robinp:
Note: this is should be diffbased on #3490. Will repush once that merged - in the mean time, please choose only the last commit in the "files changed" view.


kythe/go/serving/pipeline/beamio/entries.go
```diff
@@ -30,37 +32,56 @@ import (
...
-// Riegeli file.
-func ReadEntries(ctx context.Context, s beam.Scope, file string) beam.PCollection {
-	if p := tryReadRiegeli(ctx, s, file); p.IsValid() {
-		return p
+// given file, or files stored in a directory.  The file can be part of any
+// filesystem registered with the beam/io/filesystem package and can either be a
+// delimited protobuf stream or a Riegeli file.
+func ReadEntries(ctx context.Context, s beam.Scope, fileOrDir string) (beam.PCollection, error) {
+	if strings.HasSuffix(fileOrDir, "/") {
+		errv := beam.PCollection{}
```
schroederc: `var errv beam.PCollection`

kythe/go/serving/pipeline/beamio/entries.go
```diff
@@ -30,37 +32,56 @@ import (
...
+		if err != nil {
+			return errv, err
+		}
+		defer fs.Close()
+		files, err := fs.List(ctx, fileOrDir+"*")
+		if err != nil {
+			return errv, err
+		}
+		if len(files) == 0 {
+			return errv, fmt.Errorf("No entries found in %s. Maybe mistyped path?", fileOrDir)
```
schroederc: Please use lowercase letters to begin a `fmt.Errorf` message.

kythe/go/serving/tools/write_tables/write_tables.go
```diff
@@ -61,6 +63,7 @@ var (
 	verbose = flag.Bool("verbose", false, "Whether to emit extra, and possibly excessive, log messages")
 
 	experimentalBeamPipeline = flag.Bool("experimental_beam_pipeline", false, "Whether to use the Beam experimental pipeline implementation")
+	beamShards               = flag.Int("beam_shards", 128, "Number of shards for beam processing.")
```
schroederc: Please set the default to `0` and state that a "reasonable" default will be chosen based on the input if given a non-positive value.  Then check for non-positive numbers below and set it to 128 otherwise,.

robinp:
Thank you, fixed, will remember these.

By the way, how should the conventional-commits style preserved when merging? Should I write the PR name and desc in that style? Or keep a single squashed commit in the PR with that style?

schroederc:
> Thank you, fixed, will remember these.
> 
> By the way, how should the conventional-commits style preserved when merging? Should I write the PR name and desc in that style? Or keep a single squashed commit in the PR with that style?

Github doesn't handle force-pushing all that well, so it's better to add commits to a PR rather than squashing or amending existing commits in the PR.

It would be most convenient if the PR title and description reflected our commit style.  That makes it easier for the merger to ensure the actually merged commit follows the style.

PR #3499, opened 2019-02-14
feat(extraction): add Bazel extraction Docker image

schroederc:
Use pre-built extractor binaries to create a local Bazel repository to
use as a source of extra actions.  Add all common Kythe configuration to
the builder's .bazelrc file and use a simple script to collect the .kzip
artifacts after a Bazel execution.

### DIFF
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -1,6 +1,6 @@
-#!/bin/bash -e
+#!/bin/bash
 
-# Copyright 2015 The Kythe Authors. All rights reserved.
+# Copyright 2019 The Kythe Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


kythe/extractors/bazel/extract.sh
```diff
@@ -14,103 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Script to extract compilations from a Bazel repository.  See usage() below.
+bazel "$@"
```
shahms: This pretty radically changes the requirements for running this script.

Whereas before it would work outside of a docker image, it now *requires* the exact filesystem layout of the Docker image and that the requisite bazel build has already been run.  This is a big step backwards.

danielmoy-google: Sure maybe we can change /workspace/output to be $KYTHE_OUTPUT_DIRECTORY mimicing what I have in gcp land, which works locally.

For /kythe/kzip, I have no immediate great ideas.  I've been stuffing things into /opt/kythe where things used to install (still do, too? /hedberg), and referencing those from either inside or outside of docker.  But I have no preference for /opt/kythe vs /kythe

Could require kzip tool to live on $PATH, as discussed offline maybe.  (Or have a $KYTHE_TOOL_DIRECTORY)

shahms: I suspect we should have a design document around things like this, KYTHE_PATH falling back to PATH if that's missing for binary locations seems reasonable. The same with KYTHE_OUTPUT_DIRECTORY, although for a script like this having it be a command line flag with a suitable usage message is more user friendly.

PR #3502, opened 2019-02-14
fix: go style errors introduced in previous commit

robinp:
Ran `arc lint` manually, will see to add it to second hook. @shahms


kythe/go/serving/tools/kythefs/kythefs.go
```diff
@@ -64,10 +64,10 @@ func init() {
 		"[--server SERVER_ADDRESS]")
 }
 
-type KytheFs struct {
+type kytheFs struct {
```
creachadair: Should be `kytheFS` ideally. [https://github.com/golang/go/wiki/CodeReviewComments#initialisms]

robinp: Thank you!

robinp:
Done.

PR #3508, opened 2019-02-15
fix: downstream fix to provider syntax

danielmoy-google:
Migrate off legacy provider syntax

`ctx.attr._java_toolchain.java_toolchain` and
`ctx.attr._java_toolchain[java_common.JavaToolchainInfp]` have the same
behaviour, but the latter syntax is preferred and will be the only
supported option in the future.


shahms:
It looks like we may have to bump the minimum Bazel version for this.

danielmoy-google:
> It looks like we may have to bump the minimum Bazel version for this.

Yea cushon is trying some things internally

danielmoy-google:
> > It looks like we may have to bump the minimum Bazel version for this.
> 
> Yea cushon is trying some things internally

Ok good news it works without a newer bazel version, at least on my box.  Let's see what buildbot thinks.

PR #3509, opened 2019-02-15
feat(extraction): add generic GCB Bazel extraction config

schroederc:


kythe/extractors/gcp/bazel/README.md
```diff
@@ -0,0 +1,37 @@
+# Google Cloud Build -- Bazel Kythe extraction
```
danielmoy-google: Could also consider stuffing this in gcp/README.md, but up to you

PR #3511, opened 2019-02-15
feat(cxx_common): proto plugin to embed metadata in C++ headers

### LINKED ISSUE #3506: Support building `cc_proto_library` targets with embedded metadata
Support swapping the C++ proto toolchain with one that embeds Kythe metadata in the generated proto sources.

See https://github.com/kythe/kythe/blob/2af6297310eae2a8c95bcfbb69f722d0e338cd26/kythe/extractors/BUILD#L7 for the Java analog.

shahms:
This fixes #3506


kythe/cxx/tools/proto_metadata_plugin.cc
```diff
@@ -0,0 +1,169 @@
...
+    return "Not a metadata file";
+  }
+
+  Printer printer(stream, '\n');
+  printer.PrintRaw("/* ");
+  printer.WriteRaw(metafile.data(), metafile.size());
+  printer.PrintRaw("\n");
+  std::string metadata;
+  absl::Base64Escape(contents, &metadata);
+  printer.WriteRaw(metadata.data(), metadata.size());
```
zrlk: Some tools behave badly in the presence of very long lines. It might be worth adding line breaks to keep this to 80 columns.

justbuchanan:
I'd like to add some verifier tests for xrefs between proto and c++. Is there a way to use the new toolchain just for my tests (without passing --proto_toolchain_for_cc on the command line). Otherwise I could just have a rule that calls protoc with this plugin, but that gets more difficult when dependencies are involved.

zrlk:
Doesn't look like [cc_proto_library](https://docs.bazel.build/versions/master/be/c-cpp.html#cc_proto_library) supports overriding the toolchain, so you'll probably have to go with the genrule.

kythe/cxx/tools/proto_metadata_plugin.cc
shahms: A very good point, thanks. Done.

shahms:
PTAL; I added the option to compress the inline metadata, include code to add the parameters for the underlying code generator if they aren't present and break the emitted encoded metadata into lines.

shahms:
Note: compression does make a noticeable difference, but is not yet supported by the indexer.

kythe/cxx/tools/proto_metadata_plugin.cc
```diff
@@ -0,0 +1,247 @@
...
+ private:
+  bool compress_metadata_;
+  GeneratorContext* wrapped_;
+  // Map from header name -> metadata contents.
+  absl::node_hash_map<std::string, std::string> metadata_map_;
+  absl::flat_hash_map<std::string, std::unique_ptr<ZeroCopyOutputStream>>
+      stream_map_;
+};
+
+// Returns true if the compress_metadata paramete is present.
```
zrlk:
```suggestion
// Returns true if the compress_metadata parameter is present.
```

shahms:
Once more, should be the last.  Seems like if you're using this plugin, you want it all so actually check for and add the header guard option as well.  Also, actually *use* the compress_metadata option, rather than just stripping it :-)

PR #3515, opened 2019-02-19
fix(dev): use POSIX safe flags for tr

creachadair:
The build rules for the VName test data fail on macOS because we used the
GNU-supported --delete flag instead of POSIX-recommended -d. All the Linuces
support -d just fine with the same semantics, so this has no discernable effect
on the results.


creachadair:
Repro: `bazel test kythe/data/...` on macOS 10.14.3 (Mojave).

PR #3524, opened 2019-02-19
fix: pin gradle/maven extraction to jdk8 instead of using gcb image

danielmoy-google:
This way when [gcb's maven builder image](https://cloud.google.com/cloud-build/docs/cloud-builders) changes its default version, we won't get unpredictable behavior, because we'll be pinned to a specific version from https://hub.docker.com/_/maven.

Also some fixes to gradle examples to bring them more in-line with maven examples.


kythe/extractors/gcp/examples/gradle.yaml
```diff
@@ -24,21 +39,29 @@ steps:
...
     - 'KYTHE_ROOT_DIRECTORY=/workspace/code'
     - 'JAVAC_EXTRACTOR_JAR=/opt/kythe/extractors/javac_extractor.jar'
     - 'REAL_JAVAC=/usr/bin/javac'
     - 'TMPDIR=/workspace/out'
     - 'KYTHE_JAVA_RUNTIME_OPTIONS=-Xbootclasspath/p:/opt/kythe/extractors/javac9_tools.jar'
   volumes:
     - name: 'kythe_extractors'
-      path: '/opt/kythe/extractors/'
+      path: '/opt/kythe/extractors'
+  id: 'EXTRACT0'
```
jaysachs: Curious why the steps are named "XYZ0" and not "XYZ"?

creachadair:
Did you mean s/gcb/GCP/?

danielmoy-google:
> Did you mean s/gcb/GCP/?

gcb as shorthand for google cloud builder, the team that actually provides the maven wrapper builder image: https://cloud.google.com/cloud-build/docs/cloud-builders

I do regret the general state of acronyms, but I think gcb image is the right phrasing there.

I should update the pull request to point at the old builder though.

kythe/extractors/gcp/examples/gradle.yaml
danielmoy-google: I don't have special casing logic differentiating single-extract vs multiple-extract for naming steps.  So this is required for extracting e.g. multiple different poms that don't link in a single top-level action.  I could have logic that omits the 0 for single case, but not sure if it's worth doing.

danielmoy-google: If we feel strongly about having different logic there, we can do so.  It wouldn't be too messy I think.  And the majority of cases will probably have just one target to extract (I'd guess, could be wrong), so it would be cleaner output.

danielmoy-google:
> Did you mean s/gcb/GCP/?
> 
> Edit: Oh, this is presumably Google Cloud…Builder? I guess that's a thing now?
> 
> Edit 2: But then the folder is `/gcp/` so I'm not sure which it is _supposed_ to be.

Yea gcb is part of gcp.  So, gcp is the more general term.  I could change the commit message to gcp I suppose.

kythe/extractors/gcp/examples/gradle.yaml
danielmoy-google: ok see #3526 which special cases the default single-target to not have an index here.

PR #3526, opened 2019-02-19
fix: remove index in default maven/gradle single target extraction

danielmoy-google:
Previously we'd append index 0 to extract/preprocess steps for
maven/gradle, this makes a single-target extraction not have any index
at all, which is cleaner to read.

Split apart the setup not specific to a single target into a separate
preprocess step too, for clarity.

Also add a test for multi target extraction, to actually verify all of
this.


kythe/go/extractors/gcp/config/common.go
```diff
@@ -64,11 +63,11 @@ func commonSteps(repo string) []*cloudbuild.BuildStep {
 	}
 }
 
-func preprocessorStep(build string, buildID int) *cloudbuild.BuildStep {
+func preprocessorStep(build string, targetID string) *cloudbuild.BuildStep {
```
jaysachs: s/targetID/idSuffix/ 

?

kythe/go/extractors/gcp/config/converter.go
```diff
@@ -107,14 +108,23 @@ func KytheToBuild(conf *rpb.Config) (*cloudbuild.Build, error) {
...
+
+	build.Steps = append(build.Steps, g.preExtractSteps()...)
+
 	targets := hints.Targets
 	if len(targets) == 0 {
 		targets = append(targets, g.defaultExtractionTarget())
 	}
 	for i, target := range targets {
-		build.Steps = append(build.Steps, g.extractSteps(hints.Corpus, target, i)...)
+		id := ""
```
jaysachs: idSuffix?

danielmoy-google: ack

kythe/go/extractors/gcp/config/common.go
danielmoy-google: sure we can do that

PR #3530, opened 2019-02-20
fix(build): remove duplicate src paths in java_extract_kzip

justbuchanan:
java_extract_kzip() calls extract(), which handles passing src paths
to the underlying extractor:
https://github.com/kythe/kythe/blob/448d46e8bb30c16345d516f09b6904e92b7a05cf/tools/build_rules/verifier_test/verifier_test.bzl#L147-L157.


justbuchanan:
I think `path` is what we want. If the rule is given a generated file as input, `path` is where it actually lives relative to the working directory. Short path would be something like `kythe/java/com/...` vs `bazel-out/k8-fastbuild/kythe/java/com/...`.

PR #3531, opened 2019-02-21
feat(cxx_indexer): add tests and fix support for C++/proto metadata references

### LINKED ISSUE #3504: cxx extractor should error out if a #include is not found
I noticed that the extractor exits "successfully" with code zero even if errors are encountered during extraction. Should it error out instead?

As an example, here's what I see in bazel when I attempt to build a kzip that doesn't have all its dependencies:

```console
$ bazel build //kythe/cxx/indexer/cxx/testdata/proto:proto_genlang_kzip
INFO: Analysed target //kythe/cxx/indexer/cxx/testdata/proto:proto_genlang_kzip (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
INFO: From ExtractCompilation kythe/cxx/indexer/cxx/testdata/proto/proto_genlang_kzip/proto.cc.kzip:
kythe/cxx/indexer/cxx/testdata/proto/proto.cc:1:10: fatal error: 'kythe/cxx/indexer/cxx/testdata/proto/testdata.pb.h' file not found
#include "kythe/cxx/indexer/cxx/testdata/proto/testdata.pb.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Target //kythe/cxx/indexer/cxx/testdata/proto:proto_genlang_kzip up-to-date:
  bazel-bin/kythe/cxx/indexer/cxx/testdata/proto/proto_genlang_kzip/proto.cc.kzip
INFO: Build completed successfully, 2 total actions
```

shahms:
This PR does a few things with the goal of adding tests for cross-language references between C++ and protocol buffers.

1) Updates ProtobufMetadataSupport to accept `.pb.h` and `.proto.h` files so that inline metadata will be decoded.
2) Adds a `cc_kythe_proto_library` rule and aspect for generating protobufs with embedded metadata from within Bazel.
3) Fixes #3504 
4) Fixes a variety of verifier_test and related facilities to work better with dependencies.


kythe/cxx/indexer/cxx/testdata/proto/testdata.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
```
justbuchanan: Is this file referenced anywhere?

shahms: Ach, no, both of those were left over from previous revisions. Also looks like I have to make some fixes for older Bazel.

PR #3534, opened 2019-02-21
fix: pin gcp extraction example and tests to :stable

danielmoy-google:
I have separately tagged all of our gcr.io/kythe-public images with
stable.


creachadair:
🐎

PR #3538, opened 2019-02-21
feat(kzip_create): allow embedding details messages in kzips

shahms:
Turns out that the JSON format is more flexible in what it allows us to parse, so go with that.

### DIFF
--- a/kythe/go/platform/tools/kzip/createcmd/createcmd.go
+++ b/kythe/go/platform/tools/kzip/createcmd/createcmd.go
@@ -29,7 +29,6 @@ import (
 	"kythe.io/kythe/go/util/kytheuri"
 
 	"github.com/golang/protobuf/jsonpb"
-
 	"github.com/google/subcommands"
 
 	anypb "github.com/golang/protobuf/ptypes/any"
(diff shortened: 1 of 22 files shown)


kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -19,25 +19,31 @@ package createcmd
...
 	"flag"
 	"os"
+	"strings"
 
 	"kythe.io/kythe/go/platform/kzip"
 	"kythe.io/kythe/go/util/cmdutil"
 	"kythe.io/kythe/go/util/kytheuri"
 
+	"github.com/golang/protobuf/jsonpb"
+
```
schroederc: nit: remove blank line

kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -90,21 +97,26 @@ func (c *createCommand) Execute(ctx context.Context, fs *flag.FlagSet, _ ...inte
 		SourceFile: []string{c.source},
 	}
 
+	err = unmarshalDetails(c.details, &cu.Details)
```
schroederc: `if err := unmarshalDetails(c.details, &cu.Details); err != nil {

kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -145,3 +157,15 @@ func addFile(out *kzip.Writer, cu *apb.CompilationUnit, path string) error {
 	})
 	return nil
 }
+
+func unmarshalDetails(details string, msg *[]*anypb.Any) error {
```
schroederc: Idiomatically return `([]*anypb.Any, error)` instead of using an output parameter

kythe/go/platform/tools/kzip/createcmd/createcmd.go
shahms: with the below change, this becomes: `details, err := ...` but otherwise remains the same to keep lint happy.

Issue #3528, opened 2019-02-20
Optionally name javac-wrapper.sh's java-extractor.err/out with a tag


danielmoy-google:
For example $VERSION if passed in on commandline.

Issue #3521, opened 2019-02-19
GCB extraction should echo errors before merging kzip, or have kzip merge cleanly support empty bits


danielmoy-google:
For example see kythe/extractors/gcp/examples/mvn.yaml

There are classes of errors on extraction for maven/gradle which don't exit from `mvn`/`gradle` build step with an error, but do emit a non-empty `javac-extractor.err`.  However the next step to merge kzip currently fails if there are no .kzip files, so the resulting javac-extractor.err file won't get uploaded.

A simple and potentially sufficient workaround for this would be to just echo out the contents of `javac-extractor.err` (probably `javac-extractor.out` for good measure too, because there are certain classes of errors that show up in the .out file).

Alternatively it might be worth making kzip merge build step not die on empty input (probably the bash isn't written correctly I'd guess).  Then just rely on artifact upload for .err (possibly .out too).

Currently I'm leaning towards the former.

jsachs-google:
What about the kzip step not die silently on empty input? e.g. give a usage?

I like *also* dumping extractor.err to stdout as a step as well.

danielmoy-google:
kzip merge step doesn't die silently on empty input, though if I gather what you're suggesting, it's to make the error message more useful (since right now it only complains about zero kzip files to merge).

Maybe what would make sense is an explicit "print .err/.out, and die if .err is non-zero size or there are no kzips" step in between extract and merge.

danielmoy-google:
The more that I think about it, the more I am of the opinion we should actually have a step after extraction explicitly push .err/.out to a gs bucket, not simply print the error.  Then, don't upload it as an artifact at the end - that way if there's any failure between end of extraction and the end (e.g. merge as prompted by this issue), we'll still get log messages in a reliable place.

I'm still not settled on what to do for errors so bad they crash the extraction step itself - perhaps we just rely on log messages printed out during the course of that failure.

danielmoy-google:
Ok we now echo errors always.

PR #3539, opened 2019-02-21
fix: remove javac-extractor.err and .out printing from javac-wrapper.sh

danielmoy-google:
Since we won't be able to have all extractors and wrappers outputting
clean err/out files like this one does, we'll instead just always print
inline errors.  This also obviates the need to set TMPDIR in
maven/gradle extraction.

This "fixes" #3528 by making it obsolete - no more files, so nothing to
tag.

This starts the ball rolling for #3521, though doesn't really make the
experience around missing kzips any cleaner.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
@@ -38,7 +38,6 @@
 #
 # Other environment variables that may be passed to this script include:
 #   KYTHE_EXTRACT_ONLY: if set, suppress the call to javac after extraction
-#   TMPDIR: override the location of extraction logs and other temporary output
 
 if [[ -z "$JAVA_HOME" ]]; then
   readonly JAVABIN="$(which java)"
--- a/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
@@ -40,8 +40,6 @@
 #   KYTHE_EXTRACT_ONLY: if set, suppress the call to javac after extraction
 #   TMPDIR: override the location of extraction logs and other temporary output
 
-export TMPDIR="${TMPDIR:-/tmp}"
-
 if [[ -z "$JAVA_HOME" ]]; then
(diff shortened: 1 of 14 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
```diff
@@ -71,8 +68,8 @@ fi
 
 "$JAVABIN" "${KYTHE_JAVA_RUNTIME_OPTIONS[@]}" \
   -jar "$JAVAC_EXTRACTOR_JAR" "$@" \
-  1>>"$TMPDIR"/javac-extractor.out \
-  2>>"$TMPDIR"/javac-extractor.err
+  1> >(sed -e 's/^/EXTRACT OUT: /') \
+  2> >(sed -e 's/^/EXTRACT ERR: /' >&2)
```
jaysachs: Oh, GCB only gives you stderr? Where does stdout go?

danielmoy-google: No GCB does give you stdout too

My bash is not super strong, but my understanding is that this will just prepend 'EXTRACT OUT' in front of stdout and 'EXTRACT ERR' in front of stderr

danielmoy-google: Crucially, for certain classes of errors (like the java11 one), the useful message is printed to stdout.  I went and tested that this message shows up:

```
Step #5 - "EXTRACT": EXTRACT OUT: -Xbootclasspath/p is no longer a supported option.
```

jaysachs: Ah, didn't see why we'd want to distinguish them. Nothing else does; I'm not sure if it's the right/best thing to do to be "different" but this is fine.

danielmoy-google: Mostly I just wanted to distinguish these from other debug printing, so I started with EXTRACT ERR as a prefix in there.

Then I figured prefixing stdout messages with EXTRACT ERR felt a bit odd, so I changed that to EXTRACT OUT.

But we could easily change this later.

PR #3541, opened 2019-02-23
refactor(build): move cc proto verifier tests into a bzl macro

justbuchanan:
These build rules will be re-used for some more testing around message
set / extension xrefs.


shahms:
I think this test size needs to be `medium` to avoid the timeout.

justbuchanan:
Ok, I'll set it to medium. I'm impressed that this test takes so long... I guess there are a lot of c++ bits to process?

shahms:
The verifier scales non-linearly with the size of the graph and the C++ indexer has to churn through a lot of headers due to the proto dependency. Only the former is affected by the test timeout, though.

PR #3542, opened 2019-02-25
feat(kzip_create): allow complete specification of a compilation unit

### LINKED ISSUE #3146: Write a generic kzip generating tool
Essentially, something which will construct a suitable .kzip file from a list of files and compiler flags.  This is mostly a matter of creating a CompilationUnit protocol message and then zipping it and the provided files into a .kzip file.

shahms:
This fixes #3146. Notably still incomplete is that the contents of the .kzip file are constructed entirely from the command line arguments.  Environment is untouched, meaning the existing Starlark `extract` helper won't work (as it expects the provided `KYTHE_*` environment variables to matter.

It's also incompatible with the previous create command, but I decided it was preferable to have to be consistent with the protobuf instead.


shahms:
@creachadair

kythe/go/platform/tools/kzip/createcmd/flag.go
```diff
@@ -0,0 +1,199 @@
...
+		return []string(nil)
+	}
+	return *f
+}
+
+type repeatedEnv map[string]string
+
+// Set implements part of the flag.Getter interface and will append a new value to the flag.
+func (f *repeatedEnv) Set(s string) error {
+	eq := strings.IndexByte(s, '=')
```
schroederc: Use `strings.SplitN(s, "=", 2)`

kythe/go/platform/tools/kzip/createcmd/flag.go
```diff
@@ -0,0 +1,199 @@
...
+	if f == nil {
+		return map[string]string(nil)
+	}
+	return *f
+}
+
+// ToProto returns a []*apb.CompilationUnit_Env for the mapped environment.
+func (f *repeatedEnv) ToProto() []*apb.CompilationUnit_Env {
+	if f == nil || *f == nil {
+		return []*apb.CompilationUnit_Env(nil)
```
schroederc: `return nil`

kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -39,117 +36,147 @@ import (
...
-	fs.StringVar(&c.uri, "uri", "", "A Kythe URI naming the compilation unit")
-	fs.StringVar(&c.output, "output", "", "Path for output kzip file")
-	fs.StringVar(&c.source, "source", "", "Path for input source file")
-	fs.StringVar(&c.details, "details", "", "JSON-encoded Any messages to embed as compilation details.")
+	fs.StringVar(&c.output, "output", "", "Path for output kzip file (required)")
+	fs.Var(&c.rules, "rules", "Path to vnames.json file (optional)")
+
+	fs.Var(&c.uri, "uri", "A Kythe URI naming the compilation unit VName (required)")
+	fs.Var(&c.source, "source_file", "Repeated paths for input source files (required)")
+	fs.Var(&c.inputs, "required_input", "Repeated paths for additional required inputs (optional)")
```
schroederc: Add a note that these should not intersect with `--source_file` or there will be duplicate required inputs in the compilation.

creachadair: Ideally the tool should deduplicate them, right?

shahms: indeed; changed this to stringset flags

PR #3545, opened 2019-02-25
fix: add support for bazel BuildSteps generation

danielmoy-google:
Uses Cody's new `gcr.io/kythe-public/bazel-extractor`.  Some small diffs on the README and variable names there.  Also adopt /workspace/output for the rest of the extractors because that reads a lot nicer than /workspace/out that I was using before.


kythe/go/extractors/gcp/config/gradle.go
```diff
@@ -21,7 +21,6 @@ import (
 	"path/filepath"
 
 	"kythe.io/kythe/go/extractors/constants"
-
```
schroederc: Restore blank line

kythe/go/extractors/gcp/config/maven.go
```diff
@@ -21,7 +21,6 @@ import (
 	"path/filepath"
 
 	"kythe.io/kythe/go/extractors/constants"
-
```
schroederc: Restore blank line

danielmoy-google: :+1: 

(bazel.go too)

kythe/go/extractors/gcp/config/gradle.go
danielmoy-google: :+1:

PR #3552, opened 2019-02-27
build: use Args objects to format action arguments

schroederc:


tools/build_rules/verifier_test/cc_indexer_test.bzl
```diff
@@ -248,18 +252,18 @@ def _cc_extract_kzip_impl(ctx):
     ])
     outputs = depset([
         extract(
+            srcs = depset([src]),
```
shahms: Did you run buildifier over these? If not, they will merely be undone the next time someone else edits them.

schroederc: I did.  Everything looks good on my end with both the internally built `buildifier` as well as with `get get -u github.com/bazelbuild/buildtools/buildifier`.

Issue #3551, opened 2019-02-27
Build failure with Go 1.12


creachadair:
At head on macOS with Go 1.12 installed via Homebrew, I get the following error when trying to build:

```
# Some log output elided for brevity.
$ bazel build kythe/go/util/tools/extractschema
ERROR: /Users/michael/software/kythe/kythe/go/util/tools/extractschema/BUILD:5:1: no such package '@org_bitbucket_creachadair_stringset//': no such package '@bazel_gazelle_go_repository_tools//': failed to build fetch_repo: build cache is required, but could not be located: GOCACHE is not defined and $HOME is not defined
 and referenced by '//kythe/go/util/tools/extractschema:extractschema'
ERROR: Analysis of target '//kythe/go/util/tools/extractschema:extractschema' failed; build aborted: no such package '@org_bitbucket_creachadair_stringset//': no such package '@bazel_gazelle_go_repository_tools//': failed to build fetch_repo: build cache is required, but could not be located: GOCACHE is not defined and $HOME is not defined
```

I don't think this is specific to macOS, but the error is (slightly) different on Linux:

```
ERROR: /home/kythedev/kythe/kythe/go/util/tools/extractschema/BUILD:5:1: no such package '@org_bitbucket_creachadair_stringset//': no such package '@bazel_gazelle_go_repository_tools//': failed to build fetch_repo: build cache is required, but could not be located: GOCACHE is not defined and neither $XDG_CACHE_HOME nor $HOME are defined
 and referenced by '//kythe/go/util/tools/extractschema:extractschema'
ERROR: Analysis of target '//kythe/go/util/tools/extractschema:extractschema' failed; build aborted: no such package '@org_bitbucket_creachadair_stringset//': no such package '@bazel_gazelle_go_repository_tools//': failed to build fetch_repo: build cache is required, but could not be located: GOCACHE is not defined and neither $XDG_CACHE_HOME nor $HOME are defined
```

I imagine this is just a configuration issue with the new 1.12 release. "And yet it fails".

danielmoy-google:
I agree it is not specific to macOS.  The error message will vary each time you run.

I'm not sure what bazel does here, because GOCACHE is definitely set on my go env.

shahms:
#3548 updated the Go build rules to 0.17.1; does it work with 0.17?

creachadair:
> #3548 updated the Go build rules to 0.17.1; does it work with 0.17?

[...]

PR #3553, opened 2019-02-27
fix(dev): pin the Go version to 1.11.5

creachadair:
A workaround for #3551. It is not clear to me how the upgrade managed to pass CI, but builds fail locally on both macOS and Linux after #3548.

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -11,7 +11,10 @@ load("@io_kythe//third_party/leiningen:lein_repo.bzl", "lein_repository")
 def _rule_dependencies():
     gazelle_dependencies()
     go_rules_dependencies()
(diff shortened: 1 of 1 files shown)


schroederc:
Updating Gazelle seems like it would also fix this: https://github.com/bazelbuild/bazel-gazelle/releases/tag/0.17.0

creachadair:
> Updating Gazelle seems like it would also fix this: https://github.com/bazelbuild/bazel-gazelle/releases/tag/0.17.0

So it does. Released 2 hours ago, sounds like a hotfix to me. But I'll take it! 👍

creachadair:
Superseded by #3555.

PR #3556, opened 2019-02-27
fix: tag google cloud builds

danielmoy-google:
Include the corpus (optionally substituted at request time).  Also
include the builder used (maven, gradle, bazel).

Don't prefix with `kythe_` though, maybe there's disagreement there.


danielmoy-google:
> Maybe add a 'kythe' or 'kythe_extraction' tag to every build?

How about `kythe_extract_maven`, etc

PR #3570, opened 2019-03-01
chore: remove an unused import from analysis.proto

creachadair:
Hitherto, analysis.proto imports filecontext.proto, but no longer uses it since the merge of #3533. Morally, this change just removes the import, but there were several strict dependency violations that also needed cleaned up.

Also update the Go generated wrappers.


kythe/cxx/tools/BUILD
```diff
@@ -18,7 +18,6 @@ cc_library(
         "//kythe/proto:analysis_cc_proto",
         "//kythe/proto:buildinfo_cc_proto",
         "//kythe/proto:cxx_cc_proto",
-        "//kythe/proto:filecontext_cc_proto",
```
creachadair: Oops. This was a mistake, I did not mean to remove this. (@schroederc fixed it in #3571)

PR #3571, opened 2019-03-01
build: restore missing proto dependency

schroederc:


creachadair:
Sorry, did I accidentally clobber this, or did I miss it in my previous change?

schroederc:
#3570 removed the dependency.  Unfortunately, I don't think Bazel supports checking this and we can only catch this on import.

creachadair:
Ah, I see why it was missed: I mistakenly dropped the import from the binary target. The binary uses it only for decoding and our tests don't run the binary so it didn't show up.

PR #3584, opened 2019-03-04
build: remove manually written header include path

schroederc:
This fixes `bazel build --incompatible_remap_main_repo //...`

### DIFF
--- a/third_party/googlelog/BUILD
+++ b/third_party/googlelog/BUILD
@@ -5,5 +5,5 @@ licenses(["notice"])  # BSD
 cc_library(
     name = "config_h",
     hdrs = ["src/config.h"],
-    includes = ["src"],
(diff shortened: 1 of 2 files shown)


third_party/googlelog/BUILD
```diff
@@ -5,4 +5,5 @@ licenses(["notice"])  # BSD
 cc_library(
     name = "config_h",
     hdrs = ["src/config.h"],
+    includes = ["src"],
```
shahms: This is incorrect.  `includes` adds the include transitively to all targets which depend on this target, but the config.h file is an implementation detail which should *not* be added or included by dependents of this target.

schroederc: I used `strip_include_prefix` instead.

PR #3587, opened 2019-03-06
feat(tooling): add --build_config filter to 'kythe decor' command

schroederc:


kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -40,10 +41,10 @@ type createCommand struct {
 	rules  vnameRules
 
 	uri          kytheURI
-	source       repeatedStringSet
-	inputs       repeatedStringSet
+	source       flagutil.StringSet
+	inputs       flagutil.StringSet
 	hasError     bool
-	argument     repeatedString
+	argument     flagutil.StringList
```
shahms: These are incompatible with the existing and intended semantics of these fields, particularly "argument".

While `source` and `inputs` may reasonably (but not necessarily) expected to not contain any embedded commas, the same cannot be said of `argument`.

kythe/go/util/flagutil/flagutil.go
```diff
@@ -73,3 +75,68 @@ func UsageError(msg string) {
 func UsageErrorf(str string, vals ...interface{}) {
 	UsageError(fmt.Sprintf(str, vals...))
 }
+
+// StringList implements a flag.Value that accepts an sequence of values as a CSV.
+type StringList []string
+
+// Set implements part of the flag.Getter interface and will append a new value to the flag.
```
shahms: "append new values" (unlike repeatedStringList, which will always only append a single value).

kythe/go/util/flagutil/flagutil.go
```diff
@@ -73,3 +75,68 @@ func UsageError(msg string) {
...
+	if f == nil {
+		return []string(nil)
+	}
+	return *f
+}
+
+// StringSet implements a flag.Value that accepts an set of values as a CSV.
+type StringSet stringset.Set
+
+// Set implements part of the flag.Getter interface and will append a new value to the flag.
```
shahms: ditto on the plural here

kythe/go/platform/tools/kzip/createcmd/createcmd.go
schroederc: I reverted that flag.

PR #3589, opened 2019-03-06
fix(cxx_indexer): incorporate build_config into starting claim context

shahms:
zrlk: this does not change the whole-file claiming, but should do for sub-claims.

### DIFF
--- a/kythe/cxx/indexer/cxx/KytheGraphObserver.h
+++ b/kythe/cxx/indexer/cxx/KytheGraphObserver.h
@@ -386,7 +386,8 @@ class KytheGraphObserver : public GraphObserver {
   /// \brief Configures the starting context.
   /// \param context Context to use when the main source file is entered.
   void set_starting_context(const PreprocessorContext& context) {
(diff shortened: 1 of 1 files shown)


kythe/cxx/indexer/cxx/KytheGraphObserver.h
```diff
@@ -384,7 +386,8 @@ class KytheGraphObserver : public GraphObserver {
   /// \brief Configures the starting context.
   /// \param context Context to use when the main source file is entered.
   void set_starting_context(const PreprocessorContext& context) {
-    starting_context_ = context;
+    starting_context_ =
+        CompressString(absl::StrCat(build_config_, ":", context));
```
zrlk: As we'd discussed offline, I think this will cause the indexer to get upset because it won't be able to match its context against the context recorded by the extractor.

shahms: Fair enough :-)

PR #3595, opened 2019-03-08
build: update lein to latest version

jaysachs:


third_party/leiningen/lein_repo.bzl
```diff
@@ -67,7 +67,7 @@ lein_repository = repository_rule(
     implementation = _lein_repository,
     attrs = {
         "version": attr.string(
-            default = "2.5.1",
+            default = "2.5.3",
```
shahms: you'll need to change the sha256 from external.bzl so it's the correct one; bizarrely Bazel seems to not give an error if it's wrong, it just keeps the cached version (regardless of URL).

shahms: (I suspect this is a bug in the repository rule, though)

jaysachs: done

PR #3596, opened 2019-03-08
feat(extraction): use bazelisk in Bazel extractor Dockerfile extraction

jaysachs:
Looking for feedback. 

(Haven't been able to test it yet due to local lein breakage holding up a local build of release, which impacts building this Dockerfile.)

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -668,6 +668,19 @@ def _kythe_contributions():
         remote = "https://github.com/kythe/lang-proto",
     )
 
+def _extractor_image_dependencies():
+    """Defines external repositories necessary for extractor images."""
+    git_repository(
+        name = "com_github_philwo_bazelisk",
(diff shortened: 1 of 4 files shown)


external.bzl
```diff
@@ -668,6 +668,19 @@ def _kythe_contributions():
         remote = "https://github.com/kythe/lang-proto",
     )
 
+def _extractor_image_dependencies():
+    """Defines external repositories necessary for extractor images."""
+    git_repository(
```
shahms: generally prefer http_archive of a github .zip file, rather than git_repository.  The former will download the snapshot of that version, the latter will download the entire commit history (and then largely discard it).

jaysachs: Done for bazelisk. Is there a corresponding "archive" for go_repositories?

shahms: NAFAIK

PR #3597, opened 2019-03-08
release: remove sample web ui

jaysachs:
lein is very finicky, and requires contortions to integrate into the bazel build (things like self-update, etc).

It's currently broken for internally used JDKs. It also only supports JDK8. 

#3600

### DIFF
--- a/kythe/release/release_test.sh
+++ b/kythe/release/release_test.sh
@@ -63,7 +63,7 @@ tools/triples < "$TEST_REPOSRCDIR/kythe/testdata/test.entries" >/dev/null
 
 rm -rf "$TMPDIR/java_compilation"
 export KYTHE_OUTPUT_FILE="$TMPDIR/java_compilation/util.kzip"
-export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/p:$JAVA_LANGTOOLS"
+export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/a:$JAVA_LANGTOOLS"
 JAVAC_EXTRACTOR_JAR=$PWD/extractors/javac_extractor.jar \
   KYTHE_ROOT_DIRECTORY="$TEST_REPOSRCDIR" \
   KYTHE_EXTRACT_ONLY=1 \
--- a/kythe/release/release_test.sh
+++ b/kythe/release/release_test.sh
@@ -100,29 +100,3 @@ fi
 
 # Ensure kythe tool is functional
 tools/kythe --api srv nodes 'kythe:?lang=java#pkg.Names'
-
-tools/http_server \
-  --public_resources web/ui \
-  --serving_table srv \
-  --listen $ADDR &
-pid=$!
-trap "kill $pid; kill -9 $pid" EXIT ERR INT
-
-while ! curl -s $ADDR >/dev/null; do
-  echo "Waiting for http_server..."
-  sleep 0.5
-done
-
-# Ensure basic HTTP handlers work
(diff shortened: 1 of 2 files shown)


schroederc:
`//kythe/release:release_test` needs fixed.

creachadair:
Could you please add some more context to the commit log?

I'm not saying we shouldn't remove—but it would be nice to have rationale in the log. 

If it does work: Why not release it?
If it doesn't work: do we have intent to fix it, or should we also consider deleting it?

creachadair:
Thanks for updating the description! Is there an issue number we can track here too, around fixing and/or replacing Leiningen?

jaysachs:
Just added the issue, and updated the description.

creachadair:
Thank you!

jaysachs:
So I think I can work around the buildbot issues with something like

```
--- a/kythe/release/release_test.sh
+++ b/kythe/release/release_test.sh
@@ -63,7 +63,12 @@ tools/triples < "$TEST_REPOSRCDIR/kythe/testdata/test.entries" >/dev/null
 
 rm -rf "$TMPDIR/java_compilation"
 export KYTHE_OUTPUT_FILE="$TMPDIR/java_compilation/util.kzip"
-export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/a:$JAVA_LANGTOOLS"
+if (java -version |& head -1 | grep 1.8 >/dev/null);
+then
+  export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/p:$JAVA_LANGTOOLS"
+else
+  export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/a:$JAVA_LANGTOOLS"
+fi
 JAVAC_EXTRACTOR_JAR=$PWD/extractors/javac_extractor.jar \
   KYTHE_ROOT_DIRECTORY="$TEST_REPOSRCDIR" \
   KYTHE_EXTRACT_ONLY=1 \

```
but that seems really awful.

kythe/release/release_test.sh
```diff
@@ -63,7 +63,12 @@ tools/triples < "$TEST_REPOSRCDIR/kythe/testdata/test.entries" >/dev/null
 
 rm -rf "$TMPDIR/java_compilation"
 export KYTHE_OUTPUT_FILE="$TMPDIR/java_compilation/util.kzip"
-export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/p:$JAVA_LANGTOOLS"
+if (java -version |& head -1 | grep 1.8 >/dev/null);
+then
+  export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/p:$JAVA_LANGTOOLS"
+else
+  export KYTHE_JAVA_RUNTIME_OPTIONS="-Xbootclasspath/a:$JAVA_LANGTOOLS"
+fi
```
jaysachs: Anyone OK with this hack?

shahms: Anything involved in attempting to support multiple versions of Java across the 8/9 boundary is going to be a hack and this seems acceptable until we can remove support for < 8 entirely.

kythe/release/release_test.sh
```diff
@@ -100,29 +105,3 @@ fi
 
 # Ensure kythe tool is functional
 tools/kythe --api srv nodes 'kythe:?lang=java#pkg.Names'
-
-tools/http_server \
-  --public_resources web/ui \
```
schroederc: Only delete this line and 117.

jaysachs: done

PR #3605, opened 2019-03-11
fix(go_indexer): always emit anchor for anon value types

schroederc:
Do not miss emitting definitions for anonymous value types when a
variable declaration duplicates the anonymous type.

Whether to join the duplicate types is left to a later decision.  This
change ensures that the reference to Y2 at least has *a* definition
location.

### DIFF
--- a/kythe/go/indexer/testdata/basic/anonymous.go
+++ b/kythe/go/indexer/testdata/basic/anonymous.go
@@ -51,6 +51,25 @@ var w struct {
 //- @X ref X
 var _ = w.X
 
+var y struct {
+	//- @Y defines/binding Y
+	//- Y.node/kind variable
+	//- Y.subkind field
+	Y int
+} = struct {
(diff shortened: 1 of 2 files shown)


kythe/go/indexer/testdata/basic/anonymous.go
```diff
@@ -51,6 +51,25 @@ var w struct {
...
+	//- @Y defines/binding Y
+	//- Y.node/kind variable
+	//- Y.subkind field
+	Y int
+} = struct {
+	//- @Y defines/binding Y2
+	//- !{ @Y defines/binding Y
+	//-    @Y ref Y }
+	Y int
+	// TODO(schroederc): investigate joining the duplicate anon types
```
creachadair: Technically they're different types that just happen to be assignable because at least one of them is not named. Not that you couldn't choose to unify them, but it is likely to be more trouble than it's worth (and may not be "correct").

PR #3607, opened 2019-03-11
fix(java_extractor): initial work enabling extraction of openjdk11

shahms:
This is some of the initial groundwork necessary for extracting OpenJDK9+ (focusing on 11).  Specifically, it implements the entirety of StandardJavaFileManager interface (which is required when compiling modular sources) and falls back to the first available compiler if the system compiler isn't available (because when compiling the JDK itself the newer compiler uses a non-standard name to avoid conflicting).

Finally, as a hack in Javac9Wrapper.java, restore the  -d flag if present as it is removed by JavaCompilationUnitExtractor but is required for modular builds.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
@@ -659,7 +659,7 @@ public class JavaCompilationUnitExtractor {
 
     // We will initialize and run the Javac compiler to detect which dependencies
     // the current compilation has.
-    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    JavaCompiler compiler = findJavaCompiler();
     if (compiler == null) {
       // TODO(schroederc): provide link to further context
       throw new IllegalStateException(
@@ -930,6 +930,23 @@ public class JavaCompilationUnitExtractor {
     completeOptions.add(tempDestinationDir.toString());
 
     return completeOptions;
+  }
+
+  private static JavaCompiler findJavaCompiler() {
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    if (compiler == null) {
+      // This is all a bit of a hack to be able to extract OpenJDK itself, which
+      // uses a bootstrap compiler and a lot of JDK options to compile itself.
+      // Notably, when using modules the system compiler is inhibited and the actual compiler
+      // resides in jdk.compiler.iterim.  Rather than hard-code this, just fall back to the first
+      // JavaCompiler we can find.
+      logger.atWarning().log("Unable to find system compiler, using first available.");
+      JavaCompilationUnitExtractor.class.getModule().addUses(JavaCompiler.class);
+      ServiceLoader<JavaCompiler> sl =
+          ServiceLoader.load(JavaCompiler.class, ClassLoader.getSystemClassLoader());
+      return sl.findFirst().orElse(null);
+    }
+    return compiler;
   }
 
   /** Returns a map from a classfile's {@link URI} to its sourcefile path's basename. */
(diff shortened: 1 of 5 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -932,6 +932,23 @@ public static ClassLoader getPlatformClassLoader() {
...
+    if (compiler == null) {
+      // This is all a bit of a hack to be able to extract OpenJDK itself, which
+      // uses a bootstrap compiler and a lot of JDK options to compile itself.
+      // Notably, when using modules the system compiler is inhibited and the actual compiler
+      // resides in jdk.compiler.iterim.  Rather than hard-code this, just fall back to the first
+      // JavaCompiler we can find.
+      logger.atWarning().log("Unable to find system compiler, using first available.");
+      JavaCompilationUnitExtractor.class.getModule().addUses(JavaCompiler.class);
+      ServiceLoader<JavaCompiler> sl =
+          ServiceLoader.load(JavaCompiler.class, ClassLoader.getSystemClassLoader());
```
schroederc: Should this be `JavaCompilationUnitExtractor.class.getModule().getClassLoader()`?

kythe/java/com/google/devtools/kythe/extractors/java/standalone/Javac9Wrapper.java
```diff
@@ -114,17 +114,35 @@ protected CompilationDescription processCompilation(
...
+            classPaths,
+            bootclasspath,
+            sourcePaths,
+            processorPaths,
+            processors,
+            genSrcDir,
+            completeOptions,
+            outputDirectory);
+
+    // Restore output directory, if present. It will be stripped by the extractor but is required
```
schroederc: Why not have the extractor leave the option?

shahms: It is replaced with a temporary directory to, I suspect, avoid trampling the outputs from the actual compiler but can easily be restored when building the CompilationUnit rather than merely removed.

kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
shahms: I honestly don't know, but it's an easy enough change to make.  It also occurs to me that I'll have to do any of the calls involving modules via reflection to avoid breaking on JDK8 :-(

shahms:
Comments addressed, PTAL.

shahms:
Also doing some of the TODO's around reflective access as we actually "link" against jdk9 everywhere (even when building with jdk8).

kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -294,10 +294,12 @@ public CompilationDescription extract(
...
   }
 
-  /** Returns a new list with the same options except any destination directory options. */
+  /**
+   * Returns a new list with the same options except header/source destination directory options.
+   */
   private static List<String> removeDestDirOptions(Iterable<String> options) {
     return JavacOptionsUtils.removeOptions(
-        Lists.newArrayList(options), EnumSet.of(Option.D, Option.S, Option.H));
+        Lists.newArrayList(options), EnumSet.of(Option.S, Option.H));
```
schroederc: Is the removal and replacement of this option on line 930 still acceptable?

shahms: Yes; that is done as part of the extraction and (AFAIK) is done so that any generated classes don't clobber the actual compiler output.

PR #3612, opened 2019-03-13
fix(java_extractor): use reflection to access/override JDK bits.

shahms:
Internally we still need to support JDK 8. Which is complainy when you try to call methods that don't exist ;-)

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/UsageAsInputReportingFileManager.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/UsageAsInputReportingFileManager.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/UsageAsInputReportingFileManager.java
```diff
@@ -145,38 +146,87 @@ public void setLocation(Location location, Iterable<? extends File> path) throws
...
+      return Iterables.transform(
+          (Iterable<? extends JavaFileObject>)
+              StandardJavaFileManager.class
+                  .getMethod("getJavaFileObjectsFromPaths", Iterable.class)
+                  .invoke(fileManager, paths),
+          input -> map(input, null));
+    } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+      throw new IllegalStateException(
+          "getJavaFileObjectsFromPaths called by unsupported Java version", e);
+    }
```
jaysachs: Generally we'd want to capture the method once in a static, to restrict the reflection (other than the "invoke()") only to startup.

shahms: That's easy enough to do

PR #3620, opened 2019-03-14
fix(java_extractor): change an iterable to a collection

zrlk:


shahms:
This is an override of a JDK9 method.  It is an Iterable, not a Collection, regardless of the warning: https://docs.oracle.com/javase/9/docs/api/javax/tools/StandardJavaFileManager.html#getJavaFileObjectsFromPaths-java.lang.Iterable-

Please revert.

PR #3626, opened 2019-03-18
fix(docs): NodeKind.getSubkind returns Optional.

jaysachs:


kythe/java/com/google/devtools/kythe/analyzers/base/NodeKind.java
```diff
@@ -67,7 +67,7 @@ public final String getKind() {
     return kind;
   }
 
-  /** Returns the node's subkind Kythe GraphStore value (or {@code null}). */
+  /** Returns the node's subkind Kythe GraphStore value (or {@code Optional.empty()}). */
```
creachadair:
```suggestion
  /** Returns the node's subkind Kythe GraphStore value. */
```

creachadair: In re: the above suggestion. Arguably it isn't really necessary to document the intrinsic behaviour of the `Optional` type here. If there were some interesting condition under which the value would be absent, that might be worth noting—but in this case it's simply whether or not the property was set by the indexer.

jaysachs: I tend to agree, but it was flagged internally and it's certainly not incorrect.

jaysachs: I do prefer it. Went with the abbreviated form.

Issue #3613, opened 2019-03-13
All `tapp`s for functions should specify a receiver on `param.2`


schroederc:
All `tapp` nodes for a `fn` builtin should have at least 3 `param` edges with `param.0` being the language-specific `fn` constructor builtin, `param.1` being the function's return type (possible being a language-specific "void" type), `param.2` being the functions receiver type (possibly being a language-specific "empty" receiver type), and all subsequent `param.N` edges are the function's formal parameter types.

Go example:

```
//- @S defines/binding S
type S struct {}

//- @F defines/binding F
//- F typed FType
//- FType.node/kind tapp
//- FType param.0 FnBuiltin
//- FType param.1 BoolBuiltin
//- FType param.2 S
//- FType param.3 IntBuiltin
func (S) F(a int) bool { return false }

//- @G defines/binding G
//- G typed GType
//- GType.node/kind tapp
//- GType param.0 FnBuiltin
//- GType param.1 BoolBuiltin
//- GType param.2 EmptyTuple
//- GType param.3 IntBuiltin
func G(a int) bool { return false }
```

creachadair:
This seems like a generally good idea. Maybe instead of "EmptyTuple" we could borrow the name "void" here? An empty tuple is more like a unit type than an uninhabited type, though maybe that doesn't matter too much.

More importantly, though, it'd be nice to share that across languages.

schroederc:
I agree that it would be nice to share builtin nodes like the "empty" receiver type.  For now, all builtins are language-specific in Kythe so the first pass of this feature will follow that trend and will probably use `EmptyTuple` instead of inventing a new builtin.  Later on, I would like to look into possibly unifying builtin nodes.

creachadair:
In retrospect, we should probably have added this to the schema before we started changing indexers—but in any case, we should do that.

schroederc:
Schema documentation updated in #3628.

PR #3628, opened 2019-03-18
feat(java_indexer): add receiver type to function types

schroederc:
Part of #3613

### DIFF
--- a/kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Methods.java
+++ b/kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Methods.java
@@ -6,15 +6,27 @@ import java.io.IOException;
 //- @Methods defines/binding Class
 public class Methods {
 
+  //- ArrayBuiltin=vname("array#builtin","","","","java").node/kind tbuiltin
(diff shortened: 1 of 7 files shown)


kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/Methods.java
```diff
@@ -6,15 +6,27 @@
...
+  //- IntBuiltin=vname("int#builtin","","","","java").node/kind tbuiltin
+  //- VoidBuiltin=vname("void#builtin","","","","java").node/kind tbuiltin
+
   //- @Methods defines/binding Ctor
   //- Ctor.node/kind function
   //- Ctor childof Class
   //- Ctor typed CtorType
   //- CtorType param.0 FnBuiltin
   //- CtorType param.1 Class
+  //- CtorType param.2 Class
```
zrlk: I'm not sure this is entirely right, but I don't have any alternatives I like better.

zrlk: Also, it looks like the schema document needs updated.

PR #3629, opened 2019-03-19
fix(bazel): Extraction exits with nonzero error code if unable to create kzip

jaysachs:
Since we invoke bazel with "--keep_going", we need to defer error detection to this point. If there is any extracted compilation unit, we consider it "successful".


kythe/extractors/bazel/extract.sh
```diff
@@ -50,5 +50,6 @@
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
 find bazel-out/*/extra_actions/external/kythe_release -name '*.kzip' | \
-  xargs /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
+  xargs -r /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
+test -f "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip" || exit $?
```
schroederc: Put this last and remove the `|| exit $?`

jaysachs: Done.

PR #3632, opened 2019-03-19
fix(java_indexer): box LOOKUP_BY_PARAM MarkedSources with text

schroederc:


kythe/java/com/google/devtools/kythe/analyzers/java/MarkedSources.java
```diff
@@ -89,15 +89,19 @@ private MarkedSources() {}
     MarkedSource.Builder methodTAppBuilder =
         MarkedSource.newBuilder().setKind(MarkedSource.Kind.TYPE);
     methodTAppBuilder
+        .addChildBuilder()
+        .setKind(MarkedSource.Kind.BOX)
```
creachadair: I have been bitten so many times by the arcane rules around the expansion and rendering of marked source. I think we desperately need a combination of better documentation (right now, the implementation is the spec) and better tools for testing the outcome.

This is particularly difficult for the LOOKUP type nodes because you need to simulate not only the rendering but also the joining. Yet right now the fact that I and other core members of the team keep stumbling on this says to me we have a deep usability problem here.

(The change seems fine, by the way, I'm just using this as a focus to raise the wider point)

PR #3634, opened 2019-03-20
chore: map[*]bool, map[string]struct{} -> map[*]struct{}, stringset

danielmoy-google:
Use less memory, use better interface where possible.

### DIFF
--- a/kythe/go/indexer/emit.go
+++ b/kythe/go/indexer/emit.go
@@ -97,8 +97,8 @@ func (pi *PackageInfo) Emit(ctx context.Context, sink Sink, opts *EmitOptions) e
 		pi:       pi,
 		sink:     sink,
 		opts:     opts,
-		impl:     make(map[impl]bool),
-		anchored: make(map[ast.Node]bool),
+		impl:     make(map[impl]struct{}),
+		anchored: make(map[ast.Node]struct{}),
 	}
 
 	// Emit a node to represent the package as a whole.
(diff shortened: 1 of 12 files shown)


kythe/go/indexer/emit.go
```diff
@@ -163,9 +163,9 @@ type emitter struct {
 	pi       *PackageInfo
 	sink     Sink
 	opts     *EmitOptions
-	impl     map[impl]bool                        // see checkImplements
+	impl     map[impl]struct{}                    // see checkImplements
 	rmap     map[*ast.File]map[int]metadata.Rules // see applyRules
-	anchored map[ast.Node]bool                    // see writeAnchor
+	anchored map[ast.Node]struct{}                // see writeAnchor
```
creachadair: If you find yourself re-implementing this a lot, you may also find it useful to know that the `stringset` code is actually generated and can be re-generated for other types easily.

```shell
# N.B. This example omits the set transforms like Map, Select, etc.
# Set "transforms": true if you want them.
cat > nodeset.json <<EOF
{
  "package": "nodeset",
  "type": "ast.Node",
  "zero": "nil",
  "less": "if x.Pos() == y.Pos() { return x.End() > y.End() } else { return x.Pos() < y.Pos() }",
  "imports": ["go/ast"],
  "transforms": false
}
EOF
go run bitbucket.org/creachadair/stringset/makeset -config makeset/nodeset.json -output nodeset
```

I haven't wired this up to Bazel specifically, but it could be done pretty easily I think with a genrule. YMMV.

PR #3638, opened 2019-03-21
fix(leiningen): fail lein setup w/ incompatible JDK

shahms:
Update the Leiningen repository rule and wrapper script to:

1) Explicitly check `PATH` and` LEIN_JAVA_CMD` for the `java` executable.
2) Use that directly in the wrapper script, if available.
3) But fall back to the system path to support remote execution.


third_party/leiningen/lein_repo.bzl
```diff
@@ -71,4 +82,26 @@ lein_repository = repository_rule(
...
         "sha256": attr.string(mandatory = True),
     },
+    environ = ["PATH", "LEIN_JAVA_CMD"],
 )
+
+def _check_jdk(repository_ctx):
+    java = repository_ctx.os.environ.get("LEIN_JAVA_CMD") or repository_ctx.which("java")
+    result = repository_ctx.execute([java, "-version"])
+    if result.return_code != 0:
+        fail("Error determining JDK version: {error} (with LEIN_JAVA_CMD={java})".format(
```
salguarnieri: Does it still make sense to say LEIN_JAVA_CMD= when we get java from which("java")?

shahms: IMO, yes as we are using this as the default value of that variable if it's empty or missing.  Plus it's an easy way to document that the user can specify it explicitly if they should choose to do so.

PR #3647, opened 2019-03-25
fix(bazel): include jsr250 jar for java extraction

jaysachs:


kythe/release/release.BUILD
```diff
@@ -21,12 +21,21 @@ construct_vnames_config(
...
+)
+
+java_library(
+    name = "protobuf",
+    visibility = ["//visibility:private"],
+    exports = [
+        "@com_google_protobuf//:protobuf_java",
+    ],
+    deps = [
+        "@com_google_protobuf//:protobuf_java",
```
shahms: Does this not need jsr250 annotations as well?

jaysachs: Turns out none of the changes in this file are needed. I originally did have them, but didn't finish removing the unnecessaries.

jaysachs: So after several experiments, I conclude that this does not need the annotations, however "deps" needed to be "runtime_deps". I'm not quite certain why the presence of this helps. Perhaps it strips out the "Generated" annotation.

PR #3649, opened 2019-03-25
Remove obsolete assertion

cushon:
The referenced javac bug has been fixed:
https://bugs.openjdk.java.net/browse/JDK-8189782


danielmoy-google:
(adding buildbot, should pick it up)

PR #3653, opened 2019-03-27
release: change buildbot config to use secrets directly

shahms:
Rather than include the sed shenanigans in the deploy.sh script, read them from master.cfg directly.  Also and a cloudbuild.yaml file to allow building buildbot images using GCB.

### DIFF
--- a/kythe/release/appengine/buildbot/cloudbuild.yaml
+++ b/kythe/release/appengine/buildbot/cloudbuild.yaml
@@ -29,7 +29,7 @@ steps:
   args:
   - push
   - gcr.io/$PROJECT_ID/buildbot.$_VERSION
-timeout: 30m
+timeout: 1800s # 30 minute timeout
 images:
   - gcr.io/$PROJECT_ID/buildbot.$_VERSION
 substitutions:
--- a/kythe/release/appengine/buildbot/cloudbuild.yaml
+++ b/kythe/release/appengine/buildbot/cloudbuild.yaml
@@ -5,6 +5,8 @@ steps:
   - cp
   - gs://kythe-buildbot/secrets.tar.enc
   - secrets.tar.enc
+# Decrypt retrieved credentials.
+# https://cloud.google.com/cloud-build/docs/securing-builds/use-encrypted-secrets-credentials
 - name: gcr.io/cloud-builders/gcloud
   id: 'Decrypt secrets'
   args:
@@ -22,6 +24,11 @@ steps:
   - -t
   - gcr.io/$PROJECT_ID/buildbot.$_VERSION
   - .
+- name: gcr.io/cloud-builders/docker
+  id: 'Push image'
+  args:
+  - push
+  - gcr.io/$PROJECT_ID/buildbot.$_VERSION
 timeout: 1800s # 30 minute timeout
 images:
   - gcr.io/$PROJECT_ID/buildbot.$_VERSION
--- a/kythe/release/appengine/buildbot/cloudbuild.yaml
+++ b/kythe/release/appengine/buildbot/cloudbuild.yaml
@@ -29,7 +29,7 @@ steps:
   args:
   - push
   - gcr.io/$PROJECT_ID/buildbot.$_VERSION
-timeout: 1800s # 30 minute timeout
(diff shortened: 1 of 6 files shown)


kythe/release/appengine/buildbot/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - --ciphertext-file=secrets.tar.enc
+  - --plaintext-file=secrets.tar
+- name: gcr.io/cloud-builders/docker
+  id: 'Build image'
+  args:
+  - build
+  - -t
+  - gcr.io/$PROJECT_ID/buildbot.$_VERSION
+  - .
+timeout: 1800s # 30 minute timeout
```
danielmoy-google: actually I'm fairly certain you can just write 30m

This just gets parsed as a duration or whatever

kythe/release/appengine/buildbot/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+- name: gcr.io/cloud-builders/gsutil
+  id: 'Fetch secrets'
+  args:
+  - cp
+  - gs://kythe-buildbot/secrets.tar.enc
+  - secrets.tar.enc
+- name: gcr.io/cloud-builders/gcloud
+  id: 'Decrypt secrets'
+  args:
+  - kms
```
danielmoy-google: maybe put a comment above this step pointing to the gcb doc on kms

https://cloud.google.com/cloud-build/docs/securing-builds/use-encrypted-secrets-credentials

kythe/release/appengine/buildbot/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - decrypt
+  - --location=global
+  - --keyring=Buildbot
+  - --key=secrets
+  - --ciphertext-file=secrets.tar.enc
+  - --plaintext-file=secrets.tar
+- name: gcr.io/cloud-builders/docker
+  id: 'Build image'
+  args:
+  - build
```
danielmoy-google: does this do push automatically?

shahms: Yes, because `images` was specified below, according to https://cloud.google.com/cloud-build/docs/configuring-builds/store-images-artifacts

shahms: You can apparently do either or both, depending on needs.  I can add an explicit push, but it shouldn't be necessary.

kythe/release/appengine/buildbot/cloudbuild.yaml
shahms: Done.

kythe/release/appengine/buildbot/cloudbuild.yaml
shahms: The documentation suggests otherwise (as do all of the examples):
```
timeout
Use the timeout field in a build step to set a time limit for executing the step. If you don't set this field, the step has no time limit and will be allowed to run until either it completes or the build itself times out. timeout must be specified in seconds with up to nine fractional digits, terminated by 's'. Example: "3.5s"
```

danielmoy-google: Ah I was just looking around at random examples:

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/testing/cloudbuild/cloudbuild.yaml#L15

the underlying proto is a Duration

I would at least try it once with 30m and see if it breaks

shahms: ERROR: (gcloud.builds.submit) INVALID_ARGUMENT: Invalid value at 'build.timeout' (type.googleapis.com/google.protobuf.Duration), Field 'timeout', Illegal duration format; duration must end with 's'
- '@type': type.googleapis.com/google.rpc.BadRequest
  fieldViolations:
  - description: Invalid value at 'build.timeout' (type.googleapis.com/google.protobuf.Duration),
      Field 'timeout', Illegal duration format; duration must end with 's'
    field: build.timeout

PR #3657, opened 2019-03-29
chore: remove unused visilibity from downstream automated fix

danielmoy-google:


danielmoy-google:
Ah bummer, we'll have to fix this on our import then, since we do actually use it out here.

PR #3658, opened 2019-03-29
Merge lang-proto repo

justbuchanan:
A while ago, we discussed splitting kythe into multiple smaller per-language repos, but decided (at least for now) to keep them all in one. As part of that decision, we decided to merge the lang-proto repo back into the main one.

Git has a nifty `--allow-unrelated-histories` flag for merge, which I used here so we can keep the git history. Unfortunately none of the github issue/pr numbers in the merged commit messages will make sense anymore.


kythe/cxx/extractor/proto/proto_extractor.cc
```diff
@@ -0,0 +1,197 @@
...
+    }
+    if (opened_files_.find(canonical_path) == opened_files_.end()) {
+      opened_files_.insert(canonical_path);
+    }
+
+    return DiskSourceTree::Open(filename);
+  }
+
+  // A set of unique file paths that have been passed to Open().
+  const std::set<std::string>& opened_files() const { return opened_files_; }
```
shahms: It looks a lot like the C++ files need to be run through clang-format (or those of us with the suggested commit-hooks set up will get warning/errors).

justbuchanan: I just ran clang-format and it only found a couple things in proto_indexer.cc and index_pack_test.cc. We're using the config at `kythe/.clang-format`, right?

shahms: Yup, that's right.  This line looked a bit long, but I miscounted :-)

justbuchanan:
About to merge this. The only option on github's web ui is to squash all the commits - do we want to do an actual merge instead so we can keep the history? I can do that via command line git instead.

schroederc:
I think keeping the history is preferable, if possible.

PR #3662, opened 2019-04-02
chore: apply fixes from clang_tidy

justbuchanan:


kythe/cxx/indexer/cxx/ImputedConstructorSupport.h
```diff
@@ -32,7 +32,7 @@ namespace kythe {
 class ImputedConstructorSupport : public LibrarySupport {
  public:
   explicit ImputedConstructorSupport(
-      std::unordered_set<std::string> allowed_constructor_patterns = {
```
shahms: This should be `patterns`; if it's names elsewhere those should be changed.

kythe/cxx/indexer/cxx/ImputedConstructorSupport.h
```diff
@@ -32,9 +32,9 @@ namespace kythe {
...
```
shahms: Sorry, I meant that it should be "allowed_constructor_patterns" rather than "allowed_constructor_names" (because they are regular expressions)

PR #3665, opened 2019-04-04
do not submit - testing changes against buildbot

justbuchanan:
build: use toolchain instead of local include dirs for verifier tests


tools/build_rules/verifier_test/cc_indexer_test.bzl
```diff
@@ -240,10 +240,8 @@ cc_kythe_proto_library = rule(
 
 def _cc_extract_kzip_impl(ctx):
     cpp = find_cpp_toolchain(ctx)
-    if cpp.libc == "macosx":
-        toolchain_includes = cpp.built_in_include_directories
-    else:
-        toolchain_includes = []
+    toolchain_files = cpp.all_files if hasattr(cpp, 'all_files') else depset()
```
shahms: Should just be able to write: `getattr(cpp, "all_files", depset())`

PR #3666, opened 2019-04-04
feat: accumulate granular file counts of kzip contents

jaysachs:


kythe/go/extractors/validation/validation.go
```diff
@@ -62,6 +65,9 @@ type Coverage struct {
 	Paths []Dir
 }
 
+// Stats on contents of archive
```
danielmoy-google: Worth putting expected key space(s) in here, or just leave it freeform?

jaysachs: Done

kythe/go/extractors/validation/validation.go
```diff
@@ -28,6 +28,7 @@ import (
 	"strings"
 
 	"kythe.io/kythe/go/platform/kzip"
+	apb "kythe.io/kythe/proto/analysis_go_proto"
```
creachadair: Conventionally we group protobuf imports separately below the others.

kythe/go/extractors/validation/validation.go
```diff
@@ -81,6 +81,8 @@ type Dir struct {
 	Missing int
 }
 
+// accumulate accumulates statistics (currently granular file counts)
```
creachadair: It might also be worth noting that calling this more than once is not idempotent. It's not an exported API, but that could easily happen along multiple paths.

kythe/go/extractors/validation/validation.go
```diff
@@ -65,7 +65,7 @@ type Coverage struct {
 	Paths []Dir
 }
 
-// Stats on contents of archive
+// Stats on contents of archive, file count by corpus/extension
```
creachadair: Suggestion:

```go
// Statistics collects file count stats on the contents of an archive per corpus, a map
// from corpus to file extension to count.
```

kythe/go/extractors/validation/validation.go
jaysachs: Done

kythe/go/extractors/validation/validation.go
jaysachs: Done.

kythe/go/extractors/validation/validation.go
jaysachs: Done.

PR #3667, opened 2019-04-05
 feat(proto_indexer): add tests for oneof field xrefs

justbuchanan:
These annotations will be emitted by the protobuf compiler after https://github.com/protocolbuffers/protobuf/pull/5991.

Note that java protoc doesn't yet emit annotations for oneof fields. I'll add that in a future PR.

~~DO NOT SUBMIT - wait until the protobuf PR is merged.~~


justbuchanan:
The protobuf PR has been merged, so this should be good to go.

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/proto/testdata.proto
```diff
@@ -24,4 +24,11 @@ message Message {
     //- @nested_bool defines/binding NestedBool
     bool nested_bool = 2;
   }
+
+  //- @oneof_field defines/binding OneofField?
```
shahms: Remove the `?` ?

justbuchanan: This proto is used in both the java and c++ tests. Without the `?`, the verifier fails on the java tests because it isn't referenced there. I'll send a PR to protobuf team soon for the java annotations, but they don't exist yet.

shahms: Add `check_for_singletons=False` to the Java tests

justbuchanan: :+1:

PR #3673, opened 2019-04-05
chore: refactor JavacOptionsUtils into mostly a builder format

danielmoy-google:
Allows us to get rid of some slow list behavior, helps usability on the
interface so we don't have to remember which methods modify-in-place vs.
remove vs. return new copies vs. some combination of the above.


kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
```diff
@@ -61,177 +67,209 @@
...
+    public ModifiableOptions removeUnsupportedOptions() {
+      List<String> previous = internal;
+      internal = new ArrayList<>();
+      ImmutableList<OptionChecker> optionCheckers =
+          ImmutableList.of(JavacTool.create(), new JavacFileManager(new Context(), false, UTF_8));
+      PeekingIterator<String> it =
+          Iterators.peekingIterator(
+              Iterators.filter(
+                  previous.iterator(),
+                  Predicates.not(o -> o.startsWith("-Xlint") || o.startsWith("-Werror"))));
```
shahms: Just do this as an additional option checker?

kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
```diff
@@ -61,177 +67,209 @@
...
+      internal = new ArrayList<>(original.size());
+      for (int i = 0; i < original.size(); i++) {
+        String opt = original.get(i);
+        boolean matched = false;
+        for (Option o : opts) {
+          if (o.matches(opt)) {
+            matched = true;
+            if (o.hasArg()) {
+              // Skip the argument too.
+              i++;
```
shahms: This is nearly identical (semantically, if not syntactically) with `removeUnsupported` and should probably be unified.

kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
```diff
@@ -61,177 +67,209 @@
...
+     * Identify the paths associated with the specified option, remove them from the builder, and
+     * return them.
+     */
+    private ImmutableList<String> removeArgumentPaths(Option option) {
+      List<String> original = internal;
+      internal = new ArrayList<>(original.size());
+      ImmutableList.Builder<String> paths = ImmutableList.builder();
+      for (int i = 0; i < original.size(); i++) {
+        if (option.matches(original.get(i))) {
+          if (i + 1 >= original.size()) {
```
shahms: Ditto here on this loop.

kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
shahms: My favorite pattern for dealing with this comes from the JavaFileManager class and is essentially the interface:

interface OptionHandler {
Iterator<String> handleOption(String value, Iterator<String> remaining);
}

// remove
(opt, rest) -> rest
// w/ value
(opt, rest) -> { rest.next(); return rest; }
// keep
(opt, rest) -> Iterator.concat(ImmutableList.of(opt), rest);

etc.

kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
danielmoy-google: sounds good

kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
danielmoy-google: ehh, maybe just the one loop can do it without arbitrary handler.

But, similar logic for different handling of what to do on match / not-match / positionally-matched makes sense still imo.

Issue #74, opened 2018-05-31
Document the supported formats for cross-language linkage metadata


creachadair:
http://www.kythe.io/docs/schema/indexing-protobuf.html talks about embedding metadata in separate files, but we also allow a couple of new formats too (cf. #70 for example). We should document these in a more public fashion so that users will have a way to decide which to use.

PR #3674, opened 2019-04-06
Improve documentation for cross-language support

nbeloglazov:
Restructured the doc to describe two different approaches, one with `impute` edges
and the second one with `generates` edges.

Related to https://github.com/kythe/kythe/issues/74


nbeloglazov:
I'm going to re-read the new docs later. But for now wanted to see if overall document structure make sense. And whether there is a way to render it nicely for easier reviewing. I couldn't build kythe locally for some clang building issues so couldn't render ascii as nice html and instead used https://asciidoclive.com/ which is not complete (doesn't render graphviz).

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+
+Now we want to teach Kythe how to create cross-references between generated
+`foo.bob` file and original `foo.alice` file. The main idea is pretty simple:
+Generator has to output extra data containing mapping of elements in `foo.bob`
+to the original elements from `alice.bob`. Then when Bob Indexer is indexing
+`foo.bob` it will use that mapping to output *generates* or *imputes* edges.
+These edges connect nodes from `foo.bob` with nodes in `foo.alice`.
+
+Kythe doesn't require implementors to use one concrete approach for passing
+mapping metadata and outputting *generates* and *imputes* edges. Below we
```
shahms: Not sure if this is mentioned elsewhere, but it might be worth mentioning why one might be preferred to the other.

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
 document describes how to encode information about generated code to permit
 cross-language links.
 
+To make the discussion easier to understand let's pretend we are working with
+two languages: AliceLang and BobLang. AliceLang has `.alice` file and BobLang
```
shahms: SourceLang TargetLang?

Alice and Bob are commonly used examples, but it's not necessarily obvious which is generated from the other.  It might be beneficial to use more descriptive names?

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+==== Changes to Generator
+
+To support cross-references betwen `color.js` and `Color.java` we need to update
+Generator to output the following mapping data for `Color`, `RED` and `GREEN`
+elements.
+
+[source,json]
+-------------------------------------------------------------------------------
+{
+    "type": "kythe0",
```
shahms: We should have a schema for these values as well, rather than just guessing at the supported bits :-(

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+-------------------------------------------------------------------------------
+
+This mapping has 3 `meta` entries. The first entry for `Color`, the second for
+`RED`. Note:
+
+* Each entry doesn't contain names of elements. Each entry contains only
+  position of elements in the source (`Color.java`) and target (`color.js`)
+  files.
+* Each position is defined as byte offset inside file and not as line/column.
+  This is required because in Kythe anchors defined using byte offsets and
```
shahms: "anchors are"

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+  not line/column. In this example JavaScript indexer will process this
+  mapping and will need to output *anchor* for `Color.java` and indexer
+  doesn't have access to the `Color.java` file (it has access only to JS
+  files). Because of that JS indexer can't translate line/column to byte
+  offset.
+* Entry doesn't contain vnames of elements in `Color.java` or `color.js` and
+  instead contains positions. Vnames of nodes are internal details of each
+  indexer and subject to change. Generator usually a standalone tool that
+  doesn't know rules for producing vnames for specific language so it's
+  impossible for Generator to output vnames of nodes.
```
shahms: -ish

For things like protobuf (and mojom?) the VName generation rules are specified to make it possible to create generates edges, rather than relying on imputes.  It might be worth expanding on that a bit? (or not, that might just make it confusing).

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+==== Changes to JavaScript Indexer
+
+On JavaScript Indexer side we need to parse metadata and output *imputes*
+edges. To parse metadata indexer can check last two lines of all `.js` files
+and see if they contain `// Kythe Indexind Metadata:` and if so - parse
+the last line as JSON.
+
+For each `meta` entry indexer should do the following:
+
+1. Output an *anchor* using `source_begin` and `source_end`. `source_vname`
```
shahms: This varies by edge and "type", e.g. kythe/cxx/indexer/cxx/testdata/metadata/inline.meta

Where a reverse-generates edge should be painted from the node found at that location to the semantic node designated by the vname.

shahms: Also looking at the above, it appears as though JS and C++ might use slightly different JSON metadata? That would be unfortunate...

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: That's a good question. Which one is preferred though? :) Does Kythe team prefers `imputes` approach because it's more general and doesn't require coupling between indexers and generators?

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: The purpose of schema would be mostly to remove the need for users to come with their own schema? Because generators and indexers would still need to write code that creates/parses JSON. I don't think Kythe can provide much tools for working with that schema given that Kythe supports many languages and re-implementing these tools in all possible languages as part of Kythe repo seems heavy.

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: How would you suggest to expand? There is a whole section protobuf below. I can add something like "for another approach based on VNames see Protcol Buffers section below"?

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: 1. This approach is talking about `imputes` edges, not `generates`. In my mental picture those are two different approaches based on availability of VNames. So in this doc I'm trying to keep them separate. If that's not the right mental picture - how should I change it?

2. What is that C++ JSON metadata used for? I was only aware of `GeneratedCodeInfo` binary proto used by protobuf <=> C++. And yes, this JSON and the one I described seem to have similar fields but different structure :/ Maybe I can unify them but need to understand better C++ use case.

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: Renamed. Yes SourceLang and TargetLang are easier to understand.

nbeloglazov:
Addressed/replied to comments. Please take another look. Thanks!

kythe/docs/schema/indexing-protobuf.txt
shahms: `generates` is slightly preferred because there's less work during post-processing, `imputes` exists to cover the case where the semantic VNames of the source language aren't stable/

kythe/docs/schema/indexing-protobuf.txt
shahms: Considering that the metadata needs to be understood by the target language indexer and many code generators can target several different languages from a particular source, it seems like there would be a fair amount of benefit from standardizing and documenting a single such format, rather than having each target language use a subtly different one.

kythe/docs/schema/indexing-protobuf.txt
shahms: I think a parenthetical to that effect would be useful, yeah.

kythe/docs/schema/indexing-protobuf.txt
shahms: That's the right mental picture.  But from a metadata perspective, they're similar. Same application rules, but different "payload".

I think that JSON metadata format is used by Mojom (a Chromium thing similar to proto). I filed #3711 to document/specify the formats.  Looking a bit more closely, it appears the structure varies a lot by the "type" field.  The "anchor_anchor" type is the same(?) in both JavaScript and C++, but C++ also supports "anchor_defines", which has a different structure.

shahms: I think for here, it's enough to note that different "type" fields have a different expected structure and here we're using "anchor_anchor" which is for imputes edges.

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+  RED: 0,
+};
+
+// Kythe Indexing Metadata:
+// {"type":"kythe0","meta":[{"type":"anchor_anchor","source_begin":13,...
+-------------------------------------------------------------------------------
+
+Inlining metadata inside `color.js` has benefit of avoiding passing extra
+files to Indexer. All Indexer needs is to know that some JavaScript files can
+contain metadata on the last line and parse it.
```
robinp: For the sake of separation of concerns, the generator would like to stay independent of Kythe, and the indexer of TargetLang would like to stay independent of any SourceLang or metadata.

Indexer-independence can be achieved by the generator emitting separate entries file. But generator independence is also possible, by having stable VName convention in both Source- and TargetLang. Then an external tool could generate the entries itself, using `generates` directly, instead of `imputes`.

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+-------------------------------------------------------------------------------
+const Color = {
+  RED: 0,
+};
+
+// Kythe Indexing Metadata:
+// {"type":"kythe0","meta":[{"type":"anchor_anchor","source_begin":13,...
+-------------------------------------------------------------------------------
+
+Inlining metadata inside `color.js` has benefit of avoiding passing extra
```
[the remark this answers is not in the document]
shahms: The indexer takes a range of text in the generated file.  When it emits a "defines/binding" edge covering that range, it also emits an edge from the semantic entity it just defined at that location towards the entity specified in the metadata file.  For "imputes" edges, the semantic entity points at an anchor. For "generates" edges the target language indexer emits a reverse edge to the semantic entity which created the text range.  Of note, the indexer in question *always* emits at least one semantic node whose VName is known only to it.

In both of these cases, the post-processor has to do some work:
For imputes edges, it has to emit the appropriate generates edge pair.
For generates edges, it has to emit the forward edge.

There is no way to feed data into post-processing which doesn't come from an indexer.

kythe/docs/schema/indexing-protobuf.txt
shahms: Stable semantic VNames are infeasible for a substantial fraction of languages and this is intentional.

For some languages with intentionally constrained purposes it is reasonable to document and specify VName construction rules.  This is what was done for protobuf, for example.  It is not feasible to do so for more full-features and evolving languages like C++, Java or JavaScript.

creachadair: Unfortunately even if we had a stable naming convention on both sides (which as @shahms pointed out turns out to be trickier than it seems), we still run into the fact that most code generators don't have enough information to emit complete names for the entities they define. The indexer is able to do this because it has the dependencies resolved already, but a code generator generally works a file at a time and cannot predict how the generated file will be built (e.g., the import path for a Go package, the preprocessor settings and include state for a C translation unit). At best, most generators can give only partial information.

A further complication is that many code generators (most, in my experience) do not work in an abstract model of the output language, but simply write text to a file. The "knowledge" of what program structures are being generated is implicit in the control flow of the program, and while the author of the code generator (hopefully) knows what she is doing, adding instrumentation is therefore fiddly.

So while it's certainly true that using location as a proxy for identity is _also_ not ideal in some ways, it has the advantage that it doesn't require a new abstraction for the language to be built and maintained: Instrumenting the code generator can be done via lightweight location tracking, and all the language-specific knowledge stays in the indexer.

As you might have noted, though, there are some cases where we can do better: For example, protocol buffers already contain a well-defined addressing scheme for structures in the schema language, which are provided to both the code generator and the indexer in the descriptor messages. This works out primarily because the naming scheme is part of the language itself, rather than an addition.

kythe/docs/schema/indexing-protobuf.txt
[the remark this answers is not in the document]
robinp: So IIUC the inlining is for the indexer to emit the semantic VName of the `defines/binding` generated at the given span (annotated in the inline data). If VNames are not predictable, that sounds fair.

About having to emit the forward edge for `generates`: why can't the indexer (or an external process) emit the forward edge directly? It is just the source and target swapped, isn't it?

kythe/docs/schema/indexing-protobuf.txt
robinp: Thank you. The incomplete information argument is convincing. So assuming stable language-level VNames, the tradeoff is making the indexer somewhat more complicated (parsing inline data), vs creating an external tool which would need to depend on build-system details. Got it!

creachadair: > [T]he tradeoff is making the indexer somewhat more complicated (parsing inline data), vs. creating an external tool which would need to depend on build-system details.

Exactly. There is no adequate build-system abstraction anyway (since ~all extant build tools except possibly Nix are defined by their implementation), and we'd rather pay the extra cost on the Kythe side than in the code generator.

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: Ok, I'll address it in a follow up PR.

kythe/docs/schema/indexing-protobuf.txt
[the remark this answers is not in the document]
nbeloglazov: Yes, I think it would be possible to pass anchor-to-anchor mapping from generator to post-processing and let post-processing to find two corresponding `defines/binding` anchors and stich semantic VNames together. But as Shahms mentioned there is no way to feed data outside of indexer. Personally it makes it a little bit cleaner as you can assume only indexers output data. 

As for forward edges for `generates` - I don't know. Maybe to keep indexer a little bit simpler?

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: Robin, could you suggest a way to improve this doc to make it clearer? I thought by providing two different approaches/examples and explaining each approach it would make it clear, but probably it didn't. Should I add a section about `imputes` vs `generates`?

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: Added comment. Wondering if I should add a special section at the end of doc that compares both approaches and goes into depth of why both are needed.

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,235 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+        "edge": "/kythe/edge/imputes",
+        "source_vname": {
+            "corpus": "corpus",
+            "path": "path/to/Color.java"
+        }
+    }]
+}
+-------------------------------------------------------------------------------
+
+This mapping has 3 `meta` entries. The first entry for `Color`, the second for
```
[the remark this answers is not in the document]
nbeloglazov: Thanks. Fixed.

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: Added a sentence.

kythe/docs/schema/indexing-protobuf.txt
```diff
@@ -31,12 +31,241 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+Generator has to output extra data containing mapping of elements in `foo.target`
+to the original elements from `foo.source`. Then when Target Indexer is indexing
+`foo.target` it will use that mapping to output *generates* or *imputes* edges.
+These edges connect nodes from `foo.target` with nodes in `foo.source`.
+
+Kythe doesn't require implementors to use one concrete approach for passing
+mapping metadata and outputting *generates* and *imputes* edges. Below we
+describe two different approaches, each has its own pros and cons. But in
+both cases it is assumed that implementors can change Generator and Target
+Indexer. If possible *generates* approach is preferred as requires less
```
shahms: Add "the" after "possible" and "it" before "requires"

shahms:
A huge thanks for adding to this documentation and sorry for the long delays.

shahms:
@zrlk thoughts?

shahms:
It looks like there are some asciidoc failures as well...

kythe/docs/schema/indexing-protobuf.txt
nbeloglazov: Done.

nbeloglazov:
Addressed the grammar comment, fixed asciidoc errors and renamed file to `indexing-generated-code`.

nbeloglazov:
Yay, it passed all tests.

shahms:
Still LGTM, but I'd like to get an @zrlk comment :-)

kythe/docs/schema/indexing-generated-code.txt
```diff
@@ -31,12 +31,241 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+-------------------------------------------------------------------------------
+const Color = {
+  RED: 0,
+};
+-------------------------------------------------------------------------------
+
+=== Changes to Generator
+
+To support cross-references betwen `color.js` and `Color.java` we need to update
+Generator to output the following mapping data for `Color`, `RED` and `GREEN`
```
zrlk: I don't see any `GREEN` elements.

kythe/docs/schema/indexing-generated-code.txt
```diff
@@ -31,12 +31,241 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+contain metadata on the last line and parse it.
+
+One downside is that it adds noise to `color.js` but usually generated
+files are invisible to developers so it's not a big concern.
+
+==== Changes to JavaScript Indexer
+
+On JavaScript Indexer side we need to parse metadata and output *imputes*
+edges. To parse metadata indexer can check last two lines of all `.js` files
+and see if they contain `// Kythe Indexind Metadata:` and if so - parse
```
zrlk:
```suggestion
and see if they contain `// Kythe Indexing Metadata:` and if so - parse
```

kythe/docs/schema/indexing-generated-code.txt
```diff
@@ -31,12 +31,241 @@ implementation, as this is often difficult (or uninteresting) to read. This
...
+rednodejava -> rednode [label = "generates"];
+}
+--------------------------------------------------------------------------------
+
+This is the end state. Now tools using Kythe graph can see that Color enum
+in JS is generated by Color enum in Java and perform proper action (for example
+IDE upon clicking on `Color` in JS file will go to the definition of `Color`
+enum in java file.
+
+== Protocol Buffers with *generate* edges
```
zrlk:
```suggestion
== Protocol Buffers with *generates* edges
```

kythe/docs/schema/indexing-generated-code.txt
nbeloglazov: Thanks, it was left from original version where I had RED and GREEN colors but then decided that extra element adds too much noise and having only `Color` and `RED` entities is enough. Fixed.

shahms:
The PR and your patience are much appreciated!

nbeloglazov:
Woohoo. Thank you all for the comments and feedback! Yes it took a while,
20% and github workflow is not friendly for quick review turnaround.

PR #3675, opened 2019-04-06
feat(tools): kzip merge can optional read input files from file

jaysachs:
This is more efficient than extending merge to also merge in the contents of the existing output file, which would be required for kzip to work in general with xargs (which currently is very dependent on ensuring both -n and -s are set correctly to ensure exactly one invocation of kzip).

More explicitly, we currently do

`find bazel-out/stuff -name '*.kzip' | xargs -r kzip merge --output-file output.kzip`

but that fails when the cardinality (in input file count or input file path lengths) is too large. We could choose a `-n` and `-s` argument to `xargs` that covered "most".  A more general change would be to have `kzip merge` merge in the contents of the existing output file, if any. That would be a more intrusive change. This PR simply gives kzip merge the same functionality as xargs if no input files are specified.

### DIFF
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -59,6 +59,7 @@ fi
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
 find bazel-out/*/extra_actions/external/kythe_release -name '*.kzip' | \
-  /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
+  /kythe/tools/kzip merge --inputs_from_stdin \
+                          --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
 /kythe/fix_permissions.sh "$KYTHE_OUTPUT_DIRECTORY"
 test -f "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -59,7 +59,7 @@ fi
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
 find bazel-out/*/extra_actions/external/kythe_release -name '*.kzip' | \
-  /kythe/tools/kzip merge --inputs_from_stdin \
+  /kythe/tools/kzip merge --inputs_list_file - \
                           --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
 /kythe/fix_permissions.sh "$KYTHE_OUTPUT_DIRECTORY"
 test -f "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -59,6 +59,6 @@ fi
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
 find bazel-out/*/extra_actions/external/kythe_release -name '*.kzip' | \
-  xargs -r /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
+  /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
 /kythe/fix_permissions.sh "$KYTHE_OUTPUT_DIRECTORY"
 test -f "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
--- a/kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
+++ b/kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
@@ -18,6 +18,7 @@
 package mergecmd
 
 import (
+	"bufio"
 	"context"
 	"flag"
 	"fmt"
@@ -57,7 +58,14 @@ func (c *mergeCommand) Execute(ctx context.Context, fs *flag.FlagSet, _ ...inter
 	if c.output == "" {
 		return c.Fail("required --output path missing")
 	}
-	if err := mergeArchives(ctx, c.output, fs.Args()); err != nil {
+	archives := fs.Args()
+	if len(archives) == 0 {
+		scanner := bufio.NewScanner(os.Stdin)
+		for scanner.Scan() {
+			archives = append(archives, scanner.Text())
+		}
+	}
+	if err := mergeArchives(ctx, c.output, archives); err != nil {
 		return c.Fail("Error merging archives: ", err)
 	}
 	return subcommands.ExitSuccess
--- a/kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
+++ b/kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
@@ -37,7 +37,8 @@ import (
 type mergeCommand struct {
 	cmdutil.Info
 
-	output string
+	output    string
+	xargsMode bool
 }
 
 // New creates a new subcommand for merging kzip files.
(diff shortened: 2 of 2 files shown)


kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -57,7 +58,14 @@ func (c *mergeCommand) Execute(ctx context.Context, fs *flag.FlagSet, _ ...inter
 	if c.output == "" {
 		return c.Fail("required --output path missing")
 	}
-	if err := mergeArchives(ctx, c.output, fs.Args()); err != nil {
+	archives := fs.Args()
+	if len(archives) == 0 {
```
schroederc: Check that both STDIN and non-flag args are not given.

jaysachs: i.e. you'd like a check like in all caps below

   if len(archives == 0) { // read from stdin }
   else { // read from args
      // *VERIFY STDIN IS EMPTY*
   }

?

schroederc: 👍

jaysachs: That makes it pretty annoying to run by hand from the shell (having to redirect stdin from devnull, or hitting ^D).

schroederc: You could make reading from stdin require a flag and then check that both that flag and `flag.NArg() != 0` are not true.  I'd just like to make it harder to accidentally lose inputs without any user notice.

The flag could be more generic as well.  It could point to a file to read lines from.  If you want stdin, pass `/dev/stdin`.

jaysachs: Checked for terminal first.

jaysachs: Ok, added a "xargs_mode" flag.

schroederc: I think I'd rather have a `--read_from_file` flag.  Special-casing `xargs` behavior seems weird.

jaysachs: I think it's weird to have to specify /dev/stdin. If you want input from a file, use xargs_mode and redirect. That seems way more natural.

jaysachs: If you just want the flag named "read_inputs_from_stdin" that's fine (just wordy)

kythe/extractors/bazel/extract.sh
```diff
@@ -59,6 +59,6 @@ fi
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
 find bazel-out/*/extra_actions/external/kythe_release -name '*.kzip' | \
-  xargs -r /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
+  /kythe/tools/kzip merge --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
```
schroederc: Add `--inputs_from_stdin`

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
creachadair: Is there some reason we can't _use_ `xargs` instead of adding a flag to this tool?

creachadair: To be more specific: Can we remove the restriction of one file per invocation, without getting ourselves into stdin-handling?

https://godoc.org/bitbucket.org/creachadair/fileinput might also help here, maybe.

jaysachs: It's not possible to make xargs work in general. For any specific application, you need to specify an appropriate value for `-n` (max args per invocation of cmd) and `-s` (max size of cmdline per invocation). There's no (reasonable) values that will work in all circumstances.

If `kzip` actually read in the existing output file and merged into that, then xargs will work. I didn't want to go through that approach because, well, it was more work to write to a tmp output file, etc etc. It would also be poorly performing (n^2) as you re-merge the same data  for each invocation of kzip.

creachadair: I'm confused: It is perfectly possible to use `xargs` without specifying those things, provided the command-line works with its free arguments correctly. And if it doesn't—e.g., if we need a way to partition one batch of arguments from another—we shouldn't be using free arguments for that purpose IMO.

What am I missing?

creachadair: Perhaps more importantly: If there is a performance problem being addressed, what is the wider problem we are trying to solve here, and is there an issue/design doc around that talks about why we're doing this to solve it?

jaysachs: The impetus for this came up because in [extract.sh](https://github.com/kythe/kythe/blob/7a1d39c7320fd5132e91fecb8679de1e36c823d6/kythe/extractors/bazel/extract.sh#L61) we essentially do `find ... | xargs kzip`. That fails because xargs picks defaults for `-n` and `-s`, and that ends up essentially doing
```
kzip --output o.kzip f1 f2 f3 ...
kzip --output o.kzip f100 f101 f202 ...
kzip --output o.kizp f200 f201 f202 ...
```
... which overwrites the output each invocation. Like I said, for any particular set of files I can choose `-n` and `-s` such that there is exactly one invocation of kzip, but I can't write a general purpose script to do it.

creachadair: What about allowing the `-output` flag to take a template with an (optional) randomizing string, like the way `mktemp` works?

creachadair: Or, another alternative: An `-output_dir` that, if set instead of `-output`, generates nonce output file names in that directory?

Both of these seem like they might be simpler to implement and less brittle to use.

jaysachs: The other option, which I'm willing to consider, is to have kzip actually merge in the output file as well, if it exists. If we want to say "let's not worry about the performance of re-reading the same CUs and files multiple times", maybe OK, but I'd like to handle large-cardinality cases well.

jaysachs: We want just one output file at the end of this; I'm not seeing how having an output_dir or a an output "template" helps here.  We'll then have to do 'multiphase' merging. Why can't the tool just support it?

jaysachs: I don't think "simpler to implement" applies here. This is pretty simple. "Brittle", sure. I'll put in validation that either files to process come from stdin, or they appear as arguments.

creachadair: It appears you are trying to shoehorn a very specific use-case into a general-purpose tool, just to suit an existing script. I think a better solution would be to write a program that does what the script does, instead of polluting the API of a tool that doesn't suit that one use case.

jaysachs: I'm not sure I agree. The use case is not all that specific: I have a lot (say 100k) of kzips, and I want to merge them. How do I do that? I'm proposing "let `kzip merge`" do that.  If `kzip merge` can't do that, then it's not very good merge tool; it's just a toy. I don't the high cardinality case is all that specific.

What do you think about the other proposal: let `kzip merge` preserve the existing output file contents, and essentially "append" to it (possibly controlled by some flag)?

creachadair: It's specific in that there is no good reason to combine hundreds of thousands of kzip files into one. The fact that we are doing so in this case appears to be an artifact of a particular way this script was set up. So my argument is that as a merge tool it already works fine: If you really need to be able to merge "into" an existing file, you could simply add that feature, though as you pointed out the ZIP format isn't well-situated for that case and you might want to consider a different format tif there's truly a need for huge single-file combinations.

I am skeptical that this is a real problem in need of solving, though.

jaysachs:
Obsoleted by #3679

PR #3677, opened 2019-04-08
feat(proto_indexer): emit deprecation tags for proto fields

justbuchanan:


kythe/cxx/indexer/proto/proto_graph_builder.cc
```diff
@@ -211,4 +211,10 @@ void ProtoGraphBuilder::AddCodeFact(const VName& element,
   recorder_->AddMarkedSource(VNameRef(element), code);
 }
 
+void ProtoGraphBuilder::SetDeprecated(const VName& node_name) {
+  VLOG(1) << "Adding deprecation tag for " << StringifyNode(node_name);
+  recorder_->AddProperty(VNameRef(node_name), kythe::PropertyID::kTagDeprecated,
+                         "This field is deprecated.");
```
jaysachs: I'd leave the fact value the empty string. If there was an explicit deprecation message to capture, that'd be useful but I don't think there is for proto.

kythe/cxx/indexer/proto/testdata/basic/deprecated.proto
```diff
@@ -0,0 +1,13 @@
...
+option java_package = "io.kythe";
+
+message M {
+  //- @field1 defines/binding S
+  //- S.tag/deprecated "This field is deprecated."
+  optional string field1 = 1 [deprecated=true];
+
+  // Fields that are annotated with [deprecated=true] should be tagged as
+  // deprecated by the proto indexer.
+}
```
jaysachs: Mind adding a second non-deprecated field and asserting it's not deprecated?

PR #3679, opened 2019-04-09
feat(tools): kzip can merge into the existing output file

jaysachs:
`kzip merge` will preserve the existing compilation units and files in the output file when merging.

This is an alternative to #3675


kythe/go/platform/vfs/vfs.go
```diff
@@ -106,9 +106,9 @@ func Create(ctx context.Context, path string) (io.WriteCloser, error) {
 	return Default.Create(ctx, path)
 }
 
-// CreateTemp creates a new temp file, using the Default VFS.
-func CreateTemp(ctx context.Context, dir string, pattern string) (string, error) {
-	return Default.CreateTemp(ctx, dir, pattern)
+// TempFile creates a new temp file, using the Default VFS.
+func TempFile(ctx context.Context, dir, pattern string) (*os.File, error) {
```
creachadair: Apologies, I wasn't clear in my earlier comments.

I'm not sure it's practical to return an `*os.File` here. It works fine with the implementation that falls through to the `os` package, but these shims are here because we had to support the non-standard Google filesystem types, which do not have file descriptors. 

If we do have descriptors we can get a real `*os.File`. Otherwise you might consider defining a `File` type with analogous properties, e.g.,

```go
type File interface {
   io.ReadWriteCloser
   io.Seeker
   Name() string
   // …whatever other capabilities it needs
}
```

using a subset of the capabilities of `*os.File` as a guideline. The above appears to be all we need in this case—possibly also `io.Seeker` or `io.ReaderAt` depending on whether we pass this directly to a zip writer.

kythe/go/platform/vfs/vfs.go
```diff
@@ -63,6 +63,9 @@ type Writer interface {
 	// Create creates a new file for writing, as os.Create.
 	Create(ctx context.Context, path string) (io.WriteCloser, error)
 
+	// CreateTemp creates a new temp file returning the path to it.
+	CreateTemp(ctx context.Context, dir string, pattern string) (string, error)
```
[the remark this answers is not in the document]
jaysachs: There isn't a correctness TOCTOU issue; the file gets created, so we "own" it. Regardless, I made the changes. I think it makes vfs a little weird (the first *os.File occurrence instead of some interface). I could introduce a new interface that extends WriteCloser and gives access to the name (required to use the other vfs operations) but I think that may be overkill.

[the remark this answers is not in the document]
jaysachs: Done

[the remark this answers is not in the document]
creachadair: Adding a new type was what I had in mind (see my other comment in reply to your recent commit). It might indeed be overkill: But I think at very least if we want to return a path the documentation needs to explicitly spell out the expectations, e.g.

"An empty writable file is created and its path is returned; the caller is responsible for cleaning it up when it is no longer in use (either by removing or renaming it)." 

Also if the caller wants to rename, they'll need to make sure the eventual target directory is on the same filesystem as `dir`, otherwise the rename may not work. The implementation of `TempFile` can't enforce that, but someone using the file will need to know it.

kythe/go/platform/vfs/vfs.go
jaysachs: I'd already started on a more minimal NamedWriteCloser.

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -50,16 +52,45 @@ func New() subcommands.Command {
...
+			}
+		}
+	}
+	if err := mergeArchives(ctx, tmpOut, archives); err != nil {
 		return c.Fail("Error merging archives: ", err)
 	}
+	if err := vfs.Rename(ctx, tmpOut, c.output); err != nil {
+		return c.Fail("Error renaming tmp to output:", err)
+	}
+	tmpOut = ""
```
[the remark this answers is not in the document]
jaysachs: If you mean the `tmpOut = ""`, that was leftover; removed.

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -50,16 +52,45 @@ func New() subcommands.Command {
 // for merging kzip files.
 func (c *mergeCommand) SetFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.output, "output", "", "Path to output kzip file")
+	fs.BoolVar(&c.mergeExisting, "merge_existing", false, "Whether to merge the contents of the existing output file, if it exists")
```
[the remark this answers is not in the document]
jaysachs: I'll leave that for a separate PR. I agree; changing the default behavior might be unexpected, though.

kythe/go/platform/vfs/vfs.go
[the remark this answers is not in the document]
jaysachs: Added NamedWriteCloser

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -50,24 +52,50 @@ func New() subcommands.Command {
...
+	}()
+	if err != nil {
+		return c.Fail("Error creating temp output: ", err)
+	}
+	archives := fs.Args()
+	if c.append {
+		orig, err := vfs.Open(ctx, c.output)
+		if err == nil {
+			archives = append([]string{c.output}, archives...)
+			err := orig.Close()
```
schroederc:
```
if err := orig.Close(); err != nil {
  return ...
}
```

kythe/go/platform/vfs/vfs.go
```diff
@@ -63,6 +70,11 @@ type Writer interface {
 	// Create creates a new file for writing, as os.Create.
 	Create(ctx context.Context, path string) (io.WriteCloser, error)
 
+	// TempFile creates a new temp file returning a NamedWriteCloser to it. The
+	// name of the file is constructed from dir pattern and per
+	// ioutil.TempFile.
+	TempFile(ctx context.Context, dir, pattern string) (NamedWriteCloser, error)
```
schroederc: Should specify what dir and pattern mean.

kythe/go/platform/vfs/vfs.go
```diff
@@ -41,6 +41,8 @@ type Interface interface {
 	Writer
 }
 
+// NamedWriteCloser composes io.WriteCloser and access to its "name". For
```
creachadair: What do you think about calling this more simply `TempFile`?

```go
func (localFS) TempFile(ctx context.Context, dir, pattern string) (TempFile, error) { …
```

Not that this is inaccurate, but it's lower level when what we care about is that it's the contact surface for a temporary file. YMMV.

jaysachs: I like it, except it conflicts with your preferred function name. You OK with "CreateTempFile" ?

creachadair: It doesn't conflict—the function name is inside the type.

kythe/go/platform/vfs/vfs.go
jaysachs: Done

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
jaysachs: Done

kythe/go/platform/vfs/vfs.go
creachadair: Oh, I see, there is a top-level helper function.

Yeah, I mean that works too.

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -50,24 +52,49 @@ func New() subcommands.Command {
...
 	}
 	return subcommands.ExitSuccess
 }
 
-func mergeArchives(ctx context.Context, output string, archives []string) error {
-	out, err := vfs.Create(ctx, output)
-	if err != nil {
-		return fmt.Errorf("error creating output: %v", err)
-	}
+func mergeArchives(ctx context.Context, out vfs.TempFile, archives []string) error {
```
schroederc: `out` should just be typed `io.WriteCloser`.

schroederc: Or pass in the kzip Writer

jaysachs: Prefer to pass in the WriteCloser.

done

PR #3692, opened 2019-04-10
feat(go_indexer): add deprecation facts for go

jaysachs:


kythe/go/indexer/emit.go
```diff
@@ -983,6 +984,29 @@ func (e *emitter) writeDoc(comments *ast.CommentGroup, target *spb.VName) {
 	e.writeFact(docNode, facts.NodeKind, nodes.Doc)
 	e.writeFact(docNode, facts.Text, strings.Join(lines, "\n"))
 	e.writeEdge(docNode, target, edges.Documents)
+	e.handleDeprecation(target, lines)
+}
+
+// handleDeprecation emits a deprecated fact for the specified target if the
+// comment lines indicate it is deprecated.
+func (e *emitter) handleDeprecation(target *spb.VName, lines []string) {
```
schroederc: Please add pointer to https://github.com/golang/go/wiki/Deprecated

kythe/go/indexer/emit.go
```diff
@@ -983,6 +984,29 @@ func (e *emitter) writeDoc(comments *ast.CommentGroup, target *spb.VName) {
 	e.writeFact(docNode, facts.NodeKind, nodes.Doc)
 	e.writeFact(docNode, facts.Text, strings.Join(lines, "\n"))
 	e.writeEdge(docNode, target, edges.Documents)
+	e.handleDeprecation(target, lines)
```
schroederc: Do we want the deprecation text to be escaped in the same way of the `doc` node text?  I don't think we need to render them with interpolation in the same way.

kythe/go/indexer/emit.go
```diff
@@ -983,6 +984,29 @@ func (e *emitter) writeDoc(comments *ast.CommentGroup, target *spb.VName) {
...
+}
+
+// handleDeprecation emits a deprecated fact for the specified target if the
+// comment lines indicate it is deprecated.
+func (e *emitter) handleDeprecation(target *spb.VName, lines []string) {
+	deprecated := false
+	depMsg := ""
```
schroederc: var depMsg string

kythe/go/indexer/emit.go
```diff
@@ -983,6 +984,29 @@ func (e *emitter) writeDoc(comments *ast.CommentGroup, target *spb.VName) {
 	e.writeFact(docNode, facts.NodeKind, nodes.Doc)
 	e.writeFact(docNode, facts.Text, strings.Join(lines, "\n"))
 	e.writeEdge(docNode, target, edges.Documents)
+	e.handleDeprecation(target, lines)
+}
+
+// handleDeprecation emits a deprecated fact for the specified target if the
+// comment lines indicate it is deprecated.
+func (e *emitter) handleDeprecation(target *spb.VName, lines []string) {
+	deprecated := false
```
schroederc: var deprecated bool

kythe/go/indexer/emit.go
```diff
@@ -983,6 +984,29 @@ func (e *emitter) writeDoc(comments *ast.CommentGroup, target *spb.VName) {
...
+	depMsg := ""
+	for _, line := range lines {
+		if !deprecated {
+			if msg := strings.TrimPrefix(line, "Deprecated: "); msg != line {
```
schroederc: The space after the colon does not seem to be required.

kythe/go/util/schema/facts/facts.go
```diff
@@ -27,6 +27,7 @@ const (
 	Code         = prefix + "code"
 	Complete     = prefix + "complete"
 	ContextURL   = prefix + "context/url"
+	Deprecated   = prefix + "deprecated"
```
schroederc: This should be `"tag/deprecated"`: https://kythe.io/docs/schema/#tagdeprecated

kythe/go/indexer/emit.go
creachadair: nit: The rest of this package uses `writeXX` for functions that explicitly write facts/edges for a particular construct, or `emitXX` for functions that handle a general category of attributes (e.g., overrides) that may or may not be present. By that logic this is probably `emitDeprecation`, since it may or may not be present.

jaysachs: Thanks, I missed (and was hoping to find) that distinction.

kythe/go/indexer/emit.go
jaysachs: So my thought here is "maybe yes" ... However neither the [Java](https://github.com/cushon/kythe/blob/0ab3f6a0011e8b060247a84ef7c451bfdf6c2bd2/kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java#L158) nor [C++](https://github.com/cushon/kythe/blob/0ab3f6a0011e8b060247a84ef7c451bfdf6c2bd2/kythe/cxx/indexer/cxx/KytheGraphObserver.cc#L1041) do not. They should at least be consistent, so I'll do the Go without escaping.

kythe/go/indexer/emit.go
creachadair: Unfortunately it's not written down anywhere, and I may not have applied it entirely consistently either.

kythe/go/indexer/emit.go
```diff
@@ -983,6 +984,29 @@ func (e *emitter) writeDoc(comments *ast.CommentGroup, target *spb.VName) {
...
+				depMsg = msg
+				deprecated = true
+			}
+		} else if line == "" {
+			break
+		} else {
+			depMsg += " " + line
```
creachadair: I suggest: https://play.golang.org/p/HPtp-YVDKrF

Key points: Although we're not worried about memory use here, this simplifies both the concatenation issue and the correct handling of spaces. It also cleans up extra whitespace, and handles "morally empty" lines that have only whitespace on them.

jaysachs: I actually have just finished a similar change.

kythe/go/indexer/emit.go
jaysachs: Done.

PR #3707, opened 2019-04-30
feat(api): allow a semantic scope to be associated per decoration

schroederc:
The semantic scope for each reference corresponds to the target of an `anchor` node's `childof`edge.  Currently, these are only prescribed for `ref/call` anchors, but may be expanded to other `anchor` kinds in the future.  Experimental support for `childof` edges on other `anchor`s is guarded by the Java indexer's `--emit_anchor_scopes` flag.


schroederc:
> Can you add some more documentation on the expected semantics of this (and what it requires from indexers)? I assume something like:
> 
> ```
> type Foo struct {
>   Field int
> }
> ```
> 
> The semantic node for `Foo` would be the SemanticScope of the anchor `Field`?

I added some further documentation in the description.  Only `ref/call` anchors are currently prescribed.  No indexer changes are required for this until we expand that in the schema.

Issue #3698, opened 2019-04-17
Provide full definition spans in C++


zrlk:
C++ has `defines/binding` edges but no `defines` edges.

dwightnaylor:
It looks like the /kythe/edge/defines is missing entirely for the function when this happens. Presumably it is in the .h file where the function is declared.

dwightnaylor:
However, if we follow the /kythe/edge/completes edge, and get xrefs for that ticket, we can see that the definition for the function also has the wrong span, consisting of only the binding span, instead of the full defines span.

This doesn't seem to happen for other languages where something is defined and declared separately (I only checked the example of interfaces with java)

shahms:
C++ does not currently emit "defines" edges which are not "defines/binding".

robinp:
Apart from being a nice thing to do, is there a direct benefit? Like the serving table containing extra info, or the UI likely being to exploit in some shiny way?

One comes to mind is deriving outlines based on containment of binding anchors (so without using explicit childof relations)?

PR #3708, opened 2019-04-30
feat(cxx_indexer): provide defines spans for functions

zrlk:
Partially addresses issue #3698.

### DIFF
--- a/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
+++ b/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
@@ -541,13 +541,18 @@ clang::SourceRange IndexerASTVisitor::RangeForSingleToken(
 void IndexerASTVisitor::MaybeRecordDefinitionRange(
     const absl::optional<GraphObserver::Range>& R,
     const GraphObserver::NodeId& Id,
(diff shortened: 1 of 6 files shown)


kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -2932,9 +2938,22 @@ bool IndexerASTVisitor::VisitFunctionDecl(clang::FunctionDecl* Decl) {
     }
   }
 
-  MaybeRecordDefinitionRange(NameRangeInContext, OuterNode,
-                             BuildNodeIdForDefnOfDecl(Decl));
   bool IsFunctionDefinition = IsDefinition(Decl);
+  if (!IsFunctionDefinition || Decl->isImplicit() ||
+      Decl->getBody() == nullptr) {
+    MaybeRecordDefinitionRange(NameRangeInContext, OuterNode,
+                               BuildNodeIdForDefnOfDecl(Decl));
```
shahms: I think this would be simpler as two separate calls, where the full definition is only emitted when found, rather than repeating the same function twice with an optional parameter that isn't used anywhere else.

zrlk: Done

PR #3713, opened 2019-05-03
chore(build): add blacklist for known-bad releases (currently 0.25.0)

shahms:
Seeing as we have a specific reason we're avoiding 0.25, document that the reason is known and warn users against it for now.

### DIFF
--- a/version.bzl
+++ b/version.bzl
@@ -32,3 +32,13 @@ def check_version(min_required, max_supported):
     max = _parse_version(max_supported)
     if max < _bound_size(found_version, len(max)):
         fail("Your bazel is too new. Maximum supported version {} of bazel, found {}".format(max_supported, found))
+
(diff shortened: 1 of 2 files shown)


version.bzl
```diff
@@ -32,3 +32,13 @@ def check_version(min_required, max_supported):
     max = _parse_version(max_supported)
     if max < _bound_size(found_version, len(max)):
         fail("Your bazel is too new. Maximum supported version {} of bazel, found {}".format(max_supported, found))
+
+def blacklist_version(version, reason):
```
creachadair: In retrospect, it would have been nice to make this a list inside the `check_version` rule so that the error could include a suggestion of which versions are allowed. (It is almost certainly not worth re-doing though)

shahms: I considered doing it that way, but this seemed simpler.

PR #3716, opened 2019-05-03
fix(cxx_indexer): insist that compilation units have working directories

zrlk:
The C++ (+ friends) indexer should be hermetic modulo claiming when
reading from extracted compilation units. Currently, it will use the
indexing environment's working directory when one is not set in the
compilation unit (or if the value that was set is not an absolute path).
This PR turns this into an error condition.

### DIFF
--- a/kythe/cxx/indexer/cxx/frontend.cc
+++ b/kythe/cxx/indexer/cxx/frontend.cc
@@ -123,11 +123,10 @@ void MaybeNormalizeFileVNames(IndexerJob* job) {
 
 void UpdateJobWdirFromUnit(IndexerJob* job) {
   job->working_directory = job->unit.working_directory();
-  if (!llvm::sys::path::is_absolute(job->working_directory)) {
-    llvm::SmallString<1024> stored_wd;
-    CHECK(!llvm::sys::fs::make_absolute(stored_wd));
-    job->working_directory = stored_wd.str();
-  }
+  CHECK(!job->working_directory.empty())
+      << "Indexer jobs must have their absolute working directory set.";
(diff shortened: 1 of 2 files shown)


kythe/cxx/indexer/cxx/frontend.cc
```diff
@@ -123,11 +123,10 @@ void MaybeNormalizeFileVNames(IndexerJob* job) {
...
 void UpdateJobWdirFromUnit(IndexerJob* job) {
   job->working_directory = job->unit.working_directory();
-  if (!llvm::sys::path::is_absolute(job->working_directory)) {
-    llvm::SmallString<1024> stored_wd;
-    CHECK(!llvm::sys::fs::make_absolute(stored_wd));
-    job->working_directory = stored_wd.str();
-  }
+  CHECK(!job->working_directory.empty())
+      << "Indexer jobs must have their absolute working directory set.";
+  CHECK(llvm::sys::path::is_absolute(job->working_directory))
```
creachadair: Does this not imply the previous condition already?

zrlk: Yes, but I expect custom extractors to forget to populate the field entirely more often than to populate it incorrectly.

creachadair: Well, okay, but then should that case have a different diagnostic?

creachadair: The other problem with this solution is that the indexer will now fail on stored compilations that didn't set this value. Maybe instead we should assume "/" if it's not set, and only issue a diagnostic?

PR #3718, opened 2019-05-06
fix: update out of date script in verifier example

danielmoy-google:


kythe/docs/schema/writing-an-indexer.txt
```diff
@@ -444,7 +444,8 @@ set -o pipefail
 # Read JSON entries from standard in and pass them to the verifier.
 # The entrystream tool turns the JSON into length-delimited protocol buffers,
 # described at http://godoc.org/kythe.io/kythe/go/platform/delimited
-/opt/kythe/tools/entrystream --read_json | /opt/kythe/tools/verifier "$@"
+/opt/kythe/tools/entrystream --read_format=json | \
+/opt/kythe/tools/verifier --nofile_vnames --nocheck_for_singletons "$@"
```
creachadair: I don't think we really want to suggest disabling the singletons check as part of the default example, do we?

shahms: We should probably update the example so that it passes with this check enabled.  The shortest route is to append `?` to the node in question.

danielmoy-google: done

kythe/docs/schema/writing-an-indexer.txt
```diff
@@ -444,15 +444,16 @@ set -o pipefail
...
+/opt/kythe/tools/verifier --nofile_vnames "$@"
 ----
 
 We can write a rule file that checks whether we have any file nodes at all and
 call it `test.goals`:
 
 [source,c]
 ----
-//- FileNode.node/kind file
+//- FileNode?.node/kind file
```
creachadair: Wait—why wasn't this a singleton before? Is it because we have multiple examples concatenated in the one input?

shahms: It was a singleton before (hence the need to change the script), but the singletons check was added after the example.  The singleton check is disabled for inspected nodes (the `?`).

PR #3723, opened 2019-05-07
Create presubmit.yml for Bazel continuous integration

achew22:
This (+ a PR in https://github.com/bazelbuild/continuous-integration) will put
make Bazel changes that break Kythe visible to the Bazel team before a
release is cut.

This just builds for Ubuntu 18.04, but can be expanded to do windows,
osx, or ubuntu 16.04. For now I want to start small and see if we can get
the two to just talk to each other.


presubmit.yml
```diff
@@ -0,0 +1,8 @@
+---
+tasks:
+  ubuntu_1804:
+    platform: ubuntu1804
+    build_targets:
+    - "..."
```
shahms: Both of these should probably be "//kythe/..." rather than just "..."

shahms: Also, from the documentation shouldn't this be: `.bazelci/presubmit.yml`?

achew22: Done

achew22:
Friendly ping

PR #3728, opened 2019-05-09
build: update to Bazel 0.25.1

shahms:
Note: The test changes to account for a more zealous inclusion of JSR250 related classes during extraction will cause that same test to fail on import.

Additionally, buildbot is currently being updated so this may fail the automated checks until that is finished.

This updates the minimum and maximum supported version to 0.25.1 as maintaining compatibility with older versions is no longer feasible due to accumulated Bazel changes.


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -498,6 +498,9 @@ private void processRequiredInput(
     if ((isJarPath && jarPath.startsWith(jdkJar)) || path.startsWith(JDK_MODULE_PREFIX)) {
       return;
     }
+    if (path.contains("Generated.class")) {
```
schroederc: endsWith("javax/annotation/Generated.class")?

shahms: Oops, that was for debugging :-( I'll delete it.

PR #3741, opened 2019-05-15
feat(indexing): add binary for indexing multi-language kzips

justbuchanan:
A lot of our extraction setups end by merging all of the per-target kzips into one large one for the whole repository. These kzips are hard to process locally because language-specific indexers don't handle kzips that contain units of multiple languages. This binary is a wrapper around the set of language-specific indexers that can process a multi-language kzip.

TODO: better name. suggestions?
TODO: environment variable instead of `--release_dir` flag?
~~TODO: progress indicator? some of the kzips are enormous and it's nice to know something's happening~~
TODO: multithread - `Driver.Run()` is currently single-threaded


shahms:
Why not add a `split` subcommand to the `kzip` tool?

justbuchanan:
Yeah I can put this into a `kzip split` subcommand if that makes more sense.

I would still like to have a single command/script for indexing the merged kzips though. I could do this as a script that calls `kzip split`, then runs the indexers. Once the [analyzer/driver interface](https://github.com/kythe/kythe/pull/3339) is finalized though, we could write an efficient version that just reads the merged kzip and avoids the need to save a bunch of temporary files. Thoughts?

shahms:
We've also got interfaces defined in //kythe/go/platform/analysis/driver for this sort of thing which should likely be utilized.  An external `analyzer_driver` equivalent would be useful to have (and I thought one already existed).

creachadair:
The original aspiration was that the driver-to-indexer interface should be completely uniform: The driver would just run the binary and feed it work. Unfortunately the reality in practice is almost none of the indexers can be run without a cluster of tool-specific flags. (https://github.com/kythe/kythe/pull/3339 and its antecedent were partly an attempt to mitigate some of this, though even that isn't quite sufficient)

So I agree there are separate concerns here: Demultiplexing a kzip into per-language compilations is one part, and `kzip split` sounds like a good idea for that. But there is also a separate need (or at least desire) for a one-stop "run the indexer, no horsepucky" tool. I wrote a thing called `indexer_driver` internally that baked the various settings in—similar to what you proposed here, but using MPMs and Blaze outputs.

If we were starting over today, I would probably think in terms of treating the indexer as a container rather than a naked process: Separating the configuration concerns from the launch step would make a lot of this easier to reason about. We didn't think of that in 2014, though.

justbuchanan:
Thanks for the additional context, @creachadair. I do look forward to getting #3339 finalized and implemented - it will definitely make it easier to implement something like what I'm trying to do in this PR. I'd like it to be similar to the internal `indexer_driver`, but have it instantiate the language-specific indexers as-needed so that it can be run against a kzip with no extra setup.

For now, I'm rewriting this PR to make use of the //kythe/go/platform/analysis/driver package @shahms pointed out, which will make it easier to move towards the #3339 solution once it's figured out.

justbuchanan:
I refactored this to use the driver module and associated FileQueue.

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
+// This package provides an implementation of the CompilationAnalyzer interface
```
schroederc: This isn't a separate package from the binary, but it probably should be.  It could go in kythe/go/platform/analysis/local/local.go.

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	"kythe.io/kythe/go/platform/analysis"
+	"kythe.io/kythe/go/platform/delimited"
+	"kythe.io/kythe/go/platform/kzip"
+	apb "kythe.io/kythe/proto/analysis_go_proto"
```
schroederc: nit: blank line to separate proto import group

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+
+type localAnalyzer struct {
+	cmd     []string
+	fetcher analysis.Fetcher
+}
+
+// NewLocalAnalyzer creates an analyzer that will execute the given command when
+// Analyze() is called. The filepath to a kzip is appended to the command, so if
+// given ["my_indexer", "--index"], the final executed command will be
+// ["my_indexer", "--index", "some/file.kzip"].
```
schroederc: Add assumptions about output format.

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+func NewLocalAnalyzer(cmd []string, fetcher analysis.Fetcher) *localAnalyzer {
+	return &localAnalyzer{cmd: cmd, fetcher: fetcher}
+}
+
+func (a *localAnalyzer) Analyze(ctx context.Context, req *apb.AnalysisRequest, f analysis.OutputFunc) error {
```
schroederc: // Analyze implements the analysis.CompilationAnalyzer interface.

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+	tmpkzip, err := saveSingleUnitKzip(ctx, req.Compilation, a.fetcher)
+	if err != nil {
+		return fmt.Errorf("creating tmp kzip: %v", err)
+	}
+	defer os.Remove(tmpkzip)
+
+	// Run indexer binary
+	var errBuf strings.Builder
+	var outBuf bytes.Buffer
+	cmd := exec.Command(a.cmd[0], append(a.cmd[1:], tmpkzip)...)
```
schroederc: Check that `cmd` has the proper shape in `NewLocalAnalyzer` (adding a new error return).

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+	rd := delimited.NewReader(&outBuf)
+	for {
+		n, err := rd.Next()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return fmt.Errorf("parsing entry stream: %v", err)
+		}
+
+		f(ctx, &apb.AnalysisOutput{Value: n})
```
schroederc: Check error returned from `f`.

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	f(ctx, &apb.AnalysisOutput{FinalResult: &apb.AnalysisResult{Status: apb.AnalysisResult_COMPLETE}})
```
schroederc: Check error returned from `f`.

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return nil
+}
+
+// saveSingleUnitKzip creates a kzip file containing the given compilation unit
+// and any files it depends on. File contents are retrieved from the given
+// fetcher object.
+func saveSingleUnitKzip(ctx context.Context, cu *apb.CompilationUnit, fetcher analysis.Fetcher) (string, error) {
+	tmpfile, err := ioutil.TempFile("", fmt.Sprintf("*_%s.kzip", cu.VName.Language))
```
schroederc: Use `vfs.CreateTempFile`

kythe/go/platform/tools/indexer_wrapper/local.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	w, err := kzip.NewWriter(tmpfile)
+	if err != nil {
+		return "", err
+	}
+	defer w.Close()
```
schroederc: Check writer close error

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -1,137 +1,85 @@
-package main
-
-// This binary reads compilation units from the provided kzip(s) and indexes
-// them using language-specific indexers. Entry protos are written to stdout.
+// binary indexer_wrapper reads compilation units from the provided kzip(s) and
```
schroederc: `// Binary indexe_wrapper`

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -1,137 +1,85 @@
...
 	"os"
-	"os/exec"
 	"path/filepath"
 
-	"kythe.io/kythe/go/platform/kzip"
+	"kythe.io/kythe/go/platform/analysis"
+	"kythe.io/kythe/go/platform/analysis/driver"
+	"kythe.io/kythe/go/platform/analysis/local"
+	"kythe.io/kythe/go/platform/delimited"
+	apb "kythe.io/kythe/proto/analysis_go_proto"
```
schroederc: Separate proto import group

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,99 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+		return "", err
+	}
+
+	w, err := kzip.NewWriter(tmpfile)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if err := w.Close(); err != nil {
+			log.Printf("unable to close kzip writer: %v", err)
```
schroederc: This error should be returned.

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,99 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+// saveSingleUnitKzip creates a kzip file containing the given compilation unit
+// and any files it depends on. File contents are retrieved from the given
+// fetcher object.
+func saveSingleUnitKzip(ctx context.Context, cu *apb.CompilationUnit, fetcher analysis.Fetcher) (string, error) {
+	tmpfile, err := vfs.CreateTempFile(ctx, "", fmt.Sprintf("*_%s.kzip", cu.VName.Language))
+	if err != nil {
+		return "", err
+	}
+
+	w, err := kzip.NewWriter(tmpfile)
```
schroederc: NewWriteCloser to ensure the underlying file is also closed.

kythe/go/platform/analysis/local/local.go
justbuchanan: Good point. What's the best way to do this since there are multiple early returns?

schroederc: One way is to name the error return value:

`func saveSingleUnitKzip(ctx context.Context, cu *apb.CompilationUnit, fetcher analysis.Fetcher) (path string, err error) {`

And set it in the `defer`:

```
	defer func() {
		if closeErr := w.Close(); closeErr != nil {
                        if err != nil {
                            log.Printf("WARNING: unable to close kzip writer: %v", closeErr)
                        } else {
                            err = closeErr
                        }
		}
	}()
```

Or you can `w.Close()` on every early exit, ignoring errors, and only return the `Close` errors on the final `return`.

justbuchanan: :thumbsup:

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+	fetcher analysis.Fetcher
+}
+
+// NewLocalAnalyzer creates an analyzer that will execute the given command when
+// Analyze() is called. The filepath to a kzip is appended to the command, so if
+// given ["my_indexer", "--index"], the final executed command will be
+// ["my_indexer", "--index", "some/file.kzip"]. The indexer command should emit
+// length-delimited Entry protos to stdout as described in
+// //kythe/go/platform/delimited.
+func NewLocalAnalyzer(cmd []string, fetcher analysis.Fetcher) (*localAnalyzer, error) {
```
creachadair: A couple things:

1. This name stutters with the package name. Consider naming the type `Analyzer` and the constructor `NewAnalyzer`, so that at the usage site the reader sees `local.NewAnalyzer` and the type `*local.Analyzer`.

2. Although it is legal to do so, I would discourage you from having an exported function or method return an unexported type. Better, I think to export the type (though not its fields). 

Other comments: As an API design issue, I would probably not bother having this constructor return an error at all—and either document that `cmd` is required and panic, or define it so that a zero-valued Analyzer is ready for use (e.g., by having an empty command do nothing). Panicking should, of course, be used with care—but in this case a program that passes an empty command or a nil fetcher is already statically wrong—i.e., it's a programming error rather than a runtime property. A more expressive type system might prevent it from occurring ("`[n≥1]string` type anyone?), but in lieu of that a panic is OK.

That said: The main things I suggest fixing here are the names and the exported-ness. 🙂

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+	tmpkzip, err := saveSingleUnitKzip(ctx, req.Compilation, a.fetcher)
+	if err != nil {
+		return fmt.Errorf("creating tmp kzip: %v", err)
+	}
+	defer os.Remove(tmpkzip)
+
+	// Run indexer binary
+	var errBuf strings.Builder
+	var outBuf bytes.Buffer
+	cmd := exec.Command(a.cmd[0], append(a.cmd[1:], tmpkzip)...)
```
creachadair: It looks like you only really want stderr for logging purposes. In that case, I suggest writing

```go
out, err := exec.Command(…).Output()
if e, ok := err.(*exec.ExitError); ok {
   return fmt.Errorf("indexing kzip %s: %v (stderr=%v)", tmpkzip, err, string(e.Stderr))
} else if err != nil { 
  // …some other issue, e.g., the process didn't start…
```

The `Output` method captures stderr for you.

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+			break
+		} else if err != nil {
+			return fmt.Errorf("parsing entry stream: %v", err)
+		}
+
+		if err := f(ctx, &apb.AnalysisOutput{Value: n}); err != nil {
+			return fmt.Errorf("unable to write analysis output: %v", err)
+		}
+	}
+	if err := f(ctx, &apb.AnalysisOutput{FinalResult: &apb.AnalysisResult{Status: apb.AnalysisResult_COMPLETE}}); err != nil {
```
creachadair: Minor nit: I think this'd be more legible as

```go
	if err := f(ctx, &apb.AnalysisOutput{
		FinalResult: &apb.AnalysisResult{Status: apb.AnalysisResult_COMPLETE},
	}); err != nil {
		return fmt.Errorf("unable to write analysis output: %v", err)
	}
```

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+		return fmt.Errorf("unable to write analysis output: %v", err)
+	}
+
```
creachadair:
```suggestion
```

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+	w, err := kzip.NewWriteCloser(tmpfile)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if closeErr := w.Close(); closeErr != nil {
+			if err != nil {
+				log.Printf("WARNING: unable to close kzip writer: %v", closeErr)
```
creachadair: This seems unduly complicated—if the write failed it doesn't matter if the close failed, and if the close failed it doesn't matter if the write succeeded. I'd suggest more simply:

```
if cerr := w.Close(); err == nil {
  err = cerr
}
```

which ensures that (a) the writer always gets closed, (b) a write error is reported first if it exists, (c) a close error is reported if not.

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+		return fmt.Errorf("unable to write analysis output: %v", err)
+	}
+
+	return nil
+}
+
+// saveSingleUnitKzip creates a kzip file containing the given compilation unit
+// and any files it depends on. File contents are retrieved from the given
+// fetcher object.
+func saveSingleUnitKzip(ctx context.Context, cu *apb.CompilationUnit, fetcher analysis.Fetcher) (path string, err error) {
```
creachadair:
```suggestion
func saveSingleUnitKzip(ctx context.Context, cu *apb.CompilationUnit, fetcher analysis.Fetcher) (_ string, err error) {
```

Since you don't actually need `path` for anything in the function, but you do want `err` for the deferral.

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +145,103 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+		b, err := fetcher.Fetch(file.Info.Path, file.Info.Digest)
+		if err != nil {
+			return "", fmt.Errorf("fetching file %s: %v", file.Info.Path, err)
+		}
+		if _, err := w.AddFile(bytes.NewReader(b)); err != nil {
+			return "", fmt.Errorf("writing file to kzip %s: %v", file.Info.Path, err)
+		}
+	}
+
+	return tmpfile.Name(), err
```
creachadair: If you get here, everything succeeded: `return tmpfile.Name(), nil`.

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -0,0 +1,133 @@
...
+		return fmt.Errorf("unrecognized language: %q", lang)
+	}
+	log.Printf("analyzing compilation: {%+v}", req.Compilation.VName)
+
+	return analyzer.Analyze(ctx, req, f)
+}
+
+type driverContext struct{}
+
+func (c *driverContext) Setup(ctx context.Context, comp driver.Compilation) error {
```
creachadair:
```suggestion
func (driverContext) Setup(context.Context, driver.Compilation) error { return nil }
```

Since the struct has no fields to be modified, there is no need for a pointer. Since the implementation doesn't do anything, the parameters (including the receiver) don't need names. (The absence of names is actually useful documentation, since the reader in godoc will see the absence and know the implementation doesn't use them)

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+func (c *driverContext) Teardown(ctx context.Context, comp driver.Compilation) error {
+	return nil
+}
+
+// log any errors that happen during analysis.
+func (c *driverContext) AnalysisError(ctx context.Context, comp driver.Compilation, err error) error {
```
creachadair: Line 73, ibid.

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -0,0 +1,133 @@
...
+var (
+	releaseDir = flag.String("release_dir", "/opt/kythe", "Path to kythe release dir, which contains indexer binaries and more")
+)
+
+func init() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, `Usage: %s file.kzip...
+
+Indexes all compilation units from the provided kzip(s) by delegating to
+language-specific indexers. Entry protos are written to stdout.
```
creachadair:
```suggestion
language-specific indexers. Entry protos are written as a delimited stream to stdout.
```

kythe/go/platform/analysis/local/local.go
creachadair: Re: package names & stutter, cf. https://golang.org/doc/effective_go.html#package-names

creachadair:
Sorry, I keep forgetting that GoogleBot does not play gracefully with commits offered by non-Google contributors (even if we've signed the CLA).

kythe/go/platform/analysis/local/local.go
justbuchanan: :+1:

kythe/go/platform/analysis/local/local.go
justbuchanan: :+1:

justbuchanan:
@creachadair no worries - the cla bot is annoying, but fighting it is worth it to use the suggested changes feature.

Thanks for the feedback. I went ahead and made those changes - let me know if you see anything else. Any suggestions for a better name than 'indexer_wrapper'?

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +144,98 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if cerr := w.Close(); err == nil {
+			err = cerr
+		}
+	}()
+
+	if _, err = w.AddUnit(cu, nil); err != nil {
```
creachadair:
```suggestion
	if _, err := w.AddUnit(cu, nil); err != nil {
```

The way you have it is fine, and happens to work, but there's no need to reuse the binding here: Returning explicit values updates the named return bindings.

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -0,0 +1,122 @@
...
+func (driverContext) Teardown(ctx context.Context, comp driver.Compilation) error { return nil }
+
+// log any errors that happen during analysis.
+func (driverContext) AnalysisError(ctx context.Context, comp driver.Compilation, err error) error {
+	log.Printf("AnalysisError: %v", err)
+	return err
+}
+
+var (
+	releaseDir = flag.String("release_dir", "/opt/kythe", "Path to kythe release dir, which contains indexer binaries and more")
```
creachadair: I think the environment variable we use for this is `KYTHE_DIR`, it may be worthwhile to echo that as `--kythe_dir` here, though YMMV.

justbuchanan:
Sounds good, thanks. I guess I'll keep it as `indexer_wrapper` for now, but I'll think about it some more.

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +144,98 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+// Analyze implements the analysis.CompilationAnalyzer interface.
+func (a *Analyzer) Analyze(ctx context.Context, req *apb.AnalysisRequest, f analysis.OutputFunc) error {
+	tmpkzip, err := saveSingleUnitKzip(ctx, req.Compilation, a.fetcher)
+	if err != nil {
+		return fmt.Errorf("creating tmp kzip: %v", err)
+	}
+	defer os.Remove(tmpkzip)
+
+	// Run indexer binary
+	out, err := exec.Command(a.cmd[0], append(a.cmd[1:], tmpkzip)...).Output()
```
creachadair: Consider using `exec.CommandContext` here, and plumbing in `ctx`. That way if the caller cancels the context, the subcommand will get terminated and this will return without any extra effort on our part. 👍

kythe/go/platform/analysis/local/local.go
```diff
@@ -140,3 +144,98 @@ type kzipFetcher struct{ r *kzip.Reader }
...
+			break
+		} else if err != nil {
+			return fmt.Errorf("parsing entry stream: %v", err)
+		}
+
+		if err := f(ctx, &apb.AnalysisOutput{Value: n}); err != nil {
+			return fmt.Errorf("unable to write analysis output: %v", err)
+		}
+	}
+	if err := f(ctx, &apb.AnalysisOutput{
```
creachadair: Arguably this could as readily be:

```go
return f(ctx, &apb.AnalysisOutput{
   FinalResult: &apb.AnalysisResult{Status: apb.AnalysisResult_COMPLETE}.
})
```

The additional annotation probably doesn't add much value, since the caller already controls both the callback and the analyzer and therefore knows _why_ the error.

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -0,0 +1,122 @@
...
+type delegatingAnalyzer struct {
+	analyzers map[string]analysis.CompilationAnalyzer
+}
+
+// binPath returns the absolute path to an indexer given its name.
+func binPath(binname string) string {
+	return filepath.Join(*kytheDir, "indexers", binname)
+}
+
+func NewDelegatingAnalyzer(fetcher analysis.Fetcher) *delegatingAnalyzer {
```
creachadair: Unless you're thinking about pulling this out into a library, I recommend unexporting it. It really doesn't matter that much in package main, but I find it helpful as a reader (seeing "unexported" is good signal).

kythe/go/platform/tools/indexer_wrapper/main.go
```diff
@@ -0,0 +1,122 @@
...
+func (driverContext) Setup(context.Context, driver.Compilation) error { return nil }
+func (driverContext) Teardown(ctx context.Context, comp driver.Compilation) error { return nil }
+
+// log any errors that happen during analysis.
+func (driverContext) AnalysisError(ctx context.Context, comp driver.Compilation, err error) error {
+	log.Printf("AnalysisError: %v", err)
+	return err
+}
+
+var (
```
creachadair: [optional] When there's only one, it's fine to omit the parentheses if you like: `var kytheDir = flag.String(…)`.

kythe/go/platform/tools/indexer_wrapper/main.go
justbuchanan: :+1:

kythe/go/platform/analysis/local/local.go
justbuchanan: :+1:

kythe/go/platform/analysis/local/local.go
justbuchanan: Thanks, I wasn't aware of `CommandContext`, but that works great here.

shahms:
Justin, is this still relevant?

justbuchanan:
We can close this, although I might revisit eventually.

PR #3742, opened 2019-05-15
fix(cxx_indexer): ngrmadlize windows paths

zrlk:
The implementation of Clang's FileManager::FixupRelativePath
assumes that all paths are in the host-machine style, so it's
currently hopeless to try and thread path style around the indexer.
Instead, since the only other path style right now is Windows,
turn Windows-style paths (though probably not UNC ones) into
POSIX-style paths (/c:/foo).

This doesn't mangle the various include arguments yet, so if
a compilation specifies an absolute path to an include directory
it won't have a good time.

CUs with missing or bad working directories are now only mildly
scolded instead of being rejected outright.


kythe/cxx/indexer/cxx/IndexerFrontendAction.cc
```diff
@@ -143,6 +145,15 @@ std::string IndexCompilationUnit(
     const LibrarySupports* LibrarySupports,
     std::function<std::unique_ptr<IndexerWorklist>(IndexerASTVisitor*)>
         CreateWorklist) {
+  llvm::sys::path::Style Style;
+  if (!kythe::IndexVFS::DetectStyleFromAbsoluteWorkingDirectory(
+          Unit.working_directory(), &Style)) {
+    if (!Unit.working_directory().empty()) {
+      absl::FPrintF(stderr, "warning: could not detect path style for %s\n",
+                    Unit.working_directory().c_str());
```
shahms: .c_str() is unnecessary with absl::FPrintF

kythe/cxx/indexer/cxx/IndexerFrontendAction.cc
```diff
@@ -143,6 +145,15 @@ std::string IndexCompilationUnit(
     const LibrarySupports* LibrarySupports,
     std::function<std::unique_ptr<IndexerWorklist>(IndexerASTVisitor*)>
         CreateWorklist) {
+  llvm::sys::path::Style Style;
+  if (!kythe::IndexVFS::DetectStyleFromAbsoluteWorkingDirectory(
+          Unit.working_directory(), &Style)) {
```
shahms: return `absl::optional<Style>` ? This then becomes:

llvm::sys::path::Style Style = Detect(...).value_or(Style::posix);

(You'd lose the warning on detection failure, but could move that into detect).

kythe/cxx/indexer/cxx/KytheVFS.h
```diff
@@ -32,9 +33,17 @@ class IndexVFS : public llvm::vfs::FileSystem {
...
   /// \param working_directory The absolute path to the working directory.
   /// \param virtual_files Files to map.
+  /// \param virtual_dirs Directories to map.
+  /// \param style Style used to parse incoming paths. Paths are normalized
+  /// to POSIX-style.
   IndexVFS(const std::string& working_directory,
            const std::vector<proto::FileData>& virtual_files,
-           const std::vector<llvm::StringRef>& virtual_dirs);
+           const std::vector<llvm::StringRef>& virtual_dirs,
+           const llvm::sys::path::Style style);
```
shahms: top-level `const` is meaningless clutter on value parameters in function declarations, please omit.

creachadair:
Please fix the description before merging, too 😄

shahms:
Also, it would be great to have some unittests for these things...

PR #3750, opened 2019-05-17
feat(post_processing): allow nodes/anchors to be ranked

schroederc:


tools/arc/linter.sh
```diff
@@ -60,7 +60,7 @@ case $file in
     fi ;;
   *.h|*.cc|*.c|*.proto|*.js)
     if command -v clang-format &>/dev/null; then
-      if diff -q <(clang-format --style=file "$file") "$file"; then
+      if ! diff -q <(clang-format --style=file "$file") "$file"; then
```
salguarnieri: Was this intentional?

schroederc: Yes.

kythe/proto/serving.proto
```diff
@@ -30,6 +30,8 @@ message Node {
 
   // This node's definition location (anchor).
   ExpandedAnchor definition_location = 3;
+
+  double rank = 4;
```
creachadair: Can this (in a future change) get some more documentation? E.g., what are the units here? Is there an expected range? Is it a magnitude, a percentage, other? Is higher better? What do fractional values mean?

(I realize these things may still be up for discussion, but I haven't seen any of it so I'm operating blind here)

PR #3760, opened 2019-05-22
chore(build): allow Bazel CI to opt out of version check

### LINKED ISSUE #3757: Kythe doesn't build with development build of Bazel
Hello dear Kythe,

I see Kythe implements a Bazel version check, but this doesn't work with the development build of Bazel which we use on our downstream pipeline (https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/989#a722f124-2b90-4559-a939-7c6dd6d921dd). The build is green with the released version of Bazel, it's the only empty version string of development Bazel build that causes trouble.

Thanks!

shahms:
Fixes #3757 by replacing the import of `version.bzl` with an equivalent implementation with no-op stubs.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,8 +1,6 @@
 ---
 tasks:
   ubuntu_1804:
-    shell_commands:
-      - sed -i.bak -e 's/version.bzl/dummy_version.bzl/' WORKSPACE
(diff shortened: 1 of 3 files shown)


schroederc:
Could we just always allow dev versions and add a warning for its use?

shahms:
That seems better, thanks.

PR #3762, opened 2019-05-22
java_verifier_tests: declare deps for annotations

rohanjoyce:
Prior to this PR, importing kythe using the [intellij bazel plugin](https://github.com/bazelbuild/intellij) fails at sync time. The `java_extract_kzip` rule includes a `java_common.compile` action that will fail if the source requires any (non-jvt) deps. The failing targets used annotations that were not declared as deps, but the `java_common.compile` actions were getting pruned as the outputs were not used. The IntelliJ plugin tries to build them because the rule return a `JavaInfo` provider.

This PR renames the existing `deps` attribute to `java_verifier_test` to `verifier_deps` (as it is specifically only meant to be passed `java_verifier_test` targets) and introduces a new `deps` attribute that just gets passed through to the `java_common.compile` action.

### DIFF
--- a/tools/build_rules/verifier_test/java_verifier_test.bzl
+++ b/tools/build_rules/verifier_test/java_verifier_test.bzl
@@ -154,6 +154,7 @@ def java_verifier_test(
         name,
         srcs,
         meta = [],
+        verifier_deps = [],
         deps = [],
         size = "small",
         tags = [],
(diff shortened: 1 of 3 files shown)


tools/build_rules/verifier_test/java_verifier_test.bzl
```diff
@@ -154,6 +154,7 @@ def java_verifier_test(
         name,
         srcs,
         meta = [],
+        jvt_deps = [],
```
schroederc: Rename without the initialism.  Maybe `verifier_deps`?

schroederc:
Truly unfortunate but it's a consequence of using a macro rather than defining a proper Starlark rule.

tools/build_rules/verifier_test/java_verifier_test.bzl
rohanjoyce: Done.

PR #3769, opened 2019-05-28
Migrate Kythe for Bazel 0.27

hlopko:
This PR migrates Kythe for
https://github.com/bazelbuild/bazel/issues/7260, which we plan to flip in Bazel 0.27.


hlopko:
CC @shahms. Thanks!

PR #3775, opened 2019-05-29
feat(typescript_indexer): add `defines` anchors for functions

dwightnaylor:


kythe/typescript/testdata/function.ts
```diff
@@ -8,13 +8,15 @@ interface Num {
   num: number;
 }
 
+//- TestDef defines F
```
schroederc: Add `//- TestDef.node/kind anchor`

dwightnaylor: Done (I think)

PR #3776, opened 2019-05-29
fix(post_processing): split semantic callers within xrefs group

### LINKED ISSUE #3768: Call contexts get mashed together with experimental_beam_pipeline preprocessor.
As a followup to https://groups.google.com/forum/#!topic/kythe/hbo_hFkBuXE, I now see callsites being returned, but with the following buggy behavior:

There's only one `callers` entry is returned in `CrossReferenceSet`, with it's `anchor` set to an arbitrary (first?) call context, but its `sites` containing all call usages from other contexts as well (even ones in different files).

schroederc:
Fixes #3768


robinp:
This works, thank you!

PR #3777, opened 2019-05-29
chore(cxx_indexer): remove redundant working_directory field

shahms:


shahms:
(This ensure that the working directory in the CompilationUnit is updated and used exclusively and will eliminate the warning message from `cc_indexer_test` of a single source file, rather than bundled compilation).

PR #3778, opened 2019-05-31
fix(release): include jsr250-api-1.0.jar in release archive

justbuchanan:
the release.BUILD file references it and fails if it isn't present.


justbuchanan:
Any idea why the go build is failing? It's trying to `go get github.com/pierrec/lz4/v3`, which doesn't seem to exist.

Issue #3779, opened 2019-06-02
Add a VName specification for TypeScript Indexer


ayazhafiz:
This is a feature proposal for a well-defined VName schema used by the TypeScript indexer.

## Motivation

A project with references across multiple languages can be difficult to index, as generally each language must be indexed individually. It would be nice to have a well-defined VName schema for TypeScript declarations so that other languages that use TypeScript code do not have to re-index a TypeScript project to extract the declarations they need, instead building the VName of the declaration using the formalized specification.

For example, I am developing an Angular indexer, the templating language of which uses TypeScript identifiers declared in other source files.

### Benefits

Makes it easier to write an indexer for a superset/integrator of TypeScript. Reduces duplication of entries generated by indexers.

### Drawbacks

May make future feature development of the the TypeScript indexer more strict. Though I'm not sure this is a drawback.

ayazhafiz:
@evmar

shahms:
Why can you not use `generates` or `imputes` edges like the other languages?  Having a fixed VName format is a potentially huge impediment to future development in that language, so definitely counts as a "drawback".

Another possible approach is that taken by Java/JVM where the lower level has well-defined semantics and a correspondingly well-defined VName schema but the languages built on top of the JVM do not.

ayazhafiz:
`generates` would be a better solution, you're right. Even for this though, it would be nice to have a spec that guarantees the VName of a semantic node. For instance an Angular component has a TypeScript class and HTML template, where the desired behavior is something like

```typescript
class MyComponent {
  someProperty: number;
}
```

```html
<div>{{someProperty}}</div>
```

The desired behavior is something like

```html
@someProperty ref VName("MyComponent.someProperty", _, _, _, _)
<div>{{someProperty}}</div>
```

without the template indexer having to look at the TypeScript source file, making a "best-guess" reference.

creachadair:
Having conventions for the signature is probably fine. It's tricky to constrain the other fields, though, since they usually depend on the build environment. That said: For a source-only language like TS it is probably _less_ problematic, since the usage site has to have the source anyway. The main thing to watch out for in the spec are features the call site can't know without seeing the rest of the build environment (e.g., indirect dependencies).

ayazhafiz:
[...]

PR #3780, opened 2019-06-02
feature: Add a VName specification for TypeScript Indexer

ayazhafiz:
#3779

Discussion wanted.

In progress:
- [x] VName specification
  - [x] Signature format
    - [x] declarations
  - [x] Corpus format
  - [x] Root format
  - [x] Path format
  - [x] Language format
- [x] Specification tests
- [x] Extension of TypeScript indexer to support specification

TODO:
- [ ] Move specification to another file? Discussion wanted.
- [x] Import/export/better module specification

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -36,3 +36,4 @@ Daniel Moy <danielmoy@google.com>
 Salvatore Guarnieri <salguarnieri@google.com>
 Jay Sachs <jsachs@google.com>
 Alexander Bezzubov <alex@sourced.tech>
+Ayaz Hafiz <ayaz.hafiz.1@gmail.com>
--- a/kythe/typescript/README.md
+++ b/kythe/typescript/README.md
@@ -79,6 +79,7 @@ describing the module.
 
 -   `NamespaceImport`
 -   `ClassDeclaration`
+-   `Constructor`
 -   `PropertyDeclaration`
 -   `MethodDeclaration`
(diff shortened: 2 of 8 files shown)


ayazhafiz:
@evmar

ayazhafiz:
Adding support for references to getters and setters requires context of their usage, because an identifier with a getter and setter has two explicit declarations. A decent approach seems to be to reference all property accesses to their getters, and all assignments to their setters.

```typescript
class A {
  get foo() { return this._foo; }
  set foo(newFoo) { this._foo = newFoo; }
  
  method() {
    let f = this.foo; // ref A.foo#getter
    this.foo = f; // ref A.foo#setter
  }
}
```

ayazhafiz:
> ```ts
> class A {
>   get foo() { return this._foo; }
>   set foo(newFoo) { this._foo = newFoo; }
>   
>   method() {
>     let f = this.foo; // ref A.foo#getter
>     this.foo = f; // ref A.foo#setter
>   }
> }
> ```

The format now is refer to getters in the same manner other class members are (guaranteeing that there is a class member of form `Class.foo` for a getter `foo`) and setters via `Class.foo:setter`. Switched to a colon as to not interfere with the pound sign used for type namespace signatures (such as `Interface#type`).

evmar:
This looks like it's on the right track, but you should split the getter/setter impl from the declaration spec stuff, because they are independent from each other and each will probably need a bit of iteration.

PR #3783, opened 2019-06-04
feat(typescript_indexer): define VName specification

ayazhafiz:
Parent PR for #3779

Discussion wanted.

In progress:
- [x] VName specification
  - [x] Signature format
    - [x] declarations
  - [x] Corpus format
  - [x] Root format
  - [x] Path format
  - [x] Language format
- [x] Specification tests

Changes to the indexer that implement this spec will be applied in separate PRs referencing this one.

Should the specification be moved to another file, maybe `SPEC.md`?


ayazhafiz:
@evmar the large PR got pulled apart as you recommended; this one is just for the spec. The changes to the indexer reference this PR.

kythe/typescript/README.md
```diff
@@ -1,5 +1,241 @@
 # Kythe indexer for TypeScript
 
+## VName Specification
+
+This specification defines how the indexer expresses VNames for TypeScript
+declarations. You may find this useful if you are developing an application that
+relies on TypeScript code you don't want to re-index.
+
+[spec tests](./testdata/declaration_spec.ts)
```
evmar: I think this doesn't render in the output (?)

kythe/typescript/README.md
```diff
@@ -1,5 +1,241 @@
...
+
+```typescript
+class A {
+  public foo: string;
+}
+
+type B = A;
+```
+
+`foo` has the signature `A.foo` and `B` has the signature `B#type`.
```
evmar: Should it be some A#instance.foo, given that A can have static members too?
Maybe it's fine to just document the current behavior.

kythe/typescript/README.md
ayazhafiz: It does on GitHub

kythe/typescript/README.md
ayazhafiz: Yeah, this is actually something we need to resolve. Let's point static members to `A#type.foo` and instance members to `A.foo`, as you suggested.

ayazhafiz: Resolved in 74e1e4a2771a7c58ddbd56a00fcf0505b6c082a1

ayazhafiz:
cc @evmar this should be ready, can you review for anything missing at this time?

kythe/typescript/README.md
```diff
@@ -1,5 +1,269 @@
 # Kythe indexer for TypeScript
 
+## VName Specification
```
evmar: The other stuff in the README is like "how to run the tests" which is what I think is a good opener.  I think this belongs below the other stuff in the README, or maybe in a different file.

kythe/typescript/README.md
```diff
@@ -1,5 +1,269 @@
...
+    -   e.g. `get#$NAME` is really `get#foo` if `$NAME = foo`.
+
+The signature of a TypeScript declaration is defined by the following schema:
+
+```regex
+$PART[.$PART]*[#type]?
+```
+
+where `$PART` is a component of the enclosing declaration scope and `#type` is
+appended to the signature of type declarations. `SyntaxKind`s that are type
```
evmar: While it's true they're attached to type declarations, I think the right way to talk about it is to say "the signature of types".  (That is, remove the 'declarations' part.)  Because sometimes for a given symbol (like a class name) we talk about both its type and its value for the same declaration.

kythe/typescript/README.md
evmar: This is now A#type.foo, right?

kythe/typescript/README.md
```diff
@@ -1,5 +1,269 @@
...
+`foo` has the signature `A.foo` and `B` has the signature `B#type`.
+
+Signature components (`$PART`s in the schema) are defined below.
+
+#### File Module
+
+**Form**: `module`
+
+**Notes**: The first character of TypeScript source file binds to a VName
+describing the module.
```
evmar: Maybe mention the discussion of module names in the readme.

kythe/typescript/README.md
```diff
@@ -1,5 +1,269 @@
...
+-   `VariableDeclaration`
+-   `PropertyAssignment`
+-   `TypeAliasDeclaration`
+-   `TypeParameter`
+
+```typescript
+export class Klass {
+  //- @property defines/binding VName("Klass.method", _, _, _, _)
+  method() {
+    //- @property defines/binding VName("Klass.method.val", _, _, _, _)
```
evmar: #type

kythe/typescript/README.md
```diff
@@ -1,5 +1,269 @@
...
+}
+
+type B = A;
+```
+
+`foo` has the signature `A.foo` and `B` has the signature `B#type`.
+
+Signature components (`$PART`s in the schema) are defined below.
+
+#### File Module
```
evmar: To keep this spec machine-checked, I wonder if it's better to stuff all of this text as comments within the test.

kythe/typescript/README.md
```diff
@@ -1,5 +1,269 @@
...
+
+Project-specific, defined by the compilation unit you pass to the indexer.
+
+### VName path
+
+-   For entire source code files
+    -   the entire file path, relative to the corpus and root.
+-   For declarations with a file
+    -   the file path stripped of `.d.ts` or `.ts` extensions, relative to the
+        corpus and root.
```
evmar: This might be better offloading to the module name discussion already in the readme.

kythe/typescript/README.md
```diff
@@ -45,11 +309,11 @@ The Google TypeScript build relies heavily on TypeScript's `--declaration` flag
 to enable separate compilation. The way this works is that after compiling
 library A, we generate -- using that flag -- the "API shape" of A into `a.d.ts`.
 Then when compiling a library B that uses A, we compile `b.ts` and `a.d.ts`
-together.  The Kythe process sees the same files as well.
+together. The Kythe process sees the same files as well.
 
 What this means for indexing design is that a TypeScript compilation may see
-only the generated shape of a module, and not its internals.  For example,
-given a file like
+only the generated shape of a module, and not its internals. For example, given
```
evmar: These whitespace changes seem fine to me but maybe put them in a different change to keep the history clear.

kythe/typescript/README.md
ayazhafiz: I agree that is less brittle, but it also makes reading the specification much harder. I'll try it out though.

kythe/typescript/README.md
ayazhafiz: I worry that will make reading the spec much harder, though. The schema shouldn't have many breaking changes from when it's merged, so I don't anticipate this file having to be updated too often. We link to the test file at the top of the schema; if only GitHub supported embedded code snippets in Markdown...

ayazhafiz:
@evmar do you think we're good to merge this

ayazhafiz:
@evmar can you add the buildbot tag?

PR #3784, opened 2019-06-04
feat(typescript_indexer): Getter/Setter entries

ayazhafiz:
Child of #3783. Adds compliance for getters/setters per the spec described there. The idea is that because getters and setters give multiple declarations to the same semantic entity, setters are differentiated in their signature with a `:setter` suffix. The getter has no postfix to ensure compliance that each class member has a signature of the form `Class.member`.

```typescript
class Class {
  //- @foo defines/binding VName("Class.foo", _, _, _, _)
  get foo() { ... }

  //- @foo defines/binding VName("Class.foo:setter", _, _, _, _)
  set foo(...) { ... }
}
```

References for a getter or setter depend on the context. For a given getter/setter pair named `foo`, references to `foo` that are in an assignment context will reference the setter, and all other contexts will reference the getter.

```typescript
class Class {
  method() {
    //- @foo ref FooGetter
    this.foo;
    //- @foo ref FooSetter
    this.foo = 0;
  }
}
```


kythe/typescript/indexer.ts
```diff
@@ -363,16 +374,43 @@ class Vistor {
   }
 
   /** getSymbolName computes the VName (and signature) of a ts.Symbol. */
-  getSymbolName(sym: ts.Symbol, ns: TSNamespace): VName {
+  getSymbolName(sym: ts.Symbol, ns: TSNamespace, context?: ts.Node): VName {
```
evmar: Can you define (in the comment above) what context is?

kythe/typescript/indexer.ts
```diff
@@ -363,16 +374,43 @@ class Vistor {
...
     let vnames = this.symbolNames.get(sym);
-    if (vnames && vnames[ns]) return vnames[ns]!;
+    let declarations = sym.declarations;
+    if (!(context && declarations.length > 1) && vnames && vnames[ns])
+      return vnames[ns]!;
+    // TODO: update symbolNames table to account for context kind
```
evmar: I didn't follow this TODO, can you elaborate?

kythe/typescript/indexer.ts
```diff
@@ -363,16 +374,43 @@ class Vistor {
...
+        default:
+          // If the node is not being assigned to it cannot reference a setter.
+          declarations = declarations.filter(
+              decl => decl.kind !== ts.SyntaxKind.SetAccessor);
+          break;
+      }
+    }
 
-    const decl = sym.declarations[0];
+    const decl = declarations[0];
```
evmar: It's still possible for declarations.length > 1 here, right?

kythe/typescript/indexer.ts
```diff
@@ -996,7 +1036,10 @@ class Vistor {
           // An undeclared symbol, e.g. "undefined".
           return;
         }
-        const name = this.getSymbolName(sym, TSNamespace.VALUE);
+        // The identifier's parent is the node participating in
+        // contextual expression.
+        const expression = node.parent.parent;
```
evmar: This does a double .parent, which is not what the comment suggests.

kythe/typescript/testdata/class.ts
```diff
@@ -52,6 +67,10 @@ class Class implements IFace {
         this.member;
         //- @method ref Method
         this.method();
+        //- @mem ref GetMember
+        this.mem;
+        //- @mem ref SetMember
+        this.mem = 0;
```
evmar: What happens in code like


```
class C {
  get x(): string { return 'a' }
}

interface I {
  x: string;
}

let o: I = new C();
o.x;
```

Where o.x refers to a getter but you can't "see" it from the usage site?
I think maybe o.x refers to I.x, and then we lose the connection between I.x and C.x#getter?

kythe/typescript/indexer.ts
ayazhafiz: Yep, done. I actually had another version I forgot to merge, but I did now so some of this is expired.

kythe/typescript/indexer.ts
ayazhafiz: The VName for a symbol under some context is not currently cached in symbolNames. We may want to add a field in the map that checks for a context and returns the VName for a symbol with that context later on.

kythe/typescript/indexer.ts
ayazhafiz: Yep. The approach I thought was a "best-effort" kind of thing; the role of the context is to reduce the declarations as much as possible.

kythe/typescript/testdata/class.ts
ayazhafiz: Yep, `o.x` will refer to I.x. The same thing happens if `x` is a property or function in `class C`. It's a difficult to tell what the desired behavior here is since `o` has a type of I but a value of C. For me either resolution is acceptable, but I'll defer to your judgement.

creachadair: It's possible to have different representations of the same thing, but I recommend trying to avoid it. When some of the uses do xref with the accessors, and some of the uses do not, users will notice and will definitely complain.

There is no universal solution, but one option is indirection: For example, you could squint and pretend that the explicitly-defined accessors are overrides of implicitly-defined default ones. In that model, any assignment to a field is a ref/call to the implicit setter, and that will cluster with the explicit one via override resolution. (I'm not saying this is the _right_ solution, but I wanted to give an example of how to think about these issues).

creachadair: > @creachadair is this in regards to this interface/class example or the getter/setter idea as a whole?

It's in response to the example @evmar raised, that the indexer might not be able to tell, from the usage site `o.x` whether it dispatches to a getter (say) or is a direct access to the field.

The resulting cross-references can be confusing or even misleading if they're mixed: A developer may infer that the getter isn't being used and try to issue a fix, only to discover the bug wasn't in the code but in the index. With good tests no harm is done, but it's a waste of their time and effort. (This example may sound hypothetical, but it's a distillation of real bug reports we had on Grok over the years)

Conversely, if the indexer speculatively assumes there is always a getter, but there actually _is_ no getter, those cross-references won't link up to the field. 

So despite seeming fairly simple, this is actually an interesting design question for the indexer.

ayazhafiz: > Conversely, if the indexer speculatively assumes there is always a getter, but there actually is no getter, those cross-references won't link up to the field.

Right, I see what you're saying. The indexer assumes nothing about where identifiers are defined; it defers entirely to the TypeScript compiler which is producing the behavior here (I wish I could provide a direct link, but you can plug this into https://ts-ast-viewer.com to see).

> the indexer might not be able to tell, from the usage site o.x whether it dispatches to a getter (say) or is a direct access to the field.

> you could squint and pretend that the explicitly-defined accessors are overrides of implicitly-defined default ones.

If the desire is to pick one behavior, I think this is the better one. We can find the most-overidden (is that the right term?) accessor, but not always the implicit one:

```typescript
class C {
  get x(): string { return 'a' }
}

interface I {
  x: string;
}

let o = new C();
o.x;
```

There is now no relationship between C and I, and there is no way of knowing the definition in `I` exists.

evmar: I realized that in my original example, we already do the "wrong" thing, regardless of this change, because of exactly the example @ayazhafiz provided at the end here -- in a structural type system, you cannot ever "know" all the ways in which types interact.  So forgive my red herring there.

With some more thought, I think my objection here is that in code like
```
o.x = 3;
```
what this code is doing is "peeking in" to the definition of `o` to see whether it has a setter or not.

But per our separate compilation model, you only can peek in to files in the same compilation unit as the current file, and our compilation model is to have many small units.

So I think I've convinced myself that we should always emit the same thing on prop access, which means that for setters/getters we should just name them the same as the property.  Does that make sense?

ayazhafiz: > So I think I've convinced myself that we should always emit the same thing on prop access, which means that for setters/getters we should just name them the same as the property. Does that make sense?

It does. There's a couple of caveats here - the signature of the getter and setter will have to be different (`foo` for the getter and `foo#setter` for the setter, because otherwise there would be two identical VNames; this still means that if there is only a setter its signature won't be the same as a property, but that is a concern only for the VName specification), and the reference emitted will be the first declaration discovered. We can ask that this points to the getter, but if only a setter is defined it will have to point to that. So in code like

```typescript
class A {
  get o() {...}
  set o(...) {...}
```

`A.o`  can always reference the getter, but in 

```typescript
class B {
  set o(...) {...}
```

`B.o` can of course only reference the setter.

ayazhafiz: The play is now
1. `getter`s refer to themselves
2. `setter`s refer to themselves
3. references to a property that _has a getter_ (with or without a setter) refers to the getter
4. references to a property that _only has a setter_ refers to the setter

ayazhafiz:
To resolve issues presented by separate compilation (#3806, also see [@evmar's comment](https://github.com/kythe/kythe/pull/3784#discussion_r292216213)), when getters and setters become just one property name on a class, my approach is now the following.
- Both getters and setters are differentiated, with `:getter` and `:setter` suffixes.
- If a getter is present, it will emit an entry of the form of a class property.
- If only a setter is present, it will emit an entry of the form of a class property.

This lets separate compilation units "see" a getter/setter as just a class property and reference them as such without having to check.

evmar:
Makes sense!  Can you add some tests for the special cases (class with only a getter, class with a setter, class with both)?  Maybe in a separate test file so that we don't overfill class.ts.

evmar:
LGTM, looks great.  The test makes the behavior obvious too.

ayazhafiz:
@creachadair would you like to take another look at this? @evmar are you able to approve this PR?

creachadair:
> @creachadair would you like to take another look at this? @evmar are you able to approve this PR?

It looks fine to me, but I am unable to approve or merge the change, we'll need a core team member.

kythe/typescript/testdata/getter_setter.ts
```diff
@@ -0,0 +1,65 @@
+// Tests the behavior of getter and setter entries.
+
+// Both getters and setters
+class A {
+  prop = 0;
+
+  //- @foo defines/binding GetFoo=VName("A.foo:getter", _, _, _, _)
+  //- GetFoo.node/kind function
+  //- @foo defines/binding PropFoo=VName("A.foo", _, _, _, _)
+  //- PropFoo.node/kind function
```
schroederc: The property itself doesn't seem like a `function` node.  The Kythe team recently discussed getters/setters in another context and thought that these might better be represented as `variable` nodes with a subkind of `implicit`.

We also discussed representing the relation between getters and setters explicitly through that `variable` node with `property/{writes,reads}` edges:

```
//- GetFoo property/reads PropFoo
//- SetFoo property/writes PropFoo
```

What do you think?

ayazhafiz: I like that. I'll throw it together.

schroederc: No, it's not yet available.  I'll look into adding it to the schema soon.

ayazhafiz:
@schroederc I added the requested changes. Is this okay to be merged?

schroederc:
Ah, you should add yourself to the CONTRIBUTORS file in this PR before I merge it.

ayazhafiz:
> Ah, you should add yourself to the CONTRIBUTORS file in this PR before I merge it.

Done :) thanks for reminding me about that.

Issue #2978, opened 2018-09-04
Define schema components for modifiers


creachadair:
Certain nodes have additional Boolean attributes, which are semantically interesting independent of their identity. These often correspond to language keywords such as `static` or `public`, or may arise from metasyntactic annotations (e.g., `@Deprecated`) or conventions of documentation (e.g., the [deprecation comments in Go](https://rakyll.org/deprecated/)).

Update the schema to include a way for indexers to add such annotations uniformly, where it makes sense to do so. Access and visibility modifiers, deprecation, and linkage are good candidates, since these concepts are shared across languages.

creachadair:
The last time we discussed this topic, the leading proposal seems to have been to define a `tag/` namespace of facts, e.g., 

    A.tag/deprecated "use this other thing instead"

For tags without meaningful values (i.e., they are present or absent only), `""` suffices as a value.

    A.tag/static ""

But we would probably want to define at least some categories of these, and we might want to consider whether it's sensible to further partition the namespace by, say `tag/visibility/…` or similar.

jrtom:
Might be worth considering having two categories of tag-likes: ones that have values and ones that don't (since a lot of them don't).

jrtom:
Note that there are (in Java, at least) three different kinds of things here:
* modifiers, e.g.: "public", "static", "final"
  (all built-in, none take 
* [annotations](https://docs.oracle.com/javase/tutorial/java/annotations/predefined.html), e.g.: "@Deprecated" [capital "D"], "@SuppressWarnings("unchecked")", "@VisibleForTesting"
  (some are built into the Java API, some aren't; some take parameters, some don't)
* [Javadoc tags](https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#tag), e.g.: "@deprecated because it's terrible" [lowercase "d"] and "@since 13.0"
  (defined by the Javadoc tool; not sure whether there are examples that don't take parameters or not)

I'm not sure which of these is in scope for this request, but we may want to consider calling them different things in the schema, if only because there's nothing (that I know of) that prevents these types of things having distinct names.

creachadair:
Java-style annotations are already indexed separately, so I don't think we need to cover those explicitly here—we already record their declaration sites and their association to the entity they annotate.

[...]

PR #3785, opened 2019-06-04
feat(typescript_indexer): VName schema compliance for constructors

ayazhafiz:
Child of #3783. Adds compliance for constructors per the spec described there. Classes in TypeScript can be both types and values in expressions, so the indexer currently does the following:

- Add a VName with signature "Class#type" to the class identifier, signifying the type binding
- Add a VName with signature "Class" to the class identifier, signifying the value binding

This PR changes the behavior of the second bullet; now, for noting the class value,

- When the class has a constructor: A VName with signature "Class:ctor" is bound to the constructor, signifying the class value binding so all value references point to the constructor.
- When the class does not have a constructor: A VName with signature "Class:ctor" is bound to the class identifier. Same behavior as above, but references point to the class identifier.

```typescript
class Class {
  //- @constructor defines/binding ClassCtor=VName("Class:ctor", _, _, _, _)
  constructor() {
    //- @Class ref ClassCtor
    new Class();
  }
}

//- @CC defines/binding VName("CC:ctor", _, _, _, _)
class CC {}
```


creachadair:
> * references just point to the entire class definition.

Just a brief terminological nit: The example makes it clear that you mean just the identifier, but note that the data model distinguishes between anchors that span the whole definition (`@—defines—X`) and those that span only the name of the definition (`@—defines/binding—X`). 

For UI support you should always provide at least a `defines/binding` edge (as you've done), but that is not the "entire class definition". It's optional whether to provide a `defines` anchor spanning the whole class definition syntax (here `class CC {}` for the second case).

jaysachs:
> For UI support you should always provide at least a `defines/binding` edge (as you've done), but that is not the "entire class definition". It's optional whether to provide a `defines` anchor spanning the whole class definition syntax (here `class CC {}` for the second case).

Actually we have use cases now that make use of the "entire definition", so I'd like to consider it non-optional.

ayazhafiz:
Yep, that was a blatant mistake on my part. Thanks for the feedback @creachadair.

@jaysachs The TypeScript indexer does not currently generate `defines` anchors for class definitions. This would have to be added, you're saying?

jaysachs:
> Yep, that was a blatant mistake on my part. Thanks for the feedback @creachadair.
> 
> @jaysachs The TypeScript indexer does not currently generate `defines` anchors for class definitions. This would have to be added, you're saying?

Ah, you are correct. I was thinking for functions (which it does since #3785). Support for classes is still missing.

ayazhafiz:
@creachadair @jaysachs would you like to review these changes? anything I should update?

kythe/typescript/testdata/class.ts
```diff
@@ -33,11 +30,12 @@ class Class implements IFace {
...
-    // top-level class declaration.
     // This ctor declares a new member var named 'otherMember', and also
     // declares an ordinary parameter named 'member' (to ensure we don't get
     // confused about params to the ctor vs true member variables).
+    //- @constructor defines/binding ClassCtor
+    //- ClassCtor.node/kind function
+    //- ClassCtor.subkind constructor
     //- @otherMember defines/binding OtherMember
     //- OtherMember.node/kind variable
     //- @member defines/binding FakeMember
```
jaysachs: Wondering if we capture the visibility on member vars? And if so, can you add a test for it here?

jaysachs: Can you also test the scoping (childof) of both `member` and `otherMember`

ayazhafiz: > Wondering if we capture the visibility on member vars? And if so, can you add a test for it here?

What do you mean by visibility?

jaysachs: https://github.com/kythe/kythe/blob/master/kythe/data/schema_index.textproto#L112

defined in #2978 

but I don't think any indexers use it yet. So nevermind.

kythe/typescript/indexer.ts
```diff
@@ -885,16 +910,25 @@ class Visitor {
...
+      let classCtorAnchor;
+      const ctor = sym.members!.get(ts.InternalSymbolName.Constructor);
+      if (ctor) {
+        const decl = ctor.declarations[0];
+        const span = this.getTextSpan(decl, 'constructor');
+        classCtorAnchor = this.newAnchor(decl, span.start, span.end);
+      } else {
+        // TODO: the specific constructor() should really be the thing tagged
+        // with constructor, but we also need to handle classes that don't
+        // declare a constructor.  Fix me later.
```
creachadair: It's not clear what could be fixed in this case—if there is no constructor syntax, the choices are to choose some other syntax (here: the class declaration itself) or to omit the declaration entirely. Since the class declaration is the only thing that _has_ to be present, that seems like the only reasonable fallback.

In which case, I would suggest making that the default:

```ts
let classCtorAnchor = classAnchor;
if (ctor) {
   …
}
```

creachadair:
> @creachadair @jaysachs would you like to review these changes? anything I should update?

Looks fine to me.

kythe/typescript/indexer.ts
ayazhafiz: Sorry, that comment was an artifact before this was fixed. The behavior added in this PR is what you described, so I'll get rid of the comment :-)

kythe/typescript/testdata/class.ts
```diff
@@ -38,8 +40,10 @@ class Class implements IFace {
     //- ClassCtor.subkind constructor
     //- @otherMember defines/binding OtherMember
     //- OtherMember.node/kind variable
+    //- OtherMember childof Class
     //- @member defines/binding FakeMember
     //- FakeMember.node/kind variable
+    //- !{ FakeMember childof Class }
```
jaysachs: FakeMember should also be childof the constructor.

ayazhafiz: You're right. I'll resolve this.

ayazhafiz:
@jaysachs I added the changes you requested. Is there anything else you'd like to be done?

Issue #3786, opened 2019-06-05
Add Plugin model to TypeScript indexer


ayazhafiz:
This proposes an addition of a plugin model to the TypeScript indexer, similar to the Java and C++ ones.

## Motivation

Supersets of TypeScript need to index both TypeScript and their superset-specific code. It would be convenient for the TypeScript indexer to have a plugin model that calls superset indexers as they are needed, so that extra indexer functionality can be introduced without affecting the base TypeScript indexing operations.

### Benefits

Superset of TypeScript can be indexed alongside the existing TypeScript indexer.

### Drawbacks

- Enforces a coupling between TypeScript and its supersets. Only strict supersets of the language would find this useful.
- Plugins are limited in what they are provided and can emit by the plugin API. 
- The TypeScript indexer has to provide some formatting guarantees (#3779)

ayazhafiz:
Closed by #3787

PR #3787, opened 2019-06-05
feature: Plugin model for TypeScript indexer

ayazhafiz:
Adds plugins for extensible indexing functionality (#3786).

Each plugin has an `index` entry point that takes the same inputs the TS indexer's entry point does. The plugins emit with the strategy that the indexer does.

Designed so the caller of the TS indexer has to supply plugins; the indexer will not own any feature extensions.


kythe/typescript/indexer.ts
```diff
@@ -32,6 +32,24 @@ export interface VName {
...
+  name: string;
+  index:
+      (/**
+        * The VName for the CompilationUnit, containing compilation-wide info.
+        */
+       compilationUnit: VName,
+       /**
+        * A map of path to path-specific VName.
+        */
+       pathVNames: Map<string, VName>, paths: string[], program: ts.Program,
```
jaysachs: Instead of replicating the component pieces of the VName construction, what if instead the Plugin took a function that given a path, returned the VName?

ayazhafiz: Great idea. Thanks for the feedback.

kythe/typescript/indexer.ts
```diff
@@ -32,6 +32,24 @@ export interface VName {
...
+ * A indexer plugin adds extra functionality with the same inputs as the base
+ * indexer.
+ */
+export interface Plugin {
+  name: string;
+  index:
+      (/**
+        * The VName for the CompilationUnit, containing compilation-wide info.
+        */
+       compilationUnit: VName,
```
jaysachs: I don't see a need to pass this to the plugins anymore.

jaysachs: if in particular, the path argument to pathVNames could be undefined.

ayazhafiz: I think it's still needed; the `compilationUnit` contains the root and corpus of the project. We could get rid of `pathVNames` though, since that defers to `compilationUnit` if the path doesn't exist (per the existing model of the TS indexer):

https://github.com/kythe/kythe/blob/16bf176d3eb7d0860eed4c9c6785674470a82763/kythe/typescript/indexer.ts#L168-L178

jaysachs: True, it's needed in newFileVName, but I don't see why it needs to be passed to each plugin. In particular, if you instead had
`pathVNames: (path: string?) => VName | undefined, paths: string[]`
then passing undefined would be equivalent to querying for the compilation unit.

ayazhafiz: Okay, I see what you're saying. I agree, and I think it would be even better just to pass `newFileVName`, since the compilation unit's only use is to make path vnames.

kythe/typescript/test.ts
```diff
@@ -128,9 +128,38 @@ async function testIndexer(args: string[]) {
...
+                            .replace(/\.(d\.)?ts$/, '');
+
+        const pluginMod = {
+          ...pathToVName(relPath),
+          signature: 'plugin-module',
+          language: 'plugin-language',
+        };
+        emit!({
+          source: pluginMod,
+          fact_name: '/kythe/node/kind',
```
jaysachs: Just for ensure it's clear we're not "inheriting" the fact from the indexer proper, use some other test fact and value, and verify that.

PR #3790, opened 2019-06-06
chore: add tag/deprecated as a fact in schema.proto for easier parsing

danielmoy-google:


kythe/proto/schema.proto
```diff
@@ -122,6 +122,7 @@ enum FactName {
   TEXT_ENCODING = 17;
   VISIBILITY = 18;
   BUILD_CONFIG = 19;
+  DEPRECATED = 20;
```
schroederc: TAG_DEPRECATED

danielmoy-google:
Oh right I forgot to add in the go proto

schroederc:
I think you need to update the generated code with `update_data.sh`.

PR #3794, opened 2019-06-06
deprecation(java_indexer): change default to --emit_jvm=SEMANTIC

schroederc:


shahms:
That release test failure does seem legitimate

PR #3802, opened 2019-06-08
chore: add visibility/testonly for bazel testonly

danielmoy-google:


kythe/proto/schema.proto
```diff
@@ -123,6 +123,7 @@ enum FactName {
   VISIBILITY = 18;
   BUILD_CONFIG = 19;
   TAG_DEPRECATED = 20;
+  VISIBILITY_TESTONLY = 21;
```
creachadair: Has this already been documented in the scheme? I don't see it in https://kythe.io/docs/schema/ but it's possible there have been changes not yet pushed to the web site.

danielmoy-google: It has not, we're still deciding where to put it exactly.

danielmoy-google: (Also worth noting that a good chunk of the facts aren't documented in kythe.io/docs/schema, specifically a bunch of those specific to the bazel indexer which has yet to be open sourced - this would be another one of those)

I am hoping at some point the bazel team allocates headcount to figure out how to bull that indexer out into open source world.  It's still tricky now because of the differences with blaze I guess.

PR #3806, opened 2019-06-10
docs(typescript_indexer): document separate compilation

evmar:
This describes how separate compilation works for Google's TS builds.
It's not obvious from this code that this is how things work; for
example, in non-Google TS builds, the "compilation unit" that TS works
with is generally all source code in the app all at once.

(As you might expect, that is costly; recent versions of TypeScript have
been working on adding some more incremental compilation support, but
I'm not clear on how exactly it works.)


kythe/typescript/README.md
```diff
@@ -39,6 +39,38 @@ TypeScript projects this doesn't come up because all files are modules.
...
+The generated `.d.ts` file for it describes this getter as if it was a readonly
+property:
+
+```
+class C {
+  readonly x: string;
+}
+```
+
+In practice, what this means is that code should not assume it can to "peek into"
```
creachadair: The reality of scripting languages is that the indexer will usually wind up having to _look_ at the source for the dependencies. Unless there really is a plan to provide separate compilation, in which the compiler plumbs a well-defined type header from each compiled fragment to its dependencies, that may not be completely avoidable. (The `.d.ts` file you describe here sounds like an attempt at that direction).

For indexing purposes, the key thing to avoid is *generating output* for the dependencies. In other words, if you are indexing module A that depends on module B, it's okay—perhaps unavoidable—that you will look at the source text for B's files to discover the VNames of things. The indexer should not _index_ those declarations, however—meaning, it shouldn't emit anchors or references originating in the source files for module B.

Of course, nothing except convention prevents the indexer from doing so, but the reason for avoiding it is straightforward: If every usage of module B re-indexes its source files, the amount of duplicate data will overwhelm the processing capabilities of the indexing pipeline.

(We did this experiment with C++ in around 2012 at Google, pre-Kythe, and ran out our 100TiB storage quota for the indexers in a cell with less than 10% of the C++ code having been indexed. C++ is arguably a harder case because the disposition of templates requires source-equivalent at the usage site to correctly generate instantiations—but the basic issue applies to any language that passes dependencies in source format to a greater or lesser degree).

evmar: What I am documenting here is the current state of things, so there is no plan to provide separate compilation -- separate compilation is implemented and the only mode we support.

I wrote this document to help people *outside* of Google understand some of our design decisions, in that we don't want to break separate compilation even if some non-Google TS build doesn't rely on it.

And yeah, we've written the "reprocess B while working on A" bug like 5 times so far in getting our separate compilation work in TS going. :) In our case, it's "re-typecheck B while checking A", and it manifests as builds that get slower and slower as you get targets with more and more dependencies.  I think we've flushed all them out now though.

evmar:
@ayazhafiz would this have been helpful? or is it too specific to setters/getters?  I could come up with a non-set/get example if it's useful too...

ayazhafiz:
> @ayazhafiz would this have been helpful? or is it too specific to setters/getters? I could come up with a non-set/get example if it's useful too...

I think it's tremendously useful for these kinds of unanticipated cases. I don't think it's too specific to getters/setters.

kythe/typescript/README.md
creachadair: > I wrote this document to help people _outside_ of Google understand some of our design decisions, in that we don't want to break separate compilation even if some non-Google TS build doesn't rely on it.
> 
> And yeah, we've written the "reprocess B while working on A" bug like 5 times so far in getting our separate compilation work in TS going. :) In our case, it's "re-typecheck B while checking A", and it manifests as builds that get slower and slower as you get targets with more and more dependencies. I think we've flushed all them out now though.

It makes sense. Ultimately, all separate compilation is a special case of caching, and on a per-compilation basis it is only a constant-time improvement—but as you've observed, caching type judgements provides an _asymptotic_ improvement when "building" is viewed as an algorithm distinct from "compiling" and multiplied over a large ecosystem of interdependent modules. The larger the ecosystem, the greater the leverage. That insight is obvious in retrospect, but is still demonstrably not "received truth" in the wider community of developers. (Perhaps in part because all languages start small, and the pain doesn't accumulate till they grow)

"Building" also introduces coupling at a higher level than compilation, and language designers mostly abdicate responsibility for that plumbing. The net effect of that (fairly universal) choice is that you can't build efficiently at scale without additional configuration metadata written down _somewhere_, whether that's BUILD files or package registries or combinations thereof.

evmar:
@schroederc  can you approve/merge

PR #3819, opened 2019-06-13
fix(typescript_indexer): differentiate signatures of static and instance members

ayazhafiz:
Currently class instance and static properties generate the same type of signature, which collide when there is a static and instance member of the same name. This resolves that by scoping the signature of instance members to the class type rather than value.

old:
```typescript
class C {
  //- @prop defines/binding VName("C.prop", _, _, _, _)
  prop;
  // Oh no!
  //- @prop defines/binding VName("C.prop", _, _, _, _)
  static prop;
```

new:
```typescript
class C {
  //- @prop defines/binding VName("C#type.prop", _, _, _, _)
  prop;
  //- @prop defines/binding VName("C.prop", _, _, _, _)
  static prop;
```

Part of compliance for #3779, #3783.


ayazhafiz:
@evmar

kythe/typescript/testdata/class.ts
```diff
@@ -27,10 +27,10 @@ class Class implements IFace {
     //- !{ Member.tag/static _ }
     member: number;
 
-    //- @staticMember defines/binding StaticMember
+    //- @member defines/binding StaticMember
     //- StaticMember.tag/static _
     //- StaticMember childof Class
-    static staticMember: number;
+    static member: number;
```
evmar: There's a way in the test language to assert that StaticMember != Member, which is important here.

kythe/typescript/indexer.ts
```diff
@@ -351,6 +359,9 @@ class Visitor {
           const decl = node as ts.NamedDeclaration;
           if (decl.name && decl.name.kind === ts.SyntaxKind.Identifier) {
             let part = decl.name.text;
+            if (this.isStaticMember(startNode, decl)) {
+              part += '#type';
```
evmar: I think this might be backwards.

Given a class A, the value `A` has all the statics as properties.  Confusingly the type is `typeof A`, not `A`.  Meanwhile, *instances* of `A` (which are of type `A`) are the things that carry the members on them.

Another way to say this is like:

```
const obj = { z: 3 };
class C1 {
  z: number;
}
class C2 {
  static z: number;
}
interface HasZ {
  z: number;
}
let hasZ: HasZ = obj;  // valid
hasZ = C1;  // wrong
hasZ = C2;  // valid
hasZ = new C1;  // valid
```

And for the `HasZ` case we do name the field like `HasZ.z` or so on, so I think the class should match.


One other thing to check here:

```
class X {
  static y = class Y {
    z: number;
  };
  static a = {
    b: 3;
  };
}
```

I'm not sure what name we get for `z` and `b`.  (It could be that we do something entirely wrong already and so don't worry about it in that case, I just wanna be sure we don't do something catastrophically wrong.)

ayazhafiz: With the new changes, for z, we get a signature of `X#type.y.Y.z`. For b, it's `X.a.anon0.b` because of the anonymous block.

evmar:
BTW I haven't forgotten about this, I am just thinking through how to best express it.  For example maybe we also are naming methods wrong?  (Class with a static field and a member now conflict)

ayazhafiz:
> Class with a static field and a method with same name now conflict

They don't; the method is scoped to the class type, and the static field to the class value.

kythe/typescript/indexer.ts
ayazhafiz: I think separately from this PR, we should reconsider how both these signatures are created.

> class X {
>   static y = class Y {
>     z: number;
>   };
>   static a = {
>     b: 3;
>   };
> }

Ideally, `z` would have the signature `X.y.z` and `b` `X.a.b`. Right now the problem is we're generating extra scopes where maybe we don't want them (at assignments?)

ayazhafiz: A code like

```typescript
const a = class A {
  prop = 0;
};
```

Creates a declaration file

```typescript
declare const a: {
    new (): {
        prop: number;
    };
};
```

I'm not sure how to resolve this.

Should we move this discussion to a new issue?

ayazhafiz: Moved to a separate issue. Resolving this conversation.

evmar:
LGTM, sorry for the delay.  The test makes it more obvious that this is right.

ayazhafiz:
@kythe could you review and merge this?

ayazhafiz:
> I think you have a typo in the commit description. pop should be prop.

Resolved.

salguarnieri:
Can you resolve the conflicts, then I can merge it. Thanks!

ayazhafiz:
> Can you resolve the conflicts, then I can merge it. Thanks!

Resolved.

PR #3822, opened 2019-06-14
fix(proto): include deprecation tags for messages, enums and enum values.

jaysachs:


kythe/cxx/indexer/proto/testdata/basic/deprecated.proto
```diff
@@ -14,4 +20,44 @@ message M {
...
+    //- F4.tag/deprecated ""
+    optional string field4 = 1 [deprecated = true];
+  }
+
+
+  //- @D defines/binding EnumD
+  //- EnumD.node/kind sum
+  //- EnumD.tag/deprecated ""
+  enum D {
+    option deprecated = true;
```
justbuchanan: If the enum itself is deprecated, do we want to mark the individual values deprecated? In some uses of the enum, we see `VAL1`, but not `D`.

justbuchanan: Actually it's probably best as you have it. I don't think we propagate the parent's deprecated tag to children/fields in other languages.

creachadair: That's a really interesting idea, though—I suspect the code generator could do that, in which case the index would reflect it.

jaysachs: Though the languages themselves don't necessarily propagate these things, e.g.

```java
class A {
  @Deprecated void foo() { }
}

@Deprecated
class B {
  void foo() { }
}

public class D {
  public static void main(String[] args) {
    A a = new A();
    a.foo(); // warning here
    B b = new B(); // warning here
    b.foo();  // no warning here
  }
}
```

creachadair: Right, that's why I say the code generator would have to be involved: When generating the class (say) for a protobuf enum, the generator has the option to inject the annotation not just on the class itself, but on all the appropriate members. I don't think any of the extant generators actually does that, but that's arguably the correct place to put it if someone wanted it.

PR #3823, opened 2019-06-14
typescript: handle constraints/bounds in generics

evmar:
E.g. f<T extends Foo>, we should examine Foo.  See test.


evmar:
cc @ayazhafiz  (random CL)

kythe/typescript/testdata/generic.ts
```diff
@@ -33,6 +33,14 @@ function generic<T>(x: T, y: IFace) {
   return x;
 }
 
+// Constrained generic.
+//- @IFace ref IFace
+function constrainedGeneric<T extends IFace>() {}
```
ayazhafiz: (nit) maybe we should define the relationship between T and IFace as well?

evmar: Good idea, fixed.

kythe/typescript/indexer.ts
```diff
@@ -416,6 +416,13 @@ class Visitor {
       const kType = this.getSymbolName(sym, TSNamespace.TYPE);
       this.emitNode(kType, 'absvar');
       this.emitEdge(this.newAnchor(param.name), 'defines/binding', kType);
+      // ...<T extends A>
+      if (param.constraint) {
+        const superType = this.visitType(param.constraint);
```
jaysachs: Does this handle things like "<T, K extends keyof T>" ?

(and is there only "extends" in TS, no "super" like Java has?)

And are there recursive types, e.g. <T extends Foo<T>>, and if so does this handle those?

evmar: #1: yes, added a test
#2: right
#3: yes, added a test

evmar:
PTAL, ready for submit

kythe/typescript/testdata/generic.ts
```diff
@@ -33,6 +33,27 @@ function generic<T>(x: T, y: IFace) {
...
+//- @T defines/binding ConstrainedGenericT
+//- @IFace ref IFace
+//- ConstrainedGenericT bounded/upper IFace
+function constrainedGeneric<T extends IFace>() {}
+
+// Constrained generic with reference to type param within it.
+//- @#0T defines/binding ConstrainedGenericT2
+//- !{@#0T ref ConstrainedGenericT}
+//- @#1T ref ConstrainedGenericT2
+function constrainedGenericRef<T, K extends keyof T>() {}
```
jaysachs: Are there some assertions on K that could be made here?

kythe/typescript/testdata/generic.ts
```diff
@@ -33,6 +33,27 @@ function generic<T>(x: T, y: IFace) {
...
+
+// Recursive generic type.
+//- @#0T defines/binding ConstrainedGenericT3
+//- @#1T ref ConstrainedGenericT3
+function constrainedGenericRecursive<T extends Array<T>>() {}
+
+// Default generic.
+//- @IFace ref IFace
+function defaultGeneric<T = IFace>() {}
+
```
jaysachs: Can you supply a default when a bound is also specified, e.g.

  function f<T extends C<T> = U>() {}

kythe/typescript/testdata/generic.ts
```diff
@@ -33,6 +33,27 @@ function generic<T>(x: T, y: IFace) {
...
+function constrainedGenericRef<T, K extends keyof T>() {}
+
+// Recursive generic type.
+//- @#0T defines/binding ConstrainedGenericT3
+//- @#1T ref ConstrainedGenericT3
+function constrainedGenericRecursive<T extends Array<T>>() {}
+
+// Default generic.
+//- @IFace ref IFace
+function defaultGeneric<T = IFace>() {}
```
ayazhafiz: What is the relationship between `T` and `IFace`? Can you add a verifier goal for the binding of `T`?

kythe/typescript/testdata/generic.ts
evmar: Added some.

kythe/typescript/testdata/generic.ts
evmar: Yes, and I added an example, but I'm not what else to say.

kythe/typescript/testdata/generic.ts
evmar: IFace is the default value (er, type) of T.  I added a bit but I'm really sure what to assert.

evmar:
@kythe plz merge

PR #3825, opened 2019-06-17
build: canonicalize Bazel repos pinned to a commit

schroederc:


external.bzl
```diff
@@ -385,6 +389,7 @@ def _go_dependencies():
         name = "org_golang_x_tools",
         commit = "589c23e65e65055d47b9ad4a99723bc389136265",
         custom = "x_tools",
+        shallow_since = "1551386336 +0000",
```
creachadair: Not that it really matters, but why do the timezones vary on these? Does GitHub really return different values for different repositories?

schroederc: `¯\_(ツ)_/¯` I copied these values verbatim from Bazel's complaints.  However, Bazel does not seem to be able to use them on the Buildbot instance, so I just abandoned the change.

PR #3828, opened 2019-06-18
feat(typescript_indexer): pass module name creation logic to plugins

ayazhafiz:
Plugins would benefit from determining the name of TypeScript modules in the same manner that the TypeScript indexer does, to ensure compatibility between the TS indexer and its plugins. This PR adds this feature.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -33,11 +33,11 @@ export interface VName {
 }
 
 /**
- * A indexer host holds information about the program indexing and methods used
- * by the TypeScript indexer that may also be useful to plugins, reducing code
- * duplication.
+ * An indexer context holds information about the program indexing and methods
+ * used by the TypeScript indexer that may also be useful to plugins, reducing
+ * code duplication.
  */
-export interface IndexerHost {
+export interface IndexerContext {
   /**
    * Converts a file path into a file VName.
    */
@@ -69,10 +69,10 @@ export interface Plugin {
   name: string;
   /**
    * Indexes a TypeScript program with extra functionality.
-   * Takes a indexer host, which provides useful properties and methods that
+   * Takes a indexer context, which provides useful properties and methods that
    * the plugin can defer to rather than reimplementing.
    */
-  index(host: IndexerHost): void;
+  index(context: IndexerContext): void;
 }
 
 /**
@@ -124,19 +124,16 @@ enum Context {
 }
 
 /**
- * IndexerContext provides useful information about a TypeScript program and
- * common methods used by the TypeScript indexer and its plugins.
- * See the IndexerHost interface definition for more details.
+ * StandardIndexerContext provides the standard definition of information about
+ * a TypeScript program and common methods used by the TypeScript indexer and
+ * its plugins. See the IndexerContext interface definition for more details.
  */
-class IndexerContext implements IndexerHost {
+class StandardIndexerContext implements IndexerContext {
   /**
    * rootDirs is the list of rootDirs in the compiler options, sorted
    * longest first.  See this.moduleName().
    */
   rootDirs: string[];
-
-  /** A shorter name for the rootDir in the CompilerOptions. */
-  sourceRoot: string;
 
   constructor(
       /**
@@ -151,8 +148,8 @@ class IndexerContext implements IndexerHost {
       public paths: string[],
       public program: ts.Program,
   ) {
-    this.sourceRoot = program.getCompilerOptions().rootDir || process.cwd();
-    let rootDirs = program.getCompilerOptions().rootDirs || [this.sourceRoot];
+    const sourceRoot = program.getCompilerOptions().rootDir || process.cwd();
+    let rootDirs = program.getCompilerOptions().rootDirs || [sourceRoot];
     rootDirs = rootDirs.map(d => d + '/');
     rootDirs.sort((a, b) => b.length - a.length);
     this.rootDirs = rootDirs;
@@ -226,6 +223,9 @@ class Visitor {
    */
   anonNames = new Map<ts.Node, string>();
 
+  /** A shorter name for the rootDir in the CompilerOptions. */
+  sourceRoot: string;
+
   typeChecker: ts.TypeChecker;
 
   constructor(
@@ -233,6 +233,9 @@ class Visitor {
       private file: ts.SourceFile,
       private getOffsetTable: (path: string) => utf8.OffsetTable,
   ) {
+    this.sourceRoot =
+        this.context.program.getCompilerOptions().rootDir || process.cwd();
+
     this.typeChecker = this.context.program.getTypeChecker();
 
     this.kFile = this.newFileVName(file.fileName);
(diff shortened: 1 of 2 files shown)


ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -39,7 +39,8 @@ export interface VName {
 export interface Plugin {
   name: string;
   index:
-      (pathToVName: (path: string) => VName, paths: string[],
+      (pathToVName: (path: string) => VName,
+       moduleName: (path: string) => string, paths: string[],
        program: ts.Program, emit?: (obj: {}) => void) => void;
```
evmar: Also, as this accumulates more arguments, perhaps it could itself be a named interface, like

```
interface PluginHost {
  pathToVName(...);
  moduleName(...);
}
```

Then here:

```
index(host: PluginHost): void;
```

In fact, the `Indexer` class itself might already implement this PluginHost interface, which would mean you just pass "this" to the plugin to forward this functionality to it.

kythe/typescript/indexer.ts
```diff
@@ -79,6 +80,26 @@ function getFileVName(
   };
 }
 
+/**
+ * moduleName returns the ES6 module name of a path to a source file.
+ * E.g. foo/bar.ts and foo/bar.d.ts both have the same module name,
+ * 'foo/bar', and rootDirs (like bazel-bin/) are eliminated.
+ * See README.md for a discussion of this.
```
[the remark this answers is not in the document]
ayazhafiz: Moved this back inside a class so the sorted rootDirs are guaranteed.

kythe/typescript/indexer.ts
ayazhafiz: I like this much more, thanks Evan! I've implemented this.

kythe/typescript/indexer.ts
```diff
@@ -32,15 +32,46 @@ export interface VName {
...
   name: string;
-  index:
-      (pathToVName: (path: string) => VName, paths: string[],
-       program: ts.Program, emit?: (obj: {}) => void) => void;
+  /**
+   * Indexes a TypeScript program with extra functionality.
+   * Takes a plugin host, which provides useful properties and methods that the
+   * plugin can defer to rather than reimplementing.
+   */
+  index: (host: PluginHost) => void;
```
evmar: Same comment as before here:
  index(host: PluginHost): void;

kythe/typescript/indexer.ts
```diff
@@ -106,8 +122,83 @@ enum Context {
...
+   * emitting to stdout but users may replace it.
+   */
+  emit = (obj: {}) => {
+    console.log(JSON.stringify(obj));
+  };
+}
+
 /** Visitor manages the indexing process for a single TypeScript SourceFile. */
-class Visitor {
+class Visitor extends IndexerHost {
```
evmar: I think I'd just leave all the methods in Visitor, there's no real design benefit to separating the two.  It's unclear which properties/methods belong in which, and as the plugin needs more stuff we'd need to move them around.

kythe/typescript/indexer.ts
```diff
@@ -1275,9 +1334,11 @@ export function index(
   if (plugins) {
     for (const plugin of plugins) {
       try {
-        plugin.index(
-            (path: string) => getFileVName(path, pathVNames, vname), paths,
-            program, emit);
+        const host = new IndexerHost(vname, pathVNames, paths, program);
```
evmar: Then down here you can just pass 'visitor' as the host instead.

kythe/typescript/indexer.ts
ayazhafiz: > I think I'd just leave all the methods in Visitor, there's no real design benefit to separating the two.

I tried this, but passing the Visitor as the `PluginHost` felt a little bit hacky because the Visitor is constructed for only one source file to be indexed:

https://github.com/kythe/kythe/blob/c2e3491259dabe1fb3c5aa9dfe55efc7c58893bc/kythe/typescript/indexer.ts#L153

which means that passing a Visitor to the plugins, we would have to construct a dummy Visitor to some arbitrary TypeScript file. Because this arbitrary file is never used in the `PluginHost`, I find it a little confusing and inelegant to include it - a conscious reader may question what `files[0]` in the Visitor construction `new Visitor(vname, pathVNames, paths, program, files[0])`, and while we can add a comment that the file is irrelevant, this seems less elegant to me than decoupling the Host from the actual visitor.

Please let me know if you disagree, I want to discuss this.

kythe/typescript/indexer.ts
ayazhafiz: See previous conversation.

kythe/typescript/indexer.ts
evmar: Nope, your reasoning makes sense!

One other idea: would it make sense for Visitor to accept an IndexerHost as a parameter too?  That would mean we create one IndexerHost and pass it to both the Visitor and the Plugin.  In particular I wonder if there's a more meaningful name we can give IndexerHost, because right now it's basically "grab bag of all the stuff that Visitor and Plugin both need access to".

This LGTM as is if you don't have any better ideas, just thinking aloud.

jaysachs: I'm wondering about the name "...Host". I wonder if the more generic "Context" might be better, e.g. PluginContext, IndexerContext, etc.

ayazhafiz: How about `IndexerContext`? In this case, though, it would probably be best to get rid of the interface or rename it to something even more generic, which I'm not sure there is.

ayazhafiz: Changed the interface to `IndexerHost` and the implementation to `IndexerContext`. The Visitor and plugins now both consume a common `IndexerContext`.

kythe/typescript/indexer.ts
```diff
@@ -32,15 +32,47 @@ export interface VName {
   language: string;
 }
 
+/**
+ * A indexer host holds information about the program indexing and methods used
+ * by the TypeScript indexer that may also be useful to plugins, reducing code
+ * duplication.
+ */
+export interface IndexerHost {
```
jaysachs: "Host" just doesn't sit well. What about having this be "IndexerContext" and the implementation something (java-esque, I know) like "StandardIndexerContext"?

ayazhafiz: Works for me,

evmar: My suggestion of "Host" here is intentional, because it's what the TypeScript compiler uses.  For example as users of it we provide a CompilerHost, ModuleResolutionHost, FormatDiagnosticsHost and so on.  Looking at the TS API it looks like they define ~12 "Host" interfaces.

jaysachs: Ah. As unintuitive as it reads (to me), aligning with the compiler makes sense.

ayazhafiz: Moved the interface to `IndexerHost`. The class is still `StandardIndexerContext`. Let me know if either of you think that should be renamed too :)

ayazhafiz:
@jaysachs is this good to merge?

PR #3829, opened 2019-06-18
chore(java_common): move JDK9-compatibility shims to common base

shahms:
This will allow sharing the hacks between the indexer and extractor.


kythe/java/com/google/devtools/kythe/common/BUILD
```diff
@@ -1,5 +1,11 @@
 package(default_visibility = ["//kythe:default_visibility"])
 
+java_library(
```
schroederc: Move to kythe/java/com/google/devtools/kythe/platform/java/filemanager/

PR #3835, opened 2019-06-19
fix(typescript_indexer): point class type and values at the class name

### LINKED ISSUE #3833: TypeScript indexer: Fix handling of class values and types
Since cdee8bf8a813ddd0309dbab00110f0e972282996, the structure for class VNames is:

- Class name identifiers refer to the class type and have a node kind of a record
- Class constructors refer to the class value and have a node kind of a function, subkind of constructor. If there is no constructor, the class name takes this.

```typescript
//- @Class defines/binding Class=VName("Class#type", _, _, _, _)
class Class {
  //- @constructor defines/binding ClassCtor=VName("Class:ctor", _, _, _, _)
  constructor() {
    //- @Class ref ClassCtor
    new Class();
  }
}

//- @CC defines/binding VName("CC#type", _, _, _, _)
//- @CC defines/binding VName("CC:ctor", _, _, _, _)
class CC {}
```

This is probably backwards. The constructor should refer to the class type, and the class value should point to the class identifier. This should resolve problems when references to classes are interpreted in a value context and taken to the constructor when they should be taken to the class identifier. The class type and value will both have a node kind of a record.

We should also emit a separate entry for constructors with a signature of `Class.constructor`, for example. This makes scoping parameters and variables inside the constructor much easier, so that while the constructor and class type point to the same thing, they are not actually the same thing. The class constructor will have a node kind of a function with a subkind of a constructor.

ayazhafiz:
Revert "feat(typescript_indexer): VName schema compliance for constructors (#3785)"
Also add a 'ref/call' edge to class constructors in a `new` expression.

This reverts commit cdee8bf8a813ddd0309dbab00110f0e972282996.

Resolves #3833


ayazhafiz:
cc @evmar

evmar:
LGTM

ayazhafiz:
@kythe could you review this?

PR #3840, opened 2019-06-26
feat(kzip): add `info` command to kzip tool which returns count of compilation units and referenced corpus names.

jaysachs:


kythe/go/platform/tools/kzip/infocmd/infocmd.go
```diff
@@ -0,0 +1,91 @@
...
+		return c.Fail("empty .kzip: %v", c.input)
+	}
+
+	rd, err := kzip.NewReader(f, size)
+	if err != nil {
+		return c.Fail("error creating reader: %v", err)
+	}
+
+	corpora := stringset.New()
+	units := 0
```
shahms: var units int

PR #3841, opened 2019-06-26
fix(typescript_indexer): Do not name scopes created by rvalues

### LINKED ISSUE #3834: TypeScript indexer: Do not name scopes created by rvalues
See [this discussion](https://github.com/kythe/kythe/pull/3819#discussion_r294388875) on #3819.

I think we should reconsider how signatures are created for rvalues. In particular, I don't think rvalues should introduce a new scope, since they are anonymous to the lvalues. In a case like

```typescript
class X {
  static y = class Y {
    z: number;
  };
  static a = {
    b: 3;
  };
}
```
Right now, for z, we get a signature of `X.y.Y.z`. For b, it's `X.a.anon0.b` because of the anonymous block.
Ideally, z would have the signature `X.y.z` and b `X.a.b`.

Further complicating this, a code like

```typescript
const a = class A {
  prop = 0;
};
```

Creates a declaration file

```typescript
declare const a: {
    new (): {
        prop: number;
    };
};
```

I'm not sure how to resolve this. The least ambigious answer to me is to give `prop` a signature of `a.prop` in both cases, and we treat the constructor signature `new ()` as an rvalue.

_Edit_: Better explain the issue and proposed solution here.

ayazhafiz:
When creating the signature of a declaration, do not include scopes created by rvalue in initializers.

Closes #3834.


ayazhafiz:
cc @evmar

ayazhafiz:
@jaysachs is this ready to merge after buildbot?

PR #3846, opened 2019-06-26
fix(kzip): use proper incantation so testdata still found on import

jaysachs:


kythe/go/platform/kzip/BUILD
```diff
@@ -23,12 +23,10 @@ go_library(
 go_test(
     name = "kzip_test",
     srcs = ["kzip_test.go"],
-    data = [
-        "missing-pbunit.kzip",
-        "missing-unit.kzip",
-    ],
+    data = glob(["testdata/*.kzip"]),
```
shahms: should still enumerate the expected testdata files

PR #3851, opened 2019-06-28
feat(go extractor): allow specifying go +build tags

justbuchanan:
Some go files are conditionally included in the build according to semantic comments at the top of the file. The new `--buildtags` flag allows specifying which of these tags/constraints to enable so that the right files get included by the extractor.

https://golang.org/pkg/go/build/#hdr-Build_Constraints

TODO: is there a cleaner way to have multi-value flags in go?

### DIFF
--- a/kythe/go/extractors/cmd/gotool/gotool.go
+++ b/kythe/go/extractors/cmd/gotool/gotool.go
@@ -73,7 +73,7 @@ Options:
 	flag.StringVar(&bc.GOROOT, "goroot", bc.GOROOT, "Go system root")
 	flag.BoolVar(&bc.CgoEnabled, "gocgo", bc.CgoEnabled, "Whether to allow cgo")
 	flag.StringVar(&bc.Compiler, "gocompiler", bc.Compiler, "Which Go compiler to use")
-	flag.Var(&buildTags, "buildtags", "Comma-separated list Go +build tags to enable during extraction.")
+	flag.Var(&buildTags, "buildtags", "Comma-separated list of Go +build tags to enable during extraction.")
 
 	// TODO(fromberger): Attach flags to the build and release tags (maybe).
 }
--- a/kythe/go/extractors/cmd/gotool/gotool.go
+++ b/kythe/go/extractors/cmd/gotool/gotool.go
@@ -50,6 +50,20 @@ var (
 	canonicalizePackageCorpus = flag.Bool("canonicalize_package_corpus", false, "Whether to use a package's canonical repository root URL as their corpus")
 )
 
+// Wrapper type for using flag.Var() to collect a list of strings.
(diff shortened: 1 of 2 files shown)


kythe/go/extractors/cmd/gotool/gotool.go
```diff
@@ -50,6 +50,20 @@ var (
...
+
+func (af *arrayFlags) String() string {
+	return strings.Join(*af.array, ",")
+}
+
+func (af *arrayFlags) Set(value string) error {
+	*af.array = append(*af.array, value)
+	return nil
+}
+
```
jaysachs: I think, based on my recent, change, you'll also need a Get implementation.

Also, maybe "stringArrayFlag" as the type name?

kythe/go/extractors/cmd/gotool/gotool.go
```diff
@@ -70,6 +84,7 @@ Options:
 	flag.StringVar(&bc.GOROOT, "goroot", bc.GOROOT, "Go system root")
 	flag.BoolVar(&bc.CgoEnabled, "gocgo", bc.CgoEnabled, "Whether to allow cgo")
 	flag.StringVar(&bc.Compiler, "gocompiler", bc.Compiler, "Which Go compiler to use")
+	flag.Var(&arrayFlags{array: &bc.BuildTags}, "buildtag", "Go +build tags to enable during extraction. May be specified multiple times.")
```
jaysachs: "buildtags"?

Also, what about plumbing them through the build?

kythe/go/extractors/cmd/gotool/gotool.go
```diff
@@ -70,6 +84,7 @@ Options:
...
```
shahms: `flagutil.StringSet` from `kythe.io/kythe/go/util/flagutil`

shahms: or flagutil.StringList

justbuchanan: Awesome :+1:

kythe/go/extractors/cmd/gotool/gotool.go
justbuchanan: Renamed to "buildtags".

What do you mean by "plumbing them through the build"? The extractor does already handle these correctly, it just didn't have a way to specify them on the command line.

creachadair: This extractor doesn't actually execute a build—it only simulates one. When running `go build` explicitly one includes tags in the same way, via a flag.

When building in Bazel, tags are handled differently: Bazel permits mixing packages in a single directory, so the the rules have to do some shenanigans to virtualize the directory structure and I believe file selection is done there.

PR #3855, opened 2019-07-01
typescript: handle variables declared within for-loop initializer

evmar:
Rather than specially handling for-loops, instead change around some
other code so our existing processing picks this up.

To understand why this works, consider:
  var x = 3;
  for (var y in ...)
The first is a VariableDeclarationStatement, while both both contain
a VariableDeclaration.

Previously we handled VariableDeclarationStatement specially to extract
defines/binding.  Instead if we handle VariableDeclaration, both forms
are covered.

So this fix for for loops instead only touches VariableDeclaration, and
then to delete a bit of our VariableDeclarationStatement handling.


evmar:
CC @ayazhafiz

kythe/typescript/indexer.ts
```diff
@@ -1226,6 +1229,9 @@ class Visitor {
         return this.visitExportDeclaration(node as ts.ExportDeclaration);
       case ts.SyntaxKind.VariableStatement:
         return this.visitVariableStatement(node as ts.VariableStatement);
+      case ts.SyntaxKind.VariableDeclaration:
+        this.visitVariableDeclaration(node as ts.VariableDeclaration);
```
ayazhafiz:
```suggestion
        return this.visitVariableDeclaration(node as ts.VariableDeclaration);
```

kythe/typescript/testdata/for.ts
```diff
@@ -0,0 +1,8 @@
+export {};
+
+//- @x defines/binding X
+//- X.node/kind variable
+for (const x of [1, 2]) {
+  //- @x ref X
+  x;
+}
```
ayazhafiz: Can you add some more examples, like standard for loops?

evmar: Done.  Also it revealed some bugs!  So I fixed those too.

evmar:
@kythe plz merge

PR #3856, opened 2019-07-02
fix(go): -tags is space-separated, not comma

justbuchanan:
See `go help build` for more info.

### DIFF
--- a/kythe/go/extractors/golang/packages.go
+++ b/kythe/go/extractors/golang/packages.go
@@ -150,7 +150,7 @@ func (e *Extractor) listPackages(query ...string) ([]*jsonPackage, error) {
 	// TODO(schroederc): support GOPACKAGESDRIVER
 	args := append([]string{"list",
 		"-compiler=" + e.BuildContext.Compiler,
-		"-tags=" + strings.Join(e.BuildContext.BuildTags, ","),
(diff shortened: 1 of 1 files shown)


kythe/go/extractors/golang/packages.go
```diff
@@ -150,7 +150,7 @@ func (e *Extractor) listPackages(query ...string) ([]*jsonPackage, error) {
 	// TODO(schroederc): support GOPACKAGESDRIVER
 	args := append([]string{"list",
 		"-compiler=" + e.BuildContext.Compiler,
-		"-tags=" + strings.Join(e.BuildContext.BuildTags, ","),
+		"-tags=" + strings.Join(e.BuildContext.BuildTags, " "),
```
salguarnieri: How does this work? Why isn't each tag after the first picked up as a new arg/flag?

justbuchanan: Because these are passed to `exec.Command`, I think it essentially gets quoted like "-tags=tag1 tag2" before being passed to to `go` cmd.

I agree though that spaces vs commas is weird...

justbuchanan: If you just did something like `go build -tags tag1 tag2` on the cmdline, it would behave like you said and see "tag2" as a new flag.

PR #3859, opened 2019-07-02
fix(textproto): continue analysis even if schema comments are invalid

justbuchanan:
instead of exiting when bad schema comments are found, log an error and continue

### DIFF
--- a/kythe/cxx/indexer/textproto/analyzer.cc
+++ b/kythe/cxx/indexer/textproto/analyzer.cc
@@ -523,7 +523,7 @@ Status AnalyzeCompilationUnit(const proto::CompilationUnit& unit,
 
   Status status = analyzer.AnalyzeSchemaComments(*file_vname, *descriptor);
   if (!status.ok()) {
-    return status;
+    LOG(ERROR) << "Error analyzing schema comments: " << status;
   }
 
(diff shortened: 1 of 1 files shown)


kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -523,7 +523,7 @@ Status AnalyzeCompilationUnit(const proto::CompilationUnit& unit,
 
   Status status = analyzer.AnalyzeSchemaComments(*file_vname, *descriptor);
   if (!status.ok()) {
-    return status;
+    LOG(ERROR) << "Error analyzing schema comments: " << status;
```
shahms: Can we emit a diagnostic? At least in theory those might show up in the UI at some point.

justbuchanan: sgtm. Is there a canonical/preferred way to construct the vname for the diagnostic? I see in `KytheEntrySets.java`, there's some code to generate a signature based on the hash of the contents. I could do something similar here, but right now I'm just using "schema_comments" as the signature since there will be at most one of them.

shahms: That made sense to me; if there's a future need to change the signature it's pretty simple to hash the contents or something as you say.

justbuchanan: :+1:

PR #3861, opened 2019-07-03
fix(KZipWriterTest.java): remove previous kzip before writing new one

justbuchanan:
The test would occasionally fail due to interactions with the old file.
Error msg: "java.lang.IllegalArgumentException: MALFORMED" at java.util.zip.ZipCoder.toString(ZipCoder.java:65).

I'm not 100% sure why this is an issue. Does the kzip writer append to an existing file rather than replacing it? If so, maybe that should be changed instead.

### DIFF
--- a/kythe/javatests/com/google/devtools/kythe/platform/kzip/KZipWriterTest.java
+++ b/kythe/javatests/com/google/devtools/kythe/platform/kzip/KZipWriterTest.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package com.google.devtools.kythe.platform.kzip;
 
 import static com.google.common.truth.Truth.assertThat;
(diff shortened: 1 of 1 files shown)


kythe/javatests/com/google/devtools/kythe/platform/kzip/KZipWriterTest.java
```diff
@@ -77,8 +77,12 @@ private void testWrite(KZip.Encoding encoding) throws IOException, KZipException
     // Read in the provided kzip.
     InMemoryKZip originalKZip = readKZip(TestDataUtil.getTestFile("stringset.kzip"));
 
+    // Delete old tmp kzip (if present).
+    File tmpKZipFile = new File(getTestTempDir(), "write_test.kzip");
+    tmpKZipFile.delete();
+
     // Write out our kzip.
-    KZipWriter writer = new KZipWriter(new File(getTestTempDir(), "write_test.kzip"), encoding);
```
jaysachs: So I think the real issue is that the temp file is reused across tests. Rather than deleting it, let's make sure it has a unique name. Maybe just prepend encoding.toString() to the name?

justbuchanan: Sounds good. Done.

PR #3864, opened 2019-07-03
feat(typescript_indexer): add xrefs for declarations in objects

ayazhafiz:
Shorthand property assignments like foo in a = {foo,} were
previously not indexed at all. This emits edges for shorthand
assignments similar to other property assignments. Also adds xrefs for
computed properties in objects and ignores lvalue literals in a
declaration.

_edit_: new description

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -968,6 +968,7 @@ class Visitor {
     let vname: VName|undefined;
     switch (decl.name.kind) {
       case ts.SyntaxKind.Identifier:
+      case ts.SyntaxKind.ComputedPropertyName:
         const sym = this.getSymbolAtLocation(decl.name);
         if (!sym) {
           this.todo(
@@ -978,15 +979,16 @@ class Visitor {
         this.emitNode(vname, 'variable');
 
         this.emitEdge(this.newAnchor(decl.name), 'defines/binding', vname);
+
+        if (ts.isComputedPropertyName(decl.name)) {
+          this.visit(decl.name.expression);
+        }
         break;
       case ts.SyntaxKind.ObjectBindingPattern:
       case ts.SyntaxKind.ArrayBindingPattern:
         for (const element of (decl.name as ts.BindingPattern).elements) {
           this.visit(element);
         }
-        break;
-      case ts.SyntaxKind.ComputedPropertyName:
-        this.visit(decl.name.expression);
         break;
       default:
         this.todo(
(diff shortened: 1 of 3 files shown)


ayazhafiz:
cc @evmar

ayazhafiz:
> It's not clear whether the `foo` on L3 should xref to the `foo` on L2 or the `foo` on L4. If you write it out as `foo: foo` then the two things to anchor to are obvious.

I was just thinking about this. Turns out with the current TypeScript AST it is impossible to determine the xref to the declaration a shorthand property expression takes the value of, at least from my playing on https://ts-ast-viewer.com/

kythe/typescript/indexer.ts
```diff
@@ -984,10 +985,15 @@ class Visitor {
           this.visit(element);
         }
         break;
+      case ts.SyntaxKind.ComputedPropertyName:
+        this.visit(decl.name.expression);
+        break;
+      // Nothing meaningful can be indexed about a literal expression.
+      case ts.SyntaxKind.StringLiteral:
+      case ts.SyntaxKind.NumericLiteral:
```
evmar: Maybe you intended these for another change (?)

kythe/typescript/testdata/object.ts
```diff
@@ -1,14 +1,25 @@
...
+  [computed]: 0,
+
   //- @method defines/binding Method
   //- Method.node/kind function
-  method() {
-  }
+  method() {}
 };
 
 //- @property ref Property
```
evmar: Maybe some test of accessing shortProperty here too?

ayazhafiz:
This should clean up most `todo`s recorded when running the indexer.

kythe/typescript/indexer.ts
ayazhafiz: Sure, I'll put this in another change.

kythe/typescript/testdata/object.ts
ayazhafiz: Done. Also added xrefs to computed properties.

kythe/typescript/indexer.ts
```diff
@@ -977,6 +979,10 @@ class Visitor {
         this.emitNode(vname, 'variable');
 
         this.emitEdge(this.newAnchor(decl.name), 'defines/binding', vname);
+
+        if (ts.isComputedPropertyName(decl.name)) {
+          this.visit(decl.name.expression);
+        }
```
evmar: Maybe we should just recurse with ts.forEachChild?  Not sure.

ayazhafiz:
@kythe could you review?

PR #3865, opened 2019-07-04
feature(typescript_indexer): index binding element parameters

ayazhafiz:
Currently we entirely skip over function parameters that are defined
with binding elements, like

```typescript
function foo({a, b: {c, d}}, [e, [f]]) { ... }
```

Implementing this is fairly straightforward. The existing parameter
visitor implementation is pulled out into its own method, and when we
encounter a set of parameter declarations they are recursively traversed
for their declarations in a binding expression.

Maybe ideally, these paramaters would be named `a: 0.0`, `c: 0.1.0`,
etc. to more accurately reflect how they are processed as arguments by
JavaScript/TypeScript. But Kythe doesn't support that and for the user
it makes more sense to think about them linearly.


ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -1006,7 +1006,7 @@ class Visitor {
 
   visitFunctionLikeDeclaration(decl: ts.FunctionLikeDeclaration) {
     this.visitDecorators(decl.decorators || []);
-    let kFunc: VName|undefined = undefined;
+    let kFunc: VName;
```
evmar: Good change!

kythe/typescript/indexer.ts
```diff
@@ -1117,7 +1086,76 @@ class Visitor {
...
+    }
+  }
+
+  /**
+   * Visits a function parameters, which can be recursive in the case of
+   * parameters created via bound elements:
+   *    function foo({a, b: {c, d}}, e, f) {}
+   * In this code, a, c, d, e, f are all parameters with increasing parameter
+   * numbers [0, 4]. Ideally the parameter number would be hierarchical, with
+   * `a` having a parameter number of 0.0, but Kythe doesn't support that.
```
evmar: Why is it idea to have these specific names?  It only needs to be consistent within this module, right?

kythe/typescript/indexer.ts
```diff
@@ -1117,7 +1086,76 @@ class Visitor {
...
+              }
+            default:
+              break;
+          }
+
+          if (ts.isParameter(param) && param.type) this.visitType(param.type);
+          if (param.initializer) this.visit(param.initializer);
+        }
+
+    for (const [, element] of toArray(parameters.entries())) {
```
evmar: Isn't parameters already an array?  Why not
for (const element of parameters) {
?

kythe/typescript/indexer.ts
ayazhafiz: Oops! Removed.

kythe/typescript/indexer.ts
ayazhafiz: yeah, i dont know why i did this. thanks!

PR #3867, opened 2019-07-04
feature(typescript_indexer): index computed function names

ayazhafiz:
Symbols now exist for computed names.

### DIFF
--- a/kythe/typescript/testdata/function.ts
+++ b/kythe/typescript/testdata/function.ts
@@ -32,18 +32,18 @@ function test(a: number, num: Num): Num {
 //- @test ref F
 test(3, {num: 3});
 
-
-// Test comoputed function names
+// Test computed function names by creating a constant string and checking that
+// a name computed from it is properly referenced to its computed definition.
 //- @#0"fn" defines/binding OFnStr
 const fn = 'fn';
 let o = {
-  //- @"['ofn']" defines/binding OFn
+  //- @"[fn]" defines/binding OFn
   //- OFn.node/kind function
   //- @fn ref OFnStr
   [fn]() {},
 };
 
-//- @ofn ref OFn
+//- @fn ref OFn
 o.fn();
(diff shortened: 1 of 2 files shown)


ayazhafiz:
cc @evmar

kythe/typescript/testdata/function.ts
```diff
@@ -32,6 +32,21 @@ function test(a: number, num: Num): Num {
 //- @test ref F
 test(3, {num: 3});
 
+
+// Test comoputed function names
```
creachadair:
```suggestion
// Test computed function names
```

It may also be worth explaining _how_ this tests computed function names.

kythe/typescript/testdata/function.ts
```diff
@@ -32,6 +32,21 @@ function test(a: number, num: Num): Num {
...
+//- @#0"fn" defines/binding OFnStr
+const fn = 'fn';
+let o = {
+  //- @"['ofn']" defines/binding OFn
```
creachadair: I can't see how this anchor could possibly match—there is no corresponding text.

ayazhafiz: Doh, that's on me. Thanks!

ayazhafiz:
> x has type 'foo', not string.

I see. That's pretty cool too.

ayazhafiz:
@evmar does this look good?

creachadair:
> The only reason I didn't LG before was the commit description, which shouldn't mention constant folding. I'm not sure what description github ends up merging, but please fix before submit.

It depends on how the merge is done. On Kythe we generally squash and merge, so whoever does the commit will be offered the concatenation of the PR description and the various commit logs to combine into the final merge log. (So having an accurate description helps—otherwise whoever does the merge has to guess).

ayazhafiz:
@kythe could you review?

PR #3868, opened 2019-07-04
chore(dev): add canonical import comments to each Kythe library package

creachadair:
We expect users to import Kythe Go packages via kythe.io import paths.  In support of this, we have a go-import redirector set up to allow "go get" to work with these import paths, and Bazel enforces this naming when building the Kythe repo itself.

Since we host our packages under github.com/kythe, however, it is also possible for users to import our packages directly via that location.

Following https://golang.org/doc/go1.4#canonicalimports, this commit adds canonical import comments to each Go library package in the Kythe repository. This has no effect on users building via Bazel or via the "go" tool with the recommended kythe.io import paths. For users who have vendored Kythe packages, however, it ensures that tools like "goimports" and IDE Go integrations will get the correct import paths.

This change was generated semi-mechanically by running

    go list ./kythe/go/...

at the top level of the Kythe repository, and then applying a keyboard macro to the file in each package that carries the package comment. Package main files are not affected.


jaysachs:
Useless rant:

Ah, Go. Whatever you do, keep it "simple" so you don't have to encode of build metadata anywhere other than the source files.

Why couldn't they allow the language to just explicitly say

  `package kythe.io/kythe/go/extractors/bazel`

and have that be interpreted as "canonical"? Let the language express the build system for realz, or don't bother at all.

creachadair:
> Why couldn't they allow the language to just explicitly say
> 
> `package kythe.io/kythe/go/extractors/bazel`
> 
> and have that be interpreted as "canonical"? Let the language express the build system for realz, or don't bother at all.

I don't disagree—but this is in no way specific to Go. No language currently extant provides enough context in the source file to correctly and reproducibly build. Not one. So at worst I think Go is special in having seen this pattern repeated several times and continuing it.

To this specific issue: Spelling out the full package name or even the URL doesn't solve the problem either. The package name isn't the real story here—it's that the build tool has to know how to map package names to repositories, and adding the comment (or other annotation) does nothing of value without that. The build tool is ~always invented later, and so this kind of business has to get shoved in someplace.

"Building" is still not recognized as an essential role of the language—it's viewed, tooled, and treated as a separate concern. But building is never orthogonal to the language, because dependencies are deeply embedded in the grammar and semantics. As such, you cannot have a build tool without knowledge of the language—even if that comes in via plugins and wrapper tools, as Bazel does it.

creachadair:
> Is this something that can be checked via go lint?

Not in the sense I think you mean: The Go `vet` tool will complain if the comments are present and you import these packages _not_ by the documented path. But it will not complain about the presence or absence of the comment itself. We could probably write something to check that a non-`main` package has such a comment in one of its files (maybe someone already did? I didn't look).

creachadair:
Rebased.

PR #3870, opened 2019-07-08
feat(typescript_indexer): emit "aliases" edge on aliased types

ayazhafiz:
If a type is user-defined and not a union, we can definitely emit an
"aliases" edge for it when it is aliased in a type definition, like
`type A = B`. Built-ins (like `number`) and unions are left untouched,
because they have zero and multiple derived VNames, respectively.


kythe/typescript/indexer.ts
```diff
@@ -638,11 +638,21 @@ class Visitor {
 
     if (decl.typeParameters) this.visitTypeParameters(decl.typeParameters);
     this.visitType(decl.type);
-    // TODO: in principle right here we emit an "aliases" edge.
-    // However, it's complicated by the fact that some types don't have
-    // specific names to alias, e.g.
-    //   type foo = number|string;
-    // Just punt for now.
+
+    // Emit an "aliases" edge if the aliased type is a singleton and named in
```
creachadair: I think the meaning of "singleton" here must be unusual: Does it matter that the type has only one element? Or is this a TypeScript-ism for a type that is not a union?

ayazhafiz: > Does it matter that the type has only one element?

Yeah, this is what I mean to say. Is there a better name for this?

creachadair: I don't know if there is a term of art for this. Perhaps "If the aliased type is a user-defined, named type that is not a union"?  User-defined rules out builtins.

"Singleton" usually refers to the members of the type, rather than the structure—but I have noticed that TS uses some other type-related terms in a slightly unusual way too, so I wasn't sure if this was another case of that.

ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -638,11 +638,20 @@ class Visitor {
 
     if (decl.typeParameters) this.visitTypeParameters(decl.typeParameters);
     this.visitType(decl.type);
-    // TODO: in principle right here we emit an "aliases" edge.
-    // However, it's complicated by the fact that some types don't have
-    // specific names to alias, e.g.
-    //   type foo = number|string;
-    // Just punt for now.
+    // Emit an "aliases" edge if the aliased type is a user-defined, named
+    // type that is not a union. This has exactly one symbol, whereas other
```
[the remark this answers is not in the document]
creachadair: Sorry, the wording of my suggestion may have been wrong—I was just riffing off the original, and I don't actually know what tycons TypeScript provides.

kythe/typescript/testdata/type_alias.ts
```diff
@@ -1,13 +1,17 @@
...
 //- @B defines/binding B
 //- B.node/kind talias
+//- B aliases A
 type B = A;
+
+//- @T1 defines/binding T1
+type T1 = Array<string>;
+//- @T2 defines/binding T2
+//- !{ T2 aliases T1 }
+type T2 = Array<string>;
```
evmar: Sorry, github ate my comment.  Here, I think we wanna verify that if given

```
type T1 = Array<string>;
type T2 = Array<number>;
```

T1 and T2 don't alias the same thing.  I think they might -- the above code makes them both alias Array because it ignores the generic in the vname generation.

kythe/typescript/indexer.ts
```diff
@@ -664,24 +664,28 @@ class Visitor {
...
    * a type are types (handled here).
+   *
    * @return the VName of the type, if available.  (For more complex types,
    *    e.g. Array<string>, we might not have a VName for the specific type.)
    */
   visitType(node: ts.Node): VName|undefined {
     switch (node.kind) {
+      case ts.SyntaxKind.TypeReference:
+        const ref = node as ts.TypeReferenceNode;
+        (ref.typeArguments || []).forEach(type => this.visitType(type));
```
evmar: I think if there are any type arguments we should return undefined, for the reasoning described below.

kythe/typescript/testdata/type_alias.ts
ayazhafiz: Yeah, they do point to the same thing. Let's see if we can change the VName at all.

ayazhafiz: Nevermind, I see. We just don't emit a reference at all.

ayazhafiz:
> Also I think the test needs to cover the
> type T1 = Array;
> type T2 = Array;
> case, to ensure we don't alias them together.

Resolved by making sure

```typescript
type T1 = Array<string>;
type T2 = Array<number>;
```

each do not alias any VName.

ayazhafiz:
@evmar can you review again?

kythe/typescript/indexer.ts
```diff
@@ -682,24 +682,38 @@ class Visitor {
...
    * a type are types (handled here).
+   *
    * @return the VName of the type, if available.  (For more complex types,
    *    e.g. Array<string>, we might not have a VName for the specific type.)
    */
   visitType(node: ts.Node): VName|undefined {
     switch (node.kind) {
+      case ts.SyntaxKind.TypeReference:
+        const ref = node as ts.TypeReferenceNode;
+        const kType = this.visitType((node as ts.TypeReferenceNode).typeName);
```
evmar: I think calling visitType on the typeName here will cause emits of the type without the generic, is that what we want?  Or should this bail entirely when there are type arguments?

I guess when you see "let x: Array<string>" we still want "Array" to be clickable and to go to the defn of array, without generics, so I think I convined myself that this is correct as written.

ayazhafiz: yeah it'll emit a ref for `Array` in `Array<string>` but bail on emitting anything for `Array<string>` as a whole.

ayazhafiz:
@kythe PTAL

PR #3871, opened 2019-07-08
feat(typescript_indexer): index export equals and import equals nodes

ayazhafiz:
Adds support for indexing `import =` and `export =` statements by
extending existing import and export logic. `export =` statements
are given a VName signature of `export =`, similar to how
`export default` statements are given signatures `default`.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -867,9 +867,8 @@ class Visitor {
     }
 
     if (!clause.namedBindings) {
-      // TODO: I believe clause.name or clause.namedBindings are always
-      // present, which means this check is not necessary, but the types don't
-      // show that.
+      // TODO: I believe clause.name or clause.namedBindings are always present,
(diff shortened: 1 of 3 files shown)


ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -383,7 +383,7 @@ class Visitor {
             // named 'default'.
             parts.push('default');
           } else {
-            this.todo(node, 'handle ExportAssignment with =');
+            parts.push('export =');
```
evmar: Do we care about putting spaces within the signature here?  Maybe make it 'export='?

ayazhafiz: :+1:

ayazhafiz:
@kythe could you review?

salguarnieri:
Is there a build problem?

ayazhafiz:
> Is there a build problem?

Resolved.

PR #3872, opened 2019-07-09
feat(typescript_indexer): emit "overrides" edges for overriden methods

ayazhafiz:
Fix a long-standing TODO. Emit overrides edges for methods that override
immediately extended class and implemented interfaces.


ayazhafiz:
I'm not sure if we should also emit "overrides" edges for methods in parent classes beyond the immediately extended/implemented class/interface.

ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -1065,15 +1065,39 @@ class Visitor {
...
+        //             ^^^^^^^^^-^^^^^^^^^^^^^^^----- `HeritageClause`s
+        // Look at each type listed in the heritage clauses and traverse its
+        // members. If the type has a member that matches the method visited in
+        // this function (`kFunc`), emit an "overrides" edge to that member.
+        if (funcSym && decl.parent.heritageClauses) {
+          for (const heritage of decl.parent.heritageClauses) {
+            for (const baseType of heritage.types) {
+              const baseSym = this.getSymbolAtLocation(baseType.expression);
+              if (!baseSym) {
+                this.todo(baseType, `heritage type ${baseSym} has no symbol`);
```
evmar: I think this maybe can occur when you have stuff like

class X extends class { ... } {
}

kythe/typescript/indexer.ts
```diff
@@ -1065,15 +1065,39 @@ class Visitor {
...
+                continue;
+              }
+              const funcName = funcSym.name;
+              const funcFlags = funcSym.flags;
```
evmar: What are the circumstances where the funcFlags are relevant?

kythe/typescript/indexer.ts
ayazhafiz: Yep, exactly.

kythe/typescript/indexer.ts
ayazhafiz: We want to make sure that we're matching the same kind of member, e.g. method to method and not method to property of the same name.

kythe/typescript/indexer.ts
```diff
@@ -1139,15 +1139,39 @@ class Visitor {
...
+              const overriden =
+                  toArray((baseSym.members || new Map()).values())
```
evmar: I read this like four times before I followed.

This might be easier to follow if the missing members thing was handled first, like:

```
baseSym.members ? toArray(...).find(...) : false;
```

or with an if statement.

kythe/typescript/indexer.ts
```diff
@@ -1139,15 +1139,39 @@ class Visitor {
...
+        // a class or interface.
+        //     class X extends A implements B, C {}
+        //             ^^^^^^^^^-^^^^^^^^^^^^^^^----- `HeritageClause`s
+        // Look at each type listed in the heritage clauses and traverse its
+        // members. If the type has a member that matches the method visited in
+        // this function (`kFunc`), emit an "overrides" edge to that member.
+        if (funcSym && decl.parent.heritageClauses) {
+          for (const heritage of decl.parent.heritageClauses) {
+            for (const baseType of heritage.types) {
+              const baseSym = this.getSymbolAtLocation(baseType.expression);
```
evmar: One case where this may happen is 

```
class X extends Array<string> {}
```

I forget whether generics get their own symbol or not.

I wonder whether it's worth marking this as extending just "Array" or something in that case?  But I think that also will do a very wrong thing in the case where it's some sort of mapped type, so maybe it's better to just leave that alone for now.

ayazhafiz: Generics don't get their own symbol. This would emit an overrides edge to something in Array I think.

ayazhafiz:
@kythe could you review?

PR #3876, opened 2019-07-10
feat(build): add generic kzip_archive rule

shahms:
Add a kzip_archive target which allows direct construction of a compilation unit from given inputs.  This should facilitate writing verifier_test rules for those languages which can use the generic extractor.

### DIFF
--- a/tools/build_rules/verifier_test/kzip_archive.bzl
+++ b/tools/build_rules/verifier_test/kzip_archive.bzl
@@ -1,0 +1,122 @@
+# Copyright 2019 The Kythe Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


tools/build_rules/verifier_test/kzip_archive.bzl
```diff
@@ -0,0 +1,122 @@
...
+    args.add("--uri", _kythe_uri(
+        corpus = ctx.attr.corpus,
+        language = ctx.attr.lang,
+        path = _single_path(ctx.files.srcs),
+        root = ctx.attr.root,
+    ))
+    args.add("--output", ctx.outputs.kzip)
+    args.add("--output_key", str(ctx.label))
+    args.add_joined("--source_file", ctx.files.srcs, join_with = ",")
+    args.add_joined("--required_input", ctx.files.deps, join_with = ",")
```
salguarnieri: does required_input also need to include srcs?

shahms: No, `kzip create` assumes that source files are also required input :-)

PR #3879, opened 2019-07-12
fix(typescript_indexer): log missing signature only if name is defined

ayazhafiz:
We currently log an error message for nodes not included in a VName
signature but that have a "name" attribute. What we really care about,
though, is if the node has a name value; for instance, function
expressions like `(function())` have a name attribute, but it is always
`undefined`.

This reduces superfluous error logging.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -521,7 +521,7 @@ class Visitor {
           // If the node is actually some subtype that has a 'name' attribute
           // it's likely this function should have handled it.  Dynamically
           // probe for this case and warn if we missed one.
-          if ('name' in (node as any)) {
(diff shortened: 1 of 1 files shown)


ayazhafiz:
cc @evmar

evmar:
Hm, in the function() case, shouldn't it still be a problem if we get here?  Because for function() we should've introduced an anonymous scope?  (From the perspective of the check, we're worried about if the thing we're looking at is the *kind* of thing that could have a name, not that the given instance has a name.)

Not sure of the above, what do you think?

ayazhafiz:
> Because for function() we should've introduced an anonymous scope?

Oh yes, you're totally right. I completely overlooked that. Okay, time to introduce an anon scope for this!

PR #3880, opened 2019-07-12
fix(typescript_indexer): add anon signature part for function expressions

ayazhafiz:
Function expressions like `(function())` have no name but introduce a
scope. Add an anonymous name for them when they are encountered in
generated a VName signature.

### DIFF
--- a/kythe/typescript/testdata/function.ts
+++ b/kythe/typescript/testdata/function.ts
@@ -92,7 +92,7 @@ function bindingTest({aa, bb: {cc, dd}}, [ee, [ff]]: NestedArray) {
 //- @an defines/binding An
 let ev, an;
 (function() {
-//- !{ @ev ref Ev }
+//- !{ @ev defines/binding _Ev2=Ev }
 let ev;
 //- @an ref An
 an;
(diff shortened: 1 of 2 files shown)


ayazhafiz:
cc @evmar

evmar:
I think you can add a test like

```
const x, y;
function() { const x; y;}
```

and checks that the xs are different and ys are same (maybe the latter isn't necessary).

ayazhafiz:
> I think you can add a test like
> 
> ```
> const x, y;
> function() { const x; y;}
> ```
> 
> and checks that the xs are different and ys are same (maybe the latter isn't necessary).

done.

kythe/typescript/testdata/function.ts
```diff
@@ -84,3 +86,14 @@ function bindingTest({aa, bb: {cc, dd}}, [ee, [ff]]: NestedArray) {
...
   dd = ff;
 }
+
+// Test function expressions introducing an anonymous scope
+//- @ev defines/binding Ev
+//- @an defines/binding An
+let ev, an;
+(function() {
+//- !{ @ev ref Ev }
+let ev;
```
evmar: This is defines/binding, not ref, so I think it's always true even without your change.

ayazhafiz: Fixed.

ayazhafiz:
@evmar does this look okay?

ayazhafiz:
@kythe PTAL

PR #3881, opened 2019-07-12
feature(typescript_indexer): add Kythe typings

ayazhafiz:
Add typings for Kythe. Ported from
[@types/kythe](https://www.npmjs.com/package/@types/kythe).

### DIFF
--- a/kythe/typescript/kythe.ts
+++ b/kythe/typescript/kythe.ts
@@ -1,6 +1,21 @@
+/*
+ * Copyright 2017 The Kythe Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 /**
- * @fileoverview TypeScript typings of the Kythe schema.
- * @author Kythe Authors
+ * @fileoverview TypeScript types and values of the Kythe schema.
  */
 
 /**
--- a/kythe/typescript/kythe.ts
+++ b/kythe/typescript/kythe.ts
@@ -1,4 +1,9 @@
 /**
+ * @fileoverview TypeScript typings of the Kythe schema.
+ * @author Kythe Authors
+ */
+
+/**
  * A VName (Vector Name) for a node in the Kythe schema consists of:
  *  - `signature`: a unique, opaque signature for a node
  *  - `corpus`: a collection of related files the node is defined in
@@ -76,7 +81,7 @@ export type OrdinalEdge = string&{
 };
 export function makeOrdinalEdge(edge: EdgeKind, ordinal: number): OrdinalEdge {
   const ordinalEdge = `${edge}.${ordinal}`;
-  return <OrdinalEdge>ordinalEdge;
+  return ordinalEdge as OrdinalEdge;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ayazhafiz:
cc @evmar 

I put a function to create ordinal edges in the typings file, so it's not true `.d.ts`. Let me know if that should moved elsewhere.

kythe/typescript/kythe.ts
```diff
@@ -0,0 +1,206 @@
+/**
```
evmar: Add a @fileoverview at the top describing what this is.  Which I think is something like "exposes to TS the types and values from the kythe schema", or some such.

kythe/typescript/kythe.ts
```diff
@@ -0,0 +1,206 @@
...
+ * A VName (Vector Name) for a node in the Kythe schema consists of:
+ *  - `signature`: a unique, opaque signature for a node
+ *  - `corpus`: a collection of related files the node is defined in
+ *  - `root`: a label denoting a distinct subset of the corpus
+ *  - `path`: the relative path of the file containing the node
+ *  - `language`: programming language the node belongs to
+ */
+export interface VName {
```
evmar: Shouldn't this be used by indexer.ts?

kythe/typescript/kythe.ts
```diff
@@ -0,0 +1,206 @@
...
+ *   `${EdgeKind}.${number}`
+ * This is represented as a branded string that is incompatible with a string
+ * but can be compared to a string.
+ */
+export type OrdinalEdge = string&{
+  __ordinalBrand: 'ordinal';
+};
+export function makeOrdinalEdge(edge: EdgeKind, ordinal: number): OrdinalEdge {
+  const ordinalEdge = `${edge}.${ordinal}`;
+  return <OrdinalEdge>ordinalEdge;
```
evmar: I think
return ordinalEdge as OrdinalEdge;
The brackets syntax is obsolete

kythe/typescript/kythe.ts
```diff
@@ -0,0 +1,206 @@
...
+ * A Fact is an Entry that also has a fact `value`.
+ */
+export interface Fact extends Entry {
+  value: string;
+}
+
+/**
+ * An Edge is an Entry that also has a `target` and an edge `kind`.
+ */
+export interface Edge extends Entry {
```
evmar: Should these be used by indexer.ts too?

ayazhafiz: Yes, I'm going to add that in a separate PR.

kythe/typescript/kythe.ts
ayazhafiz: Yes, going to add that in a separate PR.

kythe/typescript/kythe.ts
ayazhafiz: Done.

kythe/typescript/kythe.ts
ayazhafiz: Done.

kythe/typescript/kythe.ts
```diff
@@ -0,0 +1,211 @@
+/**
+ * @fileoverview TypeScript typings of the Kythe schema.
```
evmar: 'typings' has a specific meaning in TS, it's about defining types for a JS library, which I think doesn't apply here.

kythe/typescript/kythe.ts
```diff
@@ -0,0 +1,211 @@
...
+ * @author Kythe Authors
```
evmar: I think this is can be omitted.

Now that I think of it, I think you need to copy the license block into this file too.
So like

/*
 * Copyright [paste from other file ]
 */

/** @fileoverview TypeScript definitions of types found in the Kythe schema. */

ayazhafiz: Done.

kythe/typescript/kythe.ts
ayazhafiz: Fixed.

ayazhafiz:
@kythe PTAL

PR #3882, opened 2019-07-12
fix(typescript_indexer): don't throw on unfound module path

ayazhafiz:
This broke development downstream. Emit a todo rather throw an error
that crashes the process.


ayazhafiz:
cc @evmar

PR #3883, opened 2019-07-13
refactor(typescript_indexer): cleanup module path resolution

ayazhafiz:
We were handling module path resolution correctly previously, but doing
it in a confusing way. Since we only care about paths that define an
entire module, just query to see if the module reference is an entire
source file. If so, return the module path to that source file;
otherwise, we can bail because there is no unique path to reference.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -753,17 +753,15 @@ class Visitor {
    * getPathFromModule(the './foo' node) might return a string like
    * 'path/to/project/foo'.  See this.moduleName().
    */
-  getModulePathFromModuleReference(sym: ts.Symbol): string {
-    const name = sym.name;
-    // TODO: this is hacky; it may be the case we need to use the TypeScript
-    // module resolver to get the real path (?).  But it appears the symbol
-    // name is the quoted(!) path to the module.
-    // TODO(ayazhafiz): Fix the above.
-    if (!(name.startsWith('"') && name.endsWith('"'))) {
-      this.todo(sym.declarations[0], `TODO: handle module symbol ${name}`);
+  getModulePathFromModuleReference(sym: ts.Symbol): string|undefined {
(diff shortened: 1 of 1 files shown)


ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -753,17 +753,15 @@ class Visitor {
...
-  getModulePathFromModuleReference(sym: ts.Symbol): string {
-    const name = sym.name;
-    // TODO: this is hacky; it may be the case we need to use the TypeScript
-    // module resolver to get the real path (?).  But it appears the symbol
-    // name is the quoted(!) path to the module.
-    // TODO(ayazhafiz): Fix the above.
-    if (!(name.startsWith('"') && name.endsWith('"'))) {
-      this.todo(sym.declarations[0], `TODO: handle module symbol ${name}`);
+  getModulePathFromModuleReference(sym: ts.Symbol): string|undefined {
+    const sf = sym.valueDeclaration;
```
evmar: Does valueDeclaration always exist/work, even for type-only modules?

ayazhafiz: What is a type only module? .d.ts files?

evmar: Sorry for being vague.  When I see "valueDeclaration" it makes me think of how a given symbol normally has a 0-1 value declarations and 0-N type declarations, and I was wondering if it's possible here for the value declaration to not exist.  It's weird because this is a module object, but normally isn't it possible for a symbol to not have a value declaration?

ayazhafiz: Apparently not. The types suggest that a symbol always has a value declaration.

danielmoy-google:
Sorry there are conflicts before this can be merged

ayazhafiz:
> Sorry there are conflicts before this can be merged

Resolved.

PR #3884, opened 2019-07-14
refactor(typescript_indexer): integrate Kythe types in indexer

ayazhafiz:
Add Kythe types from `kythe.ts` to the indexer. In cases where custom
Kythe facts and edges are used, just cast them to a `FactName` or
`EdgeKind`.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -758,8 +758,9 @@ class Visitor {
     // TODO: this is hacky; it may be the case we need to use the TypeScript
     // module resolver to get the real path (?).  But it appears the symbol
     // name is the quoted(!) path to the module.
+    // TODO(ayazhafiz): Fix the above.
     if (!(name.startsWith('"') && name.endsWith('"'))) {
-      throw new Error(`TODO: handle module symbol ${name}`);
+      this.todo(sym.declarations[0], `TODO: handle module symbol ${name}`);
     }
     const sourcePath = name.substr(1, name.length - 2);
     return this.host.moduleName(sourcePath);
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -315,6 +315,11 @@ class Visitor {
     this.emitFact(source, FactName.NODE_KIND, kind);
   }
 
+  /** emitSubkind emits a new fact entry, declaring the subkind of a VName. */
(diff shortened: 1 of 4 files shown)


ayazhafiz:
cc @evmar

kythe/typescript/indexer.ts
```diff
@@ -308,30 +303,36 @@ class Visitor {
...
     });
   }
 
   /** emitEdge emits a new edge entry, relating two VNames. */
-  emitEdge(source: VName, name: string, target: VName) {
+  emitEdge(source: VName, name: EdgeKind|OrdinalEdge, target: VName) {
     this.host.emit({
       source,
-      edge_kind: '/kythe/edge/' + name,
+      edge_kind: name,
```
evmar: This is weird, should the parameter be 'kind'?

kythe/typescript/indexer.ts
```diff
@@ -1106,7 +1117,7 @@ class Visitor {
     if (vname && decl.kind === ts.SyntaxKind.PropertyDeclaration) {
       const declNode = decl as ts.PropertyDeclaration;
       if (this.isStaticMember(declNode, declNode.parent)) {
-        this.emitFact(vname, 'tag/static', '');
+        this.emitFact(vname, '/kythe/tag/static' as FactName, '');
```
evmar: Should this be in kythe.ts?

kythe/typescript/test.ts
```diff
@@ -146,7 +147,7 @@ async function testPlugin() {
         };
         context.emit({
           source: pluginMod,
-          fact_name: '/kythe/node/pluginKind',
+          fact_name: '/kythe/node/pluginKind' as kythe.FactName,
```
evmar: Should this be in kythe.ts?

kythe/typescript/indexer.ts
ayazhafiz: This is a non-standard Kythe fact name. We can add it though.

kythe/typescript/test.ts
ayazhafiz: This definitely shouldn't be because it is only used for testing

kythe/typescript/indexer.ts
ayazhafiz: yes

ayazhafiz:
@evmar does this look good to you?

PR #3888, opened 2019-07-16
feat(typescript_indexer): add minimal support for JSX

ayazhafiz:
Adds minimal support for JSX attributes.


ayazhafiz:
cc @evmar

evmar:
As always, tests please :)

ayazhafiz:
> As always, tests please :)

Oops! Sorry. Fixed.

kythe/typescript/testdata/tsx.tsx
```diff
@@ -0,0 +1,11 @@
+// Test TypeScript JSX
+
```
evmar: I think add an export {}; to this, per README

kythe/typescript/testdata/tsx.tsx
```diff
@@ -0,0 +1,11 @@
...
+//- @#0"value" defines/binding Value
+const value = 'value';
+function render() {
+  return (
+    //- @#0"value" defines/binding vname("render.value", _, _, _, _)
```
evmar: This is weird, how is this defining a binding?

ayazhafiz: Because this "value" is now a property on the element returned by this function, so we can treat it as variable-like.

evmar: Won't this conflict with a local named 'value' within the render function?

ayazhafiz: Yes. You're totally right. Fixed.

This supports a very limited subset of tsx, but i don't think we should support much other than what customers downstream use, which is what's been added here.

kythe/typescript/testdata/tsconfig.for.tests.json
```diff
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "module": "commonjs",
         "target": "es6",
+	"jsx": "preserve",
```
evmar: Something weird with tabs vs spaces here perhaps

kythe/typescript/testdata/tsx.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export {};
+
+function render() {
+  //- @#0"value" defines/binding Value
+  const value = 'value';
+  return (
+    //- @"attr" defines/binding vname("render.div.attr", _, _, _, _)
+    //- @"value" ref Value
+    //- @+3"src" defines/binding vname("render.div.img.src", _, _, _, _)
```
evmar: I'm not sure this is right.  Look at what the output is to see what these attributes actually do semantically:

http://www.typescriptlang.org/play/?jsx=2#code/GYVwdgxgLglg9mABAJwKZgCauQCgJSIDeAUIiqlCMkjqWYgDwYwBuiAhlFMgLyEvsANiFQBfAHx16jGAFsA5ogDOyCHwHCxiAPSTpDbcxZ7EeANzFRQA

I think we should treat 'src' and 'attr' in the same way we treat keys in objects.

ayazhafiz: What do you mean by that? We currently visit JSX attributes as variable declarations, same as we do property declarations

evmar: I the attributes effectively become a `{ src: ... }` object literal.  So I think your naming here won't work if there are two of the same tags as siblings, in a layout like

```
<div>
  <img src={a}/>
  <img src={b}/>
</div>
```

I think maybe we need to use anonymous name generation for these, and ignore the tag names entirely.

ayazhafiz: Fixed by creating an anonymous names for all JSX nodes. This is not needed for some nodes, but reduces code complexity since we don't prioritize JSX anyway.

PR #3890, opened 2019-07-16
feat: Add commands necessary for tensorflow builds

salguarnieri:
tensorflow needs external setup before it can run a bazel build. It
needs to have certain Python pips present and it needs to run a
configure script before builds can be done. See https://www.tensorflow.org/install/source.

### DIFF
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -54,6 +54,22 @@ if [ -n "$KYTHE_SYSTEM_DEPS" ]; then
   apt-get clean
 fi
 
+if [ -n "$KYTHE_PIP_DEPS" ]; then
+  pip install -U $KYTHE_PIP_DEPS
+fi
+
+if [ -n "$KYTHE_PIP_NO_REC_DEPS" ]; then
+  pip install -U $KYTHE_PIP_NO_REC_DEPS --no-deps
+fi
+
+# Always use bazelisk for bazel so we have complete control over the bazel
+# version being used.
+ln -s /kythe/bazelisk /usr/local/bin/bazel
+
+if [ -n "$KYTHE_PRE_BUILD_STEP" ]; then
+  eval "$KYTHE_PRE_BUILD_STEP"
+fi
+
 /kythe/bazelisk --bazelrc=/kythe/bazelrc "$@"
 
 # Collect any extracted compilations.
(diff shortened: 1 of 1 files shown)


kythe/extractors/bazel/extract.sh
```diff
@@ -54,6 +54,22 @@ if [ -n "$KYTHE_SYSTEM_DEPS" ]; then
...
 fi
 
+if [ -n "$KYTHE_PIP_DEPS" ]; then
+  pip install -U $KYTHE_PIP_DEPS
+fi
+
+if [ -n "$KYTHE_PIP_NO_REC_DEPS" ]; then
+  pip install -U $KYTHE_PIP_NO_REC_DEPS --no-deps
+fi
+
```
jaysachs: In an ideal world, all this would be taken care of by the "KYTHE_PRE_BUILD_STEP" below. Just exactly how messy/impossible would that be? I'm not certain we want to elevate "pip" up to the level of apt.

salguarnieri: It would just be a long command in the text proto setup for tensorflow, so not too bad as long as it isn't changed often. This isn't the first time I've heard of pip though, so I thought it might be near the level of apt for python development.

@shahms Do you have an idea on how standard pip is for python?

justbuchanan: `pip` is pretty much the standard for python packaging, but on second thought, I'd be inclined to just put the `pip install` as in the pre-build step as @jaysachs said.

Longer term, I'd like our extraction setup to know less about the repo's build requirements. Instead of having the build take place inside of the kythe extraction container, we'd have kythe extraction take place inside of the project's build container.

jaysachs: I'm also thinking that maybe we could ask the TF team to submit a "pre_kythe.sh" (or whatever) script into the repository. (Maybe in a canonical location?)

jaysachs: We could possibly look at pushing the system deps into the pre-build step as well. Sometime later. The more we formalize that step (e.g. canonical location in repo?) the closer we get to Justin's articulated ideal of extracting inside the project's build.

salguarnieri: Works for me. I'll test it out after vacation and update this PR when I have a working proof of concept again.

salguarnieri: I removed the pip steps and it looks like it works locally. The pre-bazel command isn't very pretty at the moment but it works. It will be much more manageable when it can just call a pre-existing shell script in the host repo.

PR #3891, opened 2019-07-16
fixup: fail extraction after first failed command

salguarnieri:
Print out the executed commands and exit the script after the first failed
command to make debugging easier and to help avoid silent failures in an early
step that cause unexpected failures in a later step.

### DIFF
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -61,9 +61,7 @@ fi
 # It is ok if targets fail to build. We build using --keep_going and don't
 # care if some targets fail, but bazel will return a failure code if any
 # targets fail.
-set +e
-/kythe/bazelisk --bazelrc=/kythe/bazelrc "$@"
-set -e
+/kythe/bazelisk --bazelrc=/kythe/bazelrc "$@" || true
 
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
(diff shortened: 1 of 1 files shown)


kythe/extractors/bazel/extract.sh
```diff
@@ -43,6 +43,10 @@
 # Also assumes you have extractors installed as per
 # kythe/extractors/bazel/extractors.bazelrc.
 
+# Print our commands for easier debugging and exit after our first failed
+# command so we avoid silent failures.
+set -ex
+
```
jaysachs: I'm pretty sure bazel (and bazelisk) exit with non-zero code if any targets fail. But we want to tolerate some targets failing. Or at least our builds depend on that right now.

justbuchanan: Good point. I thought we were ok since we use `--keep_going` when extracting, but that does still give a nonzero exit code if any part of the build failed.

salguarnieri: I updated to turn off e for the bazel invocation

kythe/extractors/bazel/extract.sh
```diff
@@ -54,7 +58,12 @@ if [ -n "$KYTHE_SYSTEM_DEPS" ]; then
   apt-get clean
 fi
 
+# It is ok if targets fail to build. We build using --keep_going and don't
+# care if some targets fail, but bazel will return a failure code if any
+# targets fail.
+set +e
 /kythe/bazelisk --bazelrc=/kythe/bazelrc "$@"
+set -e
```
jaysachs: `bazelisk --bazelrc=... "$@" || true` might be more idiomatic.

salguarnieri: done

kythe/extractors/bazel/extract.sh
salguarnieri: done

PR #3899, opened 2019-07-23
feat(typescript_indexer): add support for indexing literal properties

ayazhafiz:
Add support for indexing string and numeric literals in property names,
and referencing string and numeric literals in a property access.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -464,11 +464,11 @@ class Visitor {
                   part = decl.name.text;
                 }
                 // Wrap literals in quotes, so that characters used in other
-                // signatures do not interfere with the signature creates by a
+                // signatures do not interfere with the signature created by a
                 // literal. For instance, a literal
                 //   obj.prop
                 // may interefere with the signature of `prop` on an object
-                // `foo`. The literal receives a signature
+                // `obj`. The literal receives a signature
                 //   "obj.prop"
                 // to avoid this.
                 if (ts.isStringLiteral(decl.name)) {
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -463,6 +463,17 @@ class Visitor {
                 } else {
                   part = decl.name.text;
                 }
+                // Wrap literals in quotes, so that characters used in other
+                // signatures do not interfere with the signature creates by a
+                // literal. For instance, a literal
+                //   obj.prop
+                // may interefere with the signature of `prop` on an object
+                // `foo`. The literal receives a signature
+                //   "obj.prop"
+                // to avoid this.
+                if (ts.isStringLiteral(decl.name)) {
+                  part = `"${part}"`;
+                }
                 // Instance members of a class are scoped to the type of the
                 // class.
                 if (ts.isClassDeclaration(decl) && lastNode !== undefined &&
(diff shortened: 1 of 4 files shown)


ayazhafiz:
cc @evmar

kythe/typescript/testdata/object.ts
```diff
@@ -21,13 +21,27 @@ const Object = {
 
   //- @method defines/binding Method
   //- Method.node/kind function
-  method() {}
+  method() {},
+
+  //- @"'sliteral'" defines/binding SLiteralProperty
+  //- SLiteralProperty.node/kind variable
+  'sliteral': 0,
```
evmar: What if you have something like

```
  'sliteral#type': 0
```

I think our name generator needs to escape characters that aren't ascii or something.

ayazhafiz: No, I think that's okay too. Kythe escapes such characters when processing it internally if that's what you're worried about, but for names it's fine. See the new test I pushed.

kythe/typescript/testdata/object.ts
```diff
@@ -21,13 +21,27 @@ const Object = {
...
+  //- NLiteralProperty.node/kind variable
+  123: 'nliteral',
 };
 
 //- @property ref Property
 //- @shortProperty ref ShortProperty
 //- @computed ref ComputedProperty
 const x = Object.property || Object.shortProperty || Object.computed;
 
+//- @"'string#literal'" ref SLiteralProperty
```
evmar: What I'm worried about is that this will generate a VName with signature
  `whatever.string#literal`
and meanwhile some other code will generate signatures that have other characters at the end with a different meaning, like `#type`.

For another example of this worry, imagine things like
  obj['string.literal'].foo
vs
  obj.string.literal.foo

ayazhafiz: Okay, I see what you're saying. What if we wrap literals in quotes?

evmar: Yeah that should be fine, I think.

ayazhafiz: Fixed.

kythe/typescript/indexer.ts
```diff
@@ -463,6 +463,17 @@ class Visitor {
                 } else {
                   part = decl.name.text;
                 }
+                // Wrap literals in quotes, so that characters used in other
+                // signatures do not interfere with the signature creates by a
```
zzmp: nit: s/creates/created

kythe/typescript/indexer.ts
```diff
@@ -463,6 +463,17 @@ class Visitor {
...
+                // literal. For instance, a literal
+                //   obj.prop
+                // may interefere with the signature of `prop` on an object
+                // `foo`. The literal receives a signature
```
zzmp: Wouldn't it interfere with an object `obj`, not `foo`?

kythe/typescript/indexer.ts
ayazhafiz: fixed

kythe/typescript/indexer.ts
ayazhafiz: fixed

ayazhafiz:
@evmar / @kythe can you add the buildbot tag?

ayazhafiz:
@evmar / @kythe can we merge?

PR #3905, opened 2019-07-24
feat(typescript_indexer): expose utf offset tables and getSymbolName to IndexerHost

zzmp:
- Adds support for UTF-16 offset lookup (from UTF-8).
- Exposes `getOffsetTable` through the `IndexerHost`. `Plugin`s can now access it.
- Exposes `getSymbolName` through the `IndexerHost`. `Plugin`s can now access it.
  - Adds `setSymbolNames` so that the `Visitor` can still sync remote `Symbol`s.

This exposes more functionality to `Plugin`s: looking up byte offsets and looking up canonical symbols. This avoids duplication in the `Plugin`, and allows `Plugin`s to integrate better with the existing indexer.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -31,6 +31,12 @@ const LANGUAGE = 'typescript';
  */
 export interface IndexerHost {
   /**
+   * Gets the offset table for a file path.
+   * These are used to lookup UTF-8 offsets (used by Kythe) from UTF-16 offsets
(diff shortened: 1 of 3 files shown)


zzmp:
cc @evmar 
cc @ayazhafiz

ayazhafiz:
Can we break this up into two PRs? One for the offset table stuff and one for `getSymbolName`.

zzmp:
See #3908 #3909.

PR #3906, opened 2019-07-25
Comment localhost:<port> vs :<port> for server "listen" specifications

kamahen:
This prevents some "connection refused" (ECONNREFUSED) errors when connecting
from another machine.


kamahen:
Fixes https://github.com/kythe/kythe/issues/3907

creachadair:
Not accepting connections from other machines is probably the correct default for example code that people are likely to copy and paste: The service does no authentication, so exposing a port on the wildcard interface is usually _not_ what you want.

I'd suggest a better solution here is to replace `BROWSE_PORT` with `BROWSE_ADDR`, with a default set to `localhost:8080` and a comment to let the reader know that if they really _do_ want to open up their service to other machines, they can swap out `localhost` for the wildcard `0.0.0.0`.

kamahen:
The Ubuntu default is for the firewall `ufw` blocking access from outside, so does using `localhost` for the server significantly increase security?

I've done a lot of TCP server programming, and at one time had practically memorized the Stevens books (this was a _long_ time ago), yet it still took me a bit to figure out that I needed to change `localhost` to `0.0.0.0`, partly because I was unfamiliar with the Go socket API, partly because I had forgotten that `INADDR_ANY` is 0, and partly because I had to figure out where the `ECONNREFUSED` was happening. Someone who hasn't done much TCP programming would probably have done a lot more head scratching.

Does anyone else have an opinion on change `localhost` to `0.0.0.0` vs add comments. (FWIW, typical example code uses `INADDR_ANY` or `0.0.0.0`, e.g. https://www.paulgriffiths.net/program/c/sockets.php which looks a lot like the Stevens code as I distantly remember it.)

kamahen:
PS: I don't mind following @creachadair's suggestion; I simply wish to see if there are other opinions before I make the changes.

creachadair:
I agree with your concern about people not knowing about the wildcard address, that's why I suggested we should add a comment to tell them what to do. Perhaps something like:

```shell
# By default, listen only on the local host. If you want other computers to be able to connect
# to your server, use "0.0.0.0" instead of "localhost".
BROWSE_ADDR="${BROWSE_ADDR:-locahost:8080}"
```

or words to that effect. Firewalls certainly help, but I have observed a lot of personal computers without firewalls set up (or set up poorly), and 8080 is a common "alternate HTTP" port that application level firewalls often exempt.

I don't think the risk is that huge, certainly, but I've seen enough copy-pasted config settings to be cautious about defaults. That said, if others don't think it's a problem, I'm fine with that.

kamahen:
It turns out that the `-listen` flag can be `:8080`, so I'll change the `0.0.0.0`s to the empty string. BTW, that's the example in the Go documentation for [ListenAndServe](https://golang.org/pkg/net/http/#ListenAndServe), so there must be a lot of copy-and-paste errors out in the world, if this is an error. ;)

If everyone agrees with my change, I'll add a comment about using `localhost` to restrict access.

shahms:
Context matters.  Referencing the Go documentation for a facilities which is used as an example to serve external requests is not really applicable to application defaults in other areas.  This is particularly true for the languageserver which is intended to run on the same machine as the editor with which it is communicating and should most assuredly not accept external connections by default.

kamahen:
Not to derail this conversation, but I'm curious why languageserver needs to be on the same machine as the editor -- is it a performance thing? Or a security thing because of no authentication?

Anyway, I'm fine with changing my PR for languageserver (again, with a comment), if that's what you want.

BTW, I'm not sure what the implications are of using locahost:0 for a multi-homed machine and I don't have an easy way to experiment. I don't recall reading any discussion of this field for _servers_ (e.g., in Stevens) ... typically the field is used for a _client_ connecting to a server but a something needs to be filled in for servers (because `struct sockaddr` is used by both `connect` and `bind`), so `INADDR_ANY` is used for servers. (Yet another reason to hate the Berkely socket API, but I digress.)

creachadair:
> Not to derail this conversation, but I'm curious why languageserver needs to be on the same machine as the editor -- is it a performance thing? Or a security thing because of no authentication?

The [spec](https://microsoft.github.io/language-server-protocol) doesn't discuss the rationale, but I think it's mostly intende to reduce coupling and complexity for the client. A language server can use secure and authenticated connections to remote tools, but the local client editor only needs to speak LSP over a socket to plug in. The language server can also do smart things with caching, which can help performance too, but I don't know whether that was a primary design consideration.

kythe/go/languageserver/bin/main.go
```diff
@@ -37,7 +37,7 @@ import (
 var (
 	pageSize = flag.Int("page_size", 0, "Set the default xrefs page size")
 
-	serverAddr = flag.String("server", "localhost:8080",
+	serverAddr = flag.String("server", "0.0.0.0:8080",
```
schroederc: Default should remain `localhost`.  Security by default.  If anything, add documentation to use `:8080` if they want to expose the server to the wider network.

kamahen:
I did a quick look through the spec and some of the implementations but couldn't spot where they configured the servers (either to `localhost` or `0.0.0.0`).
I did find this: https://en.wikipedia.org/wiki/Language_Server_Protocol#Transport_layer:

> The protocol does not make any provisions about how requests, responses and notifications are transferred between client and server. For example, client and server could be components within the same process exchanging JSON strings via method calls. They could also be different processes on the same or on different machines communicating via network sockets.

And I found https://code.visualstudio.com/api/advanced-topics/remote-extensions which seems to use a proxy from localhost to a remote host rather than directly connecting to a remote host. There's something about authentication, etc.

(Enough time spent on this, although it's been somewhat educational.) 
It seems reasonable for languageserver to restrict itself to `localhost` ... as for the other servers, is it OK for them to use `0.0.0.0`?

kythe/go/languageserver/bin/main.go
kamahen: Done

kythe/examples/bazel/kythe-index.sh
```diff
@@ -31,8 +31,9 @@ Paths to output artifacts are controlled with the --graphstore and --serving_tab
 defaults are $TMPDIR/gs.bazel and $TMPDIR/serving.bazel, respectively.
 
 The --serving_addr flag controls the listening address of the Kythe http_server.
-Its default is 0.0.0.0:8888.  Any --webui path provided will be passed to the
-http_server as its --public_resources flag.
+Its default is :8888.  (localhost:8888 allows access from only the same machine
```
shahms: localhost should be the default, the note should suggest how to enable access from additional hosts (this applies to the other locations as well).

If you want to change this default locally, writing a local wrapper is simple, but I'd far rather avoid providing instructions and defaults to users which open up a port on the local machine to the outside world.

kamahen:
I've changed everything to localhost. However ...

On my machine, specifying `--listen=localhost:8080` allows access from the same machine using `telnet 127.0.0.1 8080` but doesn't work with `telnet ::1 8080`; if I specify `--listen=:8080`, then both `127.0.0.1` and `::1` work (also `localhost`).

So, I think it's a mistake to default to `localhost`, but the decision is yours.

kythe/go/serving/tools/http_server/http_server.go
```diff
@@ -45,7 +45,7 @@ import (
 var (
 	servingTable = flag.String("serving_table", "", "LevelDB serving table")
 
-	httpListeningAddr = flag.String("listen", "localhost:8080", "Listening address for HTTP server")
+	httpListeningAddr = flag.String("listen", "localhost:8080", "Listening address for HTTP server (":<port>" allows access from any machine)")
```
shahms: The embedded quotes need to be escaped.

kamahen: Fixed. (I must have rebuilt everything in the wrong branch ...)

kythe/go/languageserver/bin/main.go
kamahen: "localhost:8080" blocks access using `telnet ::1 8080` but allows access `telnet 127.0.0.1 8080`. `nc -v ::1 8080` also failed with connection refused.

PR #3909, opened 2019-07-25
feat(typescript_indexer): expose getSymbolName through IndexerHost

zzmp:
Exposes getSymbolName through the IndexerHost. Plugins can now access it.
Adds setSymbolNames so that the Visitor can still sync remote Symbols.


zzmp:
cc @evmar 
cc @ayazhafiz

kythe/typescript/indexer.ts
```diff
@@ -30,6 +30,25 @@ const LANGUAGE = 'typescript';
  * code duplication.
  */
 export interface IndexerHost {
+  /**
+   * Computes the VName (and signature) of a ts.Symbol. A Context can be
```
ayazhafiz: nit: "(and signature)" is redundant

kythe/typescript/indexer.ts
```diff
@@ -30,6 +30,25 @@ const LANGUAGE = 'typescript';
...
+   * optionally specified to help disambiguate nodes with multiple declarations.
+   * See the documentation of Context for more information.
+   */
+  getSymbolName(sym: ts.Symbol, ns: TSNamespace, context?: Context): VName;
+  /**
+   * Sets the VNames of a ts.Symbol. Throws if a VName was already set.
+   */
+  setSymbolNames(sym: ts.Symbol, names: [VName|null, VName|null, VName|null]):
```
ayazhafiz: This could still lead to symbol pollution, as we discussed yesterday. I think the indexing host should be source of truth, with only unidirectional communication from the host to the indexer. `getSymbolName` can cache symbols.

kythe/typescript/indexer.ts
```diff
@@ -243,121 +224,98 @@ class Visitor {
   /** A shorter name for the rootDir in the CompilerOptions. */
   sourceRoot: string;
```
ayazhafiz:
```suggestion
private  sourceRoot: string;
```

kythe/typescript/indexer.ts
```diff
@@ -243,121 +224,98 @@ class Visitor {
...
 
+  /**
+   * rootDirs is the list of rootDirs in the compiler options, sorted
+   * longest first.  See this.moduleName().
+   */
+  rootDirs: string[];
+
   typeChecker: ts.TypeChecker;
```
ayazhafiz:
```suggestion
  private typeChecker: ts.TypeChecker;
```

kythe/typescript/indexer.ts
```diff
@@ -243,121 +224,98 @@ class Visitor {
   /** A shorter name for the rootDir in the CompilerOptions. */
   sourceRoot: string;
 
+  /**
+   * rootDirs is the list of rootDirs in the compiler options, sorted
+   * longest first.  See this.moduleName().
+   */
+  rootDirs: string[];
```
ayazhafiz:
```suggestion
  private rootDirs: string[];
```

kythe/typescript/indexer.ts
```diff
@@ -543,83 +501,153 @@ class Visitor {
...
 
-    return vname;
+  constructor(
+      private readonly host: IndexerHost,
+      private file: ts.SourceFile,
+  ) {
+    this.sourceRoot =
+        this.host.program.getCompilerOptions().rootDir || process.cwd();
+
+    this.typeChecker = this.host.program.getTypeChecker();
```
ayazhafiz: I think the visitor should defer entirely to the host now that the host owns a typechecker. `getSymbolAtLocation` can be a method inside the host as well. Two type checkers floating around could get confusing, and the indexer has previously tried to hide the existence of a type checker so that callers of `getSymbolAtLocation` don't need to know it exists at all.

zzmp: The typechecker is owned by the program. The typecheckers in the host and the visitor are just convenience aliases of this. I'm fine with moving `getSymbolAtLocation`, but I don't think its necessary to clean up other uses of the typechecker.

kythe/typescript/indexer.ts
zzmp: There needs to be a way to alias symbols, because that is what the `Visitor` currently does.
How do you feel about the `aliasSymbol` method that I replaced this with? I can also add a `throw` if you try to alias an already set `Symbol`, although I'm not sure that's necessary.

zzmp:
Sorry about the force-push mid-review. It was not building, and so I took the opportunity to re-org into multiple commits to try to make the changes easier to review.

kythe/typescript/indexer.ts
ayazhafiz: Well, if you move `getSymbolAtLocation` into the indexing host you won't need to expose the `typechecker` in the visitor since the host also owns the program.

kythe/typescript/indexer.ts
ayazhafiz: How do you feel about keeping a set of aliased symbols in the Visitor itself that we hit before going to `getSymbolNames`? This would keep the aliasing local to the TypeScript indexer itself.

If this is too much work (or too nitpicky) I think it's okay, as we can move that later and we will probably refactor `visitImport` where this method is used; if I remember correctly Evan wasn't always pleased with the way that imports are handled.

Do you expect to be using `aliasSymbol` for the TS->proto work?

zzmp: I think we want the aliased symbols to be in the symbolNames table, so that they are available (by their aliases) to downstream plugins.
I don't expect to use `aliasSymbol` for the TS->proto work, but I do expect to consume symbols that are aliased, so I don't want the aliases to be scoped to a `Visitor`.
I considered subclassing `IndexerHost` to add `aliasSymbol` and make it only visible to `Visitor`, but I thought that was too complex.

kythe/typescript/indexer.ts
zzmp: I'm confused as to what you're recommending. Could you clarify?

I've moved `getSymbolAtLocation` into the `IndexerHost`. the `ts.TypeChecker` is still used directly in the `Visitor` for aliasing (`this.typeChecker.getAliasedSymbol`) in two spots. I could move that into the `IndexerHost`, but it seems like overkill.
I also don't understand why it's necessary to cache the typechecker as `this.typeChecker`, on either the `IndexerHost` or the `Visitor`, as the `ts.Program` already caches it. Accessing it as `this[.host].program.getTypeChecker()` will not create a new one if one already exists.

kythe/typescript/indexer.ts
ayazhafiz: Let's leave this conversation open in case Evan has some comments, but this seems okay to me for now.

kythe/typescript/indexer.ts
ayazhafiz: That's my bad. I didn't see the other usage of `typechecker` in the Visitor. sgtm :+1:

kythe/typescript/indexer.ts
ayazhafiz: How do you feel about moving the `visitImport` logic into `getSymbolName`? We can do it in a separate PR. I'm taking a look at it implementing it now to see if it's a reasonable option.

zzmp: i.e. so the aliasing is all done within the host, and the symbolNames can be completely contained?

ayazhafiz: yep.

zzmp: I feel good about that, but I think it belongs in a separate PR. This one really just moves things around without changing much logic.

ayazhafiz: LGTM

ayazhafiz:
LGTM

kythe/typescript/indexer.ts
```diff
@@ -30,6 +30,32 @@ const LANGUAGE = 'typescript';
...
+   */
+  getSymbolAtLocation(node: ts.Node): ts.Symbol|undefined;
+  /**
+   * Computes the VName (and signature) of a ts.Symbol. A Context can be
+   * optionally specified to help disambiguate nodes with multiple declarations.
+   * See the documentation of Context for more information.
+   */
+  getSymbolName(sym: ts.Symbol, ns: TSNamespace, context?: Context): VName;
+  /**
+   * Aliases a local symbol to the VNames of a remote symbol.
```
evmar: What does "aliasing" as a verb here mean?  Records that they are the same in memory, emits something, or something else?

kythe/typescript/indexer.ts
```diff
@@ -30,6 +30,32 @@ const LANGUAGE = 'typescript';
  * code duplication.
  */
 export interface IndexerHost {
```
evmar: This change seems to be expanding the mandate of IndexerHost.  Before, it was "stuff the indexer needs from its environment to do its job, with functions like emit()".  Now, it seems to encompass indexing work as well, representing the API between indexer and plugins.  I think it's useful to have that interface as well, but I don't think it all belongs here.  Another way of putting this is that I think it is now not possible for anyone to implement IndexerHost except for this particular file.

I think it'd be fine for the plugin to just receive a pointer to the indexer itself.

zzmp: Is there a reason that IndexerHost needs to be general enough for other files to implement?

The Plugin and the indexer need to be able to access the same canonical signature - that is the point of this change. The indexer seems to just be Visitor. Could a typed Indexer "pointer" just be passed to both the Visitor and the Plugin, with VName lookup methods/capabilities? Do you think that'd be more appropriate?

ayazhafiz: I think introducing a separate indexer with VName lookup capabilities would make the ecosystem unnecessarily complicated.

Because the TS Indexer itself can be considered a plugin, another idea is to have a registry of indexers on the indexer host, similar to what Evan suggested. Plugins could then share arbitrary functionality. This may not be a nice API because it would likely have to be an array of `ts.Plugin`s that other plugins would have to search through, but it's worth considering.

kythe/typescript/indexer.ts
zzmp: Updated the comment.

kythe/typescript/indexer.ts
zzmp: As discussed, IndexerHost is a misnomer. The add'l functionality is fine there.

kythe/typescript/indexer.ts
```diff
@@ -173,7 +174,7 @@ function hasExpressionInitializer(node: ts.Node):
 }
 
 /**
- * Determines is a node is a static member of a class.
+ * Determines if a node is a static member of a class.
```
ayazhafiz: :) thanks

kythe/typescript/indexer.ts
```diff
@@ -50,8 +50,9 @@ export interface IndexerHost {
    */
   getSymbolName(sym: ts.Symbol, ns: TSNamespace, context?: Context): VName;
   /**
-   * Aliases a local symbol to the VNames of a remote symbol.
-   * Returns the aliased name.
+   * Aliases a local symbol to a remote symbol. Returns the aliased name.
+   * This is used to return a remote symbol's VNames for a local symbol when a
+   * symbol is imported into a file.
```
ayazhafiz: Can you add `See Visitor#visitImport`?

ayazhafiz:
LGTM, let's merge after builbot passes

Issue #3711, opened 2019-05-01
specify and document a single general embedded JSON metadata format


shahms:
Right now, in addition to the embedded `GeneratedCodeInfo` proto used for protobuf generated code, there are at least two different JSON metadata formats supported by various indexers.

C++ allows optionally line-wrapped, base64-encoded metadata of the form:
(from kythe/cxx/indexer/cxx/testdata/metadata/inline.meta)
```
{
  "type": "kythe0",
  "meta": [
    {
      "type": "anchor_defines",
      "begin": 187,
      "end": 190,
      "edge": "%/kythe/edge/generates",
      "vname": {
        "signature": "gsig",
        "corpus": "gcorp",
        "path": "gpath",
        "language": "glang",
        "root": "groot"
      }
    }
  ]
}
```
Where the payload depends on the type field and also supports, "anchor_anchor" (aka imputes edges) via:
```
{
  "type": "kythe0",
  "meta": [
    {
      "type": "anchor_anchor",
      "source_begin": 184,
      "source_end": 187,
      "target_begin": 517,
      "target_end": 518,
      "edge": "/kythe/edge/imputes",
      "source_vname": {
        "corpus": "sourcecorpus",
        "path": "sourcepath",
        "root": "sourceroot"
      }
    }
  ]
}
```
I think the JavaScript indexer supports at least the latter, but not line-wrapping or base64 encoding.  Regardless, the different structure and fields between the two types should be documented (and ideally normalized in some fashion).

PR #3912, opened 2019-07-26
feat(cxx_common): document and implement kythe metadata as proto

shahms:
This partially addresses #3711 by describing the schema using a common language (proto), with some accommodations and hacks to account for existing usage.


kythe/cxx/common/json_proto.h
```diff
@@ -19,20 +19,49 @@
...
+#include "absl/strings/string_view.h"
 #include "google/protobuf/any.pb.h"
 #include "google/protobuf/io/zero_copy_stream.h"
 #include "google/protobuf/message.h"
+#include "google/protobuf/util/json_util.h"
 #include "kythe/cxx/common/status_or.h"
 
 namespace kythe {
 
+/// \brief Deserializes a protobuf from a JSON text.
```
zrlk:
```suggestion
/// \brief Deserializes a protobuf from JSON text.
```

kythe/cxx/common/json_proto.h
```diff
@@ -19,20 +19,49 @@
...
+/// \param stream The input text to parse.
```
zrlk:
```suggestion
/// \param input The input text to parse.
```

kythe/cxx/common/json_proto.h
```diff
@@ -19,20 +19,49 @@
...
+/// \param message The message to parse.
+/// \return The status message result of parsing.
+Status ParseFromJsonString(absl::string_view input,
+                           google::protobuf::Message* message);
+
+/// \brief Deserializes a protobuf from a JSON text.
```
zrlk:
```suggestion
/// \brief Deserializes a protobuf from JSON text.
```

kythe/cxx/common/json_proto.h
```diff
@@ -19,20 +19,49 @@
...
+/// \param stream The input text to parse.
```
zrlk:
```suggestion
/// \param input The input text to parse.
```

kythe/cxx/common/json_proto.h
```diff
@@ -19,20 +19,49 @@
...
+
 /// \brief Deserializes a protobuf from a JSON text stream.
 /// \param stream The input stream from which to read.
 /// \param message The message to parse.
 /// \return The status message result of parsing.
 Status ParseFromJsonStream(google::protobuf::io::ZeroCopyInputStream* input,
                            google::protobuf::Message* message);
 
+/// \brief Deserializes a protobuf from a JSON text stream.
+/// \param stream The input stream from which to read.
```
zrlk:
```suggestion
/// \param input The input stream from which to read.
```

kythe/proto/metadata.proto
```diff
@@ -0,0 +1,70 @@
...
+
+  // Fields only relevant if type == anchor_defines.
+  kythe.proto.VName vname = 3;  // The semantic node in the source language
+                                // which generated the text range.
+  uint32 begin = 4;  // Beginning of the range to match in the generated text.
+  uint32 end = 5;    // End of the range to match in the generated text.
+
+  // Fields only relevant if type == anchor_anchor.
+  kythe.proto.VName source_vname =
+      6;  // Anchor node in the source language which generated the text range.
```
zrlk: I think this is the vname for the source *file*. (The code generating this anchor doesn't know how to spell its signature; also, why would we need to specify `source_begin` and `source_end` if we could name the anchor?)

kythe/proto/metadata.proto
```diff
@@ -0,0 +1,70 @@
...
+  uint32 source_begin =
+      7;  // Beginning of the anchor node in the source language.
```
creachadair: Did clang-format really do this to the file? I'd like to advocate for putting the comments above the fields rather than this—which, though admittedly a personal preference, I find really jarring.

shahms: clang-format wraps them this way, yeah.

creachadair: That's (IMO) terrible. 😛

shahms:
@zrlk ping?

zrlk:
`metadata.proto` still refers to `source_vname` as the anchor node in the generating source file, unless I'm holding GitHub wrong.

shahms:
> `metadata.proto` still refers to `source_vname` as the anchor node in the generating source file, unless I'm holding GitHub wrong.

You're not, but that is, in fact, what it is.  The only field not taken from source_vname is the signature and that's potentially questionable with how build_config currently works.

shahms:
As in this might be intended to be the file anchor, but it's not currently implemented that way.

zrlk:
Aha. It would be more accurate to say it's the generating source file's VName with the `language` field set. (Some metadata generators aren't able to spell the `signature` of a given anchor; `imputes` relies on the pipeline to associate equivalent spans.)

shahms:
Sure, but it's also dubious for the indexer to change the signature at all if it's already present, IMO.

kythe/proto/metadata.proto
shahms: Updated the comment to include a bit more detail on the signature (which is the only field currently modified).  I'm still skeptical we should be doing anything with the signature if it's present, though.

PR #3913, opened 2019-07-26
chore(build): switch to a maintained YAML package

creachadair:
From discussion on https://github.com/ghodss/yaml/issues/50 it has become apparent that the github.com/ghodss/yaml package is no longer being maintained.

The Kubernetes SIG community maintains a clean fork sigs.k8s.io/yaml that is compatible with Go modules and includes bug fixes and performance improvements since the last release of its ancestor. The fork is being used successfully by several large projects within the k8s ecosystem.

This commit switches our dependency to the SIG fork, updates the relevant BUILD files, and updates the go.mod and go.sum files. It makes no functional changes.


creachadair:
I'm investigating the buildbot failure. It looks like a configuration issue.

creachadair:
I see what's going on. Our `go_repository` wrapper does not respect the `remote` attribute of the rule it shims (I think that attribute is relatively new). I will fix that, and add it.

creachadair:
Actually, I'll fix that separately; for this change I'll just use the existing name, cf. 62115eb.

PR #3917, opened 2019-07-27
feat: use bazel query to find targets to extract

salguarnieri:
Sometimes it may be easier to specify targets to build using a bazel query
rather than listing out explicit targets. Unfortunately, bazel cannot take many
arguments on the command line, so we need to shard our results across several
bazel invocations.

Note that we we changing how the old bazel build invocation happens, so this
change is not backwards compatible.

### DIFF
--- a/kythe/extractors/bazel/extract.sh
+++ b/kythe/extractors/bazel/extract.sh
@@ -62,7 +62,16 @@ if [ -n "$KYTHE_PRE_BUILD_STEP" ]; then
   eval "$KYTHE_PRE_BUILD_STEP"
 fi
 
-sh /kythe/bazel_wrapper.sh --bazelrc=/kythe/bazelrc "$@"
+if [[ -n "$KYTHE_BAZEL_TARGET" ]]; then
+  sh /kythe/bazel_wrapper.sh --bazelrc=/kythe/bazelrc "$@" -- "$KYTHE_BAZEL_TARGET"
+else
+  # If the user supplied a bazel query, execute it and run bazel, but we have to
+  # shard the results to different bazel runs because the bazel command line
+  # cannot take many arguments. Right now we build 30 targets at a time. We can
+  # change this value or make it settable once we have more data on the
+  # implications.
+  /kythe/bazelisk query "$KYTHE_BAZEL_QUERY" | xargs -t -L 30 sh /kythe/bazel_wrapper.sh --bazelrc=/kythe/bazelrc "$@" --
+fi
 
 # Collect any extracted compilations.
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"


kythe/extractors/bazel/extract.sh
```diff
@@ -65,7 +65,14 @@ fi
 # It is ok if targets fail to build. We build using --keep_going and don't
 # care if some targets fail, but bazel will return a failure code if any
 # targets fail.
-/kythe/bazelisk --bazelrc=/kythe/bazelrc "$@" || true
+if [ -n "$BAZEL_TARGET" ]; then
+  /kythe/bazelisk --bazelrc=/kythe/bazelrc "$@" "$BAZEL_TARGET" || true
```
justbuchanan: Do we need to use `BAZEL_TARGET` in this branch of the `if`?

kythe/extractors/bazel/extract.sh
```diff
@@ -65,7 +65,14 @@ fi
...
+else
+  # If the user supplied a bazel query, execute it and run bazel, but we have to
+  # shard the results to different bazel runs because the bazel command line
+  # cannot take many arguments.
+  /kythe/bazelisk query "$BAZEL_QUERY" | xargs -t -L "${TARGETS_PER_INVOCATION}" /kythe/bazelisk --bazelrc=/kythe/bazelrc "$@" || true
```
justbuchanan: The other env vars we pass to this script are all prefixed with `KYTHE_` - we should probably do the same here for consistency.

kythe/extractors/bazel/extract.sh
salguarnieri: Yeah, I'm changing the args passed to the script in our yaml generator to not include the targets so we have to pass them explicitly.

kythe/extractors/bazel/extract.sh
salguarnieri: done

salguarnieri:
@justbuchanan can you take another look to make sure I merged your change and mine correctly?

justbuchanan:
Sorry about the merge conflicts - I didn't make it easy on you here :)

I think we probably want the same check on the query branch of the if statement as well. You should be able to structure it so the handling happens after the if statement and is shared:

```
retval=0
if [ -n $KYTHE_BAZEL_TARGET ]; then
 ...
else
 ...
fi
# handle retval
```

salguarnieri:
This is now based on #3923 which should make it easier to execute bazel in a way we want.

kythe/extractors/bazel/extract.sh
```diff
@@ -62,16 +62,15 @@ if [ -n "$KYTHE_PRE_BUILD_STEP" ]; then
...
-    exit 1
+if [[ -n "$KYTHE_BAZEL_TARGET" ]]; then
+  sh /kythe/bazel_wrapper.sh --bazelrc=/kythe/bazelrc "$@" -- "$KYTHE_BAZEL_TARGET"
+else
+  # If the user supplied a bazel query, execute it and run bazel, but we have to
+  # shard the results to different bazel runs because the bazel command line
+  # cannot take many arguments. Right now we build 30 targets at a time. We can
+  # change this value or make it settable once we have more data on the
+  # implications.
+  /kythe/bazelisk query "$KYTHE_BAZEL_QUERY" | xargs -t -L 30 sh /kythe/bazel_wrapper.sh --bazelrc=/kythe/bazelrc "$@" --
```
justbuchanan: What does the trailing `--` do?

salguarnieri: xargs will the targets after the `--` argument so it produces the same behavior as above for the non bazel query case.

PR #3918, opened 2019-07-27
feature(typescript_indexer): extend symbol cache to store contexts

ayazhafiz:
Create a proper caching instance for the TypeScript indexer that caches
VNames for a symbol under a particular TypeScript namespace and symbol
declaration context.

Cache lookups are done by querying a 2-layer map, mapping from a symbol
to a serialized namespace and context to a VName. The namespace and
context must be serialized because object instances are always unique in
JavaScript, and creating a new object for each lookup would render the
cache useless. An alternative to this would be to store an object with
namespace/context tuples:

```typescript
interface { [nsctx: [TSNamespace, Context]]: VName }
```

but TypeScript doesn't support this.

This PR also moves some blocks of code around so they make more sense in
their spatial and contextual locations.


ayazhafiz:
Let's get #3909 first before merging this.
cc @evmar, @zzmp

kythe/typescript/indexer.ts
```diff
@@ -24,6 +24,20 @@ import * as utf8 from './utf8';
 
 const LANGUAGE = 'typescript';
 
+/**
+ * A indexer plugin adds extra functionality with the same inputs as the base
```
zzmp: nit: An indexer plugin...

kythe/typescript/indexer.ts
```diff
@@ -216,18 +146,154 @@ class StandardIndexerContext implements IndexerHost {
...
+}
+
+/**
+ * Context represents the environment a node is declared in, and may be used for
+ * disambiguating a node's declarations if it has multiple.
+ */
+enum Context {
+  /** No disambiguation about a node's declarations. */
+  Any,
+
```
zzmp: Unnecessary blank lines.

kythe/typescript/indexer.ts
```diff
@@ -216,18 +146,154 @@ class StandardIndexerContext implements IndexerHost {
...
+  /** Get a symbol VName for a given namespace and context, if it exists. */
+  get(symbol: ts.Symbol, ns: TSNamespace, context: Context): VName|undefined {
+    if (this.cache.has(symbol)) {
+      const nsCtx = this.serialize(ns, context);
+      return this.cache.get(symbol)!.get(nsCtx);
+    }
+    return undefined;
+  }
+
+  /** Set a symbol VName for a given namespace and context. */
```
zzmp: This always overrides set names. Is that desired?
I'd expect it to throw if the ts.Symbol and nsCtx were already in the map, as changing it might break other references.

kythe/typescript/indexer.ts
```diff
@@ -216,18 +146,154 @@ class StandardIndexerContext implements IndexerHost {
...
+      vnameMap.set(nsCtx, vname);
+    } else {
+      // If there is no map for the symbol, create a new one with an entry of
+      // the VName for the given namepsace and context.
+      this.cache.set(symbol, new Map([[nsCtx, vname]]));
+    }
+
+    // Set the symbol VName for the given namespace and `Any` context, if it has
+    // not already been set.
+    const nsAny = this.serialize(ns, Context.Any);
```
zzmp: This is only necessary if `context != Context.Any`.
It might be more performant if you included that check, although I think the logic is simpler as-is.

kythe/typescript/indexer.ts
```diff
@@ -568,39 +634,32 @@ class Visitor {
...
       throw new Error('TODO: symbol has no declarations?');
     }
 
-    // Disambiguate symbols with multiple declarations using a context. This
-    // only applies to getters and setters currently.
-    if (contextApplies) {
+    // Disambiguate symbols with multiple declarations using a context.
+    if (sym.declarations.length > 1) {
       switch (context) {
+        // Getter and setter contexts are filtered for getter and setter
```
zzmp: This seems self-explanatory - I don't think these comments are necessary.

kythe/typescript/indexer.ts
```diff
@@ -568,39 +634,32 @@ class Visitor {
...
           break;
         case Context.Setter:
           declarations = declarations.filter(ts.isSetAccessor);
           break;
+        // Otherwise, make no disambiguation.
+        default:
+          break;
       }
     }
-    // Otherwise, if there are multiple declarations but no context is
```
zzmp: Will deleting this codepath change behavior? Does it matter?
Now that you're defaulting to Context.Any, won't it change the behavior if there are multiple declarations, from using the get accessor to using whichever decl happens to be first?

kythe/typescript/indexer.ts
```diff
@@ -844,21 +898,28 @@ class Visitor {
...
-        null;
-    const kImportType = remoteSym.flags & ts.SymbolFlags.Type ?
-        this.getSymbolName(remoteSym, TSNamespace.TYPE) :
-        null;
-    // Mark the local symbol with the remote symbol's VName so that all
+    // Attempt to create the Kythe VName for both the symbol's type and value
+    // and mark the local symbol with the remote symbol's VName so that all
     // references resolve to the remote symbol.
-    this.symbolNames.set(localSym, [kImportType, kImportValue, kImportType]);
+    let kType, kValue;
```
zzmp: Why kVarName if it's not a constant?

kythe/typescript/indexer.ts
```diff
@@ -844,21 +898,28 @@ class Visitor {
...
     const remoteSym = this.typeChecker.getAliasedSymbol(localSym);
     // This imported symbol can refer to a type, a value, or both.
-    const kImportValue = remoteSym.flags & ts.SymbolFlags.Value ?
-        this.getSymbolName(remoteSym, TSNamespace.VALUE) :
-        null;
-    const kImportType = remoteSym.flags & ts.SymbolFlags.Type ?
-        this.getSymbolName(remoteSym, TSNamespace.TYPE) :
-        null;
-    // Mark the local symbol with the remote symbol's VName so that all
+    // Attempt to create the Kythe VName for both the symbol's type and value
```
zzmp: This changes behavior by not setting the Namespace. Does it matter?

kythe/typescript/indexer.ts
```diff
@@ -844,21 +898,28 @@ class Visitor {
```
zzmp: I thought you were moving this aliasing logic into the SymbolVNameCache. Are you still planning on that, or is this the final state?

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts
ayazhafiz: Fixed.

kythe/typescript/indexer.ts
ayazhafiz: I think it's okay; it's definitely more readable this way.

kythe/typescript/indexer.ts
ayazhafiz: No, it doesn't change the behavior. This is an artifact of the work done to emit an implicit property on getters and setters:

https://github.com/kythe/kythe/blob/f8c4e9bb682adb3b2041e49d72029e6be3977e5f/kythe/typescript/indexer.ts#L1168-L1174

that we never removed.

kythe/typescript/indexer.ts
ayazhafiz: Could you elaborate on your question?

kythe/typescript/indexer.ts
zzmp: This could also be added to the docstring for `set`.

kythe/typescript/indexer.ts
zzmp: This is more of a style question: why did you name the variable `kType, kValue` instead of `type, value`? I thought the `k` prefix was specified for constants.

kythe/typescript/indexer.ts
zzmp: ?

kythe/typescript/indexer.ts
zzmp: ?

kythe/typescript/indexer.ts
ayazhafiz: `k` is prefixed on variables that are VNames, seemingly standing for `kythe`.

evmar: Your guess is right.  This naming convention of mine is dumb, I am sorry.  I'd be happy to replace it with something that makes more sense.

ayazhafiz:
> This is hard to review because it mixes code moves with code changes, can you split it?

yep, that's my bad. Fixed. Also renamed the VName cache to `SymbolVNameStore` to better represent its purpose.

kythe/typescript/indexer.ts
```diff
@@ -189,6 +191,78 @@ function todo(sourceRoot: string, node: ts.Node, message: string) {
...
+ *              TSNamespace + Context
+ *              -----------   -------
+ *              TYPE          Any
+ * ts.Symbol -> VALUE         Getter  -> VName
+ *              NAMESPACE     Setter
+ *                            ...
+ *
+ * The `Any` context makes no guarantee of symbol declaration disambiguation.
+ * As a result, unless explicitly set for a given symbol and namespace, the
+ * VName of an `Any` context is lazily set to the VName of an arbitrary context.
```
evmar: Does "lazily" here mean that it will be sensitive to order of operations?

kythe/typescript/indexer.ts
```diff
@@ -189,6 +191,78 @@ function todo(sourceRoot: string, node: ts.Node, message: string) {
...
+ */
+type NamespaceAndContext = string&{__brand: 'nsctx'};
+class SymbolVNameStore {
+  private readonly cache =
```
evmar: Still not a cache, right?  If so, grep for more 'cache' in this change too.

kythe/typescript/indexer.ts
ayazhafiz: yes. there is no guarantee of the assigned VName form.

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts
```diff
@@ -189,6 +191,78 @@ function todo(sourceRoot: string, node: ts.Node, message: string) {
...
+ * have. Each TypeScript symbol can be be of a different TypeScript namespace
+ * and be declared in a unique context, leading to a total (`TSNamespace` *
+ * `Context`) number of possible VNames for the symbol.
+ *
+ *              TSNamespace + Context
+ *              -----------   -------
+ *              TYPE          Any
+ * ts.Symbol -> VALUE         Getter  -> VName
+ *              NAMESPACE     Setter
+ *                            ...
```
zzmp: Are there other symbols besides the 3 contexts that we select on?

kythe/typescript/indexer.ts
```diff
@@ -141,13 +141,15 @@ export enum TSNamespace {
...
 /**
- * Context represents the environment a node is declared in, and only applies to
- * nodes with multiple declarations. The context may be used for disambiguating
- * node declarations. A Getter context means the node is declared as a getter; a
- * Setter context means it is declared as a setter.
+ * Context represents the environment a node is declared in, and may be used for
+ * disambiguating a node's declarations if it has multiple.
  */
 export enum Context {
+  /** No disambiguation about a node's declarations. */
```
zzmp: This is lazy, but that's not mentioned here. Could you include `@see SymbolVNameStore` here?

kythe/typescript/indexer.ts
```diff
@@ -189,6 +191,78 @@ function todo(sourceRoot: string, node: ts.Node, message: string) {
...
+ *              TYPE          Any
+ * ts.Symbol -> VALUE         Getter  -> VName
+ *              NAMESPACE     Setter
+ *                            ...
+ *
+ * The `Any` context makes no guarantee of symbol declaration disambiguation.
+ * As a result, unless explicitly set for a given symbol and namespace, the
+ * VName of an `Any` context is lazily set to the VName of an arbitrary context.
+ */
+type NamespaceAndContext = string&{__brand: 'nsctx'};
```
zzmp: I'm afraid the jsdoc may be associated with this when it is meant for SymbolVNameStore. Could you move this type above the jsdoc comment?

kythe/typescript/indexer.ts
```diff
@@ -189,6 +191,78 @@ function todo(sourceRoot: string, node: ts.Node, message: string) {
...
+
+  /**
+   * Set a symbol VName for a given namespace and context. Throws if a VName
+   * already exists.
+   */
+  set(symbol: ts.Symbol, ns: TSNamespace, context: Context, vname: VName) {
+    let vnameMap = this.store.get(symbol);
+    const nsCtx = this.serialize(ns, context);
+    if (vnameMap) {
+      // If there is already a map for the symbol, check that a VName is not
```
zzmp: Unnecessary comment - repeats the docstring.

kythe/typescript/indexer.ts
```diff
@@ -189,6 +191,78 @@ function todo(sourceRoot: string, node: ts.Node, message: string) {
...
+      // overriden. Then set the VName for the given namespace and context.
+      if (vnameMap.has(nsCtx)) {
+        throw new Error(`VName already set with signature ${
+            vnameMap.get(nsCtx)!.signature}`);
+      }
+      vnameMap.set(nsCtx, vname);
+    } else {
+      // If there is no map for the symbol, create a new one with an entry of
```
zzmp: Unnecessary comment - repeats the docstring.

kythe/typescript/indexer.ts
```diff
@@ -207,13 +281,9 @@ class StandardIndexerContext implements IndexerHost {
   private rootDirs: string[];
 
   /**
-   * symbolNames maps ts.Symbols to their assigned VNames.
-   * The value is a tuple of the separate TypeScript namespaces, and entries
-   * in it correspond to TSNamespace values.  See the documentation of
-   * TSNamespace.
+   * symbolNames is a store of ts.Symbols to their assigned VNames.
```
zzmp: Single-line this docstring (or remove it entirely).

kythe/typescript/indexer.ts
```diff
@@ -496,39 +565,29 @@ class StandardIndexerContext implements IndexerHost {
...
       }
     }
-    // Otherwise, if there are multiple declarations but no context is
-    // provided, try to return the getter declaration.
-    else if (declarations.length > 1) {
-      const getDecls = declarations.filter(ts.isGetAccessor);
-      if (getDecls.length > 0) declarations = getDecls;
-    }
 
     const decl = declarations[0];
```
zzmp: Maybe comment that you are taking the first decl, and so if the Context is Any, it will be "lazily" decided.
Then again, that's been mentioned in at least two other places, so this is unnecessary.

kythe/typescript/indexer.ts
```diff
@@ -541,33 +600,35 @@ class StandardIndexerContext implements IndexerHost {
...
     // This imported symbol can refer to a type, a value, or both.
-    const kImportValue = remoteSym.flags & ts.SymbolFlags.Value ?
-        this.getSymbolName(remoteSym, TSNamespace.VALUE) :
-        null;
-    const kImportType = remoteSym.flags & ts.SymbolFlags.Type ?
-        this.getSymbolName(remoteSym, TSNamespace.TYPE) :
-        null;
-    // Mark the local symbol with the remote symbol's VName so that all
+    // Attempt to create the Kythe VName for both the symbol's type and value
+    // and mark the local symbol with the remote symbol's VName so that all
```
zzmp: s/VName/VName(s)

kythe/typescript/indexer.ts
```diff
@@ -541,33 +600,35 @@ class StandardIndexerContext implements IndexerHost {
...
   aliasSymbol(localSym: ts.Symbol, remoteSym: ts.Symbol): VName {
     // This imported symbol can refer to a type, a value, or both.
-    const kImportValue = remoteSym.flags & ts.SymbolFlags.Value ?
-        this.getSymbolName(remoteSym, TSNamespace.VALUE) :
-        null;
-    const kImportType = remoteSym.flags & ts.SymbolFlags.Type ?
-        this.getSymbolName(remoteSym, TSNamespace.TYPE) :
-        null;
-    // Mark the local symbol with the remote symbol's VName so that all
+    // Attempt to create the Kythe VName for both the symbol's type and value
```
zzmp: s/the/a or s/VName/VNames

kythe/typescript/indexer.ts
ayazhafiz: Nope, not currently.

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts
ayazhafiz: I think it's okay to have the documentation where it already is. Not adding here unless big opposition.

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts
```diff
@@ -607,8 +604,8 @@ class StandardIndexerContext implements IndexerHost {
 
   aliasSymbol(localSym: ts.Symbol, remoteSym: ts.Symbol): VName {
     // This imported symbol can refer to a type, a value, or both.
-    // Attempt to create the Kythe VName for both the symbol's type and value
-    // and mark the local symbol with the remote symbol's VName so that all
+    // Attempt to create VNames for both the symbol's type and value and mark
+    // the local symbos with the remote symbol's VName(s) so that all
```
zzmp: typo: symbos

PR #3922, opened 2019-07-30
chore: use checkerframework @Nullable annotation

jaysachs:
This version of `@Nullable` is usable in TYPE_USE and TYPE_PARAMETER contexts (e.g. it permits declaring `List<@Nullable String>`); it also is used by the [checker framework](https://checkerframework.org) nullness analysis. .

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
@@ -48,10 +48,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import javax.annotation.Nullable;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.Name;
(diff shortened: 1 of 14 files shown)


salguarnieri:
Can you add to the description to explain why we are switching to a different nullable?

jaysachs:
> Can you add to the description to explain why we are switching to a different nullable?

Done.

kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
```diff
@@ -89,7 +89,7 @@ public VName getNode(
       Symbol sym,
       String signature,
       // TODO(schroederc): separate MarkedSource generation from JavaEntrySets
-      @Nullable MarkedSource.Builder msBuilder,
```
shahms: JFYI, this introduces a google-java-format issue with our presubmits, have you set up the commit hooks as documented at https://www.kythe.io/contributing/#linting-and-testing ?

PR #3925, opened 2019-07-30
fix: Properly handle target expressions

salguarnieri:
We shouldn't quote the target expression variable because we want the underlying
script/program to interpret each target expression individually, we don't want
the program to get a single target expression that contains white space.


salguarnieri:
Added comment in script about why the variable shouldn't be quoted

PR #3929, opened 2019-07-31
chore(typescript_indexer): lint

zzmp:


zzmp:
cc @ayazhafiz I'm not sure if this conflicts with #3090.
Feel free to just pull this commit into your branch (and clobber my attribution if you need to).

ayazhafiz:
I think this is good to be merged separately, but let's see if we can do a miss linting fix (https://github.com/google/gts) to catch everything at once

zzmp:
Note that gts and clang-format (which was specified in the package.json) format the file differently.
Is there a preference over which to use? If the preference is gts, then the package.json should be updated so that other contributors are able to autoformat correctly.

creachadair:
> Note that gts and clang-format (which was specified in the package.json) format the file differently.
> Is there a preference over which to use? If the preference is gts, then the package.json should be updated so that other contributors are able to autoformat correctly.

As a rule of thumb we've generally tried to use clang-format where it's applicable. I don't think many of the Kythe team use much TS, though. Given a choice of options I'd suggest going with whatever Google uses for TS internally, since all this code will wind up having to get imported to Google anyway.

ayazhafiz:
> If the preference is gts, then the package.json should be updated so that other contributors are able to autoformat correctly.

@zzmp Can we pull out just tslint from gts and keep clang-format as is? I agree with @creachadair that clang-format is more applicable.

Sorry if my suggestion of gts was bad, I just meant that we should use the same lint rules they use. I'm happy to do this if you want; my bad for not being more clear.

zzmp:
Yes. Hold off on any review - I'm going to redo this and force-push over the existing commits.

evmar:
Confusingly, the "gts" ("google typescript') project on github is made by a group unrelated to the Google TypeScript team.  :(

evmar:
Is this still relevant?  (Sorry, going through unmerged PRs now and saw this...)

shahms:
Given the lack of response to Evan in June, I'm going to close this. Feel free to re-open if it's still relevant. Thanks!

PR #3934, opened 2019-08-01
refactor(typescript_indexer): emit "named" edges for imports

ayazhafiz:
Due to limitations in Kythe UIs, an anchor cannot be a definition and a
reference at the same time. For an import like

```typescript
import {foo} from './bar';
```

it would be nice to have local uses of `foo` "ref" `foo` in the import
statement and for `foo` in the import statement to "ref" the definition
of `foo` in './bar'. Since this is not possible, instead all local uses
of `foo` "ref" the definition of `foo` in './bar'.

The current way this is done is by having all Symbol VNames
corresponding to the external definition of `foo` also correspond to the
local definition of `foo`. This is somewhat unfavorable because it
breaks TypeScript semantics.

This PR introduces an alternate approach to achieving the same UI xref
functionality by exploiting Kythe "named" edges.

["named"](https://kythe.io/docs/schema/#named) edges connect a semantic node (i.e. a node that
represents a language entity and not a source code anchor) and an
external name for that node. They have a nice property of folding
cross-references for a UI; that is, given a graph like
  (1) `A -ref-> D <-named- B <-defines/binding- C`
a cross-reference between A and D is generated in a UI as if the graph
were
  (2) `A -ref-> B <-defines/binding- C`

For the TypeScript indexer, "D" in graph (1) is a local import VName.
Local usages can reference the local import, and such references will
later be folded to provide a UI xref from the local usage to the remote
definition.

Now that they are semantic nodes, local import symbols emit a ["name"](https://kythe.io/docs/schema/#name)
node because they are identifiers for a remote definition.

This commit introduces a change in the functionality of the indexer's
entry emissions but _not_ in its end result UI xrefs, so the change is
best classified as a refactoring.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -906,20 +906,20 @@ class Visitor {
     // (including renaming imports) to the remote symbol.
     //
     // This is done by exploiting Kythe graph edges. A "name" node is defined
-    // for the local symbol which the remote symbol is said to "generates".
-    // "generates" is an edge between two semantic nodes that provides for
-    // folding references. That is, a graph like
+    // for the local symbol which is "named" by the remote symbol. The "named"
+    // edge between two semantic nodes provides a nice functionality of folding
+    // references. That is, a graph like
     //
-    //   LocalUsage    LocalSymbol  RemoteSymbol      RemoteAnchor
-    //        \___________/^ ^\_________/ ^\_______________/
-    //             ref         generates    defines/binding
+    //   LocalUsage    LocalSymbol  RemoteSymbol        RemoteAnchor
+    //        \___________/^ ^\_________/ ^\_________________/
+    //             ref           named       defines/binding
     //
     // will produce a UI cross-reference from LocalUsage to RemoteAnchor as if
     // the graph were
     //
-    //   LocalUsage    RemoteSymbol      RemoteAnchor
-    //        \___________/^ ^\_______________/
-    //             ref         defines/binding
+    //   LocalUsage    RemoteSymbol        RemoteAnchor
+    //        \___________/^ ^\_________________/
+    //             ref          defines/binding
 
     const localSym = this.host.getSymbolAtLocation(name);
     if (!localSym) {
@@ -931,20 +931,20 @@ class Visitor {
     const localImportAnchor = this.newAnchor(name);
 
     // The imported symbol can refer to a type, a value, or both. Attempt to
-    // create "name"s for the local symbol and "generates" edges to the remote
-    // symbol in both cases.
+    // create "name"s for the local symbol and "named" edges from the remote
+    // symbol to the local symbol in both cases.
     //
     // Also attempt to emit a "ref/imports" from the referencing node to the
-    // to the remote definition in both cases. If the referecning node is
-    // different not the local symbol node, skip emitting "name" nodes and
-    // "generates" edges, as this is already done for the local symbol node.
+    // to the remote definition in both cases. If the referencing node is
+    // different from the local symbol node, skip emitting "name" nodes and
+    // "named" edges, as this is already done for the local symbol node.
     let kRemoteValue, kRemoteType;
     if (remoteSym.flags & ts.SymbolFlags.Value) {
       kRemoteValue = this.host.getSymbolName(remoteSym, TSNamespace.VALUE);
       const kLocalValue = this.host.getSymbolName(localSym, TSNamespace.VALUE);
 
       this.emitNode(kLocalValue, NodeKind.NAME);
-      this.emitEdge(kRemoteValue, EdgeKind.GENERATES, kLocalValue);
+      this.emitEdge(kRemoteValue, EdgeKind.NAMED, kLocalValue);
 
       this.emitEdge(localImportAnchor, EdgeKind.REF_IMPORTS, kRemoteValue);
     }
@@ -953,7 +953,7 @@ class Visitor {
       const kLocalType = this.host.getSymbolName(localSym, TSNamespace.TYPE);
 
       this.emitNode(kLocalType, NodeKind.NAME);
-      this.emitEdge(kRemoteType, EdgeKind.GENERATES, kLocalType);
+      this.emitEdge(kRemoteType, EdgeKind.NAMED, kLocalType);
 
       this.emitEdge(localImportAnchor, EdgeKind.REF_IMPORTS, kRemoteType);
     }
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -15,11 +15,12 @@
  */
 
 import 'source-map-support/register';
+
 import * as fs from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 
-import {EdgeKind, FactName, JSONEdge, JSONFact, makeOrdinalEdge, OrdinalEdge, Subkind, VName} from './kythe';
+import {EdgeKind, FactName, JSONEdge, JSONFact, makeOrdinalEdge, NodeKind, OrdinalEdge, Subkind, VName} from './kythe';
 import * as utf8 from './utf8';
 
 const LANGUAGE = 'typescript';
@@ -49,12 +50,6 @@ export interface IndexerHost {
    * See the documentation of Context for more information.
    */
   getSymbolName(sym: ts.Symbol, ns: TSNamespace, context?: Context): VName;
-  /**
-   * Aliases a local symbol to a remote symbol. Returns the aliased name.
-   * This is used to return a remote symbol's VNames for a local symbol when a
-   * symbol is imported into a file; @see Visitor#visitImport.
-   */
-  aliasSymbol(localSym: ts.Symbol, remoteSym: ts.Symbol): VName;
   /**
    * scopedSignature computes a scoped name for a ts.Node.
    * E.g. if you have a function `foo` containing a block containing a variable
@@ -358,6 +353,7 @@ class StandardIndexerContext implements IndexerHost {
         case ts.SyntaxKind.EnumMember:
         case ts.SyntaxKind.FunctionDeclaration:
         case ts.SyntaxKind.InterfaceDeclaration:
+        case ts.SyntaxKind.ImportEqualsDeclaration:
         case ts.SyntaxKind.ImportSpecifier:
         case ts.SyntaxKind.ExportSpecifier:
         case ts.SyntaxKind.MethodDeclaration:
@@ -431,6 +427,19 @@ class StandardIndexerContext implements IndexerHost {
           // be scoped to the class, not the constructor.
           if (!ts.isParameterPropertyDeclaration(startNode)) {
             parts.push('constructor');
+          }
+          break;
+        case ts.SyntaxKind.ImportClause:
+          // An import clause can have one of two forms:
+          //   import foo from './bar';
+          //   import {foo as far} from './bar';
+          // In the first case the clause has a name "foo". In this case add the
+          // name of the clause to the signature.
+          // In the second case the clause has no name because its imports are
+          // named bindings. This contributes nothing to the signature.
+          const importClause = node as ts.ImportClause;
+          if (importClause.name) {
+            parts.push(importClause.name.text);
           }
           break;
         case ts.SyntaxKind.ModuleDeclaration:
@@ -549,25 +558,6 @@ class StandardIndexerContext implements IndexerHost {
     }
 
     return vname;
-  }
-
-  aliasSymbol(localSym: ts.Symbol, remoteSym: ts.Symbol): VName {
-    // This imported symbol can refer to a type, a value, or both.
-    const kImportValue = remoteSym.flags & ts.SymbolFlags.Value ?
-        this.getSymbolName(remoteSym, TSNamespace.VALUE) :
-        null;
-    const kImportType = remoteSym.flags & ts.SymbolFlags.Type ?
-        this.getSymbolName(remoteSym, TSNamespace.TYPE) :
-        null;
-    // Mark the local symbol with the remote symbol's VName so that all
-    // references resolve to the remote symbol.
-    this.symbolNames.set(localSym, [kImportType, kImportValue, kImportType]);
-
-    // The name anchor must link somewhere.  In rare cases a symbol is both
-    // a type and a value that resolve to two different locations; for now,
-    // because we must choose one, just prefer linking to the value.
-    // One of the value or type reference should be non-null.
-    return (kImportValue || kImportType)!;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


ayazhafiz:
cc @evmar, @zzmp
cc @kythe for usage of Kythe schema

shahms:
We should not generally bend the graph to make the UIs "work".  Generates edges have particular semantics, which it does not seem like this use actually follows.

ayazhafiz:
> Generates edges have particular semantics, which it does not seem like this use actually follows.

I agree, this definitely a hack. We are looking some way to describe an implicit alias for a symbol. Are there other edges you feel would be more appropriate? Maybe "named"?

shahms:
@zrlk I believe Python has similar semantics here (in that imports create a local definition which should also be treated as a reference to the imported module).

ayazhafiz:
Moved towards using "named" edges which provide the same functionality and are abide by graph semantics.

zrlk:
The Python indexer gets creative with which part of the syntax it associates with the definition--it actually links the `import` (see https://github.com/google/pytype/blob/8f90fdc6a666c5af886d332ac3a5347d21a9a750/pytype/tools/xref/testdata/import.py#L13). It doesn't add the `named` edges or `name` nodes, the use of which here seems somewhat novel. In particular, the schema says that a `name` node should represent an "external identifier", like a JVM-internal class name or a mangled C++ symbol.

ayazhafiz:
> The Python indexer gets creative with which part of the syntax it associates with the definition--it actually links the `import`

What happens in the case of multiple imports? Do they all reference `import`?

I think the Kythe schema doesn't currently have anything that properly describes the relationship expressed here.

zrlk:
Looks like the Python indexer will drop multiple definitions on top of the `import` in something like `from typing import Optional, Text`.

The closest existing relationship that we have is `aliases` (https://kythe.io/docs/schema/#aliases). @schroederc, do you know whether we use `aliases` edges for clustering?

schroederc:
> Looks like the Python indexer will drop multiple definitions on top of the `import` in something like `from typing import Optional, Text`.
> 
> The closest existing relationship that we have is `aliases` (https://kythe.io/docs/schema/#aliases). @schroederc, do you know whether we use `aliases` edges for clustering?

We used to cluster aliasing, but we turned it off because it was unexpected for C++ users.

zrlk:
@ayazhafiz, just to confirm, should references to the local alias be aggregated with references to the definition being aliased? Or should the local alias be an explicit stopping point that users need to click through (like how Python works now)?

ayazhafiz:
> @ayazhafiz, just to confirm, should references to the local alias be aggregated with references to the definition being aliased?

It would be nice to have this, yes. That's the change proposed here.

> Or should the local alias be an explicit stopping point that users need to click through (like how Python works now)?

If the former option is not recommended, we can do this. But this is not preferred because it breaks TypeScript semantics.

zrlk:
I think that, to satisfy the preferred requirements, we'll need to add a new edge kind to the schema. It would have the same clustering behavior as `generates` going from the definition of the thing being imported to the local name for that thing, such that the definition is the cluster head. Maybe `delegates`? Then the local name is an ordinary `variable` and `named`/`name` isn't used at all.

creachadair:
> Due to limitations in Kythe UIs, an anchor cannot be a definition and a
> reference at the same time. For an import like

Maybe I missed something, but there seem to be two separate claims here, and I think at least one of them may be false:

An anchor _can_ be both a definition and a reference. In terms of the data model, that's perfectly OK. Historically Google's internal CodeSearch tool insisted on making a single-click action prefer one or the other, which is maybe what you're referring to in this case? But either way, that is not a limitation of the UI, that's a design choice that particular tool made.

I agree with the others who've spoken out against hacking around that decision. The UI has enough information to give the user a choice, and should do so if the user cares.

creachadair:
Leaving aside the definition/declaration question, it sounds like the real semantic question here is about resolution: When someone writes `import foo as bar` and then references `bar.Baz` in the code, the indexer has at least options for what to connect to the anchor:

1. It may refer to the binding introduced by the import statement,
2. It may refer to the imported package.

I'd argue the indexer should record both associations if it knows them: A user who requests cross-references for the `bar` at `from foo import bar` should be able to locate all the places where that name is actually used _in the module_, since that's the alias that governs its use. But likewise, a user who requests cross-references for the `bar` at `bar.Baz` should be able to locate the original definition of `bar` _as well as_ the local binding at `import foo as bar`.

How that gets surfaced in the UI is an important but—I would argue—ancillary question.

evmar:
Another reason to prefer the more "local" target when you have chain of

a: export foo
b: export {foo} from a
c: export {foo} from b
d: import {foo} from c;  foo.whatever

If you click on the 'foo' in d and it jumps you to a, you sometimes lose how you even got to a, so recording each hop separately has value.

zrlk:
If we don't want to consider the local created by an import as ~equivalent to the thing being imported, another possible way to satisfy the requirements may be to simply emit two `ref` edges at each site where the local is used. One `ref`s the local's definition; the other `ref`s the value bound to the local by the import. Then it's up to the UI to figure out what to do. (This is just a rehash of @creachadair's comment above.)

ayazhafiz:
> An anchor _can_ be both a definition and a reference. In terms of the data model, that's perfectly OK.

Agreed. I think we should strive to emit this information in every such case.

> Historically Google's internal CodeSearch tool insisted on making a single-click action prefer one or the other, which is maybe what you're referring to in this case? But either way, that is not a limitation of the UI, that's a design choice that particular tool made.

Ah, that is what I am referring to; sorry for not being more clear. I mean it as a limitation in a user's interaction with the Kythe graph - that is, a user can't see that an anchor is both a definition and reference as expressed in the Kythe graph because the UI choice is to pick one.

You are right that this should be a feature requested from the UI. And also, I think because the main usage Kythe as an indexing service is to provide cross-references in a UI, we should strive to have a good solution that works around this design choice until UI representations of an anchor as a reference and definition are a feature. I could be wrong though -- this is just my intuition.

> When someone writes `import foo as bar` and then references `bar.Baz` in the code, the indexer has at least options for what to connect to the anchor:
> 
> 1. It may refer to the binding introduced by the import statement,
> 2. It may refer to the imported package.
> 
> I'd argue the indexer should record both associations if it knows them: A user who requests cross-references for the `bar` at `from foo import bar` should be able to locate all the places where that name is actually used _in the module_, since that's the alias that governs its use. But likewise, a user who requests cross-references for the `bar` at `bar.Baz` should be able to locate the original definition of `bar` _as well as_ the local binding at `import foo as bar`.

I agree; to me, giving information about everything we know about the import usage is the best solution presented so far. I do think that the most correct association is still that `bar` at `bar.Baz` references only the local import and that `bar` at `import foo as bar` references the original definition of `foo`. This preserves the semantics of the language, because in fact `bar` is an alias of `foo` and is therefore a unique entity.

This is a difficult argument on all sides because we are trying to mix hacks for user convenience with the most appropriate (Kythe) graph and language semantics, so it's likely my interpretation is still not the best (if there even is a least-ambiguous one here).

> a: export foo
> b: export {foo} from a
> c: export {foo} from b
> d: import {foo} from c; foo.whatever
> 
> If you click on the 'foo' in d and it jumps you to a, you sometimes lose how you even got to a, so recording each hop separately has value.

Agreed. @craigdbarber's solution works here too; it is possible to emit a reference to each hop. I would argue that doing so would be incorrect in semantic terms, because it is not the case that `foo` in (d) is defined by all `foo`s in (a, b, c, d) - each is an alias. But it does provide the more-accurate hopping information.

creachadair:
I wonder if the UI could do a better job if we marked the names introduced by imports specially—say, by giving them a subtype (`variable/imported` for the sake of argument). That would allow the client to distinguish "liminal" bindings from "primary" ones. It would also permit denormalizing one hop if necessary for performance reasons: If @—ref—P and @—defines/binding—P and P.node/kind variable/imported and P—ref/imports—Q, say, then folding across the ref/imports relation for */imported nodes could be done without collapsing the identity.

I haven't fully thought through the implications of this, but it occurs to me that might be one way to tease apart the layers.

ayazhafiz:
Summarizing and annotating on the ideas expressed so far:

@creachadair's `variable/imported` subkind:
> `variable/imported` ...  would allow the client to distinguish "liminal" bindings from "primary" ones. It would also permit denormalizing one hop if necessary for performance reasons

If I'm understanding this correctly, this seems very close to @zrlk's `delegates` edge idea:

> clustering behavior as `generates` going from the definition of the thing being imported to the local name for that thing, such that the definition is the cluster head. Maybe `delegates`?

For these two ideas it would be useful to also create a `variable/importExport` subkind (or similarly, a `delegates` edge) on an `export from` statement

```typescript
export {foo} from './fooSource'; // TS/JS-only?
```

This would create a graph like

IMG_1237

<sup>(left hand `variable/import` subkinds, right hand `delegates` edges)</sup>

that would have references to the import/export hops @evmar [mentioned](https://github.com/kythe/kythe/pull/3934#issuecomment-517864277).

A UI could use these `variable/import` subkinds or `delegates` edges to fold imports across nodes:



Alternatively, each transitive usage of `AExport` could reference both its most-immediate hop and the "primary" `AExport` definition:



This requires some anchors to be both definitions and references, which is valid in the graph but not currently supported in a UI (that is, a UI chooses an anchor to be either a definition or reference, but not both).

If anchors can be treated as both definitions and references in the UI, it is also possible to just provide hops between "local" imports/usages:

IMG_1236

This hopping mechanism can still be done in a UI that doesn't support anchors that are both definitions and references but separating the anchor defining a node and the anchor referencing from that node:

IMG_1235

This is what the [Python indexer does](https://github.com/kythe/kythe/pull/3934#issuecomment-517475274).

creachadair:
Your synopsis seems accurate to me. Only your point (3) I think may not work: CodeSearch does not actually care about the identity of anchors, only their spans within the file—since that's how it decorates the file. That may have changed since I left, but if not, having a separate anchor at the same location will not fix the separation issue in the UI. A more sophisticated UI could make use of it, though.

ayazhafiz:
In (3) the anchors are emitted at different locations:

```python
@import defines/binding LocalBar
@bar ref ExternalBar
from foo import bar
```

creachadair:
> In (3) the anchors are emitted at different locations:
> 
> ```python
> # @import defines/binding LocalBar
> # @bar ref ExternalBar
> from foo import bar
> ```

I see. But then what happens if there is more than one import?

```python
from foo import bar, baz          # no aliases
from foo import bar as baz, quux  # aliases or mixed
```

I realize Google's Python style guide says you shoudn't do this, but even in Google3 it shows up in third-party code all the time. And at least from what TS code I've seen, it's pretty common to import multiple names this way too.

shahms:
For the record, this would also happen for C++ using declarations, in that a more "correct" modelling of: `using ::namespace::Type;` would be to treat it as equivalent to the type alias `using Type = ::namespace::Type;`, but with overlapping spans on `Type` for the reference and alias definition.  We don't currently do this (we only emit the reference), but there is at least one feature request around this from what I recall.

I still think (1) is the right solution and, AFAICT, the `delegates` would essentially be a more general form of the `aliases` edge, arguably obviating the latter (and potentially suggesting that we expand `aliases` rather than introducing a new edge).

ayazhafiz:
> I see. But then what happens if there is more than one import?

The local import definitions all cluster on the `import` anchor. See [zrlk's comment](https://github.com/kythe/kythe/pull/3934#issuecomment-517480974).

> `delegates` would essentially be a more general form of the `aliases` edge, arguably obviating the latter (and potentially suggesting that we expand `aliases` rather than introducing a new edge).

I agree. Though I'm not sure if there are any other situations of value aliasing except in imports/exports. Perhaps pointer aliasing, but this is a compiler feature, not a programming language feature.

shahms:
It's only "value" aliasing because Python doesn't have anything else ;-)  There was a C++ proposal to allow `using Name = AnotherName;` to be used regardless of the kind of `AnotherName` but because it was reasonable it hasn't been adopted.

hildrum:
Regarding your [comment](https://github.com/kythe/kythe/pull/3934#issuecomment-517873227), Kythe may need to give the UI better information so the UI can do the right thing, but Kythe shouldn't try to work around UI choices. 

There's no real constraint that the UI can't support definitions and references on one anchor.  It's just that the UI needs to know what happens on a click. Right now the two options are goto the definition or open the xref panel.

ayazhafiz:
> Kythe may need to give the UI better information so the UI can do the right thing

What kind of additional information would be useful for Kythe to provide in this case? Is it possible to always open the xrefs panel when an anchor has definitions and references?

zrlk:
Expanding `aliases` to do what `delegates` is suggested to do above would cause a regression for C++ users, for whom we narrowed `aliases` in the first place.

I still like fixing the UI, adding both refs in (to the def of the local name and to the def of the thing being imported), and using the `variable/import` subkind. `delegates` is still an option, but it might rope in too many local names in this instance; the UI would have to figure out how to scope them to the local file if the user wanted (to avoid dumping out from a reference to an import of module `foo` *all* of the references to imports of `foo` everywhere).

zrlk:
We talked this out today and came up with a new suggestion. Use an `aliases` edge between the imported `variable/import` node and the thing it targets. To avoid the problem we ran into with C++[1], we're going to change the way `aliases` behaves during denormalization. Roughly, if `A aliases B`, references to `A` will roll up to `B`, but references to `B` (or `A'` where `A'` aliases `B`) won't roll down to `A`.

For imports like `from foo import bar`, `bar` should both `defines/binding` the `variable/import` and `ref` `bar` in `foo`[2]. For imports like `from foo import bar as baz`, `baz` `defines/binding` the `variable/import` node, while `bar` should still `ref` `bar` in `foo`.

What do you think?

[1] the historic problem had to do with a typedef like `using foo = double`, where someone wanted to identify places where `foo` was used, but was helpfully provided with all of those sites plus everything else that was a `double`.

[2] Pending UI changes, putting the `defines/binding` spans on the `import` for non-renamed imports should work.

ayazhafiz:
> For imports like `from foo import bar`, `bar` should both `defines/binding` the `variable/import` and `ref` `bar` in `foo`[2]. For imports like `from foo import bar as baz`, `baz` `defines/binding` the `variable/import` node, while `bar` should still `ref` `bar` in `foo`.

This is reasonable to me; it is the "optimal" solution preserving graph semantics. Where do `alias` edges fit in here, if at all?

> Pending UI changes, putting the `defines/binding` spans on the `import` for non-renamed imports should work.

This is okay to me too, but let's see if there are any objections, as this is a hack we may not want to spread across multiple indexers. Is there any estimate for the timeline on the UI changes (@hildrum)?

zrlk:
The `aliases` edges are necessary for usages of the local `baz` to appear in the imported `bar`'s cross-references (similarly for non-renaming imports).

ayazhafiz:
So in `from foo import bar as baz` `Baz aliases Bar`, and in `from foo import bar` `LocalBar aliases FooBar`? (node names arbitrary)

creachadair:
> We talked this out today and came up with a new suggestion. Use an `aliases` edge between the imported `variable/import` node and the thing it targets.

With the change you described, I think this could work pretty well. I suppose it could be a little weird in Python where aliases might hop, e.g.

```python
# foo/__init__.py
from _submodule import thingy as Whatzit

# app.py
from foo import Whatzit as Hoozimabob
```

So now `@Hoozimabob—defines/binding—[HM]—aliases—[WH]—aliases—[TH]` or words to that effect. But I expect that may mostly be Python, where only values (not bindings) have types. 

> For imports like `from foo import bar`, `bar` should both `defines/binding` the `variable/import` and `ref` `bar` in `foo`[2]. For imports like `from foo import bar as baz`, `baz` `defines/binding` the `variable/import` node, while `bar` should still `ref` `bar` in `foo`.

And presumably in this case `@baz—defines/binding—[BAZ:var/import]—aliases—[foo:bar]`?

I don't see any obvious explosive difficulties.

ayazhafiz:
We'll need to change the description of `aliases` to allow values to be aliased as well.

zrlk:
The schema changes are up to `variable` (#3979); initial serving support will come with #3980.

robinp:
Nice! For the record, Haskell's record field punning syntax introduces at the same location a reference to the field of a data type, as well a local binding for the field of a specific record. Using this alias mechanism would come handy there.

shahms:
Just another observation, when differentiating between definitions and declarations, we may want to treat the local imports as declarations rather than definitions so "jump to definition" goes to the root, rather than leaf.  But this is also a decision that can/should be deferred to the UI.

PR #3940, opened 2019-08-02
feat(cxx_common): support proto encoding of compilation units in kzips

jaysachs:


kythe/cxx/common/kzip_writer.cc
```diff
@@ -112,19 +125,39 @@ StatusOr<std::string> KzipWriter::WriteUnit(
...
     initialized_ = true;
   }
-  if (auto json = WriteMessageAsJsonToString(unit)) {
-    auto file = InsertFile(kUnitRoot, *json);
+  auto json = WriteMessageAsJsonToString(unit);
+  if (!json) {
+    return json.status();
+  }
+  auto digest = SHA256Digest(*json);
+  StatusOr<std::string> result(InternalError("unsupported encoding"));
```
shahms: prefer copy initialization (using = ) over direct initialization (with parens or braces).

kythe/cxx/common/kzip_writer.cc
```diff
@@ -112,19 +125,39 @@ StatusOr<std::string> KzipWriter::WriteUnit(
...
+  if (encoding_ == KzipEncoding::Json || encoding_ == KzipEncoding::All) {
+    auto file = InsertFile(absl::StrCat(kJsonUnitRoot, digest), *json);
     if (file.inserted()) {
       auto status = WriteTextFile(archive_, file.path(), file.contents());
       if (!status.ok()) {
         contents_.erase(file.path());
         return status;
       }
```
shahms: minor: The `if (file.inserted())` logic should be shared, likely moved to `InsertFile` (unless I'm missing calls to InsertFile which aren't immediately followed by the same logic, but they all appear to be).

It was only used in two places before, so largely unnecessary, but now there's the magic number of > 2 ;-)

I think that refactoring will also allow us to get rid of InsertionResult entirely, as all of the existing uses immediately call WriteTextFile if the insertion succeeded and then return the digest (or error).

kythe/cxx/common/kzip_writer.h
```diff
@@ -34,12 +35,21 @@ class KzipWriter : public IndexWriterInterface {
...
   /// \param path Path to the file to create. Must not currently exist.
-  static StatusOr<IndexWriter> Create(absl::string_view path);
+  static StatusOr<IndexWriter> Create(absl::string_view path) {
+    return Create(path, KzipEncoding::Json);
+  }
+  /// \brief Constructs a Kzip IndexWriter which will create and write to
+  /// \param path Path to the file to create. Must not currently exist.
+  /// \param encoding Encoding to use for compilation units.
+  static StatusOr<IndexWriter> Create(absl::string_view path,
+                                      KzipEncoding encoding);
```
shahms: nit: you can just use a default value instead of an overload here.

kythe/cxx/common/kzip_writer.h
```diff
@@ -34,12 +35,21 @@ class KzipWriter : public IndexWriterInterface {
...
   /// \brief Constructs an IndexWriter from the libzip source pointer.
   /// \param source zip_source_t to use as backing store.
   /// See https://libzip.org/documentation/zip_source.html for ownership.
   /// \param flags Flags to use when opening `source`.
+  /// \param encoding Encoding to use for compilation units.
   static StatusOr<IndexWriter> FromSource(zip_source_t* source,
+                                          KzipEncoding encoding,
```
shahms: minor: use a default value for the encoding as well?

kythe/cxx/common/kzip_writer.cc
```diff
@@ -78,31 +69,53 @@ absl::string_view Basename(absl::string_view path) {
...
 
 KzipWriter::~KzipWriter() {
   DCHECK(archive_ == nullptr) << "Disposing of open KzipWriter!";
 }
 
+// Creates entries for the three directories if not already present.
+Status KzipWriter::InitializeArchive(zip_t* archive) {
+  std::vector<std::string> dirs = {kRoot, kFileRoot};
+  if (encoding_ == KzipEncoding::Json || encoding_ == KzipEncoding::All) {
+    dirs.emplace_back(kJsonUnitRoot);
```
shahms: minor: prefer push_back to emplace_back when both work.  The latter will also call explicit constructors so should generally be reserved where the additional power is required.

kythe/cxx/common/kzip_writer_test.cc
```diff
@@ -185,7 +193,7 @@ TEST(KzipWriterTest, IncludesDirectoryEntries) {
       // We don't really care about the rest of the entries, but it's easy
       // enough to fix the order of the subdirectories and minimally harmful.
       ElementsAre(
-          "root/", "root/units/", "root/files/",
+          "root/", "root/files/", "root/units/",
```
shahms: Why did this order change?

kythe/cxx/common/kzip_reader.cc
```diff
@@ -175,29 +216,42 @@ StatusOr<IndexReader> KzipReader::FromSource(zip_source_t* source) {
...
+      return options.status();
     }
   }
   return error.ToStatus();
 }
 
-KzipReader::KzipReader(ZipHandle archive, absl::string_view root)
-    : archive_(std::move(archive)), root_(root) {}
+KzipReader::KzipReader(ZipHandle archive, absl::string_view root,
+                       KzipEncoding encoding)
```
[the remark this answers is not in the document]
jaysachs: This is effectively "pick the best one". The constructor is private, `Open` calls `Validate` which picks one and is guaranteed to not choose `All`

kythe/cxx/common/kzip_writer.cc
```diff
@@ -78,31 +69,53 @@ absl::string_view Basename(absl::string_view path) {
...
+    : archive_(archive), encoding_(encoding) {}
 
 KzipWriter::~KzipWriter() {
   DCHECK(archive_ == nullptr) << "Disposing of open KzipWriter!";
 }
 
+// Creates entries for the three directories if not already present.
+Status KzipWriter::InitializeArchive(zip_t* archive) {
+  std::vector<std::string> dirs = {kRoot, kFileRoot};
+  if (encoding_ == KzipEncoding::Json || encoding_ == KzipEncoding::All) {
```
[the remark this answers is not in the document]
jaysachs: And people complain about Java's verbosity. There's nothing at all redundant in `typename std::underlying_type<KzipEncoding>::type` ...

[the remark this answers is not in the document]
shahms: It's trivial to add a function which hides the verbosity, but the standard library doesn't have one:

```
template <typename T, typename = std::enable_if_t<std::is_enum_v<T>>>
std::underlying_type_t<T> Underlying(T t) {
  return static_cast<std::underlying_type_t<T>>(t);
}
```
(Most of the shorter aliases weren't added to the standard until C++14/17, though).

kythe/cxx/common/kzip_writer_test.cc
jaysachs: Because of the changes in `InitializeArchive`. It used to be 

`for (const auto name : {kRoot, kUnitRoot, kFileRoot})`

but now it's dynamically constructed, but `kRoot` and `kFileRoot` are always there (and at the front).

kythe/cxx/common/kzip_writer.cc
```diff
@@ -21,19 +21,22 @@
...
 namespace kythe {
 namespace {
 
-constexpr absl::string_view kRoot = "root/";
-constexpr absl::string_view kUnitRoot = "root/units/";
-constexpr absl::string_view kFileRoot = "root/files/";
+const std::string kRoot = "root/";
+const std::string kJsonUnitRoot = absl::StrCat(kRoot, kJsonUnitsDir, "/");
+const std::string kProtoUnitRoot = absl::StrCat(kRoot, kProtoUnitsDir, "/");
+const std::string kFileRoot = absl::StrCat(kRoot, "files/");
```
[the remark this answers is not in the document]
jaysachs: Switched back to constexpr absl::string_view with manual concatenation.

kythe/cxx/common/kzip_reader.cc
```diff
@@ -112,14 +111,37 @@ StatusOr<absl::string_view> Validate(zip_t* archive) {
...
   VLOG(1) << "Using archive root: " << root;
+  absl::flat_hash_set<std::string> protoUnits;
+  absl::flat_hash_set<std::string> jsonUnits;
   for (int i = 0; i < zip_get_num_entries(archive, 0); ++i) {
     absl::string_view name = zip_get_name(archive, i, 0);
     if (!absl::StartsWith(name, root)) {
       return InvalidArgumentError(
           absl::StrCat("Malformed kzip: invalid entry: ", name));
     }
+    std::vector<std::string> parts = absl::StrSplit(name, '/');
```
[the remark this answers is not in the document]
jaysachs: much nicer thanks.

kythe/cxx/common/kzip_writer.cc
jaysachs: Great idea.

kythe/cxx/common/kzip_encoding.h
```diff
@@ -0,0 +1,35 @@
...
+namespace kythe {
+
+enum class KzipEncoding {
+  Json = 1,           ///< JSON encoded
+  Proto = 2,          ///< Proto encodedo
+  All = Json | Proto  ///< All encodings
+};
+
+constexpr /* inline */ absl::string_view kJsonUnitsDir = "units";
+constexpr /* inline */ absl::string_view kProtoUnitsDir = "pbunits";
```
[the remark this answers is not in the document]
jaysachs: Effectively denormalized them.

kythe/cxx/common/kzip_encoding.h
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef KYTHE_CXX_COMMON_KZIP_ENCODING_H_
+#define KYTHE_CXX_COMMON_KZIP_ENCODING_H_
+
+#include "absl/strings/string_view.h"
```
justbuchanan: x

kythe/cxx/common/kzip_reader.cc
```diff
@@ -112,14 +114,34 @@ StatusOr<absl::string_view> Validate(zip_t* archive) {
...
+    std::set_symmetric_difference(json_units.begin(), json_units.end(),
+                                  proto_units.begin(), proto_units.end(),
+                                  std::inserter(diff, diff.end()));
+    if (diff.size() != 0) {
+      return InvalidArgumentError(absl::StrCat(
+          "Malformed kzip: multiple unit encodings but different entries"));
+    }
   }
-  return root;
+  return KzipOptions({root, encoding});
```
shahms: just `KzipOptions{root, encoding}`

kythe/cxx/common/kzip_reader.cc
```diff
@@ -175,29 +217,42 @@ StatusOr<IndexReader> KzipReader::FromSource(zip_source_t* source) {
   libzip::Error error;
   if (auto archive =
           ZipHandle(zip_open_from_source(source, ZIP_RDONLY, error.get()))) {
-    if (auto root = Validate(archive.get())) {
-      return IndexReader(
-          absl::WrapUnique(new KzipReader(std::move(archive), *root)));
+    if (auto options = Validate(archive.get())) {
+      return IndexReader(absl::WrapUnique(new KzipReader(
+          std::move(archive), (*options).root, (*options).encoding)));
```
shahms: options->root, options->encoding

kythe/cxx/common/kzip_encoding.h
shahms: You shouldn't need this include or the dependency.

shahms:
(This still looks fine with that change, though).

PR #3941, opened 2019-08-02
build: migrate to Starlark Java rules

schroederc:
For now, they are just wrappers around the native rules, but their usage will become manadatory in Bazel 1.0 (issue bazel#8741).


schroederc:
The Starlark rules are just using `native.java_*` calls so should be backwards compatible through most Bazel versions.

PR #3943, opened 2019-08-02
fix: add an ErrPermissionDenied for services to return

danielmoy-google:
Sometimes it is ok to return PERMISSION_DENIED instead of NOT_FOUND,
when protecting the name of an asset isn't necessary.

Be careful to not use this when protecting even the existince of some
thing is important - the nit would be better to still return NOT_FOUND.


schroederc:
sp in desc: `s/the nit/then it/`

PR #3945, opened 2019-08-05
docs(schema): add property/{reads,writes} edges

schroederc:


kythe/docs/schema/schema.txt
```diff
@@ -939,6 +939,36 @@ Ordinals are used::
...
 --------------------------------------------------------------------------------
 
+[[property/reads]]
+property/reads
+~~~~~~~~~~~~~~
+
+Brief description::
+  A *property/reads* B if A is a reader of the property B
+Points from::
+  semantic nodes
```
creachadair: I am guessing this probably only makes sense for `function` nodes?

PR #3948, opened 2019-08-05
fix(cxx): clang-format recently edit cxx files

jaysachs:


kythe/cxx/common/kzip_writer.cc
```diff
@@ -135,7 +135,8 @@ StatusOr<std::string> KzipWriter::WriteUnit(
     return json.status();
   }
   auto digest = SHA256Digest(*json);
-  StatusOr<std::string> result = InternalError("unsupported encoding");;
+  StatusOr<std::string> result = InternalError("unsupported encoding");
+  ;
```
justbuchanan: x

PR #3949, opened 2019-08-05
ci(buildbot): add builder to run 'arc lint'

schroederc:
Update buildbot to use latest clang/Go versions (7.0 and 1.12.7) and their formatters.


shahms:
Exception 
The program is attempting to read user input, but stdin is being piped from some other source (not a TTY).
(Run with `--trace` for a full exception trace.)

PR #3950, opened 2019-08-05
fix(typescript_indexer): perform non-null check on symbol values

ayazhafiz:
TypeScript doesn't enforce strict non-null checks inside the compiler,
so we should. Perform a non-null check on symbol value declarations,
which are typed as always existing but don't exist for type-only values
like interfaces.

See https://github.com/microsoft/TypeScript/issues/24706


ayazhafiz:
cc @evmar, @zzmp

zzmp:
Is this a bug within the compiler? If it is, could you open an issue there and link it?

ayazhafiz:
Yes, we'll need to open an issue on the TS repo. I'm on an offsite today and tomorrow so I'll look into that Thursday

kythe/typescript/indexer.ts
```diff
@@ -854,7 +854,9 @@ class Visitor {
     // This can happen in cases of importing local modules, like
     //   declare namespace Foo {}
     //   import foo = Foo;
-    if (ts.isSourceFile(sf)) {
+    // Check that the value declaration actually exists. TypeScript doesn't
+    // enforce strict non-null checks in the compiler.
+    if (sf && ts.isSourceFile(sf)) {
```
zzmp: Because each comment is describing a different condition, I think it makes sense to split this into two separate commented statements for readability:
```
// If...
if (sf) {
  // [...]
}
// Check...
if (ts.isSourceFile(sf)) {
  // [...]
}
```

kythe/typescript/indexer.ts
```diff
@@ -854,7 +854,9 @@ class Visitor {
     // This can happen in cases of importing local modules, like
     //   declare namespace Foo {}
     //   import foo = Foo;
-    if (ts.isSourceFile(sf)) {
+    // Check that the value declaration actually exists. TypeScript doesn't
```
zzmp: Maybe include a reference to the TypeScript issue as a TODO here, so that it can be backed out when that issue is resolved.

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts

ayazhafiz:
@evmar PTAL

evmar:
Still relevant?

ayazhafiz:
no

PR #3951, opened 2019-08-05
fix(typescript_indexer): log TODO rather than error in indexer

ayazhafiz:
For errors that do not impede indexing of other entities in a TypeScript
source code, log a TODO rather than throwing an error to allow indexing
of the rest of the TS program to complete.


ayazhafiz:
cc @evmar, @zzmp

kythe/typescript/indexer.ts
```diff
@@ -508,7 +508,7 @@ class StandardIndexerContext implements IndexerHost {
     // TODO: update symbolNames table to account for context kind
 
     if (!declarations || declarations.length < 1) {
-      throw new Error('TODO: symbol has no declarations?');
+      console.warn(`TODO: symbol ${sym.name} has no declarations`);
```
zzmp: Use the `todo` function here.

ayazhafiz: unfortunately there is no `ts.Node` to pass to the todo function in this case because the only context is a symbol.

zzmp: Could you add a comment explaining this for future readers?

ayazhafiz:
@evmar PTAL

ayazhafiz:
@evmar PTAL

evmar:
Still relevant?

ayazhafiz:
no

PR #3953, opened 2019-08-06
fix(gotool): make source paths relative to enclosing module

justbuchanan:
When the extractor is run in "GOPATH mode", the `Package.Root` field is populated and used to determine relative paths for each of the source files. When using modules, this field is not populated. This PR parses the `Module` field from `go list` and uses the module's `Dir` as the root instead.

I tested this out by extracting parts of the kythe repo with:

```.sh
unset GOPATH
cd kythe
bazel build //kythe/go/extractors/cmd/gotool
./bazel-bin/kythe/go/extractors/cmd/gotool/gotool \
  -v \
  -goroot=$(go env GOROOT) \
  -corpus kythe \
  -output /tmp/kythe.kzip \
  kythe.io/kythe/go/extractors/cmd/gotool
```

And here's a snippet of the compilation unit before/after this change:

before:
```
...
{
  "v_name": {
    "corpus": "kythe",
    "path": "home/justinbuchanan/src/github/kythe/kythe/go/extractors/cmd/gotool/gotool.go"
  },
  "info": {
    "path": "home/justinbuchanan/src/github/kythe/kythe/go/extractors/cmd/gotool/gotool.go",
    "digest": "fc042f80f23b7893efd9db4a07f15a80a2ddead3697833d7c9d970393b648893"
  }
}
...
```

after:
```
...
{
  "v_name": {
    "corpus": "kythe",
    "path": "kythe/go/extractors/cmd/gotool/gotool.go"
  },
  "info": {
    "path": "kythe/go/extractors/cmd/gotool/gotool.go",
    "digest": "a84ea3d66ce94c3305c977f0a753e8dd8d002d824c8ff478097529f28af1b480"
  }
}
...
```

See also: https://github.com/kythe/kythe/issues/2881


asmundak:
It needs to work for the case where the module is in one of the subdirectories rather than at the top. Say, I have a module with import path 'foo' in subdirectory foo/ of the top-level directory $TOP. The extractor should chop $TOP from the absolute paths. What should be the incantation to run the extractor?
$ cd $TOP/foo && gotool -output <kzip> foo/...
this will complete successfully, but the paths will have $TOP/foo removed, not $TOP

justbuchanan:
@asmundak thanks for testing this. Did you pass the `-corpus` flag when you ran `gotool`? The corpus flag puts all of the extracted files into the same corpus and should make their paths relative to $TOP. I'll test this out and make sure it's working with modules.

So I can replicate what you're seeing, which repo are you running the extractor on? Is this for https://android.googlesource.com/platform/build/soong?

asmundak:
Yes, I am trying to run the extractor on that repo. 
The documentation is scarce, so I didn't realize that the corpus can contain the relative path, too. If I understand you correctly, the extractor should be run like this for the module in build/soong:
cd build/soong && gotool --output=foo.kzip --corpus=<top-corpus>/build/soong ./...
Let me conduct such experiment.

asmundak:
It's not working.

justbuchanan:
I don't need this anymore, closing.

PR #3956, opened 2019-08-06
cleanup(go extractor): remove unused --local_path flag

justbuchanan:
This isn't used anywhere - should it be?

### DIFF
--- a/kythe/go/extractors/cmd/gotool/gotool.go
+++ b/kythe/go/extractors/cmd/gotool/gotool.go
@@ -41,7 +41,6 @@ var (
 	bc = build.Default // A shallow copy of the default build settings
 
 	corpus     = flag.String("corpus", "", "Default corpus name to use")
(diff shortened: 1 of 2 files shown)


creachadair:
Removing this makes sense. The "local path" option exists in https://godoc.org/go/build#Import to deal with packages that use unconventional "local imports", e.g.,

    import "./some/random/package"

instead of the more conventional host-rooted paths `"kythe.io/kythe/go/some/random/package"`.

It probably wasn't worthwhile (in retrospect) for me to support this case; imports like that are mostly an artifact of the days before the `go` tool when people kept custom build scripts and Makefiles to run `6g` directly. Anyone who does that now deserves to have their code not indexed correctly so they will have an incentive to fix it (and it's not hard to fix).

Sadly, a lot of packages on GitHub still have this kind of nonsense in them, but not a big enough percentage that I think we should care (in my dependency graph, 233 out of 355338 indexed packages).

PR #3957, opened 2019-08-06
feat(kzip): add `kzip view` command

justbuchanan:
This is mostly a straightforward port from a google3 tool.


kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package viewcmd displays the contents of compilation units stored in .kzip files.
+package viewcmd
```
shahms: Canonical import comment?

kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,179 @@
...
+	"kythe.io/kythe/go/platform/kzip"
+	"kythe.io/kythe/go/util/cmdutil"
+	"github.com/golang/protobuf/jsonpb"
+
+	// These are common detail messages used by Kythe compilations, and
+	// required for JSON (un)marshaling to work.
+	_ "kythe.io/kythe/proto/buildinfo_go_proto"
+	_ "kythe.io/kythe/proto/cxx_go_proto"
+	_ "kythe.io/kythe/proto/go_go_proto"
+	_ "kythe.io/kythe/proto/java_go_proto"
```
shahms: nit: putting these in a separate file in the same package, which just does the imports might make it easier to change these on import.

kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,179 @@
...
+	fs.StringVar(&c.extractDir, "extract", "", "Extract files to this directory")
+}
+
+// Execute implements part of subcommands.Command.
+func (c *cmd) Execute(ctx context.Context, fs *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
+	hasErrors := false
+	for _, path := range fs.Args() {
+		ext := filepath.Ext(path)
+		base := filepath.Base(strings.TrimSuffix(path, ext))
+		if ext != ".kzip" {
```
shahms: This tool is `kzip` it should assume the files are kzips without the extension, IMO.

schroederc:
This already exists as https://github.com/kythe/kythe/blob/master/kythe/go/platform/tools/viewindex/viewindex.go.

justbuchanan:
Oh oops, definitely didn't see that. Would we prefer to keep this functionality in the `viewindex` binary or make it part of `kzip` now that we're dropping kindex?

Also this version has a `--extract` option that dumps file contents in addition to the compilation units.

shahms:
> Oh oops, definitely didn't see that. Would we prefer to keep this functionality in the `viewindex` binary or make it part of `kzip` now that we're dropping kindex?
> 
> Also this version has a `--extract` option that dumps file contents in addition to the compilation units.

This should be made part of kzip rather than having all of the disparate one-off tools.

justbuchanan:
I started down the path of removing `viewindex` in favor of `kzip view` and realized it's still tied into a fair number of tests, etc. I'm happy to do the cleanup work of removing it, but probably as a separate PR. It will require porting several tests to kzip instead of kindex.

shahms:
> I started down the path of removing `viewindex` in favor of `kzip view` and realized it's still tied into a fair number of tests, etc. I'm happy to do the cleanup work of removing it, but probably as a separate PR. It will require porting several tests to kzip instead of kindex.

That would be great, and I'm happy to have it in a separate PR.

kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,169 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
+	"github.com/google/subcommands"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
+	"kythe.io/kythe/go/platform/kzip"
```
schroederc: Separate `kythe.io` deps from third_party deps.

kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// SetFlags implements part of subcommands.Command.
+func (c *cmd) SetFlags(fs *flag.FlagSet) {
+	fs.StringVar(&c.extractDir, "extract", "", "Extract files to this directory")
+}
+
+// Execute implements part of subcommands.Command.
+func (c *cmd) Execute(ctx context.Context, fs *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
+	hasErrors := false
```
schroederc: `var hasErrors bool`

kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	return cerr
+}
+
+func (c *cmd) writeFiles(fd []fileData) error {
+	if c.extractDir == "" {
+		return nil
+	}
+	var g errgroup.Group
```
schroederc: `g, ctx := errgroup.WithContext(context.Background())`

Use context in `g.Go` so that things get cancelled properly on errors.

kythe/go/platform/tools/kzip/viewcmd/viewcmd.go
```diff
@@ -0,0 +1,169 @@
...
+			log.Printf("Error opening .kzip file: %v", err)
+			hasErrors = true
+			continue
+		}
+		defer f.Close()
+
+		err = kzip.Scan(f, func(r *kzip.Reader, unit *kzip.Unit) error {
+			if err := c.writeUnit(base+"-"+unit.Digest, unit.Proto); err != nil {
+				return fmt.Errorf("writing unit: %v", err)
+			} else if fd, err := c.kzipFiles(r, unit); err != nil {
```
schroederc: If we're deprecating `viewindex` we should allow for its behavior of only printing the compilation unit without its required inputs.

justbuchanan: This is the behavior if `--extract` is unset. `writeFiles()` and `kzipFiles()` are both no-ops if an extract directory is not given.

If it'd be clearer, I can move the `extractDir` check up here in `Execute()`.

PR #3960, opened 2019-08-06
fix(kzip): use canonical digesting for kzip compilation units

jaysachs:
I couldn't find a clean way to reuse the JSON-ified CU created in unit::Digest() in kzip::AddUnit().

### DIFF
--- a/kythe/go/platform/kcd/kythe/units.go
+++ b/kythe/go/platform/kcd/kythe/units.go
@@ -18,7 +18,6 @@
 package kythe // import "kythe.io/kythe/go/platform/kcd/kythe"
 
 import (
-	"fmt"
 	"io"
 	"sort"
 	"strings"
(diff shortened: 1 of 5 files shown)


jaysachs:
I know that changing the digests as used by kcd will have (short term) performance implications (and also incurs a collision possibility). I'm not clear if the digest as used by kcd actually needs to match that used by kzip.

kythe/go/platform/kcd/kythe/units.go
```diff
@@ -94,36 +93,12 @@ func (u Unit) Canonicalize() {
 }
 
 // Digest satisfies part of the kcd.Unit interface.
-func (u Unit) Digest(w io.Writer) {
+func (u Unit) Digest(w io.Writer) error {
 	pb := u.Proto
 	if pb == nil {
 		pb = new(apb.CompilationUnit)
 	}
-	put := func(tag string, ss ...string) {
```
shahms: This algorithm is likely a better choice than just the JSON encoded proto (but we should work on documenting it and implementing it in the other languages).

shahms: Especially since we don't have any control over the JSON representation emitted, which could change at any time.

jaysachs: Ok. I was just going off what was in the kzip doc, which explicitly states a JSON encoding. What is currently in units.go is definitely not JSON encoding.

Issue #3734, opened 2019-05-10
Example of "diagnostic"?


kamahen:
There's no example of `diagnostic` in the schema, and no test case AFAICT.

I infer the following (from looking at `doc` and `documents`) ... is it correct?
```
//- @v defines/binding V
//- V.node/kind variable
//- @v tagged VDiagnostic
//- VDiagnostic.node/kind diagnostic
//- VDiagnostic.message "Unused variable"
int v;
```

In this situation, it seems more natural to instead have `V tagged VDoc`, but the schema says that `tagged` takes `file` or `anchor`.

PS: in the example in https://kythe.io/docs/schema/#documents is `int v;` extraneous?

shahms:
The `int v;` in the `documents` example does seem extraneous.  Your example seems correct, yes.

`{file, anchor} -- tagged --> { diagnostic }`

"documents" doesn't enter into it.  That node is about documenting a particular entity, `diagnostic` is about calling out (generally) automated errors or warnings during analysis.

kamahen:
Is it allowed for something other than a `file` or `anchor` to be `tagged` with a `diagnostic`?

Explanations such as `{file,anchor} — tagged → {diagnostic}` would be helpful elsewhere in the schema doc (e.g. `{doc,anchor} — documents → {...}`).

BTW, I mentioned `documents` because it seemed likely that `diagnostic` would use a similar node/edge structure (`s/documents/tagged/`).

creachadair:
> Is it allowed for something other than a `file` or `anchor` to be `tagged` with a `diagnostic`?

We intentionally didn't restrict it that way—in principle a diagnostic could be attached to anything for which it's relevant. Files and anchors are the most obvious, though, since files and file-ranges are typically what you get from a compiler, and I don't think we have any other specific types that use them right now.

kamahen:
The documentation is misleading (to me anyway) ... https://kythe.io/docs/schema/#tagged says 
Points from
    [anchor], [file]
Points toward
    [diagnostic]

creachadair:
I'd say the documentation reflects existing practice—are there other kinds you think should be included?

kamahen:
I could envision a `variable` being marked with a diagnostic such as `not initialized`, which doesn't have an obvious `anchor`.

[...]

PR #3961, opened 2019-08-07
remove some extraneous text from schema doc

kamahen:
(This came up in discussion #3734)


kythe/docs/schema/schema.txt
```diff
@@ -501,8 +501,8 @@ lines owing to heuristics in Clang's comment parser.
 
 [kythe,C++,"Comments document objects."]
 --------------------------------------------------------------------------------
-int v;   //- @"/// An empty class." documents ClassC
-         //- ClassC.node/kind record
+//- @"/// An empty class." documents ClassC
+//- ClassC.node/kind record
```
shahms: I'm not entirely sure why, but apparently this change causes the schema tests to fail, @zrlk ?

kamahen: My guess, based on re-re-re-reading the text above the example, is that the original is a way of getting the two sets of comments to be separate (the `//-` comments for the verifier and the `///` comment for the test). So, I should remove my PR and leave future generations to puzzle this out. ;)

zrlk: Without the `int v;` there, I think Clang will merge these comments with the class comment into a single block.

kamahen:
Based on that, I'll delete this PR.

PR #3962, opened 2019-08-07
Add details to schema "diagnostic"

kamahen:
Adding text from discussion #3734.


kythe/docs/schema/schema.txt
```diff
@@ -1747,6 +1751,17 @@ Facts::
...
     related group of diagnostics (optional).
 
+[kythe,C++,"Warning message from compiler."]
+--------------------------------------------------------------------------------
+//- @v defines/binding V
+//- V.node/kind variable
+//- @v tagged VDiagnostic
+//- VDiagnostic.node/kind diagnostic
+//- VDiagnostic.message "WARNING: Unused variable"
+int v;
```
shahms: Examples in schema files have to actually validate when indexed as the language as which they are tagged.

kamahen: I don't think I can convince the compiler to generate that diagnostic, so what should I do?

shahms: Your best bet is to omit the example.  The only indexer which really emits much in the way of diagnostics is the Java indexer and it only emits them when there is some issue analyzing the file, such that it doesn't really attach them to text spans.

kamahen: Doesn't Go output an error for unused variables? (I don't use Go, so don't have a way of verifying this.)

This example took me a bit of effort to figure out, so I thought it'd be useful to other people. (Maybe it's obvious to other people, but it wasn't to me.)

shahms: Regardless of what the Go compiler may do, the Go indexer does not emit diagnostics for unused variables.  The only situation the Go indexer presently diagnosis is failure to resolve the package to which a particular symbol belongs.  This requires a bug in the extractor to do, so is not really something you can represent in a text snippet.

The Java indexer potentially emits a fair number of diagnostics, but again, generally only for things which are potentially bugs in the indexer not routine code issues or thing which would cause the file to fail to compile (although we should perhaps do so in the latter case).

The C++ indexer doesn't currently emit diagnostics.

The vast majority of diagnostics are emitted by the analyzer driver when a file fails to index or by post-processing when a file or line exceeds the decoration limit.  None of these can be represented by an embedded verifier test.  The options are:

1) Don't include an example.
2) Update one of the indexers to emit a diagnostic for the case you're demonstrating.
3) Include the example without the verifier comments and a big notice saying that it's only an example, but not actually emitted.

I'd want buy-in from some other folks on (3) as it's pretty antithetical to the spirit of the examples at present.

creachadair: > Doesn't Go output an error for unused variables? (I don't use Go, so don't have a way of verifying this.)

Yes, it's a compile error to declare and not use a local variable in Go. The Go indexer doesn't emit a diagnostic for this case at present, since it is impossible if the code compiles at all. It does emit diagnostics for unresolved identifiers, though, so it might be possible to gin up an example using that.

kamahen: For #3, what markup should I use? (Is your markup documented anywhere?)
Would this work?
```
[kythe,,"Warning message from compiler."]
//- This is a theoretical example -- the C++ analyzer doesn't generate this kind of diagnostic
```

shahms: The markup is asciidoc, not sure if/where it's documented (particularly the pipeline plugin used for the examples).

@zrlk thoughts? I don't like the idea of non-functional examples, but providing an example of the diagnostic node seems useful but complicated since nothing currently emits them in a usable fashion.

kythe/docs/schema/schema.txt
```diff
@@ -1735,6 +1735,10 @@ Brief description::
   A *diagnostic* is a node with a message concerning some aspect of the related
   <<file>> or <<anchor>>.  These can often result from errors while building or
   analyzing a compilation and allow an analyzer to surface errors to end-users.
+  Each diagnostic is linked to the related <<file>> or <<anchor>> using a
+  <<tagged>> edge. In principle, a diagnostic could also be related to a semantic
```
zrlk:
```suggestion
  <<tagged>> edge (where the diagnostic node is the target of the edge). In principle, a diagnostic could also be related to a semantic
```

zrlk:
I agree with @shahms in that we probably shouldn't add unchecked examples. The example this PR adds also doesn't add very much signal past what you can get from the descriptive text. I do find that the change to the descriptive text is helpful.

zrlk:
Thanks for the PR!

PR #3963, opened 2019-08-07
fix(doc): update documentation regarding (non)empty Root

jaysachs:


kythe/docs/kythe-storage.txt
```diff
@@ -213,6 +213,13 @@ The initial definition of a VName will include the following 5 fields:
    to designate virtual collections like generated files.
    +
    +
+   An empty Root field should signify a concrete file in the corpus
+   relative to the corpus root. Typically the interpretation is that a
+   VName with an empty Root corresponds to a file under version
+   control in (one of) the repository(ies) being analyzed; a non-empty
+   Root would indicate an external dependency or a generated file.
```
shahms: external dependencies will typically have a different corpus, no?

jaysachs: Reworded a bit.

creachadair: > external dependencies will typically have a different corpus, no?

That would be my expectation. Generally the point of `root` is to allow the path namespace to contain overlays, so that concrete files with the same "virtual" path (but different physical directory locations) can be distinguished from each other. This could in principle be used for anything that shares the namespace, but the most obvious practical cases are Perforce-style branches and generated file overlays.

The specific concrete case that motivated `root` in the first place is how Bazel handles generated files: When you include a generated header in a C++ program, for example, you write something like

```c
#include "kythe/cxx/foo/bar/thingy.h"
```

The physical path corresponding to this file is something like `bazel-out/blahblah/genfiles/kythe/cxx/foo/bar/thingy.h` and that's how Bazel reports it to the action. But the compiler expects the path to _exactly match_ what is spelled in the `#include` so we needed some way to tell which part of the name was "virtualized" and which was concrete noise. In the original Grok implementation we tried a long list of path surgery tricks, none of which ever worked correctly. Adding another dimension to the VName allowed us to record the extractor's knowledge of the path layout in a reliable way.

The name "root" was kind of a default; everyone agrees it doesn't mean anything specific enough to explain its existence, but nobody had any better ideas either. `extra_degree_of_freedom_for_overlays` was voted soundly down.

I don't have a specific suggestion for this text, but that was the original rationale.

shahms:
Jay, is this still reelvant?

PR #3966, opened 2019-08-07
fix: warn (don't fail) on multiple java outputs

schroederc:


kythe/java/com/google/devtools/kythe/extractors/java/bazel/aspect.bzl
```diff
@@ -49,11 +50,14 @@ def _extract_java(target, ctx):
         processors = annotations.processor_classnames
 
     output_jar = [jar.class_jar.path for jar in info.outputs.jars]
+    if len(output_jar) > 1:
+        print("WARNING: multiple outputs for " + owner + ": " + str(output_jar))
+        output_jar = [output_jar[0]]
```
shahms: Is there an output-per-action? What happens with the extra actions in this case?

schroederc: There's some extra outputs for proto manifests internally that I'm going to avoid for right now.

PR #3967, opened 2019-08-07
fix(tests): make proto/textproto tests not rely on `jq`

justbuchanan:
unfortunately `jq` is not available in google3, so I changed the tests
back to awk for skipping nondeterministic fields and python's json.tool
for formatting. Because python's json output is different than jq's, all
our golden files had to be updated too.

### DIFF
--- a/kythe/go/util/formatjson/formatjson.go
+++ b/kythe/go/util/formatjson/formatjson.go
@@ -2,15 +2,15 @@
 package main
 
 import (
-	"os"
 	"encoding/json"
 	"log"
+	"os"
 )
 
 func main() {
 	obj := make(map[string]interface{})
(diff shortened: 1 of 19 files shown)


justbuchanan:
It turns out we can't use `python -m json.tool` internally either, so I added a `formatjson` go binary.

kythe/go/util/formatjson/formatjson.go
```diff
@@ -0,0 +1,22 @@
+// Binary formatjson reads json from stdin and writes formatted json to stdout.
```
schroederc: Needs copyright

justbuchanan: Done.

PR #3970, opened 2019-08-08
fix(java_indexer): dynamically resolve JRE 9 bootclasspath

schroederc:
If no JRE 8-style jars are found, fall-back to finding the default
bootclasspath for JRE 9+ using a StandardFileManager instance.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
+++ b/kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
@@ -42,12 +42,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
-import javax.tools.JavaFileManager;
 import javax.tools.OptionChecker;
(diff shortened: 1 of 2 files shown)


kythe/java/com/google/devtools/kythe/platform/java/JavacOptionsUtils.java
```diff
@@ -52,18 +60,64 @@
...
+        .getTask(
+            /* out = */ null,
+            /* fileManager = */ null,
+            /* diagnosticListener = */ null,
+            /* options = */ Arrays.asList("--system", JAVA_HOME.toString()),
+            /* classes = */ null,
+            /* compilationUnits = */ null,
+            context);
+    StandardJavaFileManager fileManager =
+        (StandardJavaFileManager) context.get(JavaFileManager.class);
```
shahms: Just: `javax.tools.JavaCompiler.getStandardFilemanager(null, null, null)`?

shahms: Aside from that, it's odd that we're constructing a filemanager to ask it for the platform_class_path which we're then using to construct a string argument list used to configure the filemanager with the platform_class_path.

schroederc: Yeah, that works.

PR #3973, opened 2019-08-08
feature(typescript_indexer): emit reference for "this" keyword

ayazhafiz:
Emit a reference between "this" keywords that have a particular type and
the type of the object they refer to. For instance,

```typescript
//- @C defines/binding CType
class C {
  foo() {
    //- @this ref CType
    this.foo();
  }
}
```

Nothing is done for "this" keywords that do not refer to any particular
type. In

```typescript
let obj = {
  foo() {
    this.foo();
  }
}
```

`obj` has no particular type other than `object` and so no reference is
emitted. The same applies to class expressions.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -1534,6 +1534,28 @@ class Visitor {
   }
 
   /**
+   * Emits a reference from a "this" keyword to the type of the "this" object.
+   */
+  visitThisKeyword(keyword: ts.ThisExpression) {
+    const sym = this.host.getSymbolAtLocation(keyword);
+    if (!sym) {
+      // "this" refers to an object with no particular type, e.g.
+      //   let obj = {
+      //     foo() { this.foo(); }
+      //   };
+      return;
+    }
+    if (!sym.declarations || sym.declarations.length === 0) {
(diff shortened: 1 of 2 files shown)


ayazhafiz:
cc @evmar, @zzmp

kythe/typescript/indexer.ts
```diff
@@ -1533,6 +1533,28 @@ class Visitor {
...
+    const sym = this.host.getSymbolAtLocation(keyword);
+    if (!sym) {
+      // "this" refers to an object with no particular type, e.g.
+      //   let obj = {
+      //     foo() { this.foo(); }
+      //   };
+      return;
+    }
+    if (!sym.declarations || sym.declarations.length === 0) {
+      // "this" keyword is `globalThis`, which has no declarations.
```
evmar: I believe this can also happen in like

```
class X { ... }
namespace X { ... }
```

I think we don't care about that though.

ayazhafiz:
@evmar or @kythe can we buildbot?

kythe/typescript/indexer.ts
ayazhafiz: There's a symbol declaration there, but it's not a type.

PR #3980, opened 2019-08-13
feat(serving): experimental support for xrefs indirection

schroederc:
Add `--experimental_cross_reference_indirection_kinds` flag, a multimap
from node kind to edge kinds, to Kythe xrefs server.  When handling a
CrossReferences request, if a requested node has a specified node kind,
all related nodes passing through one of the specified edge kinds will
be added the merged node set for the request.  This allows for dynamic
single-directional, top-down indirection through nodes such aliases or
universal name nodes.

### DIFF
--- a/kythe/go/serving/xrefs/xrefs.go
+++ b/kythe/go/serving/xrefs/xrefs.go
@@ -35,8 +35,10 @@ import (
 
 	"kythe.io/kythe/go/services/xrefs"
 	"kythe.io/kythe/go/storage/table"
+	"kythe.io/kythe/go/util/flagutil"
 	"kythe.io/kythe/go/util/kytheuri"
 	"kythe.io/kythe/go/util/schema/edges"
+	"kythe.io/kythe/go/util/schema/facts"
 	"kythe.io/kythe/go/util/schema/tickets"
 	"kythe.io/kythe/go/util/span"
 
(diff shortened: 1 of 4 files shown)


kythe/go/serving/xrefs/xrefs.go
```diff
@@ -55,8 +57,15 @@ import (
 
 var (
 	mergeCrossReferences = flag.Bool("merge_cross_references", true, "Whether to merge nodes when responding to a CrossReferencesRequest")
+
+	experimentalCrossReferenceIndirectionKinds flagutil.StringMultimap
 )
 
+func init() {
+	flag.Var(&experimentalCrossReferenceIndirectionKinds, "experimental_cross_reference_indirection_kinds",
+		`Comma-separated set of key-value pairs (node_kind=edge_kind) to indirect through in CrossReferences.  For example, "talias=/kythe/edge/aliases" indicates that the targets of a 'talias' node's '/kythe/edge/aliases' related nodes will have their cross-references merged into the root 'talias' node's.`)
```
zrlk: It looks like it will, but does this work with reverse edges? I think we want to follow `%aliases` for the aliases-for-import case (e.g., in `from foo import fn as gn`, `gn aliases fn`, and xrefs for `fn` should be a superset of the xrefs of `gn`).

schroederc: It will work for any edge kind, reverse or not.

PR #3982, opened 2019-08-13
typescript: update from 3.3 => 3.5

evmar:
There are no changes needed for this particular upgrade but it's good to
keep this in sync with current TypeScript in case the difference
matters.


evmar:
@kythe how can I merge this, given the lint failure?

shahms:
> @kythe how can I merge this, given the lint failure?

The lint failures don't block submission.  *However* in this case it's a false positive and I would opt out .lock files by editing `tools/arc/linter.sh` to ignore them.

evmar:
The lint failure blocks submission via github.  Do I just push using 'git push'?

PR #3988, opened 2019-08-15
fix(lint): ignore .lock files in copyright checks

ayazhafiz:
See discussion at #3982. The linter currently checks .lock files for
copyright headers, which do not need to have one.


ayazhafiz:
cc @kythe

PR #4002, opened 2019-08-19
chore: downstream static analysis fixes for ts

danielmoy-google:
Caveat: I don't know typescript at all, I'm just trusting the linters.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -230,7 +230,7 @@ class SymbolVNameStore {
   get(symbol: ts.Symbol, ns: TSNamespace, context: Context): VName|undefined {
     if (this.store.has(symbol)) {
       const nsCtx = this.serialize(ns, context);
-      return this.store.get(symbol)!.get(nsCtx);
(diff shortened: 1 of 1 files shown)


kythe/typescript/indexer.ts
```diff
@@ -230,7 +230,7 @@ class SymbolVNameStore {
   get(symbol: ts.Symbol, ns: TSNamespace, context: Context): VName|undefined {
     if (this.store.has(symbol)) {
       const nsCtx = this.serialize(ns, context);
-      return this.store.get(symbol)!.get(nsCtx);
+      return this.store.get(symbol).get(nsCtx);
```
evmar: I am surprised by this change.  What tool thinks this is right?

danielmoy-google: sending you a pointer to the downstream analyzer result

danielmoy-google: abandoning this in favor of #4003 then?

evmar: Yeah I will fix this in g3.

Issue #3894, opened 2019-07-18
provide option for extractors to canonicalize VName paths


shahms:
Currently, extractors map between the path as seen by the compiler (`required_input.info.path`) and that used by the indexer to emit nodes (`required_input.v_name.path`).  This is generally done by a simple textual replacement on the normalized relative path (but depends on the extractor).  This normalization is generally purely lexical.

The problem arises in build systems or specific corpora which indirect via symlinks during the build and when those symlinks don't  easily map back to the path as present in the source repository or may require an inordinate maintenance effort to keep the manual vnames.json mapping synchronized.

In those circumstances, we'll end up with broken cross references as the files will know themselves by a different name than will all of their dependents.  As such, I propose to allow extractors to canonicalize (rather than just normalize) file paths before mapping them to VNames.  There are some potential issues here in that the symlink may point outside the build root, thus no longer yielding relativizeable paths.  When that happens, the relative non-canonical path should be preferred.

shahms:
This could potentially be added to vnames.json as well to allow for a per-path configuration of symlink behavior.

shahms:
A rough outline of how to do this during extraction:

1) canonicalize the working directory
2) canonicalize compiler-paths
3) relativize the canonical paths
4) attempt to match those against vnames.json patterns

The biggest question for me is in step 3; if we fail to make the path relative to the canonical working directory should we try fallback options or keep it simple and use the canonical absolute path?

PR #4009, opened 2019-08-20
feat(cxx_extractor): allow resolving symlinks when creating vname paths

shahms:
This implements #3894 for C++ (which I suspect is the critical extractor).


kythe/cxx/common/path_utils.h
```diff
@@ -18,11 +18,89 @@
...
+  ///         error if the current directory cannot be determined.
+  StatusOr<std::string> Relativize(absl::string_view path) const;
+
+ private:
+  explicit PathCleaner(std::string root) : root_(std::move(root)) {}
+
+  std::string root_;
+};
+
+/// \brief PathRealized relativizes paths against a root using RealPath.
```
zrlk:
```suggestion
/// \brief PathRealizer relativizes paths against a root using RealPath.
```

kythe/cxx/common/path_utils.h
```diff
@@ -18,11 +18,89 @@
...
+class PathRealizer {
+ public:
+  /// \brief Creates a PathCleaner using the given root.
+  /// \param root The root against which to relativize.
+  ///             Will be resolved and made an absolute path.
+  static StatusOr<PathRealizer> Create(absl::string_view root);
+
+  /// \brief Transforms the resolved, absolute version of `path` into a path
+  ///        relative to the configured root.
+  /// \return A path relative to root, if `path` is, else a resolved `path` or
```
zrlk: The formatting here's weird.

kythe/cxx/common/status_or.h
```diff
@@ -58,34 +58,58 @@ class ABSL_MUST_USE_RESULT StatusOr final {
   template <typename U>
   StatusOr& operator=(StatusOr<U>&& other);
 
+  /// \brief Return this->status().ok()
```
zrlk: It returns this->status_.ok()

kythe/cxx/common/status_or.h
```diff
@@ -58,34 +58,58 @@ class ABSL_MUST_USE_RESULT StatusOr final {
...
-  // reasonable fix for this.
-  const Status& status() const { return this->status_; }
-  Status status() { return this->status_; }
-#else
+  /// \brief Return the contained status.
   const Status& status() const& { return this->status_; }
   Status status() && { return std::move(this->status_); }
-#endif
 
+  /// \brief Return this->ok()
```
zrlk: What are you intending to convey here?

kythe/cxx/extractor/cxx_extractor_bazel_main.cc
```diff
@@ -28,11 +28,15 @@
...
 #include "google/protobuf/stubs/common.h"
+#include "kythe/cxx/common/path_utils.h"
 #include "kythe/cxx/extractor/language.h"
 #include "third_party/bazel/src/main/protobuf/extra_actions_base.pb.h"
 
 DEFINE_string(build_config, "",
               "Human readable description of the build configuration.");
+DEFINE_string(canonicalize_vname_paths, "clean-only",
+              "Policy to use when canonicalization VName paths: "
+              "clean-only (default), prefer-relative, prefer-real.");
```
zrlk: It's super nitpicky, but is there a reason you went with dashes instead of underscores (like most flags use)?

shahms: Flag names typically use underscores (because they flag library requires it); flag values are substantially more varied in the separator.  Clang/GCC flags, for example, universally use dashes.  Similarly, underscores are typical when the flags in question are auto-generated from ALL_CAPS enumerators, but much less consistent when that's not the case.

The short is I went with dashes because I think they're more readable, but am not particularly wedded to it.

zrlk: SGTM. I also think dashes are easier to read.

kythe/cxx/common/path_utils.h
shahms: (Committed manually to avoid annoying googlebot).

kythe/cxx/common/path_utils.h
shahms: How so? should subsequent lines not be indented?

kythe/cxx/common/status_or.h
shahms: Which is equivalent; I would far rather documentation refer to public members of a class than private members.

kythe/cxx/common/status_or.h
shahms: That the explicit operator bool is equivalent to the `ok()` member function.

kythe/cxx/common/path_utils.h
zrlk: Sorry, looks like the wrong line is displayed. I meant to ask about the lonely "an" on the next line.

kythe/cxx/common/status_or.h
zrlk: That might be a better way to put it than copying the implementation.

kythe/cxx/common/status_or.h
zrlk: It's weird for me to see the implementation of a thing copied inexactly in a documentation comment. Can you describe what the function does in an alternate way?

kythe/cxx/common/path_utils.h
shahms: Ah, clang-format wrapped that line weirdly, fixed.

kythe/cxx/common/status_or.h
shahms: Sure, this is largely taken from the internal equivalent, though :-)

PR #4011, opened 2019-08-21
chore(cxx): switch to using Abseil flags library over gflags

shahms:
This does necessitate turning off gflags support the glog for the time being, due to conflicting exported macros between glog and absl. The extant flag-based functionality can also be achieved via environment variables, however.

In theory, there will be absl/logging at some point, but I don't know what the timeline on that is.


salguarnieri:
It looks like there are some build errors

shahms:
> It looks like there are some build errors

Oh the joy of undeclared dependencies.

shahms:
I cannot reproduce the clang-format issue from arc lint locally.

PR #4012, opened 2019-08-21
fix(kzip): make default kzip encodings explicit

jaysachs:


shahms:
I think the Java and C++ need to be formatted by their respective formatting tools.

shahms:
(google-java-format and clang-format)

PR #4017, opened 2019-08-22
feat(extraction): allow details to be added to a kindex required input

schroederc:


shahms:
Can you not switch to using `kzip create` for this?

PR #4034, opened 2019-08-27
refactor(typescript_indexer): constrain param type on `IndexerHost#emit`

ayazhafiz:
`IndexerHost#emit` should only accept JSON Kythe facts and edges. This
is a breaking change but as far as I remember, it should only impact a
small amount of (if any) users.


ayazhafiz:
@evmar PTAL

ayazhafiz:
@evmar PTAL

evmar:
Still relevant?

ayazhafiz:
nah its covered in the cleanup PR

PR #4035, opened 2019-08-27
refactor(typescript_indexer): use native `ts.HasExpressionInitializer`

ayazhafiz:
Removing defining our own `HasExpressionInitializer` type. This may have
been added when we upgraded to TypeScript 3.5.3.


ayazhafiz:
@evmar PTAL

PR #4042, opened 2019-08-29
fix(javac-wrapper): use `set -e` to propagate errors

justbuchanan:
If the extractor or javac calls in the wrapper fail, the wrapper should return an error code.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/standalone/javac-wrapper.sh
@@ -39,6 +39,8 @@
 # Other environment variables that may be passed to this script include:
 #   KYTHE_EXTRACT_ONLY: if set, suppress the call to javac after extraction
 
+set -e
+
 if [[ -z "$JAVA_HOME" ]]; then
   readonly JAVABIN="$(which java)"
 else


justbuchanan:
I was thinking we'd want to error from both the extractor and `javac`. Do we want it to continue even if the extractor fails?

I came across this because my build was indicating success, but I had things misconfigured so every extractor call was failing.

jaysachs:
Wouldn't the "keep going" be done at a higher level? Seems reasonable for the extractor to return an error.

schroederc:
`set -e` makes it so that the "javac" (as in our wrapper) fails before the actual compilation even starts (the actual `javac`) if our extractor fails.  This will cause all dependent extractions AND compilations to fail.

Without `set -e`, the extractor may fail on a compilation but all dependent extractions/compilations could still succeed in the build.  The higher level should be checking for the extractor failure.

creachadair:
IMO: The down side of `set -e` in this file is that it affects anything that sources it. This file doesn't execute anything of interest on its own. 

I'd have preferred to see the `set -e` put in to the script that actually uses these values, rather than here—this file just defines defaults and some shared behaviour that does not depend on success or failure. Whether to succeed or fail is an unrelated concern.

justbuchanan:
@creachadair I don't think anything sources this script - this is just a  drop-in replacement for `javac` that ties in the extractor.

I'll go ahead and roll this back then. I do think it would be nice if we can find a way to make it more obvious when extraction fails on some targets. It's hard to notice if there are just a few missing kzips in the output.

creachadair:
> @creachadair I don't think anything sources this script - this is just a drop-in replacement for `javac` that ties in the extractor.

Oh, oops, I expanded the file up but not down. I missed the call at the end.

So actually I think that basically invalidates my comment above.

PR #4057, opened 2019-09-04
chore(cxx_common): cleanups necessary to update to C++17

shahms:
Due to the Bazel RBE base image being Ubuntu 16.04, for which there is no C++17-ready standard library easily available, we are still stuck using C++14 for now.  However, should the base image change or a backported package be found these changes will be required to make the switch.


shahms:
It looks like the build failure was legitimate.  libmemcached still uses the now-removed `register` keyword in C++ code.  I can work around it, but the successes are false positives as the RBE toolchain is unconditionally adding -std=c++14 *after* the user-provided options, thus overriding them :-(

PR #4061, opened 2019-09-06
feat(typescript_indexer): print plugin names

nbeloglazov:


kythe/typescript/indexer.ts
```diff
@@ -88,6 +88,7 @@ export interface IndexerHost {
  * indexer.
  */
 export interface Plugin {
+  /** Name of the plugin. It will be printed in console when running plugin. */
```
zzmp: Can you specify "It will be printed to stderr when running plugin." instead?

PR #4062, opened 2019-09-06
feat(typescript_indexer): don't emit duplicate nodes for namespaces

nbeloglazov:


kythe/typescript/testdata/namespace.ts
```diff
@@ -17,3 +17,13 @@ namespace N {
...
 //- @obj ref NObj
 N.obj;
+
+// Repeat same namespace to make
+// sure we don't output duplicate
+// facts.
+//- SecondNdef defines NValue
+//- SecondNdef.loc/start @^"namespace"
+namespace N {
+	export const boo = 1;
```
evmar: Fix whitespace here.

kythe/typescript/testdata/namespace.ts
```diff
@@ -17,3 +17,13 @@ namespace N {
...
+  //- SecondNdef.loc/end @$"}"
+}
```
evmar: What about something like:

```
namespace N.z {
}
```

Do we emit `N` again here?

kythe/typescript/testdata/namespace.ts
nbeloglazov: Done.

kythe/typescript/testdata/namespace.ts
nbeloglazov: Yes. We emit `defines` edge but not `defines/binding`. Added test for N.Z.

PR #4064, opened 2019-09-08
Clarify documentation of vname root and path

kamahen:
This documents some conventions with vname root and path that are assumed by browsers and which I presume are followed by all the existing indexers.


kythe/docs/kythe-storage.txt
```diff
@@ -240,12 +241,21 @@ The initial definition of a VName will include the following 5 fields:
...
 
  * *Language.* The language this name belongs to. The schema defines specific
    labels for each supported language, so we don't wind up with a confusion of
    names like "cxx", "cpp", "C\+\+", etc. As a rule of thumb, we will use the
    common name of the language, in lowercase ("c++", "python", "elisp",
    "objectivec").
+   +
+   +
+   The *language* is empty (that is, "") for some nodes, such as
+   link:schema#file[file].
```
kamahen: Is this the right markup for the link?

kythe/docs/kythe-storage.txt
```diff
@@ -240,12 +241,24 @@ The initial definition of a VName will include the following 5 fields:
...
    directory, it's usually a good sign that a separate Root label should be
    assigned.
+   +
+   +
+   The components of a path are separated by "/"s, even if the
+   underlying operating system uses a different separator. The path
+   must not start with a "/" -- that is, it is relative to the *Root*.
+   If the root is empty, it is treated as meaning the top of the
+   corpus (e.g., the same as "/" if the corpus maps directly to the
+   file system).
```
shahms: VName paths are not necessarily file system paths.  They should generally *map* to them, but that mapping is not guaranteed to be direct.  This is independent of root.  We do generally try to maintain the semantics that an empty root corresponds to a concrete file in the corpus (as opposed to a generated or other build artifact), but it is not guaranteed that `(cd ${VNAME[corpus]}; cat ${VNAME[path]})` will do anything meaningful.

kamahen: I was trying to clarify what "path-structured label" means -- namely a list of names separated by "/"s. 

Is that correct? If so, I'll generalize the "mapping" language to avoid implying that there must be a file system.

(I don't think we need to introduce any concept of hierarchy, such as the Unix or Windows file systems; but it might be desirable in future because sometimes it seems to make sense to point to a "directory" rather than a file, e.g. for Python's `from foo import *`. But that should probably be left to a separate discussion because it has wider implications.)

creachadair: It looks like the closest thing we have to a definition right now is in the [Kythe URI spec](https://kythe.io/docs/kythe-uri-spec.html). I agree it'd make sense for them to be in sync, even if only by reference.

kamahen: The URI seems to allow extra "/"s in the path and doesn't say anything about "." and "..". Is this intentional or an oversight? My guess is that it's intentional because we should allow non-canonical queries; but the vnames should be in a canonical form to make "joins" work with simple string matching.

I'll add a pointer to the URI spec (and pct-encoding isn't needed for paths, assuming that they're UTF-8) and also add some language about "cleaning", "." and "..".

creachadair: Yeah, agreed, the URI spec has some URI-specific bits (like the escaping) that don't really apply to the abstract VName. 

The more I think about it, the more I think maybe these kinds of rules belong in the schema rather than the data format docs. As you pointed out, the format (purposely) allows non-canonical data, but in the schema we may implicitly rely on more constraints.

kamahen: I don't want to get involved with the structure of Kythe's documentation, so I've made what I hope are minimalist changes.

I think we need to clarify where normalization is done -- e.g., when Kythe facts are output by an indexer, can they be denormalized (containing ".", "..", in the Path, for example) or must the indexer normalize them? Similarly, can queries be denormalized?

BTW, Python has [posix.path.normpath](https://docs.python.org/3/library/os.path.html#os.path.normpath); presumably similar functions exist in C++, Java, Go, etc.

kythe/docs/schema/schema.txt
```diff
@@ -42,7 +42,7 @@ VName conventions
 -----------------
 
 By default, assume that the VNames of nodes should be chosen according to the
-following rules:
+following rules (more details are given in the link:kythe-storage#TermVName[storage model]):
```
zrlk:
```suggestion
following rules (more details are given in the link:../kythe-storage.html#TermVName[storage model]):
```

kythe/docs/kythe-uri-spec.txt
```diff
@@ -65,6 +65,11 @@ path-segment = 1*{unreserved | pct-encoded | "/"}  -- RFC 3987
 Note that the order of the attributes (the `attrs` production) is fixed, to
 ensure that a Kythe URI has a canonical string encoding.
 
+For queries, path-segment is resolved as specified in
+http://tools.ietf.org/html/rfc3986#section-5.2.4[RFC 3986 Section 5.2.4 (Remove Dot Segments)].
+
+See also link:kythe-storage-text#TermVName[Vector-Name (*VName*)]
```
zrlk:
```suggestion
See also link:kythe-storage.html#TermVName[Vector-Name (*VName*)]
```

kythe/docs/kythe-storage.txt
```diff
@@ -234,18 +239,35 @@ The initial definition of a VName will include the following 5 fields:
...
 
  * *Language.* The language this name belongs to. The schema defines specific
    labels for each supported language, so we don't wind up with a confusion of
    names like "cxx", "cpp", "C\+\+", etc. As a rule of thumb, we will use the
    common name of the language, in lowercase ("c++", "python", "elisp",
    "objectivec").
+   +
+   +
+   The *language* is empty (that is, "") for some nodes, such as
+   link:schema#file[file].
```
zrlk:
```suggestion
   link:schema/schema.html#file[file].
```

kythe/docs/kythe-storage.txt
```diff
@@ -179,7 +179,11 @@ data. Updating existing VNames to a new projection is a trivial mechanical
 rewriting process, particularly when the new projection is an extension of the
 old one.
 
-The initial definition of a VName will include the following 5 fields:
+See also link:kythe-uri-spec[Kythe URI Specification],
```
zrlk:
```suggestion
See also link:kythe-uri-spec.html[Kythe URI Specification],
```

kythe/docs/kythe-uri-spec.txt
zrlk: This is still a necessary change.

kamahen: (Gotta love github's UI. Thanks for catching this.)

kythe/proto/storage.proto
```diff
@@ -22,6 +22,7 @@ option go_package = "storage_go_proto";
 option java_package = "com.google.devtools.kythe.proto";
 
 // VName is a proto representation of a vector name.
+// See also https://kythe.io/docs/kythe-storage.html#_a_id_termvname_a_vector_name_strong_vname_strong
```
kamahen: This seems to have upset a linter: "Error  (S&RX) clang-format" but the messages doesn't say what's wrong.
Is it a too-long line? (How can I run the linter stand-alone to check?)

zrlk: `arc lint` will run the linter provided you have Arcanist installed. It looks like you just have to run `clang-format` on the proto file.

PR #4068, opened 2019-09-11
chore: update max bazel version; remove redundant flag

shahms:


version.bzl
```diff
@@ -2,7 +2,7 @@
 MIN_VERSION = "0.28.0"
```
schroederc: `--experimental_allow_incremental_repository_updates` is still needed for 0.28.0

PR #4077, opened 2019-09-13
chore: remove remaining indexpack support

jaysachs:


kythe/cxx/tools/static_claim_main.cc
```diff
@@ -264,43 +261,17 @@ int main(int argc, char* argv[]) {
...
-              CompilationUnit unit;
-              CHECK(pack.ReadCompilationUnit(file_id, &unit, &error_text))
-                  << "Error reading unit " << file_id << ": " << error_text;
-              tool.HandleCompilationUnit(unit);
-              return true;
-            },
-            &error_text)) {
-      absl::FPrintF(stderr, "Error scanning index pack: %s\n", error_text);
-      return 1;
+  while (getline(std::cin, next_index_file)) {
```
shahms: as long as you're here, this should be `std::getline` (it's using ADL right now).

jaysachs: Done

PR #4084, opened 2019-09-16
feat(kzip): add function for merging multiple KzipInfos together

justbuchanan:
This is useful when extraction results are spread among multiple
kzips and we want to know the combined details.


kythe/go/platform/kzip/info/info_test.go
```diff
@@ -0,0 +1,53 @@
+package info
```
shahms: Missing copyright header

PR #4086, opened 2019-09-16
fix(extractor): Set cxx extractor kzip times to a constant

salguarnieri:
Don't include the kzip creation time in cxx kzips so kzips can be compared on only on contents, not on when they were created.

### DIFF
--- a/kythe/cxx/common/kzip_writer.cc
+++ b/kythe/cxx/common/kzip_writer.cc
@@ -41,7 +41,7 @@ constexpr absl::string_view kProtoUnitRoot = "root/pbunits/";
 constexpr absl::string_view kFileRoot = "root/files/";
 // Set all file modified times to 0 so zip file diffs only show content diffs,
 // not zip creation time diffs.
-const time_t kModTime = 0;
+constexpr time_t kModTime = 0;
 
 std::string SHA256Digest(absl::string_view content) {
   std::array<unsigned char, SHA256_DIGEST_LENGTH> buf;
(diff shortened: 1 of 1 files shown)


kythe/cxx/common/kzip_writer.cc
```diff
@@ -39,6 +39,9 @@ constexpr absl::string_view kRoot = "root/";
 constexpr absl::string_view kJsonUnitRoot = "root/units/";
 constexpr absl::string_view kProtoUnitRoot = "root/pbunits/";
 constexpr absl::string_view kFileRoot = "root/files/";
+// Set all file modified times to 0 so zip file diffs only show content diffs,
+// not zip creation time diffs.
+const time_t kModTime = 0;
```
shahms: needs to be constexpr

kythe/cxx/common/kzip_writer.cc
```diff
@@ -52,8 +55,16 @@ Status WriteTextFile(zip_t* archive, const std::string& path,
                      absl::string_view content) {
   if (auto source =
           zip_source_buffer(archive, content.data(), content.size(), 0)) {
-    if (zip_file_add(archive, path.c_str(), source, ZIP_FL_ENC_UTF_8) >= 0) {
-      return OkStatus();
+    if (auto idx =
+            zip_file_add(archive, path.c_str(), source, ZIP_FL_ENC_UTF_8);
+        idx >= 0) {
+      if (idx == 0) {
```
shahms: Why is zero being treated special? It shouldn't be different than any other index.

kythe/cxx/common/kzip_writer.cc
```diff
@@ -52,8 +55,16 @@ Status WriteTextFile(zip_t* archive, const std::string& path,
                      absl::string_view content) {
   if (auto source =
           zip_source_buffer(archive, content.data(), content.size(), 0)) {
-    if (zip_file_add(archive, path.c_str(), source, ZIP_FL_ENC_UTF_8) >= 0) {
-      return OkStatus();
+    if (auto idx =
+            zip_file_add(archive, path.c_str(), source, ZIP_FL_ENC_UTF_8);
+        idx >= 0) {
```
shahms: if-init is a C++17 feature and we're using C++14 due to some issues with the Bazel runtime build environment.

Even if this works, we should restrict to the specified version of C++ for now.

kythe/cxx/common/kzip_writer.cc
salguarnieri: done

kythe/cxx/common/kzip_writer.cc
salguarnieri: I think 0 is only returned in the case of zip_file_replace, but it should be harmless to handle 0 the same as positive indexes since otherwise the case would do something unexpected silently.

kythe/cxx/common/kzip_writer.cc
salguarnieri: done

PR #4087, opened 2019-09-17
fix(go): Change action go bazel extractor runs on

salguarnieri:
The go bazel rules we use now produce a GoCompilePkg action instead of a
GoCompile action.

### DIFF
--- a/kythe/extractors/BUILD
+++ b/kythe/extractors/BUILD
@@ -81,7 +81,7 @@ extractor_action(
     ],
     data = ["//external:vnames_config"],
     extractor = "@io_kythe//kythe/go/extractors/cmd/bazel:bazel_go_extractor",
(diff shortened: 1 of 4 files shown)


kythe/extractors/BUILD
```diff
@@ -81,7 +81,7 @@ extractor_action(
     ],
     data = ["//external:vnames_config"],
     extractor = "@io_kythe//kythe/go/extractors/cmd/bazel:bazel_go_extractor",
-    mnemonics = ["GoCompile"],
+    mnemonics = ["GoCompilePkg"],
```
schroederc: Mention that we only support Bazel rules_go 0.19.0 and up.

salguarnieri: done

PR #4090, opened 2019-09-17
chore(cxx_extractor): migrate many of the extractor tests to C++

shahms:
This is part of the effort in removing support for kindex files and the correspondingly obsolete kindex_tool (which also supports kzip files largely for these tests).

Additionally, it's gets rid of some complicated shell machinery.


shahms:
There remain additional tests to be migrated, but I figured this was a good start.

PR #4092, opened 2019-09-19
typescript: index inputs that contain errors

evmar:
This change makes us ignore TypeScript errors and index the parts
of the program that still have type information.  Previously we would
abort.

TypeScript in general has lots of accomodations around programs with
errors, so it's no trouble for TypeScript to do this.  (For example,
the compiler still wants to provide jump-to-definition features even
if you have a type error elsewhere in the same file.)

Within Google we don't run programs that have errors so code in this
state is rare, with one major exception: Clutz
(https://github.com/angular/clutz) output is invalid when its input
JavaScript is also invalid.  So this change has the main effect of
making Kythe able to index Clutz output.

We already had a test that exercised what happens when you have
bad input, so I adjusted that slightly to exercise this adjusted
logic.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -1809,31 +1809,20 @@ class Visitor {
 export function index(
     vname: VName, pathVNames: Map<string, VName>, paths: string[],
     program: ts.Program, emit?: (obj: {}) => void, plugins?: Plugin[],
-    readFile: (path: string) => Buffer = fs.readFileSync) {
+    readFile: (path: string) => Buffer = fs.readFileSync): ts.Diagnostic[] {
   // Note: we only call getPreEmitDiagnostics (which causes type checking to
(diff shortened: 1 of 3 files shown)


evmar:
@kythe/kythe  for review

kythe/typescript/indexer.ts
```diff
@@ -1863,6 +1852,8 @@ export function index(
       }
     }
   }
+
+  return diags;
```
salguarnieri: Is anything done with these diagnostics? Like logging or emitting kythe diagnostic nodes?

evmar: Not here.  We can in the consumer of this API if we want.  I'm not sure it's useful for us to log anything, because if the compile fails we should already know about it via other means.

salguarnieri: As long as we know about it somehow for when we are triaging a bug report, that sounds good to me.

PR #4095, opened 2019-09-19
fix(kzip info): determine unit corpus and file language more robustly

justbuchanan:
* unit vnames typically omit the corpus, so determine it from the
contained source files instead.
* file vnames omit the language, so determine it from the unit vname.

To test, I ran `kzip info` on a recent kzip of the kythe repo before and after this code change. Here's a partial diff of the output:

```diff
     "com_google_protobuf": {
       "files": {
-        "": 1862
+        "c++": 39,
+        "java": 1816,
+        "protobuf": 7
       },
       "units": {
-        "c++": 1
+        "c++": 1,
+        "java": 2
       }
     },
     "com_google_re2j_re2j": {
       "files": {
-        "": 2
+        "java": 2
       },
       "units": {}
     },
     "cstdlib": {
       "files": {
-        "": 27034
+        "c++": 27034
       },
       "units": {}
     },
     "github.com/kythe/kythe": {
       "files": {
-        "": 22511
+        "c++": 17012,
+        "java": 1380,
+        "jvm": 3441,
+        "protobuf": 68,
+        "typescript": 610
       },
       "units": {
         "c++": 150,
+        "java": 146,
+        "protobuf": 35,
         "typescript": 8
       }
     },
```

### DIFF
--- a/kythe/go/platform/kzip/info/info.go
+++ b/kythe/go/platform/kzip/info/info.go
@@ -54,8 +54,8 @@ func KzipInfo(f kzip.File, scanOpts ...kzip.ScanOption) (*apb.KzipInfo, error) {
 
 		for _, ri := range u.Proto.RequiredInput {
 			kzipInfo.TotalFiles++
-			// File VNames don't generally specify a language, so we determine
-			// it from the unit VName.
+			// Determine language from the unit VName (note that file VNames are
+			// forbidden from specifying a language).
 			corpusInfo(ri.GetVName().GetCorpus()).Files[u.Proto.GetVName().GetLanguage()]++
 			if srcs.Contains(ri.Info.Path) {
 				srcCorpora.Add(ri.GetVName().GetCorpus())
--- a/kythe/go/platform/kzip/info/info.go
+++ b/kythe/go/platform/kzip/info/info.go
@@ -19,6 +19,9 @@ package info // import "kythe.io/kythe/go/platform/kzip/info"
 
 import (
 	"fmt"
+	"log"
+
+	"bitbucket.org/creachadair/stringset"
 
 	"kythe.io/kythe/go/platform/kzip"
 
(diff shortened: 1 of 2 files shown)


kythe/go/platform/kzip/info/info.go
```diff
@@ -43,10 +46,30 @@ func KzipInfo(f kzip.File, scanOpts ...kzip.ScanOption) (*apb.KzipInfo, error) {
...
+
+		// The unit VName generally does not specify a corpus, so we have to
+		// determine corpus from the source files.
+		var srcCorpora stringset.Set
+		srcs := stringset.New(u.Proto.SourceFile...)
+
 		for _, ri := range u.Proto.RequiredInput {
 			kzipInfo.TotalFiles++
-			corpusInfo(ri.GetVName().GetCorpus()).Files[ri.GetVName().GetLanguage()]++
+			// File VNames don't generally specify a language, so we determine
```
shahms: File VNames are actually forbidden from specifying a language.

kythe/go/platform/kzip/info/info.go
```diff
@@ -43,10 +46,30 @@ func KzipInfo(f kzip.File, scanOpts ...kzip.ScanOption) (*apb.KzipInfo, error) {
 
 	err := kzip.Scan(f, func(rd *kzip.Reader, u *kzip.Unit) error {
 		kzipInfo.TotalUnits++
-		corpusInfo(u.Proto.GetVName().GetCorpus()).Units[u.Proto.GetVName().GetLanguage()]++
+
+		// The unit VName generally does not specify a corpus, so we have to
+		// determine corpus from the source files.
```
shahms: If it does we should defer to that, no?

kythe/go/platform/kzip/info/info.go
schroederc: The VNames in the graph for `file` nodes mustn't have a language, but there's no prohibition of the VName for the required input having no language.  It could potentially be useful to specify in cross-language compilations.

kythe/go/platform/kzip/info/info.go
schroederc: It's probably more useful to detect that all source files share a corpus and use that for validation/informational purposes.

kythe/go/platform/kzip/info/info.go
shahms: Given the structure and format of a CompilationUnit, I don't think cross-language compilations in the same CU are viable.  As such, I think the risk of accidentally producing invalid file VNames in the graph abides; required_input VNames should not have a language.

kythe/go/platform/kzip/info/info.go
shahms: Sure, but given the documented semantics of this we should actually use those semantics when determining the source file corpus, rather than just ignoring the CompilationUnit corpus entirely.

kythe/go/platform/kzip/info/info.go
schroederc: We already have "cross-language" compilations in the form of JVM-interop compilations (Java, Kotlin, etc.) where the required input class files may be from different source languages.  It may be useful to tag those for a consuming indexer.  The language VName field seems like a potentially appropriate location.  We could instead use a details message, though.

kythe/go/platform/kzip/info/info.go
schroederc: I don't think we're ignoring a CompilationUnit corpus.  It's empty, and therefore, not set in proto3-land.  In this case, all of the source files have the same corpus, however, and the CompilationUnit corpus really should have that same corpus (or at the very least set a differing corpus).  I think we're just covering for bad extraction here.

We could add a flag for this behavior.  Call it "--inference" or something, but I'm not sure where would want the previous behavior of attributing the compilation unit to no corpus.

justbuchanan: I simplified this a bit and put all mentions of the corpus (from unit vname and source vnames) into a stringset. At the end, if there's one value in the set, we have a consistent vname. Otherwise a warning is logged. This way there's no priority/precedence between unit and source file vnames, but if specified in both places, we do check that they're consistent

shahms: @justbuchanan that works, thanks!

> I don't think we're ignoring a CompilationUnit corpus. It's empty, and therefore, not set in proto3-land. In this case, all of the source files have the same corpus, however, and the CompilationUnit corpus really should have that same corpus (or at the very least set a differing corpus). I think we're just covering for bad extraction here.
> 
> We could add a flag for this behavior. Call it "--inference" or something, but I'm not sure where would want the previous behavior of attributing the compilation unit to no corpus.

The documented algorithm for determining a file VName corpus is:
 `unit.required_input[path].v_name.corpus OR unit.v_name.corpus`

The CompilationUnit VName corpus is not *necessarily* empty and, if present, should be used as a fallback corpus for file VNames.  It's also somewhat relevant to the discussion we had the other day about "build corpus" vs "attribute corpus" as the CU corpus is the former and file corpus the latter.

kythe/go/platform/kzip/info/info.go
justbuchanan: I updated the comment. Is there anything else to do here for this PR? Sounds like recording a language per-file could be useful for cross-language cases and this function should be updated if/when we do that.

PR #4102, opened 2019-09-26
fix: specify both GoCompile and GoCompilePkg mnemonics

jaysachs:
It's unclear if GoCompilePkg is needed, but GoCompile certainly is for extraction to work on gh/kythe/kythe/go/...

### DIFF
--- a/kythe/release/release.BUILD
+++ b/kythe/release/release.BUILD
@@ -147,7 +147,10 @@ extractor_action(
     ],
     data = [":vnames_config"],
     extractor = ":bazel_go_extractor",
-    mnemonics = ["GoCompilePkg"],
+    mnemonics = [
+        "GoCompile",
+        "GoCompilePkg",
+    ],
     output = "$(ACTION_ID).go.kzip",
 )
 


salguarnieri:
I'm surprised this works because I thought the extractor code name checked the mnemonic. So I would have thought you'd need to add it to all the places changed in https://github.com/kythe/kythe/pull/4087.

jaysachs:
> I'm surprised this works because I thought the extractor code name checked the mnemonic. So I would have thought you'd need to add it to all the places changed in #4087.

Yes, that's somewhat surprising. Here's an excerpt from the logs though:

```
Step #2 - "extract": INFO: From Executing extra_action @kythe_release//:extract_
kzip_go_extra_action on //kythe/go/platform/kzip:kzip_test:
Step #2 - "extract": 2019/09/26 03:14:18 Read 13613 bytes from extra action file
 "bazel-out/k8-fastbuild/extra_actions/external/kythe_release/extract_kzip_go_ex
tra_action/kythe/go/platform/kzip/04aaae54c1233793b89b7280f2a1ea46.xa"
Step #2 - "extract": 2019/09/26 03:14:18 Extracting XA for "//kythe/go/platform/
kzip:kzip_test" with 226 inputs
Step #2 - "extract": 2019/09/26 03:14:22 Finished reading required inputs [3.307
354507s elapsed]
```

which were not present without this change.

salguarnieri:
I'd like to better understanding why the extractor does more work, but doesn't appear to log fatal like it seems like it should. But if you want to get this in to kick off a pipeline run over night, that's fine by me.

jaysachs:
Unsure why this didn't trigger errors, but it's not required.

creachadair:
Which extractor binary was running in that example? The `bazel_go_extractor` binary does the mnemonic check, but `extract_kzip` does not.

jaysachs:
> Which extractor binary was running in that example? The `bazel_go_extractor` binary does the mnemonic check, but `extract_kzip` does not.

Tracing it from `kythe/extractors/bazel/extractors.bazelrc` to `kythe/release/release.BUILD`, it sure seems like it was `bazel_go_extractor`

PR #4105, opened 2019-09-26
feat(typescript): Support AngularTemplateCompile mnemonic

salguarnieri:
The ng_module produces actions with the AngularTemplateCompile mnemonic (not
TypeScriptCompile). The two actions should both produce extra action data that
can be processed by the same extraction config.


evmar:
I'm not sure AngularTemplateCompile is available anywhere public, but it seems harmless to me to include.

salguarnieri:
@evmar I'm not sure what you mean by public, but it looks like it is here: https://github.com/angular/angular/blob/master/packages/bazel/src/ng_module.bzl#L399

PR #4114, opened 2019-10-01
feat(extraction): add proto definition for VName rewrite rules

schroederc:


kythe/go/util/vnameutil/rewrite.go
```diff
@@ -163,13 +131,17 @@ func (p *pattern) UnmarshalJSON(data []byte) error {
 // both ends.  The template strings may contain markers of the form @n@, that
 // will be replaced by the n'th regexp group on a successful input match.
 func ParseRules(data []byte) (Rules, error) {
-	var rr []*rewriteRule
+	var rr []*spb.VNameRewriteRule
 	if err := json.Unmarshal(data, &rr); err != nil {
```
shahms: Shouldn't this use the jsonpb unmarshalling rather than "encoding/json"? Is the proto generated code guaranteed to be compatible with "encoding/json"?

schroederc: This is probably better.

PR #4136, opened 2019-10-08
fix(java_indexer): use Path to resolve metadata (fixes #4135)

### LINKED ISSUE #4135: fix improperly "relativized" paths in JavaIndexer
`JavaFileStoreBasedFileManager` disregarded the compilation unit working_directory and had poor support for absolute file paths and as a result there is a variety of code which expects to be able to make a "relative" path by removing the leading "/" if any.

When using `CompilationUnitPathFileManager` with a non-root working_directory, these lookups fail as they end up with the CU working directory specified twice.

shahms:


shahms:
The arc-lint failure appears to be google-java-format version skew? I can't reproduce.

kythe/java/com/google/devtools/kythe/platform/shared/ProtobufMetadataLoader.java
```diff
@@ -81,10 +83,11 @@ public ProtobufMetadataLoader(
   /** @return a function that looks up the VName for some filename in the given CompilationUnit. */
   private static Function<String, VName> lookupVNameFromCompilationUnit(CompilationUnit unit) {
     HashMap<String, VName> map = new HashMap<>();
+    Path root = Paths.get(unit.getWorkingDirectory().equals("") ? "/" : unit.getWorkingDirectory());
```
schroederc: `.isEmpty()`

schroederc: Or would `Paths.get("/", unit.getWorkingDirectory())` work?

shahms: That should work for any valid (non-relative) working directory, done.

PR #4140, opened 2019-10-09
fix: EntrySet::emitted becomes volatile

jaysachs:
We've seen warnings emitted from finalize() in non-error cases, suspicion is that it may be caused by memory barrier propagation (finalize runs in a different thread).

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/base/EntrySet.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/base/EntrySet.java
@@ -60,7 +60,7 @@ public final class EntrySet {
   private final ImmutableMap<String, byte[]> properties;
 
   /** Used to detect when an {@link EntrySet} was not emitted exactly once. */
-  private boolean emitted;
+  private volatile boolean emitted;
 
   protected EntrySet(
       VName source,


kythe/java/com/google/devtools/kythe/analyzers/base/EntrySet.java
```diff
@@ -60,7 +60,7 @@
   private final ImmutableMap<String, byte[]> properties;
 
   /** Used to detect when an {@link EntrySet} was not emitted exactly once. */
-  private boolean emitted;
+  private volatile boolean emitted;
```
creachadair: It'd be useful to document _why_ you made this change. I'm assuming you have multiple threads sharing a reference to this value and hit some oddball corner writing it out. Since it's clearly designed to act as an immutable type, this bit of mutable state is weird enough that it deserves some documentation IMO.

Arguably it would be better to cache emitted references in the emitter, there at least you already have side-effects. Pretending to be immutable and then squirreling away a mutable bit is a very C++ approach.

jaysachs: I tried to explain why in the PR description. Basically, we saw the warnings in the logs, and all creations of EntrySet were directly followed by emit calls (or directly enough that exceptions thrown would suggest catastrophic failure, e.g. OOM, and no sign was seen of those). So my hypothesis is that the finalize() is running in a different thread, and without any synchronization (or volatile), the mutation to `emitted` may not have propagated across the memory barrier. Once I have confirmation (or lack thereof) I'll come back and comment this or remove it.

PR #4150, opened 2019-10-18
chore(oss): Record required and source files

salguarnieri:
Record both the number of required and source files so we can better understand the difference between two kzips


kythe/go/platform/kzip/info/info.go
```diff
@@ -47,29 +52,34 @@ func KzipInfo(f kzip.File, scanOpts ...kzip.ScanOption) (*apb.KzipInfo, error) {
...
+		kzipInfo.TotalSourceFiles += int32(srcs.Len())
+		corpusInfo(cuCorpus).Units[u.Proto.GetVName().GetLanguage()]++
 
 		for _, ri := range u.Proto.RequiredInput {
-			kzipInfo.TotalFiles++
+			kzipInfo.TotalRequiredInputFiles++
 			// Determine language from the unit VName (note that file VNames are
 			// forbidden from specifying a language).
-			corpusInfo(ri.GetVName().GetCorpus()).Files[u.Proto.GetVName().GetLanguage()]++
+			corpusInfo(ri.GetVName().GetCorpus()).RequiredInputFiles[u.Proto.GetVName().GetLanguage()]++
```
schroederc: This is weird.  We're counting source files toward the compilation unit's corpus, but we're counting required inputs toward their specific VName corpus.  Nowhere do we record the number of required inputs for the compilation unit's corpus and the difference isn't documented anywhere.

salguarnieri: It is odd. I tried to make it less odd by having a field that counts the CUs and the sources and required inputs in them and a different field to count the required inputs by their own corpus in their VName.

kythe/proto/analysis.proto
```diff
@@ -188,16 +188,30 @@ message KzipInfo {
...
+    message CompilationUnits {
+      // Number of compilation units that have a specific corpus in their VName.
+      int32 count = 1;
+      // Number of source files mentioned in these compilation units.
+      int32 sources = 2;
+      // Number of required inputs mentioned in these compilation units.
+      int32 required_inputs = 3;
+    }
+
+    map<string, CompilationUnits> compilation_units = 3;
```
schroederc: Comment the shape of the keys to this map.

kythe/proto/analysis.proto
```diff
@@ -188,16 +188,30 @@ message KzipInfo {
...
+    // Required input files that have their corpus set to this in their VName.
+    // It is possible for a required input to have a different corpus than the
+    // compilation unit it is mentioned in.
+    map<string, int32> required_inputs = 4;
   }
 
   map<string, CorpusInfo> corpora = 1;
```
schroederc: Comment the shape of the keys to this map.

kythe/proto/analysis.proto
```diff
@@ -188,16 +188,30 @@ message KzipInfo {
...
+      int32 sources = 2;
+      // Number of required inputs mentioned in these compilation units.
+      int32 required_inputs = 3;
+    }
+
+    map<string, CompilationUnits> compilation_units = 3;
+    // Required input files that have their corpus set to this in their VName.
+    // It is possible for a required input to have a different corpus than the
+    // compilation unit it is mentioned in.
+    map<string, int32> required_inputs = 4;
```
schroederc: Comment the shape of the keys to this map.

kythe/go/platform/kzip/info/info.go
salguarnieri: discussed offline and cleaned up to only count corpus and language from the CU.

kythe/proto/analysis.proto
salguarnieri: done

kythe/proto/analysis.proto
salguarnieri: done

kythe/proto/analysis.proto
salguarnieri: done

PR #4156, opened 2019-10-23
chore: Count sources by their required input vname corpus

salguarnieri:


kythe/proto/analysis.proto
```diff
@@ -188,25 +188,27 @@ message KzipInfo {
...
-      // Number of required inputs mentioned in these compilation units.
-      int32 required_inputs = 3;
     }
 
     // Map from compilation unit language to stats about compilation units with
     // that language.
     map<string, CompilationUnits> language_compilation_units = 3;
 
     message RequiredInputs {
+      // TODO(salguarnieri) Add count of unique inputs for this corpus?
```
justbuchanan: If we de-duplicate the inputs to get a unique count, is there reason to keep the old counter around too?

If we only need the one number, could we replace the `RequiredInputs` messages with just an int?

kythe/proto/analysis.proto
```diff
@@ -188,25 +188,27 @@ message KzipInfo {
     // The count of data present in compilation units for a given corpus and
     // language.
     message CompilationUnits {
```
justbuchanan: Do we need this wrapper message anymore?

salguarnieri: I'd like to keep it so if we decide we want to record more info about the CU in the future we don't have to make a new map again

salguarnieri: But if you prefer I can get rid of it

justbuchanan: :+1:

PR #4164, opened 2019-10-27
refactor(typescript_indexer): allow to pass multiple files in indexer test

nbeloglazov:
Today TS test runs test files one at a time. This makes it impossible to
test scenarios that check references between multiple files. For example
when TS file requires module declared in `.d.ts` file. In this scenario
both `.ts` and `.d.ts` files must be passed togeter to indexer and
verifier.

This commit changes the way how test files are scanned by test. Now
test files passed as array instead of single file. All `.ts` files
immediately places inside testdata/ are still passed as array of single
files. Files in subfolders of testdata/ are passed as single array, one
per subfolder.

### DIFF
--- a/kythe/typescript/test.ts
+++ b/kythe/typescript/test.ts
@@ -25,6 +25,7 @@
 
 import * as assert from 'assert';
 import * as child_process from 'child_process';
+import * as fs from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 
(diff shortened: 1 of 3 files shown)


nbeloglazov:
@evmar can you take a look?

nbeloglazov:
I updated node typings to newer and now all tests pass.

PR #4169, opened 2019-10-28
chore: run `npm update`

justbuchanan:
this gets us mem>=4.0.0, which should silence the security alert on the main page


salguarnieri:
There are some buildbot errors here

justbuchanan:
I fixed most of it by installing eslint and eslint-plugin-compat, but it's still not quite happy.

PR #4173, opened 2019-10-30
docs: Add cmake extraction example

salguarnieri:


salguarnieri:
#4173 for the mkdir problem

#4175 for hiding cmake problems

Updated the comments about the env. vars

I thought about adding the merge step, but since the other examples also don't merge, I left it out.

PR #4177, opened 2019-10-30
feat(schema): support the generates relation between file nodes

schroederc:


kythe/docs/schema/schema.txt
```diff
@@ -610,7 +610,8 @@ and target languages do not share Kythe indexers, it is still semantically
 useful to connect the nodes they emit. For example, one might want to list
 all the $$C++$$ and Java uses of a particular service call, starting at the
 specification of that service. The specification and its generated artifacts
-may be joined by the *generates* edge.
+may be joined by the *generates* edge.  Both semantic nodes and their enclosing
```
zrlk: I think that files can generate files and semantic nodes can generate semantic nodes, but I'm not sure if we should relate files with semantic nodes in this way.

kythe/docs/schema/schema.txt
```diff
@@ -610,8 +610,8 @@ and target languages do not share Kythe indexers, it is still semantically
 useful to connect the nodes they emit. For example, one might want to list
 all the $$C++$$ and Java uses of a particular service call, starting at the
 specification of that service. The specification and its generated artifacts
-may be joined by the *generates* edge.
-
+may be joined by the *generates* edge.  Either both the specification and its
+generated artifact are file nodes or both are semantic nodes.
```
hildrum: Note that for java, at least, sometimes a particular message in a proto is what generates a particular file.

zrlk: With this change, the message generates the class that implements the message. The file the message is contained in generates the file containing that class definition.

schroederc: In the future, we may investigate loosening up this constraint so that we paint edges from semantic nodes to their specifically generated files, but for now, we're just adding support for the file-to-file edges.

hildrum: SGTM.  The file-to-file edges are definitely the more interesting ones.

PR #4179, opened 2019-10-30
fix(jvm extractor): use class files as srcs, not jar

justbuchanan:
Previously the jar file was listed as the `source_file`, but had no corresponding entry in `required_input`. The jvm extractor scans the jar and records each contained `.class` file as a `required_input`. This change lists each `.class` file in `source_file` instead of the jar.


justbuchanan:
Tested with:

```shell
bazel build \
  --experimental_extra_action_top_level_only \
  --experimental_action_listener=//kythe/extractors:extract_kzip_jvm \
  //third_party/javac:javac
```

PR #4192, opened 2019-11-07
feat(bazel extractor): put everything in the same corpus by default

justbuchanan:
Provide a flag for using the old behavior, which maps external projects to their own corpora:

`--define kythe_assign_external_projects_to_separate_corpora=true`

I'm open to better names for this.

### DIFF
--- a/kythe/data/simple_vnames.json
+++ b/kythe/data/simple_vnames.json
@@ -1,13 +1,6 @@
 [
   {
-    "pattern": "bazel-out/[^/]+/bin/external/[^/]+/(.+)",
-    "vname": {
-      "corpus": "CORPUS",
-      "path": "@1@"
-    }
-  },
-  {
-    "pattern": "bazel-out/[^/]+/([^/]+)/external/[^/]+/(.+)",
+    "pattern": "bazel-out/[^/]+/([^/]+/external/[^/]+)/(.+)",
     "vname": {
       "corpus": "CORPUS",
       "root": "bazel-out/@1@",
--- a/kythe/data/simple_vnames.json
+++ b/kythe/data/simple_vnames.json
@@ -1,0 +1,32 @@
+[
+  {
+    "pattern": "bazel-out/[^/]+/bin/external/[^/]+/(.+)",
+    "vname": {
+      "corpus": "CORPUS",
+      "path": "@1@"
+    }
+  },
+  {
+    "pattern": "bazel-out/[^/]+/([^/]+)/external/[^/]+/(.+)",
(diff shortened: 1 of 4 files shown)


kythe/data/simple_vnames.json
```diff
@@ -0,0 +1,32 @@
+[
+  {
+    "pattern": "bazel-out/[^/]+/bin/external/[^/]+/(.+)",
+    "vname": {
+      "corpus": "CORPUS",
+      "path": "@1@"
```
schroederc: These files do not exist in the repository (and may have conflicting paths) so we need to put a `root` on them.  It looks like the next rule will cover this case and we can just delete this rule.

kythe/data/simple_vnames.json
```diff
@@ -0,0 +1,32 @@
...
+    }
+  },
+  {
+    "pattern": "bazel-out/[^/]+/([^/]+)/external/[^/]+/(.+)",
+    "vname": {
+      "corpus": "CORPUS",
+      "root": "bazel-out/@1@",
```
schroederc: Make it `bazel-out/@1@/external/@repo_name@` so they don't conflict with the repository's generated files and other external repo files.

kythe/data/simple_vnames.json
justbuchanan: Removed.

kythe/release/release.BUILD
```diff
@@ -11,14 +11,28 @@ exports_files([
...
     name = "vnames_config",
-    srcs = [
-        "vnames.cxx.json",
-        "vnames.go.json",
-        "vnames.java.json",
-        "vnames.json",
-    ],
+    srcs = select({
+        "//conditions:default": [
+            # by default, the simple vname rules are used, which map everything
```
schroederc: The defaults feel a little weird.  It seems like we actually default to the other branch because kythe_assign_external_projects_to_separate_corpora is defaulted to true.

Maybe switch this around?  (Make the `"//conditions:default"` be the actual default of the preexisting behavior and then have a new `kythe_singleton_corpus` variable that defaults to false and trigger the simple_vnames.json usage.)

justbuchanan: kythe_assign_external_projects_to_separate_corpora should be defaulted to false/unset.

I think most people that use this will want the single-corpus behavior, so I'd have a slight preference for that behavior being the default if no extra flags are specified. Thoughts?

If you think a `kythe_singleton_corpus` flag is better, I'm fine with changing to that.

PR #4194, opened 2019-11-07
feat(java extractor): Include system modules in the compilation unit

salguarnieri:
When system modules are specified with -system <dir> there are two files that
the java compiler needs to read: dir/lib/jrt-fs.jar and dir/lib/modules.
Capture these files in the compilation unit so the indexer can access
them.

This clones a similar change by @asmundak and adds a test


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -252,6 +252,7 @@ public CompilationDescription extract(
       Iterable<String> sourcepath,
       Iterable<String> processorpath,
       Iterable<String> processors,
+      Optional<String> systemDir,
```
shahms: I'd really, really like to avoid the proliferation of arguments in this already-too-long argument list.

Can you just make this a public member function on the JavaCompilationUnit extractor instead?

`useSystemDirectory()` or similar?

salguarnieri: done

PR #4202, opened 2019-11-13
chore(llvm): update to 5b9e4daef06dcfefc786737a32c8bbb5bd0fc5c4

shahms:
This also "fixes" 4/5 failing ObjectiveC unit tests by adding `ignore_dups = True`.  The remaining failure I'm unsure of, but would be much obliged if you could look into.


shahms:
Remaining tests fixed by correcting the SourceRange used in ObjCMessageExpr.  Before, if the expression wasn't brace-enclosed (like with a property access) we would generate a range (<begin>, <invalid>).  We now adjust the end range in the braced-case and allow ExpandedRangeInCurrentContext to adjust the unbraced case.

PR #4204, opened 2019-11-14
feat(jvm_indexer): allow JVM nodes to be refined by corpus/root/path

schroederc:
It may be desirable to separate JVM nodes by corpus (and possibly
root/path).  This is important when mixing two similar corpora that
defines many of the same JVM entities that should not be considered
identical.  It is left up to the indexer whether to incorporate a
corpus/root/path into their emitted JVM nodes.


schroederc:
> When would a corpus/root be insufficient? What's the use case for permitting path?

Duplicate classes within the same corpus.  Maybe a build is using multi-release jars or has lite/uber version of a class for different clients.  We don't have to use paths, but there's no reason to exclude them.

PR #4207, opened 2019-11-15
release: build proto metadata plugin from source

justbuchanan:
The c++ code generated by the plugin #includes some headers from the
protobuf "runtime", which is required to be of the same protobuf version
as the generated code. Previously, we shipped the plugin in binary form,
which required clients to use the same version of protobuf as our
release was compiled at. This version constraint ended up being a
hassle, so we're shipping the plugin in source form instead so it will
be built against the same version of protobuf as the client project.

This unfortunately requires editing the client WORKSPACE file to add the
kythe_release repository and run an included bazel macro for retrieving
dependencies.

```
local_repository(
    name="kythe_release",
    path="/path/to/kythe/kythe-v0.0.35",
)

load("@kythe_release//:external.bzl", "kythe_release_dependencies")
kythe_release_dependencies()
```


salguarnieri:
Can you add a script to our release that adds those lines to a workspace file?

kythe/cxx/tools/BUILD
```diff
@@ -41,3 +41,10 @@ cc_binary(
         "@com_google_protobuf//:protoc_lib",
     ],
 )
+
+filegroup(
+    name = "proto_metadata_plugin_source",
+    srcs = [
+        "proto_metadata_plugin.cc",
```
shahms: you should just be able to `exports_files(["proto_metadata_plugin.cc"])`

kythe/release/BUILD
```diff
@@ -69,6 +70,7 @@ genrule(
         "$(locations misc)",
         "--cp $(location release.BUILD) BUILD",
         "--cp $(location release.WORKSPACE) WORKSPACE",
+        "--cp $(location external.bzl) ./",
```
shahms: we should ensure that these are kept in sync with the Kythe dependencies, rather than being duplicated.

shahms:
They should only need to run that if they don't already depend on `com_google_protobuf` (which they will if they're using proto) and `absl` (I think protobuf should go this dependency as well in the not too distant future). I can remove the glog dependency easily enough (although, again, I think that's being moved to absl in the not too distant future).

justbuchanan:
I spent some more time on this hoping to find a way around having clients modify their WORKSPACEs in order to run extraction. I'm not really seeing any good solutions here though. I went ahead and removed the glog dependency so it's slightly lighter.

As @shahms points out, client projects only need to modify their WORKSPACE if they don't already have the abseil and protobuf as dependencies. With the gvisor project for example, no changes are needed to the extract command or the WORKSPACE file since they already have those dependencies.

Additionally, extracting bazel projects that don't use protobuf will not require adding these deps.

kythe/release/BUILD
justbuchanan: What's the best way to keep things in sync? I could change the main repo's external.bzl to pull in this new one, but I think it gets a little hairy with some of the order-dependent calls to things like `rules_proto_dependencies()`, `rules_proto_toolchains()`, `protobuf_deps()`.

shahms: I think the easiest way is to source this (or an equivalent) file from //:external.bzl and call `kythe_release_dependencies` from there. Likely with a comment around what sorts of things it pulls in.

shahms:
Looking into this a bit more, I think #4234 is another possible solution

justbuchanan:
Closing in favor of https://github.com/kythe/kythe/pull/4234.

PR #4210, opened 2019-11-18
chore(java indexer): log filemanager exceptions

salguarnieri:
The filemanager sometimes logs errors to stderr without any context and that makes it more difficult than it should be to debug problems. Add an explicit log invocation in all methods called via reflection. We could add the log to non-reflective calls as well, but that would be a more invasive change.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/java/filemanager/ForwardingStandardJavaFileManager.java
+++ b/kythe/java/com/google/devtools/kythe/platform/java/filemanager/ForwardingStandardJavaFileManager.java
@@ -297,6 +297,10 @@ public class ForwardingStandardJavaFileManager
       String methodName, ReflectiveOperationException error, Class<IOException> declaredType)
       throws IOException {
     if (error instanceof InvocationTargetException) {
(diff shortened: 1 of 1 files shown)


kythe/java/com/google/devtools/kythe/platform/java/filemanager/ForwardingStandardJavaFileManager.java
```diff
@@ -297,6 +297,10 @@ private static IllegalStateException propagateInvocationTargetErrorIfPossible(
       String methodName, ReflectiveOperationException error, Class<IOException> declaredType)
       throws IOException {
     if (error instanceof InvocationTargetException) {
+      // Log the exception because the propagated destination may not provide a nice error log.
+      logger.atWarning().withCause(error).log(
```
shahms: this should be `error.getCause()`.  `InvocationTargetException` is the type used to wrap non-reflective exceptions originating in the underlying method.

salguarnieri: done

PR #4211, opened 2019-11-18
chore(java indexer): Handle null pointer

salguarnieri:
methodDecl.sym might be null; log if it is and avoid a NPE.


shahms:
JFYI, I have seen a few cases of these NPEs as a result of incompatibilities around modules as well as differences between JRE system modules and those being compiled.

PR #4221, opened 2019-11-24
fix(java indexer): fix startsWith check

salguarnieri:
The temporary directory may be passed in as a relative directory, so convert both parts of the startsWith check to an absolute path.


kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
```diff
@@ -194,9 +194,11 @@ private Path getPath(String path, String... rest) {
     // If this is a path underneath the temporary directory, use it. This is required for --system
     // flags to work correctly.
     Path local = Paths.get(path, rest);
-    if (temporaryDirectory != null && local.startsWith(temporaryDirectory)) {
-      logger.atFine().log("Using the filesystem for temporary path %s", local);
-      return local;
+    if (temporaryDirectory != null) {
+      if (local.toAbsolutePath().startsWith(temporaryDirectory.toAbsolutePath())) {
+        logger.atInfo().log("Using the filesystem for temporary path %s", local);
+        return local;
```
shahms: We should just store the temporaryDirectory as an absolute path, rather than doing this conversion every time.

PR #4223, opened 2019-11-25
chore(java indexer): Use optional for possibly empty values

salguarnieri:
A few values from compilation details may be null so return an optional so it
is easier to make sure the clients use them correctly.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/java/JavaCompilationDetails.java
+++ b/kythe/java/com/google/devtools/kythe/platform/java/JavaCompilationDetails.java
@@ -180,8 +180,8 @@ public class JavaCompilationDetails implements AutoCloseable {
   }
 
   /** Returns the AST for the current analysis target. */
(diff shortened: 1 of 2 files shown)


kythe/java/com/google/devtools/kythe/platform/java/JavaCompilationDetails.java
```diff
@@ -179,8 +180,8 @@ public JavacTask getJavac() {
   }
 
   /** Returns the AST for the current analysis target. */
-  public Iterable<? extends CompilationUnitTree> getAsts() {
-    return asts;
+  public Optional<Iterable<? extends CompilationUnitTree>> getAsts() {
```
jaysachs: Do we really need to distinguish "not there" from "empty iterable"? Why not

  return asts == null ? ImmutableList.of() : asts;

salguarnieri: 👍

PR #4233, opened 2019-12-02
fix(bazel extractor): give external paths a root

justbuchanan:
Paths under 'external/' correspond to external dependencies pulled in
via the WORKSPACE file. Because these files are not present in the main
repo, they must be given a 'root'.

As an example, string_view.h from abseil was previously given a vname like:

    path: "external/com_google_absl/absl/strings/string_view.h"

After this change, the vname is:

    root: "external/com_google_absl"
    path: "absl/strings/string_view.h"

### DIFF
--- a/kythe/data/simple_vnames.json
+++ b/kythe/data/simple_vnames.json
@@ -16,6 +16,14 @@
     }
   },
   {
+    "pattern": "external/([^/]+)/(.*)",
+    "vname": {
+      "corpus": "CORPUS",
+      "root": "external/@1@",
+      "path": "@2@"
+    }
+  },
+  {
     "pattern": "(.*)",
     "vname": {
       "corpus": "CORPUS",


jaysachs:
Should we just make one root for all the external repos?

justbuchanan:
@jaysachs I'd have a slight preference for keeping the external repo name somewhere in the vname because it makes it easier to trace where a file came from and might prevent vname collisions if the same path shows up in multiple repos.

jaysachs:
> @jaysachs I'd have a slight preference for keeping the external repo name somewhere in the vname because it makes it easier to trace where a file came from and might prevent vname collisions if the same path shows up in multiple repos.

I was imagining preserving that portion in the path.

justbuchanan:
That sounds fine to me. So for the string_view.h example, the vname would look like:

    root: "external"
    path: "com_google_absl/absl/strings/string_view.h"

salguarnieri:
I think path has generally been the same as what you'd see in a c++ include, would making the path not relative to the root break anything?

jaysachs:
> I think path has generally been the same as what you'd see in a c++ include, would making the path not relative to the root break anything?

Maybe. At this point I think I'm fine with the code as-is (root = "external/com_google_absl"). However we may end up coming back and revisiting/refining this.

PR #4234, opened 2019-12-02
fix(extractors): use distinct proto meta files for cc_proto_library

shahms:
It appears as though the `--cc_proto_library_header_suffixes` flag works for the .meta files (after manual testing), thus we can use distinct .meta files generated by whatever the source tree in question expects, rather than the version used to compile our release.

### DIFF
--- a/kythe/extractors/BUILD
+++ b/kythe/extractors/BUILD
@@ -45,6 +45,21 @@ proto_lang_toolchain(
     runtime = "@com_google_protobuf//:protobuf",
 )
 
+# Alternatively, if the plugin doesn't work you can use the default code generator
+# to output the metadata into a separate file.  This needs to be invoked with:
+#
+# bazel build \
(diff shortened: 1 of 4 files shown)


justbuchanan:
Thanks! I like this approach much better than requiring client repos to compile the plugin from source.

justbuchanan:
Do we still need the plugin for anything if this approach works?

justbuchanan:
The new toolchain needs to be added to `kythe/release/release.BUILD` as well.

shahms:
> Do we still need the plugin for anything if this approach works?

Users should not, but the plugin is used by some unittests as you can't customize bazel flags from bazel rules :-)

The embedded metadata is still my preferred approach, but can't be done without generating the entire file with the current structure of the proto rules and proto compiler.

shahms:
Update, PTAL?

PR #4240, opened 2019-12-03
fix(cxx_tools): replace sandboxed directory with bazel execroot

### LINKED ISSUE #4238: Non-existing directory in the compilation database
I'm trying to use Kythe in order to generate compilation database JSON file from Bazel, so I could use `clang-tidy`.

I'm building target with the following option:
```bash
bazel build --experimental_action_listener=@io_kythe//kythe/cxx/tools/generate_compile_commands:extract_json //:hello-world
```
Then I run the following:
https://github.com/kythe/kythe/blob/77e1360beca7903bc97728097d2c5e1b82204092/tools/cpp/generate_compilation_database.sh#L14-L19

I'm tying to analyze file with:
```bash
clang-tidy -p "$(bazel info execution_root)" hello-world.cc
```
taken from:
https://github.com/kythe/kythe/blob/77e1360beca7903bc97728097d2c5e1b82204092/tools/clang_tidy/clang_tidy.sh#L11
but get the following error:
```
Skipping /opt/test/hello-world.cc. Compile command not found.
```

The `compile_commands.json` file generated by Kythe looks as follows:
```json
[{
"file":"hello-world.cc",
"directory":"/root/.cache/bazel/_bazel_root/87dce07284a1eb76b6a67b97c6a39dcd/sandbox/processwrapper-sandbox/232/execroot/test",
"command":"..."
}]
```
It seems that this file points to a `.../processwrapper-sandbox/$PID/` directory that doesn't exist after Bazel finished compiling.

shahms:
Fixes #4238

### DIFF
--- a/kythe/cxx/tools/generate_compile_commands/extract_compile_command.cc
+++ b/kythe/cxx/tools/generate_compile_commands/extract_compile_command.cc
@@ -67,20 +67,6 @@ std::string JoinCommand(const std::vector<std::string>& command) {
   return output;
 }
 
-std::string GetBuildDirectory() {
-  std::string build_dir;
-  {
-    char cwd[MAXPATHLEN];
-    build_dir = getcwd(cwd, MAXPATHLEN);
-  }
-  auto sandbox_start = build_dir.find("/bazel-sandbox/");
-  if (sandbox_start == std::string::npos) return build_dir;
-  auto sandbox_end =
-      build_dir.find("/", sandbox_start + strlen("/bazel-sandbox/"));
-
-  return build_dir.erase(sandbox_start, sandbox_end - sandbox_start);
-}
-
 std::string FormatCompilationCommand(const std::string& source_file,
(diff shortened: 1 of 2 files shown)


kythe/cxx/tools/generate_compile_commands/extract_compile_command.cc
```diff
@@ -89,7 +75,7 @@ std::string FormatCompilationCommand(const std::string& source_file,
   writer.Key("file");
   writer.String(source_file.c_str());
   writer.Key("directory");
-  writer.String(GetBuildDirectory().c_str());
+  writer.String("@BAZEL_ROOT@");
```
salguarnieri: Does this impact the cmake extractor? https://kythe.io/examples/#extracting-cmake-based-repositories

shahms: No, this command is an extra action used only by the generate_compile_commands.sh script.

PR #4242, opened 2019-12-04
fix(java_common): fixes and utilities for exploded system directories

shahms:
This lays the ground work for actually having the extractor expand the system directory into the compilation unit, rather than embedding the `lib/jrt-fs.jar` and `lib/modules` files, but I'm holding off actually submitting that due to the indexer bug fixed in this PR.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/SystemExploder.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/SystemExploder.java
@@ -57,20 +57,6 @@ final class SystemExploder {
     return walkSystemModules(wrap(fileManager));
   }
 
-  public static void main(String[] args) {
-    try {
-      if (args.length == 2) {
-        copySystemModules(args[0], args[1]);
-      } else if (args.length == 1) {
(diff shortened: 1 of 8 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/SystemExploder.java
```diff
@@ -0,0 +1,129 @@
...
+  // Returns a stream of paths from the specified --system directory, beginning with the appropriate
+  // modules subdirectory.
+  public static Stream<Path> walkSystemModules(String systemDir) throws IOException {
+    StandardJavaFileManager fileManager =
+        ToolProvider.getSystemJavaCompiler().getStandardFileManager(null, null, null);
+    fileManager.handleOption("--system", Iterators.singletonIterator(systemDir));
+    return walkSystemModules(wrap(fileManager));
+  }
+
+  public static void main(String[] args) {
```
salguarnieri: Move the main method to the bottom of the file near the other main-only methods.

PR #4247, opened 2019-12-08
fix(typescript_indexer): emit refs to TYPE nodes in export statements

nbeloglazov:
Some symbols like interfaces don't have VALUE nodes and only have TYPE
nodes. This commit fixes refs for such symbols.

Tested:
  unit test

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -1241,14 +1241,28 @@ class Visitor {
           console.error(`TODO: export ${exp.name} has no symbol`);
           continue;
         }
-        // TODO: import a type, not just a value.
         const remoteSym = this.typeChecker.getAliasedSymbol(localSym);
-        const kExport = this.host.getSymbolName(remoteSym, TSNamespace.VALUE);
-        this.emitEdge(this.newAnchor(exp.name), EdgeKind.REF, kExport);
-        if (exp.propertyName) {
-          // Aliased export; propertyName is the 'as <...>' bit.
-          this.emitEdge(
-              this.newAnchor(exp.propertyName), EdgeKind.REF, kExport);
+        const anchor = this.newAnchor(exp.name);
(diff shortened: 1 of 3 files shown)


kythe/typescript/indexer.ts
```diff
@@ -1241,14 +1241,25 @@ class Visitor {
           console.error(`TODO: export ${exp.name} has no symbol`);
           continue;
         }
-        // TODO: import a type, not just a value.
         const remoteSym = this.typeChecker.getAliasedSymbol(localSym);
+        const anchor = this.newAnchor(exp.name);
         const kExport = this.host.getSymbolName(remoteSym, TSNamespace.VALUE);
```
evmar: It's not guaranteed that the export has a value.  For example, interface and type alias.

You can check:
`(flags & ts.SymbolFlags.Value) !== 0`

Be careful to check on the actual symbol, and not the alias of it.

kythe/typescript/indexer.ts
```diff
@@ -1241,14 +1241,25 @@ class Visitor {
...
-        this.emitEdge(this.newAnchor(exp.name), EdgeKind.REF, kExport);
+        const kExportType =
+            this.host.getSymbolName(remoteSym, TSNamespace.TYPE);
+        this.emitEdge(anchor, EdgeKind.REF, kExport);
+        const flags =remoteSym.flags;
+        const isAlsoType = (flags & ts.SymbolFlags.Interface) ||
+            (flags & ts.SymbolFlags.Class) || (flags & ts.SymbolFlags.Enum);
```
evmar: I think this misses some cases of types, such as type aliases.

I think you want just
`(flags & ts.SymbolFlags.Type) !== 0`

nbeloglazov: Thanks! That's much better. Also added reexporting of type to the test.

nbeloglazov:
@evmar can you take another look?

PR #4254, opened 2019-12-10
feat(textproto): index contents of google.protobuf.Any fields

justbuchanan:


kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -217,6 +236,121 @@ Status TextprotoAnalyzer::AnalyzeMessage(
   return OkStatus();
 }
 
+namespace {
```
shahms: just put this in the extant anonymous namespace at the top of the file rather than opening a new one?

shahms: Although you probably just want to use a (possible injected) https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.type_resolver for this instead.

justbuchanan: Removed the anonymous namespace - the existing anonymous namespace actually already includes this and most of the rest of the file.

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -217,6 +236,120 @@ Status TextprotoAnalyzer::AnalyzeMessage(
   return OkStatus();
 }
 
+// Given a type url that looks like "type.googleapis.com/example.Message1",
+// returns "example.Message1".
+std::string ProtoMessageNameFromAnyTypeUrl(absl::string_view type_url) {
+  std::vector<std::string> parts = absl::StrSplit(type_url, "/");
+  if (parts.size() != 2) {
+    LOG(ERROR) << "Invalid Any type_url: '" << type_url << "'";
```
shahms: This will end up failing for Kythe type urls (which is our own fault, but still).  I would do something like:
// Return the substring from after the last '/' to the end or an empty string.
// If there is no slash, returns the entire string.
return std::string(type_url.substr(std::min(type_url.size(), type_url.rfind('/') + 1)));

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -217,6 +236,120 @@ Status TextprotoAnalyzer::AnalyzeMessage(
...
+// ref to the proto message descriptor. Ideally this information would be
+// provided in the ParseInfoTree generated by the textproto parser, but since
+// it's not, we do our own "parsing" with a regex.
+Status TextprotoAnalyzer::AnalyzeAnyTypeUrl(
+    const proto::VName& file_vname, const Descriptor& msg_descriptor,
+    TextFormat::ParseLocation field_loc) {
+  // Note that line is 1-indexed; a value of zero indicates an empty location.
+  if (field_loc.line == 0) return OkStatus();
+
+  re2::StringPiece sp(textproto_content_.begin(), textproto_content_.size());
```
shahms: should likely be `textproto_content_.data()`

kythe/cxx/indexer/textproto/analyzer.cc
justbuchanan: I tried changing the type url in one of my test cases to "kythe.io/proto/any_type_test.Message2" and it fails to parse:
`Error parsing text-format any_type_test.Message: 28:22: Expected "]", found "/"`.

It looks like the url we're using isn't actually legal textproto?

justbuchanan: (that error's from TextFormat::Parser, not the analyzer)

kythe/cxx/indexer/textproto/analyzer.cc
justbuchanan: done

kythe/cxx/indexer/textproto/analyzer.cc
shahms: It technically is, but also frequently fails to work with various combinations of language and format (it works from Go, for example, but fails without a custom TypeResolver in C++ when parsing a json proto, etc.).  In practice, it was a poor choice but technically fine.  Still, we should try to support it in our tools when possible ;-)

shahms: Sorry, "it technically is" should be "technically is fine".  As documented, any scheme-less URL should be fine there, including one that has more than a single slash, it's just that the C++ parser fails to follow the spec :-(

justbuchanan: Works for me. I swapped in your code snippet to take everything after the last slash.

justbuchanan: I'll need to adjust my regex to handle this too.

justbuchanan: Actually since I'm using TextFormat::Parser and it fails to parse with extra slashes, I can't really support it anyways :/

PR #4257, opened 2019-12-12
chore(java_common): allow for lazily reading file contents

shahms:
Rather than reading all file contents upfront, initiate the read on creation, but don't wait for it to complete until necessary.  This more closely follows the behavior of the legacy JavaFileObjects

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/shared/filesystem/ByteBufferByteChannel.java
+++ b/kythe/java/com/google/devtools/kythe/platform/shared/filesystem/ByteBufferByteChannel.java
@@ -23,7 +23,6 @@ import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.NonWritableChannelException;
 import java.nio.channels.SeekableByteChannel;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
@@ -31,10 +30,6 @@ import java.util.concurrent.Future;
 class ByteBufferByteChannel implements SeekableByteChannel {
   private Future<byte[]> buffer;
   private long position;
-
-  public ByteBufferByteChannel(byte[] data) {
-    this(CompletableFuture.completedFuture(data));
-  }
 
   public ByteBufferByteChannel(Future<byte[]> data) {
     buffer = data;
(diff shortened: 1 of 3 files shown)


kythe/java/com/google/devtools/kythe/platform/shared/filesystem/ByteBufferByteChannel.java
```diff
@@ -23,13 +23,20 @@
...
+import java.util.concurrent.Future;
 
-/** Read-only {@link SeekableByteChannel} backed by byte[] data. */
+/** Read-only {@link SeekableByteChannel} lazily backed by byte[] data. */
 class ByteBufferByteChannel implements SeekableByteChannel {
-  private byte[] buffer;
+  private Future<byte[]> buffer;
   private long position;
 
   public ByteBufferByteChannel(byte[] data) {
```
jaysachs: I don't think this constructor is used. I'd remove it, and the reference to CompetableFuture.

kythe/java/com/google/devtools/kythe/platform/shared/filesystem/ByteBufferByteChannel.java
```diff
@@ -55,7 +60,11 @@ public int read(ByteBuffer dst) throws IOException {
       return -1;
     }
     wanted = (int) Math.min(possible, wanted);
-    dst.put(buffer, (int) position, wanted);
+    try {
+      dst.put(buffer.get(), (int) position, wanted);
+    } catch (InterruptedException | ExecutionException exc) {
```
jaysachs: We could be "smarter" here and use get-with-timeout, and return 0 if it times out or interrupted. WDYT?

shahms: I don't want to be in the choosing arbitrary timeouts business here.

jaysachs: Ok, but at some level, all channels need to do that. In fact, you are choosing an arbitrary timeout -- +inf. But OK.

shahms: It's buried pretty far in the stack, doesn't have easy access to command line flags, can be explicitly canceled via `close` and any non-infinite timeout we might choose is going to be too short for some case or indistinguishable from infinite anyway.

Issue #4154, opened 2019-10-21
Java indexer and extractor should look for implicit .meta files


zrlk:
The Java indexer uses `javax.annotation.Generated` annotations to associate metadata to Java source files. The compilations for these source files depend on the JSR250 annotations (as Generated is included in that set). In some circumstances, adding this dependency is not desirable; it is also not always possible to add the dependency conditionally (e.g., only during extraction).

To allow metadata to be associated with Java source in these restricted environments, the Java indexer should look for a .meta file whenever it opens a .java source file. For example, when loading `d/A.java`, the indexer should check for a `d/A.java.meta`. Java extractors should also include these metadata files whenever they are present. Go metadata support works the same way.

If a Java source file still uses the `Generates` annotation, the metadata that the annotation specifies should still be loaded. If it's the same file as was loaded by the implicit metadata check, the file does not need to be loaded twice.

The `javax.annotation.processing.Generated` annotation was introduced in JDK9. Unfortunately, JDK9 is not available everywhere, and it is not clear whether using it would also require an additional dependency.

shahms:
As it's unspecified, `javax.annotation.processing.Generated` should also work, when it exists.

shahms:
We're also somewhat close to the indexer and extractor requiring JDK9+ as maintaining compatibility with JDK8 is a large PITA at this point.  While I don't have concrete plans for this at the moment, it's something I expect to happen sometime in 2020.

PR #4262, opened 2019-12-14
feat(java_indexer): support implicit protobuf metadata

zrlk:
This implements #4154 for protobuf metadata, as we're still unable to use
javax.annotation.processing.Generated.


kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
```diff
@@ -1376,23 +1380,54 @@ private JavaNode emitDiagnostic(TreeContext ctx, String message, String details,
...
+    URI uri = filePositions.getSourceFile().toUri();
+    String name = Paths.get(uri.getPath()).getFileName().toString();
+    try {
+      String fullPath = resolveSourcePath(name + ".pb.meta");
+      if (metadataFilePaths.contains(fullPath)) {
+        return;
+      }
+      FileObject file = Iterables.getOnlyElement(fileManager.getJavaFileObjects(fullPath), null);
+      if (file == null) {
+        // This is an expected case.
```
shahms: getJavaFileObjects() should throw or return the same number of elements passed in.  If it doesn't return exactly one element, something has gone seriously wrong and should probably throw.

kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -791,6 +796,17 @@ private AnalysisResults runJavaAnalysisToExtractCompilationDetails(
             public void finished(TaskEvent e) {
               if (e.getKind() == TaskEvent.Kind.PARSE) {
                 compilationUnits.add(e.getCompilationUnit());
+                try {
+                  String annotationPath =
+                      e.getCompilationUnit().getSourceFile().toUri().getPath() + ".pb.meta";
+                  for (JavaFileObject file :
+                      fileManager.getJavaFileForSources(Arrays.asList(annotationPath))) {
```
shahms: prefer the interface method: getJavaFileObjects

kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -791,6 +796,17 @@ private AnalysisResults runJavaAnalysisToExtractCompilationDetails(
...
+                    ((UsageAsInputReportingJavaFileObject) file).markUsed();
+                    ((UsageAsInputReportingJavaFileObject) file).markOptional();
```
shahms: This whole extra annotation is unnecessary if you just check whether or not the file exists before marking it as used.

kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
zrlk: It's my understanding that [getOnlyElement](https://guava.dev/releases/20.0/api/docs/com/google/common/collect/Iterables.html#getOnlyElement-java.lang.Iterable-T-) will throw if there is more than one element.

kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
zrlk: What's the recommended way to test this? FileObject#getLastModified?

shahms: `Files.exists(annotationPath)` I would think.

shahms: Or `Files.exists(Paths.get(annotationPath))`.  When we can actually use JDK9+ in source it gets a bit simpler as we don't have to round-trip through String/URI any more.

kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
shahms: That's fine too; you're converting exactly one string into a FileObject.  Returning either more or fewer than one FileObject indicates a serious problem which should likely throw.

shahms: https://guava.dev/releases/25.0-jre/api/docs/com/google/common/collect/Iterables.html#getOnlyElement-java.lang.Iterable-

Throws NoSuchElement on empty, IllegalArgumentException if more than one.

zrlk: The overload I'm using here returns null on empty.

shahms: Because you're passing it the null to return on empty, just omit that parameter and the null check.

zrlk: Omitting the null check causes tests to fail when getJavaFileObjects returns a single null value.

PR #4281, opened 2019-12-19
chore: update go dependencies, make it easier to automate

shahms:


tools/build_rules/shims.bzl
```diff
@@ -2,31 +2,14 @@ load("@bazel_gazelle//:deps.bzl", _go_repository = "go_repository")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 load("@io_bazel_rules_go//go:def.bzl", _go_binary = "go_binary", _go_library = "go_library", _go_test = "go_test")
 
-def go_repository(name, importpath, commit = None, custom = None, custom_git = None, tag = None, **kwargs):
+def go_repository(name, **kwargs):
     """Macro wrapping the Gazelle go_repository rule.  Works identically, except
```
schroederc: Update doc

external.bzl
```diff
@@ -612,6 +695,389 @@ def _go_dependencies():
...
         ],
     )
+    go_repository(
+        name = "co_honnef_go_tools",
+        importpath = "honnef.co/go/tools",
+        sum = "h1:3JgtbtFHMiCmsznwGVTUWbgGov+pVqnlf1dEJTNAXeM=",
+        version = "v0.0.1-2019.2.3",
+    )
+    go_repository(
+        name = "com_github_bazelbuild_rules_go",
```
schroederc: Why does our code need a copy of `io_bazel_rules_go`?  Ditto for the rest of these additional indirect deps.  Can we prune them since we probably don't need them all for a `//...` build?

BUILD
```diff
@@ -27,3 +28,9 @@ sh_test(
         ".bazelversion",
     ],
 )
+
+# gazelle:build_file_name BUILD
+# gazelle:exclude kythe
+# gazelle:exclude third_party
```
schroederc: Can we add more exclusions/patches to remove the spurious warnings gazelle emits?  Things like:

```
org_golang_x_tools: gazelle: /home/schroederc/.cache/bazel/_bazel_schroederc/ddace01a1ad38df359eb57115be59fad/external/org_golang_x_tools/cmd/fiximports/testdata/src/old.com/bad/bad.go: error reading go file: /home/schroederc/.cache/bazel/_bazel_schroederc/ddace01a1ad38df359eb57115be59fad/external/org_golang_x_tools/cmd/fiximports/testdata/src/old.com/bad/bad.go:2:43: expected 'package', found 'EOF'
```

external.bzl
shahms: I wrote a dummy rule which will cause gazelle to ignore the go.mod dependency.

tools/build_rules/shims.bzl
shahms: done.

BUILD
shahms: Not here, but I've added it to the appropriate rule.

external.bzl
shahms: Note that the indirect dependencies are still present, presumably because they are used by the build files gazelle will generate for those (and thus need to have build files themselves).

external.bzl
```diff
@@ -272,334 +275,424 @@ def _java_dependencies():
...
         name = "com_github_golang_protobuf",
         build_file_proto_mode = "disable_global",
-        custom = "protobuf",
         importpath = "github.com/golang/protobuf",
         patch_args = ["-p1"],
-        patches = ["@io_bazel_rules_go//third_party:com_github_golang_protobuf-extras.patch"],
-        tag = "v1.3.1",
+        patches = [
+            "@io_bazel_rules_go//third_party:com_github_golang_protobuf-extras.patch",
+            "@//third_party/go:new_export_license.patch",
```
achew22: Due to the leading `@` this will pull the patch from the repo that loads kythe in. This would force someone who loads kythe as a dependency to vendor this patch file into the repo. It might be preferable to omit the `@` and just load it from` @io_kythe`.

Here and elsewhere

shahms: https://github.com/kythe/kythe/pull/4282

PR #4284, opened 2019-12-20
fix(java_indexer): enable readahead in new filemanager

shahms:
This appears to address the remaining performance differences between the new filemanager and old.

Notes on implementation decisions:

- Rather than caching indefinitely, the ListenableFuture is removed from the cache when read.  This is because both the underlying FileObject and analyzer_driver cache the file data, so adding a third copy seemed wasteful.
- This could be made optional as it does over-estimate the files needed.  That overzealous reading is fairly minor, however, given the pruning done by the extractor, so I opted to avoid the additional complexity of plumbing through a flag, but can do so if desired.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
+++ b/kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
@@ -137,19 +137,19 @@ public final class CompilationUnitPathFileManager extends ForwardingStandardJava
   @Override
   public Iterable<? extends JavaFileObject> getJavaFileObjectsFromFiles(
       Iterable<? extends File> files) {
-    return Iterables.transform(super.getJavaFileObjectsFromFiles(files), this::readAhead);
+    return readAhead(super.getJavaFileObjectsFromFiles(files));
   }
 
   @Override
   @SuppressWarnings({"IterablePathParameter"})
   public Iterable<? extends JavaFileObject> getJavaFileObjectsFromPaths(
       Iterable<? extends Path> path) {
-    return Iterables.transform(super.getJavaFileObjectsFromPaths(path), this::readAhead);
+    return readAhead(super.getJavaFileObjectsFromPaths(path));
   }
 
   @Override
   public Iterable<? extends JavaFileObject> getJavaFileObjectsFromStrings(Iterable<String> names) {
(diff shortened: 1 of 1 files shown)


kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
```diff
@@ -273,4 +349,42 @@ private void setSystemOption(String value) throws IOException {
...
+    void readAhead(String path, String digest) {
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(digest), "digest is empty");
+      cache.computeIfAbsent(digest, k -> provider.startLookup(path, digest));
+    }
+
+    @Override
+    public ListenableFuture<byte[]> startLookup(String path, String digest) {
+      Preconditions.checkArgument(!Strings.isNullOrEmpty(digest), "digest is empty");
+      return cache.containsKey(digest) ? cache.remove(digest) : provider.startLookup(path, digest);
+    }
```
jaysachs: So I find the near-duplication of readHead/startLookup a little odd. I get that you're trying to eagerly empty the cache. 

Is it a safe assumption that most files are read no more than once?

As an alternative to the remove on the call to startLookup, I think I'd rather see a callback added to the future, and have it removed from the cache on future completion.

That in itself points out that this should probably have some synchronization included to manage the cache.

shahms: It depends on what you mean by "files are read no more than once".  The current indexer implementation reads a compilation unit's "source_files" at least once.

The callback approach seems to add unnecessary complexity to me; can you expand on the expected benefit?  There really isn't much code that can be shared between these two tiny functions without triggering an "error prone" failure for ignoring a returned future.

jaysachs: Adding the Callback is a more idiomatic use of Futures. As for the ErrorProne warning, I'd actually prefer the @SuppressWarning to the 2nd method.

Regardless, of that, though, you need to handle concurrency here.

shahms: Sure, I wasn't questioning the concurrency (although there isn't actually any concurrency involved outside of the Future, it's better to be safe).

PR #4298, opened 2020-01-08
fix(java_extractor): preserve dummy output path for modular builds

### LINKED ISSUE #3671: support java modules
#3607 was a first pass at this, but does so in a way that kills caching behavior on compilation units.

shahms:
Fixes #3671 for extraction (at least partially)

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
@@ -328,10 +328,25 @@ public class JavaCompilationUnitExtractor {
    * Returns a new list with the same options except header/source destination directory options.
    */
   private static ImmutableList<String> removeDestDirOptions(Iterable<String> options) {
-    // TODO(#3671): Option.D needs to remain in for module support, fix either here or in indexing.
+    // Option.D needs to remain for module support, but shouldn't point to a concrete path
+    // as those are often unique paths which destroy caching.
+    // In order to preserve the existing semantics for non-modular compilations and remain
+    // source-compatible with Java 8, check for the support and presence of a
+    // --module-source-path option.
+    EnumSet<Option> remove =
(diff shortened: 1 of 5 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -328,12 +328,27 @@ public CompilationDescription extract(
...
+    // source-compatible with Java 8, check for the support and presence of a
+    // --module-source-path option.
+    EnumSet<Option> remove =
+        hasModuleSourcePath(options)
+            ? EnumSet.of(Option.S, Option.H)
+            : EnumSet.of(Option.D, Option.S, Option.H);
     return ModifiableOptions.of(options)
-        .removeOptions(EnumSet.of(Option.D, Option.S, Option.H))
+        .removeOptions(remove)
+        .replaceOptionValue(Option.D, "/dev/null")
```
schroederc: Please document what the `/dev/null` does.

shahms: You mean aside from:
```
    // Option.D needs to remain for module support, but shouldn't point to a concrete path
    // as those are often unique paths which destroy caching.
```

schroederc: Yes, it's not clear what passing a file to a flag that requires a directory does.

shahms: Digging in a bit more, I think this needs to be handled in the file manager itself.  "Hilariously" javac check that each od -d, -h, and -s either:

1) aren't specified
2) are specified but don't exist
3) are specified, exist and are a directory

The existence and directory checks are done on the *real* filesystem unconditionally which means they shouldn't be present in the command line.

PR #4299, opened 2020-01-09
fix(java_indexer): use in-memory class_output path for modular builds

shahms:
Part of addressing #3671 in the indexer.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
+++ b/kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
@@ -25,6 +25,8 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Maps;
 import com.google.common.flogger.FluentLogger;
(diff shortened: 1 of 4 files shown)


kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
```diff
@@ -200,6 +205,8 @@ public boolean handleFileManagerOption(Option unused, String value) {
   @Override
   public void close() throws IOException {
     super.close();
+    fileSystem.close();
```
jaysachs: Hygienically these should all be in try/finallys, e.g.
```
boolean threw = true;
try { super.close(); threw = false; }
finally {
  try { Closeables.close(fileSystem, threw); threw = false; }
  finally { Closeables.close(memFileSystem, threw); }
}

but that's ugly and possibly overkill.

shahms: That was my first thought, but after your comment I decided to see if Guava had any helpful utilities for accomplishing this et voila!

PR #4303, opened 2020-01-17
feat(kzip_create): relativize input paths

shahms:


kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -229,6 +231,28 @@ func (cb *compilationBuilder) done() error {
 	return cb.out.Close()
 }
 
+func (cb *compilationBuilder) tryMakeRelative(path string) string {
```
schroederc: Document "relative to `WorkingDirectory` or $PWD" or returns the full path if not relative.

PR #4307, opened 2020-01-22
fix(java_indexer): only attempt to load implilcit metadata if it exists

shahms:
Without this the logs are filled with:
```
Jan 22, 2020 9:20:12 AM com.google.devtools.kythe.analyzers.java.KytheTreeScanner loadImplicitAnnotationsFile
WARNING: Can't read metadata for x-kythe-compilation://....
```
for *every* file.


kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
```diff
@@ -1423,7 +1423,7 @@ private void loadImplicitAnnotationsFile() {
         return;
       }
       FileObject file = Iterables.getOnlyElement(fileManager.getJavaFileObjects(fullPath));
-      if (file == null) {
+      if (file == null || !isFileReadable(file)) {
```
schroederc: Can you comment why a non-existent file may be returned from the file manager?

PR #4309, opened 2020-01-22
chore(kzip): expose Encoding of Reader

salguarnieri:


kythe/go/platform/kzip/kzip.go
```diff
@@ -197,6 +197,7 @@ func NewReader(r io.ReaderAt, size int64) (*Reader, error) {
 	if err != nil {
 		return nil, err
 	}
+	log.Printf("Reading kzip with %v encoding", pref)
```
schroederc: This seems like log spam.  Maybe expose the encoding as a method and log at the point of interest?

salguarnieri: done

PR #4311, opened 2020-01-23
typescript: depend on node v10 definition of types

evmar:
Node 10 is the current LTS version, and test.ts already depends on some
of its functionality.

### DIFF
--- a/kythe/typescript/yarn.lock
+++ b/kythe/typescript/yarn.lock
@@ -6,9 +6,10 @@
   version "2.8.8"
   resolved "https://registry.yarnpkg.com/@types/jasmine/-/jasmine-2.8.8.tgz#bf53a7d193ea8b03867a38bfdb4fbb0e0bf066c9"
 
-"@types/node@^7.0.4":
-  version "7.0.4"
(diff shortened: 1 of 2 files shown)


kythe/typescript/yarn.lock
```diff
@@ -6,9 +6,10 @@
   version "2.8.8"
   resolved "https://registry.yarnpkg.com/@types/jasmine/-/jasmine-2.8.8.tgz#bf53a7d193ea8b03867a38bfdb4fbb0e0bf066c9"
 
-"@types/node@^7.0.4":
-  version "7.0.4"
-  resolved "https://registry.yarnpkg.com/@types/node/-/node-7.0.4.tgz#9aabc135979ded383325749f508894c662948c8b"
+"@types/node@^10.15.3":
+  version "10.17.13"
```
jaysachs: 10.15.3 / 10.17.13 ? Is that expected?

isker: `^10.15.3` means "anything semver compatible with `10.15.3`".  So, this is not unexpected.

PR #4318, opened 2020-01-24
chore(cmake): log the command we are running

salguarnieri:
Logging the command we are running makes it easier to understand error messages that come from cmake.


kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
```diff
@@ -133,5 +133,6 @@ func cleanBuild(dir string) {
 func runIn(cmd *exec.Cmd, dir string) error {
 	cmd.Dir = dir
 	cmd.Stderr = os.Stderr
+	log.Printf("Running: %q in %q", cmd.String(), cmd.Dir)
```
shahms: .String() is unnecessary

salguarnieri: done

PR #4319, opened 2020-01-24
chore(cmake): Tell cmake we are clang

salguarnieri:
Set the compiler id in cmake to clang so repos that generate different commands based on the compiler ID will generate the correct commands for us.

### DIFF
--- a/kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
+++ b/kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
@@ -107,7 +107,9 @@ func (c *cmakeCommand) Execute(ctx context.Context, fs *flag.FlagSet, args ...in
 		return c.Fail("Unable to create build directory: %v", err)
 	}
 
-	if err := runIn(exec.CommandContext(ctx, "cmake", "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON", sourceDir), buildDir); err != nil {
+	// The Kythe extractor is clang based, so tell cmake that we are Clang so the right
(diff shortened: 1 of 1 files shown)


kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
```diff
@@ -107,7 +107,9 @@ func (c *cmakeCommand) Execute(ctx context.Context, fs *flag.FlagSet, args ...in
 		return c.Fail("Unable to create build directory: %v", err)
 	}
 
-	if err := runIn(exec.CommandContext(ctx, "cmake", "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON", sourceDir), buildDir); err != nil {
+	// The Kythe extractor is clang based, so tell cmake that we are Clang so the right
+	// commands are generated.
+	if err := runIn(exec.CommandContext(ctx, "cmake", "-DCMAKE_CXX_COMPILER_ID=Clang", "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON", sourceDir), buildDir); err != nil {
```
shahms: Does this actually effect out CMake *picks* a compiler or just which compiler it asserts is in use?   I couldn't find it documented, but I worry that if it's just the latter the project may fail to compile due to a mismatch.

salguarnieri: switched to tell cmake to actually use clang rather than just report our ID as clang. This should be safer and we can only support builds that would normally work with clang anyway, so assuming they have a clang binary installed seems ok.

PR #4323, opened 2020-01-27
feat(java_extractor): allow passing search paths as a map

shahms:
This changes semantics slightly, so I want to put it into a small and targeted change to keep the Chesterton's Fence damage radius small.  Notable semantic differences:

1) Single-empty-element search paths will be set on the file manager (previously, both [] and [""] would be excluded).
2) Paths are only set on the fileManager, rather than being set on the fileManager and injected into the command line.  This means that if those flags are present on the command line, they will override the specified values.  This shouldn't be a problem as non-Bazel extractors currently pull these search paths out of the command line and then re-inject them, but is something to watch out for.
3) Uses configured fileManager for the processor search path, rather than the iterables directly, but the semantics should be the same.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
@@ -292,7 +292,7 @@ public class JavaCompilationUnitExtractor {
     Preconditions.checkNotNull(options);
     Preconditions.checkNotNull(outputPath);
 
-    AnalysisResults results =
+    final AnalysisResults results =
         Iterables.isEmpty(sources)
             ? new AnalysisResults()
             : runJavaAnalysisToExtractCompilationDetails(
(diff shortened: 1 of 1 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -929,40 +943,12 @@ public void close() {
...
   private static ImmutableList<String> completeCompilerOptions(
-      UsageAsInputReportingFileManager fileManager,
-      Iterable<String> rawOptions,
-      Iterable<String> classpath,
-      Iterable<String> bootclasspath,
-      Iterable<String> sourcepath,
-      Iterable<String> processorpath,
-      Path tempDestinationDir)
-      throws ExtractionException {
+      Iterable<String> rawOptions, Path tempDestinationDir) throws ExtractionException {
```
jaysachs: put tempDestinationDir on new line?

shahms: Then I will get lint erros from google-java-format

PR #4337, opened 2020-01-30
feat(go_indexer): add generates edges for proto-generated files

jaysachs:


kythe/go/indexer/cmd/go_indexer/go_indexer.go
```diff
@@ -33,27 +33,30 @@ import (
...
 var (
-	doJSON         = flag.Bool("json", false, "Write output as JSON")
-	doLibNodes     = flag.Bool("libnodes", false, "Emit nodes for standard library packages")
-	doCodeFacts    = flag.Bool("code", false, "Emit code facts containing MarkedSource markup")
-	doAnchorScopes = flag.Bool("anchor_scopes", false, "Emit childof edges to an anchor's semantic scope")
-	metaSuffix     = flag.String("meta", "", "If set, treat files with this suffix as JSON linkage metadata")
-	docBase        = flag.String("docbase", "http://godoc.org", "If set, use as the base URL for godoc links")
-	verbose        = flag.Bool("verbose", false, "Emit verbose log information")
-	contOnErr      = flag.Bool("continue", false, "Log errors encountered during analysis but do not exit unsuccessfully")
+	deriveFileMetaEdgeKinds flagutil.StringList
```
schroederc: When would this be anything other than the empty list or `["/kythe/edge/generates"]`?

jaysachs: Don't have a specific case; just felt that a boolean was too restrictive, but happy to clamp it down to a boolean.

schroederc: I don't think even the `bool` is needed.  Why gate the behavior?

jaysachs: For C++ and Java, I was able to restrict the derivation of file-file generates edges to just the proto metadata. For Go, that wasn't possible (there's no proto metadata implementation externally).

I'm not attached to the gating, though.

schroederc: I don't think this should be a gated feature.

kythe/go/indexer/emit.go
```diff
@@ -933,6 +944,22 @@ func (e *emitter) writeRef(origin ast.Node, target *spb.VName, kind string) *spb
 		} else {
 			e.writeEdge(target, rule.VName, rule.EdgeOut)
 		}
+
+		if e.opts.emitDerivedLinkageFileGeneratesEdges() {
```
schroederc: This seems like the wrong location for this logic.  We're going to unnecessarily write duplicate edges between files because this loop is per reference.

kythe/go/indexer/emit.go
```diff
@@ -933,6 +944,22 @@ func (e *emitter) writeRef(origin ast.Node, target *spb.VName, kind string) *spb
...
+			if rule.VName.Path != "" && target.Path != "" {
+				var ruleVName = rule.VName
```
schroederc: This doesn't copy the VName.  We're mutating the underlying rule.

kythe/go/indexer/testdata/meta.go
```diff
@@ -13,4 +14,5 @@ func Foobar() {}
 //- FA.loc/end   25
 //- FA defines/binding Foobar
 //- Foobar.node/kind function
-//- _Alt=vname(gsig, gcorp, groot, gpath, glang) generates Foobar
+//- _Alt=vname(gsig, gcorp, groot, gpath, glang) generates Foobar?
+//- vname("", gcorp, groot, gpath, "") generates vname("", kythe, _, "go/indexer/metadata_test", "")
```
schroederc: This test should be expanded to include multiple definitions to catch edge duplication.

kythe/go/indexer/emit.go
jaysachs: Took the same approach as C++: kept the logic here but record that a file is visited.

kythe/go/indexer/cmd/go_indexer/go_indexer.go
jaysachs: Removed

kythe/go/indexer/emit.go
jaysachs: Fixed.

kythe/go/indexer/emit.go
```diff
@@ -933,6 +935,22 @@ func (e *emitter) writeRef(origin ast.Node, target *spb.VName, kind string) *spb
 		} else {
 			e.writeEdge(target, rule.VName, rule.EdgeOut)
 		}
+		if !e.fmeta[file] {
+			e.fmeta[file] = true
```
schroederc: I think we still want to do this only when `rule.EdgeOut == edges.Generates`.

jaysachs: Done

kythe/go/indexer/emit.go
```diff
@@ -38,7 +38,7 @@ import (
 	"golang.org/x/tools/go/types/typeutil"
 
 	cpb "kythe.io/kythe/proto/common_go_proto"
-	spb "kythe.io/kythe/proto/storage_go_proto"
+	stpb "kythe.io/kythe/proto/storage_go_proto"
```
schroederc: `spb` is the standard package name we use for this proto.

kythe/go/indexer/emit.go
```diff
@@ -935,7 +935,7 @@ func (e *emitter) writeRef(origin ast.Node, target *spb.VName, kind string) *spb
 		} else {
 			e.writeEdge(target, rule.VName, rule.EdgeOut)
 		}
-		if !e.fmeta[file] {
+		if rule.EdgeOut == "/kythe/edge/generates" && !e.fmeta[file] {
```
schroederc: Use the "kythe.io/kythe/go/util/schema/edges" package

PR #4340, opened 2020-01-31
Reuse splitters in KytheURI by making the private static final 

hildrum:


shahms:
can you run google-java-format on that file?

PR #4341, opened 2020-01-31
feat: allow excluding modules from openjdk11 extraction

shahms:
There are a few problematic modules at the moment, specifically: jdk.internal.vm.compiler, jdk.aot, and jdk.internal.vm.compiler.management.  While work is ongoing to fix them, in the meantime it we can start indexing the rest of the modules.


shahms:
Buildbot has successfully completed these checks but is not reporting status?

PR #4343, opened 2020-01-31
chore: only print errors if there are any

shahms:


kythe/extractors/openjdk11/extract/extract.go
```diff
@@ -296,5 +296,7 @@ func logCollectedErrors(stderr io.Reader) {
 	if err != nil {
 		log.Println(err)
 	}
-	log.Println("Error extracting modules:\n\t", strings.Join(errors, "\n\t"))
+	if len(errors) > 0 {
+		log.Println("Error extracting modules:\n\t", strings.Join(errors, "\n\t"))
```
salguarnieri: Are format strings supported for Println (not printf)?

salguarnieri: never mind, you didn't use a format string

shahms: No, but they also aren't in use here.

PR #4344, opened 2020-02-03
typescript: Make the typescript signatures more specific to their module

johnedmonds:
Made a few changes to the Typescript Indexer so it's easier to index code across different modules:

- Module signatures are now the module name instead of the word `module`.
- Anchor signatures now include the file path instead of just having the offsets
- Symbol signatures now include their module
- Module names are now relative to `KYTHE_ROOT_DIRECTORY`
- Corpus is now set to `KYTHE_CORPUS`

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -113,11 +113,11 @@ function toArray<T>(it: Iterator<T>): T[] {
 }
 
 /**
- * stripExtension strips the .d.ts or .ts extension from a path.
+ * stripExtension strips the .d.ts, .ts, or .js extension from a path.
  * It's used to map a file path to the module name.
  */
 function stripExtension(path: string): string {
-  return path.replace(/\.(d\.)?ts$/, '');
+  return path.replace(/\.(d\.)?(ts|js)$/, '');
 }
 
 /**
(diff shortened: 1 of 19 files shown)


shahms:
`Anchor signatures now include the file path instead of just having the offsets`

This is unnecessary; the anchor VName already includes the file path.  Putting it in the signature is redundant (and potentially confusing to humans).

kythe/typescript/indexer.ts
```diff
@@ -684,7 +706,12 @@ class Visitor {
 
     this.typeChecker = this.host.program.getTypeChecker();
 
-    this.kFile = this.newFileVName(file.fileName);
+    if (process.env.KYTHE_ROOT_DIRECTORY) {
+      this.kFile = this.newFileVName(
+          path.relative(process.env.KYTHE_ROOT_DIRECTORY, file.fileName));
```
shahms: These environment variables are only provided during extraction, not indexing.  The indexer should limit itself to command line parameters for global options and to the specific compilation unit for compilation-unit specific things, including corpus and especially anything VName related.

kythe/typescript/indexer.ts
```diff
@@ -1957,14 +1996,16 @@ function main(argv: string[]) {
 
   // This program merely demonstrates the API, so use a fake corpus/root/etc.
   const compilationUnit: VName = {
-    corpus: 'corpus',
+    corpus: process.env.KYTHE_CORPUS || 'corpus',
```
shahms: This should come from the compilation unit required_input VName for the source file(s).

shahms:
Cleaning up old PRs, feel free to re-open and address the comments if this is still relevant.

PR #4347, opened 2020-02-03
feat(tooling): add vnames utility for handling rewrite rules

schroederc:


kythe/go/util/tools/vnames/apply.go
```diff
@@ -0,0 +1,114 @@
...
+		return cmdErrorf("reading %q: %v", c.rulesPath, err)
+	}
+	s := bufio.NewScanner(os.Stdin)
+	m := jsonpb.Marshaler{OrigName: true}
+	for s.Scan() {
+		line := s.Text()
+		v, ok := rules.Apply(line)
+		if ok || c.allowEmpty {
+			if err := m.Marshal(os.Stdout, v); err != nil {
+				return cmdErrorf("writing VName: %v", err)
```
jaysachs: any use case to also permit proto (text or binary) output as well?

schroederc: I didn't add them because they're not really well-supported as stream formats.

PR #4351, opened 2020-02-04
fix(tooling): keep track of added source file paths

schroederc:


kythe/go/platform/tools/kzip/createcmd/createcmd.go
```diff
@@ -211,6 +211,7 @@ func (cb *compilationBuilder) addFile(ctx context.Context, root string) ([]strin
 				Digest: digest,
 			},
 		})
+		files = append(files, path)
```
jaysachs: Is there a unit test that could easily be added/updated to catch this?

schroederc: Not easily in this commit since it's a subcommand working on the filesystem.

PR #4352, opened 2020-02-04
fix(tooling): ensure VNameRewriteRule JSON encoding is canonical

schroederc:


kythe/go/util/vnameutil/rewrite_test.go
```diff
@@ -100,6 +101,66 @@ func TestParseConsistency(t *testing.T) {
...
+		{"^$", ""},
+		{"^a$", "a"},
+		{"^a", "a"},
+		{"a$", "a"},
+		{`\$`, `\$`},
+		{`\^\$`, `\^\$`},
+		{`\^`, `\^`},
+		{`\^^$\$`, `\^^$\$`},
+		{`\\$`, `\\`},
+		{`\\\\$`, `\\\\`},
```
jaysachs: What about including

  '^abc[^def$]ghi$'

PR #4357, opened 2020-02-07
feat(example): recommend a way to handle TypeScript/&c-style modules

zrlk:


shahms:
This doesn't match my recollection of the consensus achieved during the schema meeting.

kythe/docs/schema/modules.txt
```diff
@@ -0,0 +1,148 @@
...
+//- @module ref ModuleImport
+if (random()) module = otherModule;
+//- @module ref ModuleImport
+module.foo();
+----
+
+== Caveats
+
+Downstream consumers of the graph should tolerate overlapping references.
+Variables defined by imports should have subkind `import`. Consumers may
```
schroederc: Should this be the `subkind` fact?  Wouldn't this be awkward for imports of, for example, classes since they already have a subkind?

kythe/docs/schema/modules.txt
```diff
@@ -0,0 +1,148 @@
...
+`module` *ref* the node representing the module's definition instead of the
+definition of the local alias, which may not even have a sensible syntactic
+location to anchor to:
+
+[source,javascript]
+----
+//- @"'./module'" ref/imports Module
+import './module';
+//- @module ref Module
+module.foo();
```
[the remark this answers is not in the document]
mtaran-google: +1, this is confusing here. In general the syntax in line 1 is for [side effects imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Import_a_module_for_its_side_effects_only)

kythe/docs/schema/modules.txt
```diff
@@ -0,0 +1,148 @@
...
+
+Many languages provide top-level mechanisms for organization that make
+writing large systems easier. These have explicit syntactic forms for
+expressing which components of a *module* are to be *exported* and which
+components are to be *imported* from other modules. It may also be possible
+to refer to the module itself as a first-class value in a program.
+
+A program using these features might look something like:
+
+[source,javascript]
```
mtaran-google: There's another kind of import/export possible in JS, a [default import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Importing_defaults). You'd use it as `import foo from './module'`. The local identifier for the default export is arbitrary. Internally, default imports are the only way to import some kinds of things, and in those cases it's definitely preferable for the usages of `foo` to have a ref back to the original default-exported entity (or e.g. a protobuf message if it was generated, etc.).

mtaran-google:
> This doesn't match my recollection of the consensus achieved during the schema meeting.

This seems pretty close to me. What part seems different to you?

kythe/docs/schema/modules.txt
zrlk: This treatment uses `variable` nodes with `import` subkind to represent the local binding to the thing that was imported (e.g., a class).

zrlk:
Yeah, there are some refinements to what we discussed (in particular, the double-reference behavior, which records strictly more information). Could you elaborate on what doesn't look right?

shahms:
> Yeah, there are some refinements to what we discussed (in particular, the double-reference behavior, which records strictly more information). Could you elaborate on what doesn't look right?

Hopefully I'll be able to take a closer look tomorrow, but have been swamped today

kythe/docs/schema/modules.txt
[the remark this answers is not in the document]
zrlk: I'd misunderstood the syntax here. Update incoming.

kythe/docs/schema/modules.txt
```diff
@@ -0,0 +1,148 @@
...
+----
+//- "./module'" ref/imports Module
+//- @mod_imp defines/binding ModImp
+//- ModImp.subkind import
+import * as mod_imp from './module';
+//- @value ref/imports Value
+//- @renamedValue defines/binding RenamedValue
+//- RenamedValue.subkind import
+import {value as renamedValue} from './module';
+//- @mod_imp ref Module
```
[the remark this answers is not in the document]
zrlk: In the proposed representation, places where programmers give novel names to imports are treated both as transparent references to the imported thing and as explicit bindings in the local scope.

kythe/docs/schema/modules.txt
zrlk: PTAL at the newest push.

kythe/docs/schema/modules.txt
```diff
@@ -0,0 +1,163 @@
...
+//- @mod_imp ref/imports Module
+//- ModImp.subkind import
+import * as mod_imp from './module';
+//- @"'./module'" ref/imports Module
+//- @value ref/imports Value
+//- @renamedValue defines/binding RenamedValue
+//- RenamedValue.subkind import
+import {value as renamedValue} from './module';
+//- @defaultExport ref DefaultExportDef
+//- @defaultExport ref DefaultExport
```
shahms: My recollection was that we'd agreed not to emit duplicate `ref` nodes, but using alias at the import site to accomplish this.

kythe/docs/schema/modules.txt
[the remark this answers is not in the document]
shahms: evmar's expectation matches my recollection of the discussed solution.

kythe/docs/schema/modules.txt
shahms: That seems somewhat reasonable.  I might prefer making `import` it's own `kind`.

At the very least we'd need a suggested node for when the indexer doesn't have a better kind to give the import?

zrlk: I think that a `variable` with the `import` subkind will work in the general case. I don't think it's different enough to warrant a new top-level kind.

kythe/docs/schema/modules.txt
zrlk: (merging this thread with the `aliases` suggestion; I need to check that this has the behavior we want wrt clustering)

kythe/docs/schema/modules.txt
shahms: In languages with precise information on the kind of an imported name and more specific node kind (such as `talias`) do we want to recommend they use that and only fall back to `variable+import`?

kythe/docs/schema/modules.txt
shahms: My recollection and meeting notes suggest that the behavior we want is:

1) Clicking on a reference in a file should jump to
  a) The import line, if renamed
  b) The exported entity, if not
2) Clicking on a definition should show all uses, seeing through the renaming

I believe we'd settled on using a combination of aliases, clustering and `target_definition` to accomplish those, such that indexers would:

1) emit `ref/imports` at the import site
2) {name} -- defines/binding -> {alias} -- aliases -> {original} for renames (at least, I think doing this for not-renames is also fine when combined with the rest of the changes)
3) clustering would include follow reverse alias edges (and that accurate counts don't matter)
4) 1.b from above would be accomplished by using the imported node as the `target_definition` in `DecorationsReply` without necessarily changing the schema.

zrlk:
I want to avoid using the `aliases` relationship at first because I think it will cause clustering to confuse things. We can always add it later, though. With that in mind, do you think this is worth trying out?

shahms:
My recollection is that we'd specifically decided to avoid double references in favor of aliases, `experimental_cross_reference_indirection_kinds` and consistent use of `target_definition` in `DecorationsReply`.

zrlk:
OK; I've dropped the double references. PTAL.

kythe/docs/schema/modules.txt
zrlk: variable+import is precisely a name introduced as part of an import (that may point to something of a different kind).

schroederc:
> My recollection is that we'd specifically decided to avoid double references in favor of aliases, `experimental_cross_reference_indirection_kinds` and consistent use of `target_definition` in `DecorationsReply`.

I don't think we did.  `--experimental_cross_reference_indirection_kinds` isn't available in production.

zrlk:
The determination we made was that we could deal with this in production if we weakened guarantees about counts.

zrlk:
It looks like we've been suggesting people use a single-byte range at the beginning of the file, but I think it would make more sense to have the file itself defines/binding the module. I'll address this in a follow-up PR.

shahms:
I don't think we've had a concrete suggestion before, but had started suggesting using a zero-width span at the start of the file to indicate an "whole-file" span, due to uncertain support for defines/binding edges coming from anything other than an anchor.

PR #4363, opened 2020-02-11
feat(cxx_indexer): add test for proto/c++ xrefs with included proto

jaysachs:


kythe/cxx/indexer/cxx/testdata/proto/proto3.cc
```diff
@@ -0,0 +1,18 @@
...
+    Container msg;
+
+    //- @contained ref CxxGetContainedField
+    msg.contained();
+}
+//- ContainerMessageField generates CxxGetContainedField
+//- ContainerMessage generates CxxContainerMessage
+
+// required since mentioned in testdata3a.pb.h, transitively included
+//- ThingMessage.node/kind record
```
shahms: random facts shouldn't be required, can you clarify the comment a bit?
or just omit `testdata3a.proto` from `proto_srcs` entirely and its facts shouldn't be considered.

jaysachs: Can't remove it from proto_srcs, it won't build (which disturbs me but I couldn't work it out).

The issue is that the verifier sees the goal `//- @Thing defines/binding ThingMessage` in testdata3a.proto, but complains:
```
kythe/testdata/indexers/proto/testdata3a.proto:7.28-39: singleton variable ThingMessage used only here
```
I was able to just remove the goal in testproto3a.proto (it's not part of the test in its current incarnation)

shahms: That's fair, thanks. I was confused because the comment mentioned `testdata3a.pb.h` which shouldn't have any goals, let alone goals which where actually checked by the verifier.

PR #4364, opened 2020-02-12
update examples.md

pseyfert:
From the source of triples.go and the output of `triples --help` istm that `--graphstore` does not need to be separated by ` -- `.


pseyfert:
sorry for the force push, the second commit was wrong understanding of an error on my side.

PR #4366, opened 2020-02-13
feat(kzip_merge): allow applying vname rules during merge

shahms:


kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -165,6 +168,9 @@ func mergeInto(ctx context.Context, wr *kzip.Writer, path string, filesAdded str
 					return fmt.Errorf("error closing file: %v", err)
 				}
 			}
+			if vname, match := c.rules.Apply(ri.Info.Path); match {
+				proto.Merge(ri.VName, vname)
```
jaysachs: I'm curious if instead this should be assignment, rather than merge?

shahms: IMO, no.  I think it makes more sense to leave the unspecified fields alone rather than clear them.

shahms: (Which, in this case, would be "corpus")

schroederc: I don't think this should be a merge.  At least `corpus` and `root` should be treated as a unit.

kythe/go/platform/tools/kzip/mergecmd/flag.go
```diff
@@ -0,0 +1,54 @@
...
+	"kythe.io/kythe/go/util/vnameutil"
+)
+
+// vnameRules is a path-valued flag used for loading VName mapping rules from a vnames.json file.
+type vnameRules struct {
+	filename string
+	vnameutil.Rules
+}
+
+// Set implements part of the flag.Value interface and will append a new value to the flag.
```
creachadair: It's possible but not very common in Go to have repeated flags accumulate this way. Do we need more than one at all, and can we do it with (say) comma-separated filenames instead?

Relatedly: It's a little weird to keep only the last filename. Do we need the filename at all?

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
shahms: Due to the use of proto3, this actually has to be assignment or there is no way to set an empty root :-(

kythe/go/platform/tools/kzip/mergecmd/flag.go
shahms: This comment was mistaken (and copied from `flag.go` in `createcmd`).  It just sets a new filename to parse.  The value of the flag is the filename, we just read it as part of parsing the flags.

As to the other comment, accumulating flags *is* used by the `create` because it otherwise very quickly becomes incredibly unwieldy to actually use. ISTR a whole review thread on that.

creachadair: 👍 Re: Accumulating flags. It's not necessarily always a bad idea, just something to be deliberate about. I don't intend to re-litigate the other flag. 😀

I would probably have realized it was a copy issue if I'd actually read the code more carefully 👀.

shahms: No worries, I'm just glad you're still paying attention :-)

PR #4370, opened 2020-02-13
fix: increase the coverage of the proto/cc indexing tests

jaysachs:


kythe/cxx/indexer/cxx/testdata/proto/proto3.cc
```diff
@@ -1,7 +1,15 @@
 #include "kythe/testdata/indexers/proto/testdata3.pb.h"
 
-//- vname("",_, "", "kythe/testdata/indexers/proto/testdata3.proto","") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3.pb.h", "")
-//- vname("",_, "", "kythe/testdata/indexers/proto/testdata3a.proto","") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3a.pb.h", "")
+//-    vname("",_, "", "kythe/testdata/indexers/proto/testdata3.proto","") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3.pb.h", "")
+//-    vname("",_, "", "kythe/testdata/indexers/proto/testdata3a.proto","") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3a.pb.h", "")
+//-    vname("",_, "", "kythe/testdata/indexers/proto/testdata3b.proto","") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3b.pb.h", "")
+
+//- !{ vname("",_, "", "kythe/testdata/indexers/proto/testdata3b.proto","") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3a.pb.h", "") }
```
creachadair: Out of curiosity, what's the motivation for all the negative assertions here? Is there a specific issue these are trying to guard against? (If so: Can you document it somewhere nearby in comments?)

jaysachs: Internally, I'm seeing a given .proto.h being generated by a whole host of proto files. As I try to construct a repro, I thought I'd include the various tests.

Added a comment.

jaysachs:
> Do you want to include some negative assertions that we aren't emitting references suggesting that the file in question was also generated by a dependency?

I'm probably not understanding your question, but I think those assertions are here. Can you be more specific in your suggestion?

shahms:
> > Do you want to include some negative assertions that we aren't emitting references suggesting that the file in question was also generated by a dependency?
> 
> I'm probably not understanding your question, but I think those assertions are here. Can you be more specific in your suggestion?

I just missed them (there are a lot of assertions)

kythe/cxx/indexer/cxx/testdata/proto/proto3.cc
creachadair: Thanks, that makes sense. I haven't tested this, but I wonder if we could get at it more compactly, by writing something like:

```
X=vname("", _, "", "kythe/testdata/indexers/proto/testdata3a.proto", "") generates vname("", _, "bazel-out/bin", "kythe/testdata/indexers/proto/testdata3a.pb.h", "")
!{ X generates vname("", _, _, "kythe/testdata/indexers/proto/testdata3b.pb.h") }
!{ X generates vanme("", _, _, "kythe/testdata/indexers/proto/testdata3.pb.h") }
 …
```

This is maybe a little easier to read, and reduces (at least for me) the likelihood of copy-paste-update errors.

(Also in this example, I've not-cared the root, since we presumably shouldn't generate those edges regardless of where they may be)

creachadair: (you can stick evars inside the vname too, but I couldn't find a good way to apply equality constraints on those—maybe @zrlk can suggest something smarter)

jaysachs: Great ideas. Took it further and added evars for both sides.

PR #4375, opened 2020-02-15
fix(vfs): use correct FileInfo in globWalker

shahms:


kythe/go/platform/vfs/vfs_test.go
```diff
@@ -34,13 +39,13 @@ func TestGlobWalker(t *testing.T) {
 	testutil.FatalOnErrT(t, "local and glob walks differ", testutil.DeepEqual(bylocal, byglob))
 }
 
-func collectFiles(walk Walker, root string) ([]string, error) {
-	var files []string
-	return files, walk.Walk(context.Background(), root, func(path string, info os.FileInfo, err error) error {
+func collectFiles(walk Walker, root string) ([]walkResult, error) {
+	var results []walkResult
+	return results, walk.Walk(context.Background(), root, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
```
jaysachs: I'm sure this is a go-ism, but returning the results in the same statement in which they're mutated in a lambda is a bit spooky for me.

PR #4379, opened 2020-02-19
fix(cxx_indexer): address marked-source differences on builtin functions

shahms:


shahms:
This should fix the test failures folks have been seeing when running locally.

schroederc:
Works locally for me.  Thanks!

PR #4385, opened 2020-02-20
feat(build): switch to an autoconfigured ubuntu 18.04 image and C++17

shahms:


schroederc:
Buildkite is still broken due to brotli

schroederc:
Buildkite is still broken due to brotli

shahms:
Yeah, but it's broken because it doesn't like the warning option name. Fix in flight.

schroederc:
There's also some new looking warnings from `com_google_absl` and `se_haxx_curl`.

shahms:
PTAL

PR #4387, opened 2020-02-20
chore: update npm managed dependencies

shahms:


schroederc:
Update `kythe/typescript/*` to match.

shahms:
I have no idea how to go about doing that, the instructions from the README just result in a an exception from the yarn command.

evmar:
Can you paste the error message?

shahms:
Error: Cannot find module 'babel-runtime/helpers/asyncToGenerator'

some interneting suggests I need to set a proper NODE_PATH

evmar:
What command did you run to get that error?  I have never seen that one before...

shahms:
It was essentially the same as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=933229
(`apt-get install yarnpkg` => yarnpkg fail, etc.)

evmar:
How does it make sense that the bug is "resolved" when the binary doesn't work unless you set some obscure environment variable argh.

shahms:
PTAL

PR #4388, opened 2020-02-21
docs: add instructions for extracting gradle projects

justbuchanan:


kythe/web/site/examples.md
```diff
@@ -113,6 +113,90 @@ $ /opt/kythe/tools/runextractor cmake \
...
+  {
+    "pattern": "(.*)",
+    "vname": {
+      "corpus": "MY_CORPUS",
+      "path": "@1@"
+    }
+  }
+]
+```
+
```
schroederc: Maybe comment about using the new `vnames` tool to manually test the file?

kythe/web/site/examples.md
```diff
@@ -113,6 +113,90 @@ $ /opt/kythe/tools/runextractor cmake \
...
+2. VName configuration
+
+Next, you will need to create a vnames.json mapping file, which tells the
+extractor how to assign vnames to files based on their paths. A basic vnames
+config for a gradle project looks like:
+
+```json
+// vnames.json
```
shahms: can we ship a reasonable default file or template for this?

justbuchanan: I agree we should. I think a good default template would look just like this, but have the java extractor replace the corpus value with something that can be specified via environment variable or flag. We currently do that for bazel, but I don't think the java extractor supports it (yet).

shahms: It looks like the Java extractor only uses `KYTHE_CORPUS` is `KYTHE_VNAMES` is absent, but should probably use `KYTHE_CORPUS` if the corpus resulting from applying the vnames is empty.

justbuchanan: I filed this as a separate issue: https://github.com/kythe/kythe/issues/4389.

justbuchanan: Also #4390.

kythe/web/site/examples.md
justbuchanan: Done.

justbuchanan:
Website is updated: https://kythe.io/examples/#extracting-gradle-based-repositories.

creachadair:
> Website is updated: https://kythe.io/examples/#extracting-gradle-based-repositories.

It's a fairly minor thing, but there seems to be some asciidoc wonkiness in that section: All the instruction points are numbered "1." and their content isn't co-indented with them. Maybe we should switch to subsection headers instead?

justbuchanan:
Thanks for catching that. Not sure why it doesn't like my numbers, but subsection headers work for me. Here's a fix: https://github.com/kythe/kythe/pull/4391.

PR #4391, opened 2020-02-21
fix(docs): improve formatting of gradle extraction example

justbuchanan:


kythe/web/site/examples.md
```diff
@@ -130,14 +130,13 @@ allprojects {
 }
 ```
 
-2. VName configuration
```
creachadair: I think the issue may have been that the subparagraphs weren't indented far enough. Markdown expects multiples of 4 spaces (though some implementations are flexible). In any case, if you do want to keep the numbers I think you could write, say

```
1.  VName configuration

    Next, you will need to create a `vnames.json` mapping file …
```
i.e., indenting the stuff you want to live inside the number-bulleted item.

But subheadings are fine too. 🙂

justbuchanan: Thanks, I think I like the numbers a little better. I previewed the site locally and the indenting does fix it.

PR #4393, opened 2020-02-24
fix(runextractor): set --build_file default to build.gradle

justbuchanan:
this is the file name most projects use and is recommended by:
https://guides.gradle.org/creating-new-gradle-builds/


BFergerson:
@justbuchanan, just an FYI, the description still says `The config file for a gradle repo, defaults to 'gradle.build` and I believe there a few more spots that refer to things as `gradle.build` instead of `build.gradle` (like here: https://github.com/kythe/kythe/blob/master/kythe/go/extractors/config/preprocessor/README.md#gradle).

justbuchanan:
Thanks, sent a followup in https://github.com/kythe/kythe/pull/4398.

PR #4397, opened 2020-02-25
feat(runextractor): improved debug info for maven subcommand

justbuchanan:
* print out `mvn` command
* print diff of changes made to pom.xml
* send mvn subcommand's stderr/stdout to `runextractor`'s stderr/stdout instead of /dev/null


kythe/go/extractors/config/runextractor/backup/backup.go
```diff
@@ -40,7 +40,7 @@ import (
 // file.  The Restore method replaces the contents of the original with the
 // backup, overwriting any changes that were made since the backup was created.
 type File struct {
-	orig, tmp string // file paths
+	Orig, Tmp string // file paths
```
shahms: it doesn't look like these need to be exported?

kythe/go/extractors/config/runextractor/mavencmd/mavencmd.go
```diff
@@ -82,10 +84,27 @@ func (m *mavenCommand) Execute(ctx context.Context, fs *flag.FlagSet, args ...in
 	if err := modifier.PreProcessPomXML(m.pomXML); err != nil {
 		return m.Fail("error modifying maven pom XML %s: %v", m.pomXML, err)
 	}
-	if err := exec.Command("mvn", "clean", "install",
+
+	// Print diff to show changes made to pom.xml.
+	log.Printf("Modified pom.xml. Diff:")
```
shahms: Can this be put behind a -verbose flag or something?

kythe/go/extractors/config/runextractor/backup/backup.go
shahms: Sorry, I missed that this was used for printing the differences.  Perhaps unexport these but include a function that reports the differences directly?

I think you might also want to use https://godoc.org/github.com/google/go-cmp/cmp rather than shelling out to the `diff` command.

justbuchanan: Done.

kythe/go/extractors/config/runextractor/mavencmd/mavencmd.go
justbuchanan: Done.

PR #4400, opened 2020-02-25
fix(cxx_indexer): emit ref to class from ctor

### LINKED ISSUE #4381: include class ctors in references for the class itself
When looking at references and definitions of a class, Kythe currently returns:
```
struct [[Foo]] {
   Foo();
   Foo(const [[Foo]]&);
   ~[[Foo]]();
};
```
This is suboptimal if one is looking to rename a class, for example.  Following the example with the destructor, the constructor should have a ref to the class name which overlaps the defines/binding edge for the ctor.  I believe these were inhibited prior due to poor UI/client support for overlapping ranges, but this handling is much better now.

shahms:
This fixes #4381.


kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1077,6 +1077,21 @@ void IndexerASTVisitor::VisitRecordDeclComment(
   }
 }
 
+bool IndexerASTVisitor::TraverseCXXConstructorDecl(
+    clang::CXXConstructorDecl* CD) {
+  LOG(ERROR) << DumpString(*CD);
```
jaysachs: remove

shahms: Gah, thanks!

PR #4405, opened 2020-02-27
chore: remove redudnant genSrcDir parameter and ask fileManager

shahms:


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -707,13 +694,15 @@ private void processRequiredInput(
...
-    if (genSrcDir.isPresent()
-        && !isJarPath
-        && requiredInput.getKind() == Kind.SOURCE
-        && Paths.get(relativePath).startsWith(genSrcDir.get())) {
+    // Identify generated sources by checking if the source file is under the SOURCE_OUTPUT
+    // directory.
+    if (isGeneratedSource(fileManager, requiredInput)) {
       results.explicitSources.add(strippedPath);
-      results.newSourcePath.add(genSrcDir.get().toString());
+      Iterables.addAll(
```
schroederc: This looks like we're adding all the generated sources every time we handle a required input instead of just once.

shahms: `newSourcePath` is a set and this mirrors the old behavior.  We're adding all of the generated source directories to the `newSourcePath` set for each generated source we encounter.

schroederc: That shouldn't change between generated sources, though.  We should just create the set somewhere once to be more explicit.

shahms:
I missed that you'd commented further, sorry.  I'm not sure what you mean by that.  The behavior of the current extractor is that it adds `genSrcDir` to the `newSourcePath` set whenever it encounters a generated file.  That behavior is preserved, but the source of the path is changing.

This isn't the only place we modify `results.newSourcePath` so we can't just construct it in one place.

PR #4407, opened 2020-02-27
fix(cxx_indexer): refactor TypeLoc visitation into separate class

shahms:
Rather than using the current hack for visiting types when the type-as-written differs from the resolved type, factor that logic into a separate visitation hierarchy that visits the type-as-written in parallel with the resolved type.

This is closer to the old logic, but implemented using the same visitation approach as the rest of RecursiveASTVisitor.

### DIFF
--- a/kythe/cxx/indexer/cxx/recursive_type_visitor.h
+++ b/kythe/cxx/indexer/cxx/recursive_type_visitor.h
@@ -84,13 +84,13 @@ class RecursiveTypeVisitor : public clang::RecursiveASTVisitor<Derived> {
   bool TraverseTypePair(clang::TypeLoc TL, clang::QualType T);
 
   // Declare Traverse*() for all concrete TypeLoc classes.
-  // Note: We're using TypeNodes.def as QualifiedTypeLoc needs to be handled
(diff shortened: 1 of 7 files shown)


kythe/cxx/indexer/cxx/recursive_type_visitor.h
```diff
@@ -0,0 +1,382 @@
+/*
+ * Copyright 2018 The Kythe Authors. All rights reserved.
```
zrlk: \robinwilliams{What year is it?}

kythe/cxx/indexer/cxx/recursive_type_visitor_test.cc
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2018 The Kythe Authors. All rights reserved.
```
zrlk: Please update copyright year.

kythe/cxx/indexer/cxx/recursive_type_visitor.h
shahms: lol, in fairness I'm pretty sure that's when I originally wrote this file (I have a bit of a backlog of refactorings).

PR #4415, opened 2020-03-02
fix(cxx_indexer): refactor TypeLoc visitation into separate class

shahms:
This includes a fix and test for the problem that was affecting production, some additional defense-in-depth nullptr checks and an additional reference for dependent sized arrays features `sizeof...`


shahms:
PTAL

PR #4418, opened 2020-03-03
feat(cxx_indexer): allow using compilation corpus instead of empty

shahms:
Relatedly, this also changes named namespaces to use the default
    corpus, rather than the file corpus, so that files from different
    corpora which open a namespace for extension (such as std::hash)
    refer to the same node.


shahms:
@zrlk if you could double check that the USR and namespace-related changes make sense I'd appreciate it.

zrlk:
Please note in the change description that this change affects namespaces.

zrlk:
(The USR/namespace stuff LGTM)

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
```diff
@@ -1363,23 +1364,33 @@ const KytheClaimToken* KytheGraphObserver::getAnonymousNamespaceClaimToken(
...
+      namespace_tokens_.emplace(file_token, NamespaceTokens{});
+  if (inserted) {
+    // Named namespaces belong to the same corpus as structural types due to
+    // their use as extension points which may be opened from a file in any
+    // corpus, but should still refer to the same node.
+    iter->second.named.mutable_vname()->set_corpus(
+        type_token_.vname().corpus());
+    iter->second.named.set_rough_claimed(file_token->rough_claimed());
+    // Anonymous namespaces are unique to the translation in which they're
+    // defined, which we approximate by using the file.
```
jaysachs: "the file's corpus" or "the file_token's corpus" or "the file's vname's corpus".

kythe/cxx/indexer/cxx/testdata/basic/namespace_vnames.cc
```diff
@@ -5,17 +5,16 @@
 //- @namespace ref vname(_,"bundle","","test.cc","c++")
 namespace { }
 
-// Named namespaces drop path and root but keep corpus.
-//- @ns ref vname(_,"bundle","","","c++")
+// Named namespaces drop path and root and use the default corpus.
```
jaysachs: Is this test any weaker because the default corpus is "" ? Can/should we make a distinct non-empty default corpus in play for the tests?

shahms: I don't believe so, no.  There is a specific test that the default corpus behaves as expected.

PR #4419, opened 2020-03-03
chore: update bazelisk

justbuchanan:

PR #4420, opened 2020-03-03
fix(build): add dep for new version of bazelisk

justbuchanan:
* I broke this in https://github.com/kythe/kythe/pull/4419


justbuchanan:
I broke this in #4419

Issue #4401, opened 2020-02-25
bazel-extractor docker image should support c++17


justbuchanan:
Our docker image is failing to extract the kythe repo since we [switched to c++17](https://github.com/kythe/kythe/commit/509d7c618a5203682566fbcf5fc0572a87dddef2).

The relevant packages we have in the image are: `libstdc++-5-dev` and `clang-6.0`.

Our docker image is based on the official bazel image: https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/bazel/Dockerfile, which uses ubuntu 16.04.

We may need a ppa to get a newer version of the c++ stdlib.

shahms:
We should just be able to base in on 18.04 instead (which supports C++17 and is what I had to do to get our RBE image to do so).

justbuchanan:
Sounds good to me. I don't think we get a whole lot from extending the bazel image - I can copy the bits we need into our dockerfile.

PR #4422, opened 2020-03-04
fix(bazel-extractor): update docker image for c++17

justbuchanan:
we can't get a new enough clang/libstdc++ using the gcr.io bazel image,
so I switched to debian sid.


justbuchanan:
Fixes #4401.

PR #4424, opened 2020-03-05
fix(bazel extractor): add missing deps to docker image

justbuchanan:


kythe/extractors/bazel/Dockerfile
```diff
@@ -14,9 +14,9 @@
 
 FROM debian:sid
 
-# Install C++ compilers for cc_* rule support.
+# Install C++ compilers and other deps
 RUN apt-get update && \
-    apt-get install -y git clang-8 build-essential && \
+    apt-get install -y git clang-8 build-essential zip python openjdk-8-jdk && \
```
salguarnieri: Can you note what jdk 8 is needed for so it is easier to clean up later when we don't need jdk 8 any more?

justbuchanan: Done.

PR #4425, opened 2020-03-05
fix(bazel_go_extractor): record canonical importpath for archives

justbuchanan:
This should fix cross-module linking for go code extracted through bazel.

Go package dependencies are pulled in as static archives (.a files) and are mentioned in the arguments like:

`-arc "github.com/google/subcommands=github.com/google/subcommands=bazel-out/k8-fastbuild/bin/external/com_github_google_subcommands/linux_amd64_stripped/go_default_library%/github.com/google/subcommands.a="`

This change parses the `-arc` arguments and builds a map of filepath -> importpath, which is then used to attach `GoPackageInfo` details messages to the corresponding required_inputs in the compilation unit. The indexer later picks these up and refers to package dependencies by the correct vname/importpath.

### DIFF
--- a/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
+++ b/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 The Kythe Authors. All rights reserved.
+ * Copyright 2020 The Kythe Authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
+++ b/kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
@@ -200,8 +200,8 @@ type compileArgs struct {
 
 func parseCompileArgs(args []string) *compileArgs {
 	c := &compileArgs{
-		original:  args,
-		importMap: make(map[string]string),
+		original:         args,
+		importMap:        make(map[string]string),
 		archiveImportMap: make(map[string]string),
 	}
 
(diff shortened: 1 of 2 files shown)


kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 The Kythe Authors. All rights reserved.
+ * Copyright 2020 The Kythe Authors. All rights reserved.
```
creachadair: FYI modifying an existing file generally doesn't invalidate the copyright date. The date here is when the copy right _begins_, so moving it later is arguably misleading—it still applies.

Not a big deal, just something I've seen a few times in passing.

justbuchanan: The arc linter complained about this file and seemed to indicate the top of the file. After changing the copyright date, I realized I just needed to run gofmt. Changed back to 2016.

kythe/go/extractors/cmd/bazel/bazel_go_extractor/bazel_go_extractor.go
```diff
@@ -207,12 +223,23 @@ func parseCompileArgs(args []string) *compileArgs {
...
 			if len(ps) == 2 && ps[0] != ps[1] {
 				c.importMap[ps[0]] = ps[1]
 			}
+		case "arc":
+			// Example -arc arg:
+			//   "github.com/google/subcommands=github.com/google/subcommands=bazel-out/k8-fastbuild/bin/external/com_github_google_subcommands/linux_amd64_stripped/go_default_library%/github.com/google/subcommands.a=",
+			//
+			// Only record the mappings that change something.
+			ps := strings.SplitN(strings.TrimSuffix(arg, "="), "=", 3)
+			if len(ps) == 3 {
```
creachadair: Maybe `if len(ps) == 3 && ps[1] != ps[2] {` if you want to be really safe.

justbuchanan: From what I can tell, `ps[2]` should always be a path to a `.a` file, while `ps[0]` should always be an import path.

The `-arc` flag comes from:
https://github.com/bazelbuild/rules_go/blob/8ea79bbd5e6ea09dc611c245d1dc09ef7ab7118a/go/tools/builders/compile.go#L43

PR #4435, opened 2020-03-11
feat(indexing): add ref/id edge

zrlk:
This PR adds a new edge, `ref/id`, meant to help return more complete
lists of use sites for the identifiers for nodes. In particular, it
should help automated refactoring tools.

Other names that have been proposed for this edge include `ref/nominal`
and `ref/syntactic`.


kythe/docs/schema/schema.txt
```diff
@@ -1201,6 +1201,33 @@ public class E {
 }
 --------------------------------------------------------------------------------
 
+[[refid]]
+ref/id
+~~~~~~~~~~~
+Brief description::
+  A *ref/name* B if A does not *ref* or *define* B but A's contents are
```
shahms: `ref/id`?

kythe/docs/schema/schema.txt
```diff
@@ -1201,6 +1201,33 @@ public class E {
...
+  semantic nodes
+Ordinals are used::
+  never
+
+[kythe,C++,"Constructors are named after their classes."]
+--------------------------------------------------------------------------------
+//- @C defines/binding ClassC
+class C {
+  //- @C defines/binding CtorC
+  //- // @C ref/id ClassC
```
shahms: I assume the secondary comment is because the C++ indexer doesn't currently implement this?

kythe/docs/schema/schema.txt
```diff
@@ -1201,6 +1201,33 @@ public class E {
...
+  C(int i);
+};
+
+//- @"C(1)" ref/call CtorC
+//- // @C ref/id ClassC
```
shahms: `//- @C ref CtorC`

kythe/docs/schema/schema.txt
jaysachs: If the edge is "ref/id", then A does "ref" B in some form. Also "contents" reads a bit weird, but I assume you mean "the bytes of the anchor span"?

kythe/docs/schema/schema.txt
```diff
@@ -1201,6 +1201,33 @@ public class E {
...
+void f() { auto c = C(1); }
+--------------------------------------------------------------------------------
+
 [[refincludes]]
```
jaysachs: Do we want to include something about the implicit constructors here, or would that be a separate example?

shahms: That would be a different edge (`defines/implicit`).

kythe/docs/schema/schema.txt
zrlk: Yeah, we can remove it once it's implemented.

kythe/docs/schema/schema.txt
zrlk: Changed the wording a bit.

kythe/docs/schema/schema.txt
zrlk: I think that's a separate issue.

kythe/docs/schema/schema.txt
zrlk: Whoops, this went through a couple of revisions.

kythe/docs/schema/schema.txt
```diff
@@ -1201,6 +1201,34 @@ public class E {
...
+  A *ref/id* B if A does not otherwise *ref* or *define* B but the bytes of A's
+  span are determined by the identifier of B.
+Points from::
+  anchors
+Points toward::
+  semantic nodes
+Ordinals are used::
+  never
+
+[kythe,C++,"Constructors are named after their classes."]
```
evmar: Do you have a second example?  (Why isn't this just 'ref/ctorOf'?)

jaysachs: what about "Constructors and destructors are named after their classes." ?

zrlk: This is a more general relationship than just constructor-/destructor-of.

evmar: Right, I get that it's intended to be more general, but with only a single example I am not sure what the generalization is.  When would I use id instead of some other relationship?  Is it meant as a catch-all for cases where you don't have constructor-of?  Etc.

shahms: At the moment, no other use cases have arisen but we try not to paint ourselves into a corner.  Edges and nodes are generally given names that describe the semantics specifically enough to differentiate them, but without being overly language specific.

evmar: Just making sure I follow, I should be using this for any connections that aren't covered by the existing schema then?

PR #4448, opened 2020-03-23
Support TypeScript 3.8 AST

rrdelaney:
TypeScript 3.8 updated the AST so that a `ts.ExportDeclaration` can now either have a `NamedExports` or `NamespaceExport` as child ([example](https://ts-ast-viewer.com/#code/KYDwDg9gTgLgBAKjgQwM5wHboGZQgWzgHIsiBuAKFEljgG8BfOXA40soA)). This PR updates the current exports handing to only traverse into named exports and leave namespace exports alone. This is only meant to unblock building and running Kythe with TypeScript 3.8 and it's API's.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -1249,7 +1249,7 @@ class Visitor {
    * and that case is handled as part of the ordinary declaration handling.
    */
   visitExportDeclaration(decl: ts.ExportDeclaration) {
-    if (decl.exportClause) {
+    if (decl.exportClause && ts.isNamedExports(decl.exportClause)) {
       for (const exp of decl.exportClause.elements) {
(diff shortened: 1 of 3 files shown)


kythe/typescript/indexer.ts
```diff
@@ -1249,7 +1249,7 @@ class Visitor {
    * and that case is handled as part of the ordinary declaration handling.
    */
   visitExportDeclaration(decl: ts.ExportDeclaration) {
-    if (decl.exportClause) {
+    if (decl.exportClause && ts.isNamedExports(decl.exportClause)) {
```
evmar: Can you change the package.json / lockfile too?

evmar: Or is this function in ts3.7

rrdelaney: Updated!

PR #4451, opened 2020-03-24
chore: generate partial compile_commands.json

shahms:
Rather than re-generating compile_commands.json all in one go, use `--compile_one_dependency` to insert new entries and reformat.

### DIFF
--- a/tools/cpp/generate_compilation_database.sh
+++ b/tools/cpp/generate_compilation_database.sh
@@ -1,25 +1,27 @@
 #!/bin/bash
 
 # Generates a compile_commands.json file at $(bazel info execution_root) for
-# your Clang tooling needs.
+# the given file path.
 
 set -e
 
+FILENAME=${1:?Missing required source path}
 bazel build \
   --experimental_action_listener=//kythe/cxx/tools/generate_compile_commands:extract_json \
-  --noshow_progress \
-  --noshow_loading_progress \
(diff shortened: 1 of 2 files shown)


tools/cpp/generate_compilation_database.sh
```diff
@@ -1,25 +1,29 @@
...
-  if ((COUNT++)); then
-    echo ',' >> compile_commands.json
-  fi
   sed -e "s|@BAZEL_ROOT@|$BAZEL_ROOT|g" < "$fname" >> compile_commands.json
+  echo "" >> compile_commands.json
 done
-echo "]" >> compile_commands.json
+# Decompose, insert and keep the most recent entry for a given file, then
+# recombine.
+sed 's/\(^[[]\)\|\([],]$\)//;/^$/d;' < compile_commands.json \
```
salguarnieri: I'll take your word for it. If we need to update this though, I'd be in favor of writing a custom binary so it is a little easier to read.

shahms: It would be much simpler if JSON allowed trailing commas on the last element in a list, but noooooo.

evmar: FWIW, the 'jq' tool: https://stedolan.github.io/jq/
is really great and I highly recommend it.

PR #4452, opened 2020-03-24
chore: remove spurious file data log messages

schroederc:


shahms:
While these warnings are noisy, the indexer uses the digest to find the vname (rather than the path) so having multiple paths with the same digest can be problematic.

schroederc:
> While these warnings are noisy, the indexer uses the digest to find the vname (rather than the path) so having multiple paths with the same digest can be problematic.

Sharing a digest signifies either that the files have the same content or there is a hash collision.  It's almost always the first case and these warnings are practically useless in the indexer.  If anything, the extractors should handle this case.

shahms:
> > While these warnings are noisy, the indexer uses the digest to find the vname (rather than the path) so having multiple paths with the same digest can be problematic.
> 
> Sharing a digest signifies either that the files have the same content or there is a hash collision. It's almost always the first case and these warnings are practically useless in the indexer. If anything, the extractors should handle this case.

The extractor does handle this case: it includes both files.  It's perfectly reasonable for dependent files to share the same content.  Anyway, I definitely support removing this warning just a heads up that it exists to highlight a potential problem.

PR #4453, opened 2020-03-25
chore: fix lint warnings

salguarnieri:


salguarnieri:
> It looks like the proto changes anger clang-format (or maybe trigger it making other changes).

clang-format-7 vs clang-format issue. I reverted to clang-format-7 since that's what arc uses.

PR #4456, opened 2020-03-26
fix: require flex version 2.6 or newer

shahms:
Fixes #4455 (by requiring a version of flex that omits the `register` keyword).

Issue #4455, opened 2020-03-26
bazelisk build //kythe/release broken on macos


salguarnieri:
```
$ bazel build //kythe/release
...
ERROR: .../kythe/kythe/cxx/verifier/BUILD:25:1: C++ compilation of rule '//kythe/cxx/verifier:lexparse' failed (Exit 1) wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG '-std=c++11' -iquote . -iquote ... (remaining 37 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
bazel-out/darwin-fastbuild/bin/kythe/cxx/verifier/lexer.yy.cc:867:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register yy_state_type yy_current_state;
        ^~~~~~~~~
bazel-out/darwin-fastbuild/bin/kythe/cxx/verifier/lexer.yy.cc:868:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register char *yy_cp, *yy_bp;
        ^~~~~~~~~
bazel-out/darwin-fastbuild/bin/kythe/cxx/verifier/lexer.yy.cc:868:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register char *yy_cp, *yy_bp;
        ^~~~~~~~~
bazel-out/darwin-fastbuild/bin/kythe/cxx/verifier/lexer.yy.cc:869:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register int yy_act;
        ^~~~~~~~~
bazel-out/darwin-fastbuild/bin/kythe/cxx/verifier/lexer.yy.cc:938:4: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
                        register YY_CHAR yy_c = yy_ec[YY_SC_TO_UI(*yy_cp)];
                        ^~~~~~~~~
bazel-out/darwin-fastbuild/bin/kythe/cxx/verifier/lexer.yy.cc:1345:6: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register char *dest = YY_CURRENT_BUFFER_LVALUE->yy_ch_buf;
        ^~~~~~~~~
[...]

PR #4457, opened 2020-03-26
docs: install flex via Homebrew for macOS

creachadair:
This is a follow-up to #4455 and #4456. Because macOS ships with flex 2.5 and
homebrew will not link it to avoid confusing the system build tools, this
change also adds a FLEX environment variable to allow it to be overridden
without diddling the user's PATH.

### DIFF
--- a/tools/build_rules/lexyacc/lexyacc.bzl
+++ b/tools/build_rules/lexyacc/lexyacc.bzl
@@ -86,7 +86,7 @@ def lexyacc_toolchain(name, lex, yacc):
     )
 
 def _check_flex_version(repository_ctx, min_version):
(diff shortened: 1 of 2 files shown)


tools/build_rules/lexyacc/lexyacc.bzl
```diff
@@ -86,7 +86,7 @@ def lexyacc_toolchain(name, lex, yacc):
     )
 
 def _check_flex_version(repository_ctx, min_version):
-    flex = repository_ctx.which("flex")
+    flex = repository_ctx.os.environ.get("FLEX", repository_ctx.which("flex"))
```
shahms: Need to declare this in the rule definition below as well.

PR #4461, opened 2020-04-02
fix(cxx_indexer): emit zero-width spans for entities in macro bodies

shahms:


shahms:
I repeat my motto: "Mo' macros, mo' problems"

PR #4462, opened 2020-04-02
fix(cxx_indexer): emit ref and ref/init for designated inits

shahms:


shahms:
@googlebot rescan

kythe/cxx/indexer/cxx/testdata/rec/rec_designated_initializer.c
```diff
@@ -2,18 +2,24 @@
 
 //- @field defines/binding Field
 struct s { int field; };
-//- @field ref/init Field
+//- @field ref Field
+//- @"0" ref/init Field
```
schroederc, line 6: This `ref/init` doesn't seem useful, and I don't think we do this for any other assignment where the variable/field is syntactically present.  What's the use-case here?

shahms: I also don't think `ref/init` is useful; but this matches the documented semantics.  We also do this for Go: https://www.kythe.io/docs/schema/#refinit

shahms: @creachadair Do you recall why `ref/init` was added at all?

shahms: In theory to support positional init, IIRC:
```
// Positional
//- @"17" ref/init Field
var _ = S{17}
```
but kept for the designated case for consistency?

creachadair: Yeah, I think that was the crux of the idea. We introduced it in a1b05205.

creachadair: I believe the issue was code like this:

```go
type S struct { Z int }
 x := 25
s := S{x}
```

Without distinction, the jump-to-definition logic doesn't know whether to prefer the variable or the field, since arguably that anchor serves both roles. As with many of our other schema changes, I believe this was to support CodeSearch's refusal to support a more sophisticated UI choice on click—and someone complained that it went the "wrong" place. The other option was to omit the ref to the field (which is what I had done previously), which led to just complaints about violating the Update Principle. Things have probably changed a bit since 2017.

shahms: Thinking about it a bit more, I believe we also wanted to support the general case that initializers like that are uses of the field, but don't use the name so if you want to list all "uses" of something, those would be missing without some relevant edge, but it needed to be distinguished from regular "refs".  Thanks the reminder!

creachadair: I fear a lot of these discussions may be lost. I wonder if we could attach rationale sections to the schema without polluting it too much.

jaysachs: > I fear a lot of these discussions may be lost. I wonder if we could attach rationale sections to the schema without polluting it too much.

That's an excellent idea. @zrlk and I were just discussing something related to this.

creachadair: I would be happy to contribute toward that, even if only as a reviewer.

PR #4463, opened 2020-04-03
fix(ts_indexer): bug related to overriden functions.

nbeloglazov:
This commit has 2 changes:
1. Properly emit `overrides` edges for functions when class/interface being extended/implemented is exported from another file.
2. Emit `extends` edges from class/interface to all classes/intefaces in its heritage clause.

Tested:
  tests


nbeloglazov:
This fix is related to internal bug: http://b/146173865

kythe/typescript/indexer.ts
```diff
@@ -43,6 +43,10 @@ export interface IndexerHost {
    * not yet null-safe, so it hasn't been annotated with full types.)
    */
   getSymbolAtLocation(node: ts.Node): ts.Symbol|undefined;
+  /**
+   * getSymbolAtLocation is the same as ts.TypeChecker.getTypeAtLocation.
```
evmar, line 47: Should we just expose the typeChecker instead then?

kythe/typescript/indexer.ts
```diff
@@ -1441,9 +1468,9 @@ class Visitor {
         if (funcSym && decl.parent.heritageClauses) {
           for (const heritage of decl.parent.heritageClauses) {
             for (const baseType of heritage.types) {
-              const baseSym =
-                  this.host.getSymbolAtLocation(baseType.expression);
-              if (!baseSym || !baseSym.members) {
+              const type =
+                  this.host.getTypeAtLocation(baseType.expression);
+              if (!type || !type.symbol || !type.symbol.members) {
```
evmar, line 1473: I don't understand this change, can you explain?

(Also, this function is getting extremely long, is there any reasonable way to refactor it a bit as you add this?)

nbeloglazov: My understanding that for 

```ts
import {IFace} from './interface';

class Implementation implements IFace {}
```

When we call `getSymbolForLocation` it gives us location of the IFace symbol in the import statement, not the `IFace` symbol in `interface.ts`. But getting type works as type contains the main symbol from `interface.ts`. If there are more canonical ways to get main symbol of imported alias - I can switch to that.

evmar: Ah interesting.  You can check if the symbol is an alias and then call .getAliasedSymbol() or whatever it's called, but this is fine I think.

nbeloglazov:
> Why does this fix the bug
In the situation of imported interface:
```ts
import {IFace} from './interface';

class Implementation implements IFace {}
```
When we call `getSymbolForLocation` on `IFace` in `implements` clause - it gives us location of the `IFace` symbol in the import statement, not the `IFace` symbol in` interface.ts`. But getting type works as type contains the main symbol from interface.ts.

kythe/typescript/indexer.ts
nbeloglazov: Removed this method. We already have access to the typeChecker in indexer. If some plugin in future needs it - then we can consider adding it back.

kythe/typescript/indexer.ts
```diff
@@ -329,6 +329,10 @@ class StandardIndexerContext implements IndexerHost {
     return this.typeChecker.getSymbolAtLocation(node);
   }
 
+  getTypeAtLocation(node: ts.Node): ts.Type|undefined {
+    return this.typeChecker.getTypeAtLocation(node);
+  }
```
evmar, line 334: I think this is unused now

PR #4467, opened 2020-04-06
chore: normalize and fix schema formatting

shahms:
Given that this found a bug in the inline test as well, I'm confident this will address the problem with `ref/id` missing from the TOC.


shahms:
https://f82a988f067789c5bdddb0f368be6272c00ad15b-dot-kythe-repo.appspot.com/docs/schema/

PR #4468, opened 2020-04-07
fix(cxx_indexer): properly emit ref/init edges in all cases

shahms:
This matches the Go behavior and that documented in the schema.  Prior to this, we would only emit ref/init for the designated initializer case, which is less useful because we also emit a regular ref from the field name in the designated initializer itself.

This PR cuts out a redundant traversal and emits a ref/init edge to all fields initialized from a given initializer list, excluding those without an explicit initializing expression.

### DIFF
--- a/kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
+++ b/kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
@@ -1,14 +1,34 @@
 // Tests that aggregate initialization references struct.
 
 //- @S defines/binding StructS
-struct S {};
+struct S {
+  //- @a defines/binding FieldA
+  int a;
+  //- @b defines/binding FieldB
+  long b;
+};
+
+//- @U defines/binding UnionU
+union U {
+  //- @x defines/binding FieldX
+  int x;
+  //- @y defines/binding FieldY
+  long y;
+};
 
 template <typename...T>
 void fn(T&&...);
 
 void f() {
   //- @S ref StructS
-  auto s = S{};
+  //- @"1" ref/init FieldA
+  //- !{ _ ref/init FieldB }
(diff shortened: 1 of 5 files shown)


kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
```diff
@@ -1,14 +1,34 @@
...
 
 template <typename...T>
 void fn(T&&...);
 
 void f() {
   //- @S ref StructS
-  auto s = S{};
+  //- @"1" ref/init FieldA
+  //- !{ _ ref/init FieldB }
+  auto s = S{1};
```
schroederc, line 26: Probably want another test to check that multiple fields can be `ref/init`'d

shahms: Done.  Also include a test of base-class aggregates.

shahms: Spoke too soon, working on it.

kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
```diff
@@ -22,12 +36,21 @@ void fn(T&&...);
...
+  //- @T ref StructT
+  //- @"1" ref/init FieldA
+  auto t = T{1};
+
+  //- @T ref StructT
+  //- @"{1,2}" ref/init StructS
+  //- @"3" ref/init FieldJ
+  //- @"4" ref/init FieldL
+  //- @"5" ref/init FieldM
+  auto full = T{{1,2},3,4,5};
```
schroederc, line 50: Check `FieldA` and `FieldB`

PR #4476, opened 2020-04-16
feat(cxx_indexer): emit ref/id to class when class and ctor overlap

shahms:


shahms:
This is a follow-up on #4435 actually making use of the new edges in C++.

PR #4478, opened 2020-04-18
fix(ts_indexer): emit ref/call edges from calls.

nbeloglazov:
For each call (function, constructor calls) we emit `ref/call` to the
function node or class value. Also emit `childof` to containing functions
so that callgraph can be constructed.

There is still some work to be done as we probably not handling
anonymous functions well. Also need to emit ref/call to constructor
nodes rather than class value.

Tested:
  unit test


nbeloglazov:
@evmar

kythe/typescript/indexer.ts
```diff
@@ -965,6 +993,60 @@ class Visitor {
     return undefined;
   }
 
+  getSymbolAndVNameForFunctionDeclaration(node: ts.FunctionLikeDeclaration):
+      [ts.Symbol|undefined, VName|undefined] {
```
evmar, line 997: Possibly:

```
...): {sym?: ts.Symbol, vname?: VName|} { ...
```

and then on the caller side 

```
const {sym, vname} = ...
```

kythe/typescript/indexer.ts
```diff
@@ -965,6 +993,60 @@ class Visitor {
...
+    }
+    if (node.name) {
+      const sym = this.host.getSymbolAtLocation(node.name);
+      if (!sym) { return [undefined, undefined]; }
+      const vname =
+          this.host.getSymbolName(sym, TSNamespace.VALUE, context);
+      return [sym, vname];
+    } else {
+      // TODO: choose VName for anonymous functions and return symbol
+      return [undefined, this.newVName('TODO', 'TODOPath')];
```
evmar, line 1012: My above suggestion would make this just

```
return {vname: ...}
```

kythe/typescript/indexer.ts
```diff
@@ -768,6 +768,34 @@ class Visitor {
...
+    const symbol = this.host.getSymbolAtLocation(node.expression);
+    if (!symbol) { return; }
+    const name = this.host.getSymbolName(symbol, TSNamespace.VALUE);
+    if (!name) { return; }
+    this.emitEdge(callAnchor, EdgeKind.REF_CALL, name);
+
+    // Each call should have a childof edge to its containing function
+    // scope.
+    const containingFunction = this.getContainingFunctionNode(node);
+    let containingVName : VName|undefined;
```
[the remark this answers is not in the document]
evmar: (This is something I'll figure out separately, no worries.)

kythe/typescript/indexer.ts
```diff
@@ -965,6 +993,60 @@ class Visitor {
...
+          kind === ts.SyntaxKind.MethodSignature) {
+        return node as ts.FunctionLikeDeclaration;
+      }
+    }
+    return node as ts.SourceFile;
+  }
+
+  getSyntheticFileInitVName(): VName {
+    return this.newVName(
+        'fileInit:synthetic', this.host.moduleName(this.file.fileName));
```
evmar, line 1047: What is a fileInit:synthetic?  Is this a Kythe thing?

kythe/typescript/indexer.ts

kythe/typescript/indexer.ts
nbeloglazov: It's arfiticial function node. It's needed so that each `ref/call` anchor can be a parent of a function even if if `ref/call` anchor is called on top-level. Name doesn't matter, I just used it as I think it can't conflict with any other TS name.

PR #4510, opened 2020-05-06
chore(java_indexer): guard log line with --verbose flag

schroederc:


schroederc:
https://github.com/kythe/kythe/pull/4510/files#diff-bbce36275aafa80232d5a695975c78f9R139

PR #4512, opened 2020-05-12
chore(cxx_indexer): update LLVM to 123bee602a26

shahms:


kythe/cxx/indexer/cxx/recursive_type_visitor.h
```diff
@@ -232,6 +232,16 @@ DEF_TRAVERSE_TYPEPAIR(DependentVectorType, {
 DEF_TRAVERSE_TYPEPAIR(ExtVectorType, {
   return getDerived().TraverseType(TL.getTypePtr()->getElementType());
 });
+DEF_TRAVERSE_TYPEPAIR(ConstantMatrixType, {
+  return getDerived().TraverseType(TL.getTypePtr()->getElementType());
+});
+DEF_TRAVERSE_TYPEPAIR(DependentSizedMatrixType, {
+  if (T->getRowExpr())
```
salguarnieri, line 239: Does this mutate state in T? And that's why you have the same if block below?

shahms: This was taken from the internal change and it was wrong.  Cases like this should probably just delegate to `TraverseTypeLoc`, but 🤷‍♂️

shahms: (Pushing the fix now)

PR #4514, opened 2020-05-12
Use the root directory instead of the current directory.

johnedmonds:


shahms:
Justin, can you take a look at this?

justbuchanan:
Sorry for taking so long to look at this, it completely slipped past me.

The proto extractor is intended to function as a drop-in for `protoc` in terms of how it handles import search paths, etc. `protoc` does add the current working directory to the path, so I think we want to match that here. `KYTHE_ROOT_DIRECTORY` in general just affects how the paths in the vnames are relativized before being recorded in the kzip.

I'll go ahead and close this, but please re-open if it's still needed or I'm missing something.

PR #4515, opened 2020-05-12
Fix memory leak in http_server.

johnedmonds:


kythe/go/serving/xrefs/columnar.go
```diff
@@ -282,6 +282,7 @@ func (c *ColumnarTable) CrossReferences(ctx context.Context, req *xpb.CrossRefer
 			return nil, err
 		}
 		it, err := c.DB.ScanPrefix(ctx, prefix, &keyvalue.Options{LargeRead: true})
+		defer it.Close()
```
schroederc, line 285: This should go after the `err` check.

johnedmonds: Done.

PR #4516, opened 2020-05-14
fix(cxx_indexer): don't emit ref/init for unspecified fields

shahms:
This would've been much easier if Clang consistently gave imaginary expressions invalid locations, but nooooo.

### DIFF
--- a/kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
+++ b/kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
@@ -1,11 +1,19 @@
 // Tests that aggregate initialization references struct.
 
+struct Default {
+  Default();
+};
+
 //- @S defines/binding StructS
 struct S {
   //- @a defines/binding FieldA
   int a;
   //- @b defines/binding FieldB
   long b;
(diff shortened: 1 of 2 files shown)


kythe/cxx/indexer/cxx/testdata/rec/rec_agg_init.cc
```diff
@@ -38,6 +54,9 @@ void fn(T&&...);
 void f() {
   //- @S ref StructS
   //- @"1" ref/init FieldA
+  //- !{ @"}" ref/init FieldB }
+  //- !{ @"}" ref/init FieldC }
+  //- !{ @"}" ref/init FieldD }
   auto s = S{1};
 
   //- @T ref StructT
```
jaysachs, line 62: Is it worth adding some negative assertions on T as well, or is that sufficiently covered by those in f() above?

shahms: That should be covered by the S (above) and V case (below).

> Is it worth adding some negative assertions on T as well, or is that sufficiently covered by those in f() above?

Those should be sufficiently covered by the S case (above) and V case (below).

PR #4528, opened 2020-05-27
Remove layering_check and parse_headers from cxx/intexer/proto package

hlopko:
Bazel will start supporting layering check in the next version and then
this package will start failing
(https://github.com/bazelbuild/bazel/issues/11501). This PR removes
enabled  freatures from this packages until the violations are fixed.


justbuchanan:
Thanks, great to hear layering check is coming soon!

Issue #4288, opened 2019-12-23
http_server can produce invalid proto JSON


isker:
1. When analyzing Java code, `KytheTreeScanner` can emit a fact mapped to an empty string: https://github.com/kythe/kythe/blob/33bbb90c2284676d6c5acc76f76fe3e41a22c445/kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java#L655-L657

2. http_server returns proto JSON with a null map value for such a fact.  (This presumably is because the empty string is the nil string value in go.)
```
$ kythe -json -api http://localhost:8080 nodes -filters '/kythe/tag/static' kythe://kythe?lang=java?path=com.foo.ClassWithAStaticField.java#someFieldLabel
{"kythe://kythe?lang=java?path=com.foo.ClassWithAStaticField.java#someFieldLabel":{"facts":{"/kythe/tag/static":null}}}
```
3. When deserializing such JSON, the Java protobuf library rejects this as an invalid encoding.
```
 com.google.protobuf.InvalidProtocolBufferException: Map value cannot be null.
	at com.google.protobuf.util.JsonFormat$ParserImpl.mergeMapField(JsonFormat.java:1693)
	at com.google.protobuf.util.JsonFormat$ParserImpl.mergeField(JsonFormat.java:1659)
	at com.google.protobuf.util.JsonFormat$ParserImpl.mergeMessage(JsonFormat.java:1502)
	at com.google.protobuf.util.JsonFormat$ParserImpl.merge(JsonFormat.java:1460)
	at com.google.protobuf.util.JsonFormat$ParserImpl.parseFieldValue(JsonFormat.java:2001)
	at com.google.protobuf.util.JsonFormat$ParserImpl.mergeMapField(JsonFormat.java:1688)
	at com.google.protobuf.util.JsonFormat$ParserImpl.mergeField(JsonFormat.java:1659)
	at com.google.protobuf.util.JsonFormat$ParserImpl.mergeMessage(JsonFormat.java:1502)
	at com.google.protobuf.util.JsonFormat$ParserImpl.merge(JsonFormat.java:1460)
	at com.google.protobuf.util.JsonFormat$ParserImpl.merge(JsonFormat.java:1342)
	at com.google.protobuf.util.JsonFormat$Parser.merge(JsonFormat.java:477)
```
[...]

PR #4529, opened 2020-05-27
chore: migrate most jsonpb usages to proto v2 API

schroederc:


shahms:
Thank you for doing this! Can you check if this fixes #4288 ?

Issue #4520, opened 2020-05-18
Document usage of `runextractor compdb`


justbuchanan:
We can add a basic example to https://kythe.io/examples/ alongside the cmake and bazel examples.

PR #4536, opened 2020-06-02
docs(runextractor): add more detail on using runextractor

justbuchanan:


justbuchanan:
Fixes https://github.com/kythe/kythe/issues/4520

PR #4541, opened 2020-06-03
chore: partially migrate kythe/go/util to proto v2 API

schroederc:


shahms:
(although I think you'll need to update the go deps?)

PR #4542, opened 2020-06-03
feat: Allow kzip protobuf encoding to be configurable

Arm1stice:
The popular third-party Rust protobuf crates do not support reading json-type protobufs. Therefore, the Rust extractor needs to be able to create kzips that use the proto format. This PR makes that a configurable option. It also enables this new option on the Rust extractor.


kythe/go/extractors/bazel/extractor.go
```diff
@@ -68,10 +68,11 @@ import (
...
 type Config struct {
-	Corpus   string          // the default corpus label to use
-	Language string          // the language label to apply
-	Rules    vnameutil.Rules // rules for rewriting file VNames
-	Verbose  bool            // whether to emit verbose (per-file) logging
+	Corpus      string          // the default corpus label to use
+	Language    string          // the language label to apply
+	Rules       vnameutil.Rules // rules for rewriting file VNames
+	Verbose     bool            // whether to emit verbose (per-file) logging
+	ProtoFormat string          // output format of compilation unit (json or proto)
```
jaysachs, line 75: Why use a string here and not the enum?

kythe/go/extractors/bazel/settings.go
```diff
@@ -74,6 +74,8 @@ func (s *Settings) SetFlags(f *flag.FlagSet, prefix string) func() {
 		"Only match source paths within the target package")
 	f.BoolVar(&s.Verbose, p("verbose"), false,
 		"Enable verbose (per-file) logging")
+	f.StringVar(&s.ProtoFormat, p("proto_format"), "json",
```
jaysachs, line 77: There's https://github.com/kythe/kythe/blob/master/kythe/go/platform/tools/kzip/flags/encoding.go ... maybe it needs a more central location first.

kythe/go/extractors/cmd/bazel/extract_kzip/extract_kzip.go
```diff
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"kythe.io/kythe/go/extractors/bazel"
-	"kythe.io/kythe/go/extractors/bazel/extutil"
```
jaysachs, line 28: Is there a corresponding BUILD dependency to prune?

Arm1stice: Ah you're right. Removed unused dependency

kythe/go/extractors/bazel/extractor.go
Arm1stice: Good point. Changed this to use `kzip.Encoding`

justbuchanan: I'd rename the field to something like "KzipEncoding" over "ProtoFormat".

Arm1stice: Renamed fields

Arm1stice:
After internal discussions, closing this in favor of #4547

PR #4551, opened 2020-06-11
Don't attempt to extract compilation from assembly

sfreilich:
CppCompile actions (created by cc_* rules or Starlark rules using
cc_common.compile) can accept a variety of types of files in srcs,
including assembly (".s" or ".asm"). However, those source files
don't actually result in compiler jobs, so there's nothing to
extract. Attempting to run the extractor on e.g. cc_* targets
with only ".s" srcs results in an error like:

error: unable to handle compilation, expected exactly one compiler job in ''

Instead, the extractor should write an empty output in those cases and
exit, since there are no C/C++ cross-references to extract from those
sources.


shahms:
Can you run clang-format on the file? (Or just switch the system header include order manually).

PR #4554, opened 2020-06-15
feat(extraction): generate build metadata kzip in bazel-extractor docker image

justbuchanan:


kythe/extractors/bazel/extract.sh
```diff
@@ -86,5 +86,14 @@ fi
 mkdir -p "$KYTHE_OUTPUT_DIRECTORY"
 find bazel-out/*/extra_actions/external/kythe_release -name '*.kzip' | \
   xargs -r /kythe/tools/kzip merge --append --encoding "$KYTHE_KZIP_ENCODING" --output "$KYTHE_OUTPUT_DIRECTORY/compilations.kzip"
+
+# Record the timestamp of the git commit in a metadata kzip.
+METADATA_KZIP="buildmetadata.kzip"
```
jaysachs, line 91: Curious what the value of having this be an env var is?

justbuchanan: Not much, I removed it. Before I re-arranged this a bit, there was more distance between the creation of the metadata kzip and the merge step, so it made more sense then.

kythe/extractors/bazel/Dockerfile
```diff
@@ -39,7 +39,7 @@ ADD kythe/extractors/bazel/bazel_wrapper.sh /kythe/
 ADD kythe/release/base/fix_permissions.sh /kythe/
 
 # Bazelisk
-ADD external/com_github_bazelbuild_bazelisk/linux_amd64_stripped/bazelisk /kythe/
+ADD external/com_github_bazelbuild_bazelisk/bazelisk_/bazelisk /kythe/
```
jaysachs, line 42: Is this required in this CL or could/should this be a separate PR?

justbuchanan: I can split it out - bazel changed the path name, so this fix is needed for the image to build.

justbuchanan: https://github.com/kythe/kythe/pull/4555

PR #4556, opened 2020-06-16
feat(cxx_indexer): blame variable references on the surrounding context

zrlk:
This PR makes variable references similar to function calls in that they will have a `childof` edge to their most specific surrounding context.


shahms:
> High-level comment/question: maybe use "ownership" or "enclosing" instead of "blame"? (I hesitate to suggest "scope" because we're not capturing all scopes)

"parent"  (seeing as the edge is "childof")?

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -2018,6 +2018,22 @@ bool IndexerASTVisitor::VisitDeclRefOrIvarRefExpr(
     auto StmtId = BuildNodeIdForImplicitStmt(Expr);
     if (auto RCC = RangeInCurrentContext(StmtId, Range)) {
       GraphObserver::NodeId DeclId = BuildNodeIdForRefToDecl(FoundDecl);
+      if (ShouldHaveBlameContext(FoundDecl)) {
+        if (Job->BlameStack.empty()) {
+          if (auto FileId = Observer.recordFileInitializer(RCC.value())) {
+            Observer.recordBlameLocation(
+                RCC.value(), FileId.value(),
```
shahms, line 2025: nit: prefer unary-operator* over `.value()` when the instance has already been verified to contain a value (as these have been, and below).

zrlk:
We've used 'blame' in documentation (eg https://kythe.io/docs/schema/callgraph.html) and in other parts of the code (https://github.com/kythe/kythe/blob/b06edfe3f72991d315d989e0b027cac001de32cf/kythe/cxx/indexer/cxx/indexer_worklist.h#L94). I think the use here is consistent, but we could consider changing it in a different PR.

PR #4561, opened 2020-06-22
Cleanup TS indexer types and build file

ayazhafiz:
- Types kinds of nodes to be emitted as `NodeKind` rather than string
- Removes a long-standing todo
- Since we only build the indexer through bazel now (as it has proto
  dependencies in common/proto), update script in package.json to use
  bazel commands.


ayazhafiz:
cc @evmar

PR #4562, opened 2020-06-22
TS indexer: marked source for binding elements

ayazhafiz:
If the bound element has an initializer that we can narrow (i.e. the binding pattern
has an object-like literal initializer), the initializer of the bound element is that.
Otherwise the initializer is the entire variable declaration initializer, if any.

Everything else is identical to how we generate marked source for non-bound
variable declarations.


ayazhafiz:
ptal @evnmar

kythe/typescript/indexer.ts
```diff
@@ -1549,12 +1565,31 @@ class Visitor {
    *     ((local var)|const|let) <name>: <type>( = <initializer>)?
    * where `(local var)` is the declaration of a variable in a catch clause.
    */
-  emitVariableDeclarationCode(decl: ts.VariableDeclaration, declVName: VName) {
+  emitVariableDeclarationCode(decl: ts.VariableDeclaration|ts.BindingElement, declVName: VName) {
```
evmar, line 1568: Is there a better name for this function?

kythe/typescript/indexer.ts
```diff
@@ -1549,12 +1565,31 @@ class Visitor {
    *     ((local var)|const|let) <name>: <type>( = <initializer>)?
    * where `(local var)` is the declaration of a variable in a catch clause.
```
evmar, line 1566: Does this need an update for the new binding support?

kythe/typescript/indexer.ts
```diff
@@ -1549,12 +1565,31 @@ class Visitor {
...
+      // this is always a BindingPattern; the parent of the BindingPattern is
+      // another BindingPattern, a ParameterDeclaration, or VariableDeclaration.
+      // We handle ParameterDeclarations in `visitParameters`, so here we only
+      // care about the declaration from a VariableDeclaration.
+      bindingPath.push(bindingElemIndex(decl));
+      varDecl = decl.parent.parent;
+      while (!ts.isVariableDeclaration(varDecl)) {
+        if (ts.isParameter(varDecl)) return;
+        bindingPath.push(bindingElemIndex(varDecl));
+        varDecl = varDecl.parent.parent;
```
evmar, line 1583: Is it possible there's some unpredictable layout here that will cause this loop to never terminate, or to run into an undefined?

kythe/typescript/indexer.ts
evmar: Ah I guess your other CL renames it...

kythe/typescript/indexer.ts
ayazhafiz: No, the types save us from that. The type of `BindingElement#parent#parent` is `VariableDeclaration|BindingElement|ParameterDeclaration`. So eventually we will run into a variable or param decl, both of which are handled here. I'll add an undefined check just in case though.

kythe/typescript/indexer.ts
ayazhafiz: No because the form of the fact is the same. We don't emit information other than this

kythe/typescript/indexer.ts
ayazhafiz: Yeah i'll defer to the other CL

kythe/typescript/indexer.ts
```diff
@@ -729,6 +729,22 @@ function makeMarkedSource({
...
+ * Returns the property "index" of a bound element in a binding pattern.
+ * For example,
+ * - `1` has index `2` in the binding pattern `[3, 2, 1]`
+ * - `c` has index `c` in the binding pattern `{a, b, c}`
+ * - `calias` has index `c` in the binding pattern `{a, b, c: calias}`
+ */
+function bindingElemIndex(elem: ts.BindingElement): string|number{
+  const bindingPat = elem.parent;
+  if (ts.isObjectBindingPattern(bindingPat)) {
+    return (elem.propertyName || elem.name).getText();
```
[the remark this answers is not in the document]
ayazhafiz: Added better handling for computed properties

evmar:
This is still marked as 'draft' in the github UI, not sure if you meant it

ayazhafiz:
Yeah I was fixing the property names to grab the `text` property correctly. Do you mind reviewing again @evmar ?

ayazhafiz:
@evmar / @kythe can this be merged?

ayazhafiz:
@kythe could this be merged?

ayazhafiz:
@kythe can this be merged

PR #4563, opened 2020-06-23
TS Marked Source for object properties and class members

ayazhafiz:
This is a simple extension of the support we already have for marked
source on variable declarations. Class members and object properties are
given the context `(property)`, matching what tsserver does.


ayazhafiz:
ptal @evmar

PR #4564, opened 2020-06-24
feat(rust): Change Rust indexer FileProvider

Arm1stice:
This PR changes how the FileProvider struct in the Rust indexer functions. Previously, it would work with mostly String contents, however we discovered that this is not viable going forward. Because of this, we are changing the FileProvider to work with binary contents instead.

Other small changes in this PR:
- The UnknownError has been removed to avoid compatibility issues in the future.
  - Boxed error traits do not work with most CLI error-handling libraries. I plan to use [anyhow](https://crates.io/crates/anyhow) in the open source CLI, and have confirmed that it does not support boxed errors.
  - Additionally, while working on the indexer, I have not found a scenario where an UnknownError would need to be returned, because all of the failure scenarios are covered by an existing error type. If it becomes necessary in the future, I can look at re-adding an `UnknownError` type using an alternate method, but there is no benefit in keeping it in its current form.
- The Bazel rules have been updated to the latest version


Arm1stice:
Hi @filmil, can you take a look at this PR? Thanks!

filmil:
> Hi Filip Filmar, can you take a look at this PR? Thanks!

On it.

kythe/rust/indexer/src/providers.rs
```diff
@@ -56,12 +58,34 @@ impl KzipFileProvider {
...
+                // Last parent will be blank path so break when p is empty. path will contain
+                // root
+                if p == Path::new("") {
+                    break;
+                }
+                path = p;
+            }
+            root_name = path.to_str().unwrap().into();
+        }
+        // Safe to unwrap because kzip read would have failed if the internal paths
```
filmil, line 86: This comment looks like should be above line 84.

kythe/rust/indexer/src/providers.rs
```diff
@@ -91,26 +115,28 @@ impl KzipFileProvider {
 impl FileProvider for KzipFileProvider {
     /// Given a file hash, returns whether the file exists in the kzip.
     fn exists(&mut self, file_hash: &str) -> bool {
-        let name = format!("files/{}", file_hash);
+        // root_name contains a trailing forward-slash, so it's needed in the format
+        // string.
```
filmil, line 119: Consider giving an example here as well, if you think this is worth commenting.

kythe/rust/indexer/src/providers.rs
```diff
@@ -91,26 +115,28 @@ impl KzipFileProvider {
...
+        let name = format!("{}files/{}", self.root_name, file_hash);
         self.zip_archive.by_name(&name).is_ok()
     }
 
-    /// Given a file hash, returns the string contents of the file from the
+    /// Given a file hash, returns the vect of bytes of the file from the
```
filmil, line 124: Feel free to spell out vector. :)  Helps with code search later too.

kythe/rust/indexer/tests/providers.rs
```diff
@@ -50,7 +50,9 @@ fn test_kzip_provider() {
     // Check the `contents` function
     let contents_result = kzip_provider.contents(file_hash);
     assert!(!contents_result.is_err());
-    assert_eq!(contents_result.unwrap(), "Test\n", "File contents did not match expected contents");
+    let contents_string =
+        String::from_utf8(contents_result.unwrap()).expect("File contents was not valid UTF-8");
+    assert_eq!(contents_string, "Test\n", "File contents did not match expected contents");
```
filmil, line 55: This assert is actually better than the one on line 58 below.  If that one fails it won't tell you what the contents are.  This one, though, will tell you exactly what is wrong.

kythe/rust/indexer/src/error.rs
```diff
@@ -38,10 +38,5 @@ quick_error! {
         }
         // The KytheWriter encounters an error
         WriterError {}
-        // An unknown error occured
```
filmil, line 41: Would be nice to say in the commit description why this is removed.  It's not obvious.

external.bzl
```diff
@@ -31,7 +31,7 @@ def _rule_dependencies():
     rules_proto_dependencies()
     py_repositories()
     bazel_toolchains_repositories()
-    rust_repositories(version = "1.43.1")
+    rust_repositories(version = "nightly", iso_date = "2020-06-23", dev_components = True)
```
filmil, line 34: Hm you normally don't want to depend on tags that can be moved.  *If* you can do with a hash instead, it would probably be better.

Arm1stice: Unfortunately `iso_date` is the most specific field that can be set, and we require the nightly builds to get the rustc_dev component. However, the hosting for nightly builds is stable because a lot of the community depends on them for their builds. For example, the packages for this nightly build are hosted at https://static.rust-lang.org/dist/2020-06-23

kythe/rust/indexer/src/providers.rs
Arm1stice: Ah you're right

kythe/rust/indexer/tests/providers.rs
Arm1stice: One thing I can do for the assertion on line 58 is to try including the contents in the assertion error message. The issue is that I would need to unwrap the binary contents (which shouldn't cause a panic because we would assume the result is Ok) and attempt to parse a UTF-8 string from it, which might fail and cause a panic. One mitigation might be to just print out the Vec<u8> contents instead of trying to parse as a string. However, another option would be to turn that part of the test into more complex if/else logic and then just `panic!` with a specific error message depending on the resulting logical branch. Do any of these seem like good ideas?

kythe/rust/indexer/src/error.rs
Arm1stice: Ah, my PR message had a typo. I have fixed that and also provided more in-depth reasoning about why the removal was necessary

kythe/rust/indexer/src/providers.rs
Arm1stice: Good point, I went ahead and added an example to the comment

kythe/rust/indexer/src/providers.rs
Arm1stice: Oh, I believe this was a typo

kythe/rust/indexer/tests/providers.rs
filmil: Printing vec contents is quite enough if possible.  There's no need to be super-elaborate: simply revealing the information one needs for debugging would be enough.

Arm1stice: Okay, the test will now print out the vector contents in the emitted error message

Arm1stice:
@filmil I believe I have addressed your comments and made fixes. Please let me know if I missed anything

kythe/rust/indexer/src/providers.rs
```diff
@@ -91,26 +115,28 @@ impl KzipFileProvider {
...
     /// kzip.
     ///
     /// # Errors
     ///
     /// If the file does not exist in the kzip, a
     /// [FileNotFoundError][KytheError::FileNotFoundError] is returned.
-    /// If the file cannot be read into a string, a
-    /// [FileReadError][KytheError::FileReadError] is returned.
-    fn contents(&mut self, file_hash: &str) -> Result<String, KytheError> {
+    /// If the file cannot be read, a FileReadError][KytheError::FileReadError]
```
filmil, line 131: nit: here you can simply say `if the file can not be read, an error is returned`.  We know what error it is from the type signature.  This saves you some typing, as the `KytheError::FileReadError` will be hyperlinked automatically by any IDE worth its salt, or by rustdoc.

Arm1stice: So keep the hyperlink, but change the phrasing and just make the word `error` be hyperlinked to the error type instead?

filmil: You don't need to hyperlink anything here.  Case in point this would be enough:

```
If an error is returned, that means that the file could not be read.
```

PR #4570, opened 2020-06-30
Drop dependencies from TS indexer package.json

ayazhafiz:
These dependencies are picked up from the top-level package.json, so
there is no need to list them in the indexer-specific package.json.
Keeping the TS dependency also conflicts with the tsserver version
picked up by a langauge service.

The only thing left remaining in the local package.json is bazelisk for
yarn commands.


ayazhafiz:
cc @evmar

evmar:
I don't think package.json works like this -- the outer one is a totally separate project from this one.  For example each will have its own node_modules.

ayazhafiz:
Only the outer one is used in the blaze build, so keeping the build deps in the inner package.json doesn't serve a purpose AFAICT

evmar:
Hm, do you mean the bazel build?  blaze doesn't care about any of these AFAIK

ayazhafiz:
Yeah, I mean bazel. Sorry, the names always confuse me..

kythe/typescript/package.json
```diff
@@ -9,15 +9,8 @@
...
-  "dependencies": {
-    "source-map-support": "^0.4.11",
-    "typescript": "~3.5.3"
-  },
   "devDependencies": {
-    "@bazel/bazelisk": "^1.5.0",
-    "@types/jasmine": "^2.8.8",
-    "@types/node": "^10.15.3",
-    "jasmine": "^3.1.0"
+    "@bazel/bazelisk": "^1.5.0"
```
evmar, line 13: Why keep this one?

kythe/typescript/yarn.lock
```diff
@@ -6,115 +6,3 @@
   version "1.5.0"
```
evmar, line 6: I think we should delete this file

kythe/typescript/package.json
ayazhafiz: So that `npm bazel` (and the package.json commands) work in this dir. I guess we could move it up to the root, WDYT?

kythe/typescript/yarn.lock
ayazhafiz: done

kythe/typescript/package.json
evmar: Yeah, I think having two overlapping sets of dependencies is confusing

evmar: If we're using bazel, I guess we should just doc the bazel commands to run/test, rather than trying to wrap them in npm.

shahms:
Ping? Is this still relevant?

ayazhafiz:
Yes but I am no longer at Google, cannot meaningfully contribute. Closing.

PR #4576, opened 2020-07-06
Convert PackAny's parameter to absl::string_view.

mwyman:
Modernizing the API by replacing `const char *` parameter type with `string_view`.


shahms:
I think you'll need to change the implementation to explicitly call the `Any::set_type_url(const char* value, size_t size)` overload.

mwyman:
Updated to use `Any::set_type_url(const char* value, size_t size)`.

PR #4588, opened 2020-07-16
Visit JSDocs for types and interfaces

ayazhafiz:


ayazhafiz:
@kythe can this be merged

kythe/typescript/testdata/jsdoc.ts
```diff
@@ -69,3 +69,13 @@ interface String {
   /** @deprecated */
   big: () => string;
 }
+
+//- @:76"PromiseAlias" defines/binding PromiseAlias
+//- PromiseAlias.tag/deprecated "dont use me"
+/** @deprecated dont use me */
```
creachadair, line 75: One thing to watch out for when visiting markup in comments is that types mentioned in the docs might not have corresponding imports—e.g., this package doesn't depend on type T, but the doc mentions T. The code may already be robust against unresolved types, but it's probably worth double-checking and maybe adding a test case to verify it doesn't explode. YMMV.

ayazhafiz: That's a good point, currently the TS indexer does not visit identifiers mentioned in comments. This should be kept in mind in case we ever do such a thing.

PR #4590, opened 2020-07-17
Emit marked source for function parameters

ayazhafiz:
Luckily this is also a trivial extension of the marked source we already
emit for other variables.


ayazhafiz:
@kythe please take a look

ayazhafiz:
@schroederc @zrlk shahms mentioned you two are the most familiar with marked source on another PR, do you mind looking at this one too?

ayazhafiz:
@kythe could this be reviewed?

kythe/typescript/indexer.ts
```diff
@@ -1596,8 +1597,11 @@ class Visitor {
     let declKw;
     if (ts.isVariableDeclaration(varDecl)) {
       declKw = initializerList.kind === ts.SyntaxKind.CatchClause ?
-          '(local var)' :
-          initializerList.flags & ts.NodeFlags.Const ? 'const' : 'let';
+                                                       '(local var)' :
+          initializerList.flags & ts.NodeFlags.Const ? 'const' :
+                                                       'let';
+    } else if (ts.isParameter(varDecl)) {
+      declKw = '(parameter)';
```
evmar, line 1604: BTW, do you think it is valuable to include this info in the output?  I feel like as a user of a variable I don't really care if it's a let vs a var vs a param or whatever.

ayazhafiz: I think it will be valuable when we get to things like functions and classes, where the declaration context provides more insight than in this case. I also think it worth providing as much information as possible via marked source, and then the UI can decide what is displayed to the user.

kythe/typescript/indexer.ts
```diff
@@ -1575,19 +1577,18 @@ class Visitor {
...
       // this is always a BindingPattern; the parent of the BindingPattern is
       // another BindingPattern, a ParameterDeclaration, or VariableDeclaration.
-      // We handle ParameterDeclarations in `visitParameters`, so here we only
-      // care about the declaration from a VariableDeclaration.
       bindingPath.push(this.bindingElemIndex(decl));
       varDecl = decl.parent.parent;
-      while (!ts.isVariableDeclaration(varDecl) && varDecl !== undefined) {
+      while (!ts.isVariableDeclaration(varDecl) && !ts.isParameter(varDecl) &&
+             varDecl !== undefined) {
         if (ts.isParameter(varDecl)) return;
```
schroederc, line 1584: Dead code?

ayazhafiz: Yep, thanks. Removed.

kythe/typescript/testdata/marked_source/function.ts
```diff
@@ -0,0 +1,44 @@
+//- @p1 defines/binding A_p1
+//- A_p1 code A_p1_code
+//- A_p1_code child.0 A_p1_context
```
schroederc, line 3: To match the rest of the indexers, the context of a parameter's MarkedSource should be the qualified name of its enclosing function.

ayazhafiz: I see. My impression was the context was the declaration context; for example, for mutable local variable we currently say the context is `let` and for constants it is `const`. For a local variable like `let a = 1` should the context be something else?

shahms:
Ping? Is this still relevant?

ayazhafiz:
Yes but I am no longer at Google, cannot meaningfully contribute. Closing.

PR #4591, opened 2020-07-17
Do not group files in nested test directories by default

ayazhafiz:
Previously test files for the TS indexer that were in a common
subdirectory would be tested together. This is useful in a lot of cases
but isn't the behavior we want for marked source tests, which are put in
a separate directory because they test a lot of code that is specific to
marked source and whose files are independent from each other.

To ensure that files for marked source tests are treated independently,
this commit groups tests only when the nested test directory ends with
`_group`.


ayazhafiz:
@kythe PTAL

kythe/typescript/test.ts
```diff
@@ -175,28 +181,30 @@ function filterTestFileGroups(
...
+      const start = new Date().valueOf();
+      process.stdout.write(`${testName}: `);
+      try {
+        await verify(host, config.options, testFiles, plugins);
+      } catch (e) {
+        console.log('FAIL');
+        throw e;
+      }
+      const time = new Date().valueOf() - start;
+      console.log('PASS', time + 'ms');
```
shahms, line 206: JFYI, this generates a warning during import that would be nice to clean up at some point.
"Operands of '+' operation must either be both strings or both numbers or both bigints, but found number + "ms". Consider using template literals."

ayazhafiz: thanks for the pointer, see #4593

PR #4594, opened 2020-07-17
feat: Add library for generating Rust save_analysis files

Arm1stice:
This PR adds a new library that exposes an interface to generate Rust save_analysis files. This library will be used by the new Bazel Rust extractor, which I will create a PR for after this is merged.


Arm1stice:
@filmil can you take a look at this PR?

kythe/rust/extractor/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    let mut callback_shim = CallbackShim::new(output_dir);
+
+    // Enable the output of compilation errors to stderr
+    rustc_driver::install_ice_hook();
+
+    rustc_driver::catch_fatal_errors(|| {
+        run_compiler(&rustc_arguments, &mut callback_shim, None, None)
+    })
+    .map(|_| ())
+    .map_err(|_| ())?;
```
filmil, line 44: This looks like it is swallowing any error in running the compiler.  Is there any chance to log the error at least?  Otherwise it may be quite confusing.

kythe/rust/extractor/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+use rustc_driver::run_compiler;
+use rustc_driver::{Callbacks, Compilation};
+use rustc_interface::{interface, Queries};
+use rustc_save_analysis::DumpHandler;
+use std::path::PathBuf;
+
+/// Generate a save_analysis in `output_dir`
+///
+/// `rustc_arguments` is a `Vec<String>` containing Rust compiler arguments. The
+/// first element must be an empty string.
```
filmil, line 31: If the first element must be an empty string, how about adding an explicit check and erroring out if it is not?

kythe/rust/extractor/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/// Handles compiler callbacks to enable and dump the save_analysis
+#[derive(Default)]
+struct CallbackShim {
+    output_dir: PathBuf,
+}
+
+impl CallbackShim {
+    pub fn new(output_dir: PathBuf) -> Self {
```
filmil, line 56: Always document the `pub` API.  This is what users will want to know the usage for.

kythe/rust/extractor/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+        // `rls_data::config::Config` and ours, even though we use the same version.
+        // This forces us to use the environment variable method of configuration
+        // instead.
+        std::env::set_var(
+            "RUST_SAVE_ANALYSIS_CONFIG",
+            r#"{"output_file":null,"full_docs":true,"pub_only":false,"reachable_only":false,"distro_crate":false,"signatures":false,"borrow_data":false}"#,
+        );
+
+        // Perform the save_analysis and dump it to the directory
+        // The JSON file is saved at {self.output_dir}/save-analysis/{crate_name}.json
```
filmil, line 88: Have you tested this?

I ask because the other day I learned that rustc ignores the `output_dir` parameter if the `-o` has an absolute file path.  It may be a feature of `rustc`, not the `rustc_driver` but still seems worth while to check.

kythe/rust/extractor/src/lib.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 The Kythe Authors. All rights reserved.
```
filmil, line 1: Can you add a test?

kythe/rust/extractor/src/lib.rs
Arm1stice: https://github.com/kythe/kythe/blob/8127f0e701acbe46ba3fa60691455657486f69b6/kythe/rust/extractor/src/lib.rs#L38 tells the compiler to output any compiler errors to stderr on its own. The only error type that gets returned with an `Err` result is [ErrorReported](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/struct.ErrorReported.html) which is not an actual error type.

filmil: I think it is still useful to at least log the error, if not propagate it upwards.  The type signature here could change or `ErrorReported` may gain additional useful state.  In general, never replace errors with "".

kythe/rust/extractor/src/lib.rs
Arm1stice: https://github.com/kythe/kythe/blob/8127f0e701acbe46ba3fa60691455657486f69b6/kythe/rust/extractor/src/lib.rs#L95 is how we configure where to save the save_analysis. In my testing, this method is unaffected by any output parameters passed to the Rust compiler

filmil: Adding a test seems like an ideal way to put this concern to rest.

kythe/rust/extractor/src/lib.rs
Arm1stice: The compiler errors already get logged without needing to manually log messages, and we are not provided with a copy of the compiler errors that get logged. Our function also returns an `Err` result to the caller, letting them know that the process failed. 
https://github.com/rust-lang/rls/blob/dd341d53075411cf0f71c5853c98facda35112ef/rls-rustc/src/lib.rs#L80  
also uses the same method, so there is precedent for this. `ErrorReported` is a bit of a misnomer, because it is not a struct that contains the error that was reported but rather indicates that an error *was* reported during compilation. It is empty struct by design.

Arm1stice: As a compromise I have made the function return a String error containing information about what error occurred and have made it return `"A compiler error has occurred"` when the compilation fails. This also addresses your comment above about ensuring the first argument is an empty string

kythe/rust/extractor/extractor_test.bzl
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+
+def _rust_extractor_test_impl(ctx):
+    test_binary = ctx.executable.src
+
+    lib = ctx.files.lib
+    sysroot = ctx.files.sysroot
+    # Get the parent directories for the library and sysroot files
+    lib_path = "/".join(lib[0].path.split("/")[:-1])
+    sysroot_path = "/".join(sysroot[0].path.split("/")[:-1])
```
shahms, line 22: Use the skylib `paths` module:
```
load("@bazel_skylib//lib:paths.bzl", "paths")
```

kythe/rust/extractor/extractor_test.bzl
```diff
@@ -0,0 +1,68 @@
...
+            mandatory = True,
+            allow_files = True,
+            doc = "A label for the filegroup containing the rustc library files"
+        ),
+        "sysroot": attr.label(
+            mandatory = True,
+            allow_files = True,
+            doc = "A label for the filegroup containing the rustc sysroot files"
+        ),
+        "test_file": attr.label(
```
shahms, line 61: What is being tested? The binary or the contents of the file?

kythe/rust/extractor/extractor_test.bzl
```diff
@@ -0,0 +1,68 @@
...
+        "lib": attr.label(
+            mandatory = True,
+            allow_files = True,
+            doc = "A label for the filegroup containing the rustc library files"
+        ),
+        "sysroot": attr.label(
+            mandatory = True,
+            allow_files = True,
+            doc = "A label for the filegroup containing the rustc sysroot files"
+        ),
```
shahms, line 60: Are lib/sysroot going to vary per-test or are they logically attributes of the toolchain/rule itself?  If every test has to provide the same value for these attributes (especially if that value is a `select()`) you should consider making them internal attribute (prefixed with a "_") pointing to a default label that is an `alias` rule itself containing the select (or use the Rust rules `toolchain` instead)

Arm1stice:
@filmil I have added tests to this PR now

kythe/rust/extractor/extractor_test.bzl
Arm1stice: The binary is being tested, but that binary compiles the Rust file provided by test_file. Should I change that attribute name?

shahms: That depends, the semantics of the rule aren't entirely clear.  Are both `test_file` and `src` likely to vary per-test or only one of them?

Arm1stice: `test_file` will probably not be configured. I have changed the rule to write the file itself to avoid using an attribute

kythe/rust/extractor/extractor_test.bzl
Arm1stice: Ah that makes sense! I have switched them to aliases and used internal attributes with default labels as you suggested.

kythe/rust/extractor/rustfmt.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
zrlk, line 1: Is there a reason this file (and its config) is in kythe/rust/extractor? A better place might be `/tools/rust`.

Arm1stice: That's true. I have created a universal script in `/tools/rust` and removed the scripts from the extractor and indexer directories

kythe/rust/extractor/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+use rustc_save_analysis::DumpHandler;
+use std::path::PathBuf;
+
+/// Generate a save_analysis in `output_dir`
+///
+/// `rustc_arguments` is a `Vec<String>` containing Rust compiler arguments. The
+/// first element must be an empty string.
+/// The save_analysis JSON output file will be located at
+/// {output_dir}/save-analysis/{crate_name}.json
+pub fn generate_analysis(rustc_arguments: Vec<String>, output_dir: PathBuf) -> Result<(), String> {
```
shahms, line 33: I missed this before, but the only analysis the extractor should be doing is determining what files are necessary for the build.  The Java and C++ extractors run the corresponding compiler for related reasons.  In Java's case the naive estimate of required files tends to vastly overestimate what's actually required, so it runs the compiler to determine the set of files actually inspected during compilation (and handle plugins/annotation processors).  C++ is similar and we need to run the preprocessor to actually determine the necessary files.

This might be true for Rust, but we should *not* be doing any more analysis than that.

PR #4596, opened 2020-07-20
chore: update TS deps

shahms:
Address the lodash vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,17 +1,17 @@
 {
   "devDependencies": {
-    "@bazel/jasmine": "^1.5.0",
-    "@bazel/labs": "^1.5.0",
-    "@bazel/typescript": "^1.5.0",
-    "@commitlint/cli": "^8.3.5",
+    "@bazel/jasmine": "^1.7.0",
+    "@bazel/labs": "^1.7.0",
+    "@bazel/typescript": "^1.7.0",
+    "@commitlint/cli": "^9.1.2",
     "@commitlint/config-conventional": "^8.3.4",
-    "@types/jasmine": "^3.5.10",
-    "@types/node": "^12.12.34"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,17 +1,17 @@
...
+    "@types/jasmine": "^3.5.11",
+    "@types/node": "^12.12.50"
   },
   "dependencies": {
-    "google-protobuf": "^3.12.1",
-    "source-map-support": "^0.5.16",
-    "typescript": "^3.8.3"
+    "google-protobuf": "^3.12.2",
+    "source-map-support": "^0.5.19",
+    "typescript": "^3.9.7"
```
evmar, line 14: This last line is changing TS language versions (they don't obey semver).  It's not clear to me whether the TS compilation bits of your build infra obeys this package.json, but I wouldn't expect it to.  So I think you should revert this line in particular.

shahms: I tried doing this, but the version in `package-lock.json` remains unchanged. Is that a problem?

Also, the BUILD rules should obey this file (from WORKSPACE):
```
npm_install(
    name = "npm",
    package_json = "//:package.json",
    package_lock_json = "//:package-lock.json",
)
```

evmar: The version pin here (`^3.9.7`) I think allows for any "semver-compatible" version.  I think the right way to express our current dep is
`"~3.8"`
to say we accept patchlevels on that verison (not sure, I just read the docs now).

I believe https://github.com/kythe/kythe/blob/ade706d5409a9a997f85d14a1251033577f1f5a1/kythe/typescript/BUILD#L3 controls which compiler is used, but I am not too familiar with these build rules.

shahms: Thanks!

shahms:
This look good Evan?

PR #4602, opened 2020-07-23
feat: Create new Rust Bazel extractor

Arm1stice:
This PR adds a new Bazel extractor for Rust to replace the previous implementation. The extractor gathers the compiler arguments and source files using the extra action API and uses that information to generate a save_analysis JSON file and create a kzip.

The extractor can be tested with the following command:
```
bazelisk build --experimental_action_listener=@io_kythe//kythe/extractors:extract_kzip_rust //kythe/rust/examples/hello_world --action_env=KYTHE_CORPUS=test_corpus
```

The extractor is an MVP and currently does not support the `vnames_config` functionality. Support will be added in a future PR.


kythe/extractors/BUILD
```diff
@@ -171,17 +171,15 @@ extractor_action(
...
     data = ["//external:vnames_config"],
-    extractor = "@io_kythe//kythe/go/extractors/cmd/bazel:extract_kzip",
+    extractor = "@io_kythe//kythe/rust/extractor",
     mnemonics = [
         "Rustc",
     ],
     output = "$(ACTION_ID).rust.kzip",
+    env = {
+        "LD_LIBRARY_PATH": "external/rust_linux_x86_64/lib"
+    }
```
Arm1stice, line 184: Need to add a select in here so that it works on macOS

kythe/rust/extractor/src/bin/extractor.rs
```diff
@@ -0,0 +1,230 @@
...
+    source_files: Vec<String>,
+    arguments: Vec<String>,
+    build_output_path: String,
+    required_input: Vec<CompilationUnit_FileInput>,
+) -> IndexedCompilation {
+    let mut compilation_unit = CompilationUnit::new();
+
+    let mut vname = VName::new();
+    // TODO(Arm1stice): Determine proper Corpus/Root/Path for VName
+    vname.set_corpus("kythe".into());
```
shahms, line 135: The `kythe` corpus is reserved and shouldn't be used during extraction.  It should generally be read from a `KYTHE_CORPUS` environment variable.  The C++ indexer defaults to using the corpus of the main source file if the environment variable is unset or empty.

kythe/rust/extractor/src/bin/save_analysis.rs
```diff
@@ -0,0 +1,57 @@
...
+use anyhow::Result;
+use std::path::PathBuf;
+
+/// Generates a save analysis in `output_dir`
+///
+/// * `arguments` - The Bazel arguments extracted from the extra action protobuf
+/// * `output_dir` - The base directory to output the save_analysis
+pub fn generate_save_analysis(arguments: Vec<String>, output_dir: PathBuf) -> Result<()> {
+    let rustc_arguments = generate_arguments(arguments, &output_dir)?;
+    let _input_files = kythe_rust_extractor::generate_analysis(rustc_arguments, output_dir)
```
shahms, line 24: I must have missed this earlier, but the extractor shouldn't be doing any analsysis other than determining what source files are required.

zrlk: Unfortunately, we need to do some extra work during extraction because intermediate build results are tied to the exact compiler revision used to build them. This is like the problem we have with modules in Clang/PCHs, but worse because there's no fallback for compatibility.

kythe/rust/extractor/BUILD
```diff
@@ -28,10 +29,29 @@ rust_library(
...
+        "//kythe/rust/extractor/cargo:tempdir",
+        "//kythe/rust/extractor/cargo:rust_crypto",
+    ] + PROTO_COMPILE_DEPS,
+)
+
 rust_clippy(
     name = "clippy",
     deps = [
         ":kythe_rust_extractor",
+	":extractor",
```
zrlk, line 54: Please run this through buildifier.

kythe/rust/extractor/cargo/remote/winapi-0.3.9.BUILD
```diff
@@ -39,12 +39,18 @@ rust_library(
     version = "0.3.9",
```
zrlk, line 39: How do we end up depending on the Windows API?

kythe/rust/extractor/src/bin/extractor.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    println!("Generated {:?}", &config.output_path);
+    Ok(())
+}
+
+// Inspired by https://gist.github.com/Igosuki/27e18369fec5a05d4019ac5d321e1779
+const SPAWN_INFO: protobuf::ext::ExtFieldOptional<
+    ExtraActionInfo,
+    protobuf::types::ProtobufTypeMessage<SpawnInfo>,
+> = protobuf::ext::ExtFieldOptional { field_number: 1003, phantom: ::std::marker::PhantomData };
```
zrlk, line 101: Please use a named constant instead of `1003`, preferably somewhere from the generated protobuf code.

kythe/rust/extractor/src/bin/extractor.rs
```diff
@@ -0,0 +1,230 @@
...
+    compilation_unit.set_required_input(protobuf::RepeatedField::from_vec(required_input));
+    compilation_unit.set_output_key(build_output_path);
+
+    let mut indexed_compilation = IndexedCompilation::new();
+    indexed_compilation.set_unit(compilation_unit);
+    indexed_compilation
+}
+
+/// Generate sha256 hex digest of a vector of bytes
+fn sha256digest(bytes: &[u8]) -> String {
```
zrlk, line 149: The result of this function should always be lowercase.

Arm1stice: The tests for the rust-crypto crate ensure that the `.result_str()` function always returns a lowercase hex string

kythe/rust/extractor/src/bin/extractor.rs
Arm1stice: The generated Rust code strips all field numbers, but I have created a constant variable stating that it is a field number

kythe/rust/extractor/src/bin/save_analysis.rs
shahms: I agree this is the expedient solution (and that Wyatt should proceed), but it's neither the only solution nor a particularly robust solution as the generated JSON is *also* tightly versioned (and potentially insufficient).

kythe/rust/extractor/src/bin/extractor.rs
Arm1stice: What would you recommend as a temporary corpus until the vname_config functionality is implemented?

kythe/rust/extractor/src/bin/save_analysis.rs
Arm1stice: The difference with the generated JSON is that we control the versioning with the version of `rustc_save_analysis` that we use in the extractor which is what controls the `save_analysis` version, compared to the `.rlib` versioning being controlled by a team's choice of their Rust compiler version. The `rls_data` crate, which houses the `Analysis` data structure, has also been stable since April 2019.

But I also agree that I should be thinking about a way to transition off of `save_analysis` for the future, especially if the feature will eventually be deprecated.

kythe/rust/extractor/src/bin/extractor.rs
shahms: I would read from the `KYTHE_CORPUS` environment variable unconditionally and probably emit an error if it is unset.

Arm1stice:
@zrlk can you use the command in the PR description to make sure that it works on macOS?

zrlk:
It seems like the extractor isn't getting passed some required environment variables for xcode. If I change the example command to

`bazelisk build --sandbox_debug --experimental_action_listener=@io_kythe//kythe/extractors:extract_kzip_rust //kythe/rust/examples/hello_world --action_env=DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer --action_env=KYTHE_CORPUS=hello --action_env=SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk`

it runs. I got DEVELOPER_DIR from `xcode-select --print-path` and SDKROOT from `xcodebuild -sdk -version`.

Since this is xcode-specific and I don't believe you have any reasonable way to access an OS X machine, I think this is okay to land as it is. I'll take an issue to fix it myself as I have a vested interest in getting it working on my laptop.

zrlk:
OK, I've filed #4604 and self-assigned it.

salguarnieri:
I think https://github.com/kythe/kythe/blob/master/kythe/cxx/extractor/xcode_wrapper.sh shows how we handle getting those values for the cxx extractor.

PR #4603, opened 2020-07-24
Add "web UI manual build note" to documentation

MinaToma:
Hi,

I am new to Kythe and I was following the documentation [here](https://kythe.io/examples/) and I faced some problems with web/ui and visualizing cross-references and found that it needs to be built manually (after some time).

I think it would be great if the documentation mentions that web/ui needs to be built from source first before visualizing.


MinaToma:
@googlebot I signed it!

shahms:
This is fine, but the reason web/ui was removed from the distribution is that it is unsupported and unmaintained.

PR #4607, opened 2020-07-28
feat(release): use statically linked C++ release binaries

### LINKED ISSUE #4598: Kythe releases should use fully static binaries
Right now the binaries for the Kythe release are "mostly static" which introduces a dependency on platform-specific C/C++ libraries.  We could increase the portability of our binaries by linking them entirely statically.

shahms:
Fixes #4598

### DIFF
--- a/kythe/cxx/extractor/BUILD
+++ b/kythe/cxx/extractor/BUILD
@@ -5,6 +5,7 @@ package(
 
 load("//:visibility.bzl", "PUBLIC_VISIBILITY")
 load(":toolchain.bzl", "cxx_extractor_toolchain")
+load("//tools/cpp:cc_release_binary.bzl", "cc_release_binary")
 
 cc_library(
     name = "cxx_details",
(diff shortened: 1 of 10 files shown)


shahms:
The test failure there is a simple naming issue, but fixing that reveals that the verifier itself does not like being a fully static binary:

```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x0000000000782b71 in std::locale::_S_initialize() ()
#2  0x0000000000782be9 in std::locale::locale() ()
#3  0x000000000077ff23 in std::ios_base::Init::Init() ()
#4  0x0000000000403b60 in __cxx_global_var_init () at /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/iostream:74
#5  0x0000000000403e89 in _GLOBAL__sub_I_verifier_main.cc ()
#6  0x0000000000829ea4 in __libc_csu_init ()
#7  0x0000000000829752 in __libc_start_main ()
#8  0x0000000000406c7a in _start ()
```
That is not an encouraging stacktrace.  For now, I can extra triple check that the other binaries work fine and keep the verifier as a mostly-static binary.

shahms:
Given the problems with verifying fully static binaries, I think we'll need a different approach.

Issue #4606, opened 2020-07-28
Compilation extractor for Rust on Fuchsia OS


filmil:
This is a feature request for a Fuchsia-specific Rust compilation extractor.

(Please assign me to this issue, as I'm working on it.)

The idea is to reuse the in-development indexer from https://github.com/kythe/kythe/pull/4605.  Fuchsia's extractor is a bit specific as Fuchsia has its own non-cargo Rust toolchain based on GN.

At the moment I believe that it is sufficiently different from other extractors because of this to be its own separate binary. However, it can not live inside Fuchsia's tree because Kythe is complex enough that importing it in tree re licensing isn't feasible.  So the idea is, at the moment, to work on Fuchsia's Rust indexer inside the Kythe tree.  Let me know if this is not desired for any reason.

The rough approach is as follows:
* I'd like to produce `kzip`s such that @Arm1stice 's Rust indexer CLI can reuse it right off the bat.
* I will reuse the C++ `kzip` writer library, with a C API wrapper and FFI into Rust.
* Custom `CompilationUnit` population code will be needed.
* There will be followup work in Fuchsia tree.

PR #4608, opened 2020-07-28
KZIP Writer C API (and Rust FFI)

filmil:
This will allow me to reuse the C++ kzip writing code for
the rust indexer instead of redoing that work in rust.

Issue #4606

### DIFF
--- a/kythe/cxx/common/BUILD
+++ b/kythe/cxx/common/BUILD
@@ -66,7 +66,7 @@ cc_library(
     hdrs = ["path_utils.h"],
     visibility = [PUBLIC_VISIBILITY],
     deps = [
-        ":status",
+        ":common_status",
         ":status_or",
         "@com_github_google_glog//:glog",
         "@com_google_absl//absl/container:flat_hash_map",
@@ -447,7 +447,9 @@ cc_test(
 )
 
 cc_library(
-    name = "status",
+    # The name can not be just 'status' as it conflicts with absl's status
+    # library at link time.  Weird, but true.
+    name = "common_status",
     srcs = ["status.cc"],
     hdrs = ["status.h"],
     visibility = [PUBLIC_VISIBILITY],
@@ -475,7 +477,7 @@ cc_library(
     srcs = ["libzip/error.cc"],
     hdrs = ["libzip/error.h"],
     deps = [
-        ":status",
+        ":common_status",
         "@com_google_absl//absl/status",
         "@org_libzip//:zip",
     ],
--- a/kythe/cxx/common/BUILD
+++ b/kythe/cxx/common/BUILD
@@ -421,6 +421,7 @@ cc_library(
         ":kzip_writer",
         ":status_or",
         "//kythe/proto:analysis_cc_proto",
+        "@com_github_google_glog//:glog",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/strings",
         "@com_google_protobuf//:protobuf",
@@ -440,6 +441,7 @@ cc_test(
         "//kythe/proto:go_cc_proto",  # Used in stringset.kzip.
         "//third_party:gtest",
         "//third_party:gtest_main",
+        "@com_github_google_glog//:glog",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/strings",
         "@org_libzip//:zip",
@@ -447,8 +449,9 @@ cc_test(
 )
 
 cc_library(
-    # The name can not be just 'status' as it conflicts with absl's status
-    # library at link time.  Weird, but true.
+    # Having this library be named "status" just as that of absl status
+    # affects linking this library with Rust FFI code.
+    # See for details: https://github.com/bazelbuild/rules_rust/issues/385.
     name = "common_status",
     srcs = ["status.cc"],
     hdrs = ["status.h"],
--- a/kythe/cxx/common/BUILD
+++ b/kythe/cxx/common/BUILD
@@ -410,6 +410,43 @@ cc_test(
 )
 
 cc_library(
+    name = "kzip_writer_c_api",
+    srcs = ["kzip_writer_c_api.cc"],
+    hdrs = ["kzip_writer_c_api.h"],
+    visibility = [
+        "//visibility:public",
+    ],
+    deps = [
+        ":index_writer",
+        ":kzip_writer",
+        ":status_or",
+        "//kythe/proto:analysis_cc_proto",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/strings",
+        "@com_google_protobuf//:protobuf",
+    ],
+)
+
+cc_test(
+    name = "kzip_writer_c_api_test",
+    srcs = ["kzip_writer_c_api_test.cc"],
+    data = [
+        "//kythe/testdata/platform:stringset.kzip",
+    ],
+    deps = [
+        ":kzip_reader",
+        ":kzip_writer_c_api",
+        ":libzip/error",
+        "//kythe/proto:go_cc_proto",  # Used in stringset.kzip.
+        "//third_party:gtest",
+        "//third_party:gtest_main",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/strings",
+        "@org_libzip//:zip",
+    ],
+)
+
+cc_library(
     name = "status",
     srcs = ["status.cc"],
     hdrs = ["status.h"],
--- a/kythe/cxx/common/kzip_writer.h
+++ b/kythe/cxx/common/kzip_writer.h
@@ -39,12 +39,6 @@ class KzipWriter : public IndexWriterInterface {
   /// \param encoding Encoding to use for compilation units.
   static StatusOr<IndexWriter> Create(
       absl::string_view path, KzipEncoding encoding = DefaultEncoding());
-  /// \brief Constructs a Kzip IndexWriter which will create and write to
-  /// \param path Path to the file to create. Must not currently exist.
-  /// \param encoding Encoding to use for compilation units.
-  /// Same as `Create` except it returns a pointer.
-  static StatusOr<std::unique_ptr<IndexWriter>> CreateUnique(
-      absl::string_view path, KzipEncoding encoding = DefaultEncoding());
   /// \brief Constructs an IndexWriter from the libzip source pointer.
   /// \param source zip_source_t to use as backing store.
   /// See https://libzip.org/documentation/zip_source.html for ownership.
--- a/kythe/cxx/common/kzip_writer.h
+++ b/kythe/cxx/common/kzip_writer.h
@@ -39,6 +39,12 @@ class KzipWriter : public IndexWriterInterface {
   /// \param encoding Encoding to use for compilation units.
   static StatusOr<IndexWriter> Create(
       absl::string_view path, KzipEncoding encoding = DefaultEncoding());
+  /// \brief Constructs a Kzip IndexWriter which will create and write to
+  /// \param path Path to the file to create. Must not currently exist.
+  /// \param encoding Encoding to use for compilation units.
+  /// Same as `Create` except it returns a pointer.
+  static StatusOr<std::unique_ptr<IndexWriter>> CreateUnique(
+      absl::string_view path, KzipEncoding encoding = DefaultEncoding());
   /// \brief Constructs an IndexWriter from the libzip source pointer.
   /// \param source zip_source_t to use as backing store.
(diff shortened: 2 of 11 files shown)


kythe/cxx/common/kzip_writer_c_api_test.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "kythe/proto/go.pb.h"
+
+namespace kythe {
+namespace {
+using ::testing::ElementsAre;
+using ::testing::Values;
+
+absl::string_view TestTmpdir() {
+  return absl::StripSuffix(std::getenv("TEST_TMPDIR"), "/");
+}
```
shahms, line 43: This should just be `::testing::TempDir()`

kythe/cxx/common/kzip_writer_c_api.h
```diff
@@ -0,0 +1,96 @@
...
+/// \brief Closes the writer.
+/// Must be called before destroying the object.
+int32_t KzipWriter_Close(struct KzipWriter* writer);
+
+/// \brief Writes a piece of content into the kzip, returning a digest.
+///
+/// The content is
+/// specified using both the pointer to the beginning and the size.  The caller
+/// must provide a buffer to put the digest into, and a buffer size.  Nonzero
+/// status is returned in case of an error writing.
```
shahms, line 71: Make sure to document specifically how "insufficient space in output buffer" is reported here and elsewhere.

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+  }
+  std::unique_ptr<IndexWriter>& writer = *writer_or;
+  KzipWriter* writer_ptr = reinterpret_cast<KzipWriter*>(writer.get());
+  writer.release();
+  return writer_ptr;
+}
+
+void KzipWriter_Delete(KzipWriter* writer) {
+  // We assume that `writer` was created by `KzipWriter_Create`.
+  std::unique_ptr<IndexWriter> _delete(reinterpret_cast<IndexWriter*>(writer));
```
shahms, line 80: Just call `delete reinterpret_cast<IndexWriter*>(writer);`

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+  StatusOr<std::unique_ptr<IndexWriter>> writer_or =
+      kythe::KzipWriter::CreateUnique(path_view, kzip_encoding);
+  if (!writer_or.ok()) {
+    *create_status = CodeFromStatus(writer_or.status());
+    return nullptr;
+  }
+  std::unique_ptr<IndexWriter>& writer = *writer_or;
+  KzipWriter* writer_ptr = reinterpret_cast<KzipWriter*>(writer.get());
+  writer.release();
+  return writer_ptr;
```
shahms, line 75: I think you likely just want:
```
StatusOr<kythe::IndexWriter> writer = kythe::KzipWriter::Create(...);
if (!writer.ok()) {
  *create_status = CodeFromStatus(writer.status());
  return nullptr;
}
return reinterpret_cast<::KzipWriter*>(new kythe::IndexWriter(*std::move(writer)));
```

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+KzipWriter* KzipWriter_Create(const char* path, const size_t path_len,
+                              int32_t encoding, int32_t* create_status) {
+  *create_status = 0;
+  KzipEncoding kzip_encoding = ToKzipEncoding(encoding);
+  const absl::string_view path_view(path, path_len);
+
+  StatusOr<std::unique_ptr<IndexWriter>> writer_or =
+      kythe::KzipWriter::CreateUnique(path_view, kzip_encoding);
+  if (!writer_or.ok()) {
+    *create_status = CodeFromStatus(writer_or.status());
```
shahms, line 69: I think dropping the message loses a lot of relevant information.  I'm not sure creating a full-fledged C-API status type is warranted, but at least logging the status may be?  Given that you'd want to do this consistently it may make sense to have CodeFromStatus (or a differently named equivalent) do so.

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+  return static_cast<int32_t>(s.code());
+}
+
+int32_t KzipWriter_WriteFile(KzipWriter* writer, const char* content,
+                             const size_t content_length, char* digest_buffer,
+                             size_t buffer_length,
+                             size_t* resulting_digest_size) {
+  // We assume that `writer` was created by `KzipWriter_Create`.
+  IndexWriter* w = reinterpret_cast<IndexWriter*>(writer);
+  absl::string_view view(content, content_length);
```
shahms, line 95: nit: I wouldn't bother with the local here, just:

```
StatusOr<std::string> digest = w->WriteFile({content, content_length});
```

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+  auto digest_or = w->WriteFile(content);
```
shahms, line 96: prefer to include the full type for StatusOr<T> and omit the suffix.

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+  if (!digest_or.ok()) {
+    return CodeFromStatus(digest_or.status());
+  }
+  const auto& digest = *digest_or;
+  if (digest.length() > buffer_length) {
+    return KZIP_WRITER_BUFFER_TOO_SMALL_ERROR;
+  }
+  memcpy(digest_buffer, digest.c_str(), digest.length());
```
shahms, line 104: It might be nice to leave the resulting string null-terminated, if possible.

kythe/cxx/common/kzip_writer_c_api.cc
```diff
@@ -0,0 +1,132 @@
...
+  return 0;
+}
+
+int32_t KzipWriter_WriteUnit(KzipWriter* writer, const char* proto,
+                             size_t proto_length, char* digest_buffer,
+                             size_t buffer_length,
+                             size_t* resulting_digest_size) {
+  //// We assume that `writer` was created by `KzipWriter_Create`.
+  IndexWriter* w = reinterpret_cast<IndexWriter*>(writer);
+  const std::string_view proto_bytes(proto, proto_length);
```
shahms, line 115: `std::string` also has a (char*, len) constructor so you can do the same thing here as above:

unit.ParseFromString({proto, proto_length});

or better:

unit.ParseFromArray(proto, proto_length);

kythe/cxx/common/BUILD
```diff
@@ -410,7 +410,46 @@ cc_test(
...
+        "//third_party:gtest_main",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/strings",
+        "@org_libzip//:zip",
+    ],
+)
+
+cc_library(
+    # The name can not be just 'status' as it conflicts with absl's status
+    # library at link time.  Weird, but true.
```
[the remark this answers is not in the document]
filmil: This issue summarizes my observations: https://github.com/bazelbuild/rules_rust/issues/385
cc_library name matters for rust linking.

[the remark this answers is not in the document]
shahms: Thanks! That does seem to be a bug in the Rust rules.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Done, thank you for the hint.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: I would advise against that. 

asciiz is a problem in C APIs for buffer overruns, so in C APIs I prefer passing the string length explicitly.  

And NUL-termination makes it `resulting_buffer_size` value confusing to document.  You must always keep remembering whether the reported size is with or without the final byte. As an author, I could write docs that explain, but as a user I will never be quite sure which one it was.

And since the point of the C API is to wrap the resulting buffer and make a string view of it, the terminating character isn't really useful.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Done.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Done.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Added logging.  We won't be swallowing the text message, though in the cases I've seen, the message isn't really very useful.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Done.

kythe/cxx/common/kzip_writer_c_api.h
filmil: Done.

kythe/cxx/common/kzip_writer_c_api_test.cc
filmil: It should be, but it won't work as is.

It seems that the [default gtest behavior][dtb] is not modified in kythe, so we'll always get `/tmp` as the temp dir. In bazel we want to get the sharded temp dir from`TEST_TMPDIR`; but this doesn't seem to be the default on Linux.

[dtb]: https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L6561

kythe/cxx/common/kzip_writer.h
```diff
@@ -39,6 +39,12 @@ class KzipWriter : public IndexWriterInterface {
   /// \param encoding Encoding to use for compilation units.
   static StatusOr<IndexWriter> Create(
       absl::string_view path, KzipEncoding encoding = DefaultEncoding());
+  /// \brief Constructs a Kzip IndexWriter which will create and write to
+  /// \param path Path to the file to create. Must not currently exist.
+  /// \param encoding Encoding to use for compilation units.
+  /// Same as `Create` except it returns a pointer.
+  static StatusOr<std::unique_ptr<IndexWriter>> CreateUnique(
+      absl::string_view path, KzipEncoding encoding = DefaultEncoding());
```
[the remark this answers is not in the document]
filmil: Thanks for the suggestion. I added a new function because I got a bit lost in the excavation of `IndexWriter` into `unique_ptr`.  Removed and ironed things up.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Done.

filmil:
:/ When did this start failing?

```
bazel test -c opt //...
```

fails on `buildbot/bazel-release`.

But this works:

```
bazel test -c opt //... --copt=-fPIC
```

filmil:
PTAL, when able.

kythe/cxx/common/kzip_writer_c_api.cc
shahms: That's fair, I would just make sure to explicit document that the `digest_buffer` is *not* null terminated in that case.

kythe/cxx/common/kzip_writer_c_api.h
```diff
@@ -0,0 +1,98 @@
...
+                             const size_t content_length, char* digest_buffer,
+                             size_t buffer_length,
+                             size_t* resulting_digest_size);
+
+/// \brief Writes a compilation unit into the kzip, returning a digest.
+//
+/// The caller must provide a buffer to put the digest into, and a buffer size.
+/// Nonzero status is returned in case of an error while writing.
+int32_t KzipWriter_WriteUnit(struct KzipWriter* writer, const char* proto,
+                             size_t proto_length, char* digest_buffer,
```
shahms, line 90: Document the lack of null-termination here (and in the other output buffers)

filmil: Done.

kythe/cxx/common/kzip_writer_c_api.cc
filmil: Done.

filmil:
@shahms Could you also merge the pull request once you're happy with it? I am not permitted to. Thanks!

PR #4609, opened 2020-07-28
fix(rust): Update author for Rust example package

Arm1stice:
This PR fixes the author field in the Rust example package to be "The Kythe Authors"


Arm1stice:
Made obsolete by #4611

PR #4611, opened 2020-07-28
feat(rust): Unify Kythe Rust dependencies

Arm1stice:
This PR places all of the Kythe Rust dependencies in one directory: `//kythe/rust/cargo`. Previously, each component had its own directory that contained the component's dependencies. However, this quickly became a messy system, leading to file bloat and dependency duplication. This folder can only hold one version of each dependency, though. If a situation arises where a Kythe crate needs a different version of a dependency than what is in this directory, that version can be placed in the crate's directory instead.


filmil:
> If a situation arises where a Kythe crate needs a different version of a dependency than what is in this directory, that version can be placed in the crate's directory instead.

Thanks for fixing this!  And if a situation like this arises, I think the best course of action is to ensure a single version of each dep is added.

PR #4612, opened 2020-07-29
feat(rust): Add Bazel rule for running Rust indexer tests

Arm1stice:
This PR creates a Bazel rule for running tests on the Rust indexer using the Kythe verifier. It is based on the existing `go_indexer_test` rule, and utilizes the `go_verifier_test` macro. This PR also includes a small Rust file to sanity-test the rule until the indexer starts emitting data.


tools/build_rules/verifier_test/rust_indexer_test.bzl
```diff
@@ -0,0 +1,226 @@
...
+    extract_script = " ".join(
+        ["export LD_LIBRARY_PATH=external/rust_linux_x86_64/lib:external/rust_darwin_x86_64/lib;"] +
+        ["export KYTHE_CORPUS=test_corpus;"] +
+        [ctx.executable._extractor.path] +
+        ["--extra_action=%s" % extra_action_file.path] +
+        ["--output=%s" % output.path]
+    )
+    ctx.actions.run_shell(
+        mnemonic = "RustExtract",
+        command = extract_script,
```
shahms, line 67: Prefer `ctx.actions.run` and specifying the env directly.

Arm1stice:
@shahms I changed the rules to pull the information directly from the toolchain, but the build is still failing on Buildbot. The error appears to be associated with CPU architecture, but I'm not sure what's going on.

shahms:
> @shahms I changed the rules to pull the information directly from the toolchain, but the build is still failing on Buildbot. The error appears to be associated with CPU architecture, but I'm not sure what's going on.

One thing to try is not generating a script file.  While that should work, you're not setting a shebang which may confuse things.  Rather than building up and writing out the script just use:
```
args = ctx.actions.args()
args.add(...)
...
ctx.actions.run(
  executable = xa_maker,
  arguments = [args],
  outputs = [extra_action_file],
  tools = [xa_maker],
)
```

Arm1stice:
Gotcha! Changed to run the executable directly when possible. I also switched back to using `run_shell` for the final part that runs the indexer since it runs `set -e; set -o pipefail`. I copied that from the go verifier test macro. If you think it isn't necessary, I can remove that part and convert it to a regular `ctx.actions.run`.

tools/build_rules/verifier_test/rust_indexer_test.bzl
```diff
@@ -0,0 +1,230 @@
...
+        "--output=%s" % extra_action_file.path,
+        "--owner=%s" % ctx.label.name,
+        "--crate_name=%s" % ctx.attr.crate_name,
+        "--sysroot=%s" % paths.dirname(rust_lib[0].path),
+        "--linker=%s" % linker_path
+    ])
+    ctx.actions.run(
+        executable = xa_maker,
+        arguments = [
+        "--src_files=%s" % ",".join([f.path for f in ctx.files.srcs]),
```
shahms, line 58: `arguments = [xa_maker_args]` (no need to repeat these and only use one)

PR #4617, opened 2020-08-01
feat(rust): Support emitting function definitions

Arm1stice:
This PR completes the foundation of the Rust indexer and adds support for emitting package nodes, file nodes, and function definitions (with anchor and documentation nodes)


Arm1stice:
All relevant verifier tests have been added for this PR. A file node verifier test will be added in the future because it requires generating childof edges between files and crates (and possibly the modules they are inside of)

Arm1stice:
@filmil could you take a look at this PR? The files in `kythe/rust/indexer/src/indexer` are the relevant ones. The tests are in `kythe/rust/indexer/src/indexer/testdata` and are run through the Kythe verifier.

filmil:
Looking.

kythe/rust/cargo/BUILD
```diff
@@ -43,6 +43,11 @@ alias(
     actual = "@raze__quick_error__1_2_3//:quick_error",
     tags = ["cargo-raze"],
 )
+alias(
```
filmil, line 46: Note, I will skip checking BUILD file changes and what seem to be automated edits.

I really recommend making cargo raze changes separate pull requests.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -0,0 +1,270 @@
+// Copyright 2020 The Kythe Authors. All rights reserved.
```
filmil, line 1: Unit tests, here and elsewhere?

kythe/rust/cargo/BUILD
Arm1stice: I agree, I'll do that in the future

kythe/rust/indexer/src/indexer/analyzers.rs
Arm1stice: I'll definitely add some integration tests for `entries.rs`, but `analyzer.rs` has coverage with the verifier tests in `testdata/` (the `rust_verifier_test` targets in the BUILD file)

filmil: I'm talking about unit tests here, we normally cover each mod with unit tests, so we can be confident that there are no integration mismatches.

Arm1stice: Ah, I see what you are saying. I believe I mixed up a non-inline test with an integration test. I have gone ahead and added a unit test file for the entries module. I will add some extra unit tests later today

Arm1stice:
There seems to be a strange interaction between Bazel and the way that save_analysis files are loaded, so I'm having trouble creating an exact unit test for it. However, if there was an issue, the verifier tests would fail and say that the save_analysis files failed to be loaded, which I think it valid coverage of the functionality.

filmil:
I would warmly recommend spending some time to make the tests work.  Do you
have some details, maybe I can help?

Integration tests are helpful, but it is even more helpful to have unit
tests since it will also help evolving this code in the future.

F

Arm1stice:
For some reason, the `load_analysis` function cannot see (and does not try to load) the save_analysis test file I have in the tests directory. However, running a directory walk in the same directory it shows that the file is there.

filmil:
I'll try to patch your change in and test it out.

The issue may somehow be that in the bazel sandbox used for tests, the save analysis test file is a symlink not the original file verbatim.  Not sure how that would be relevant for the loader but that's one difference that would be consistent with the issues you are observing: being unable to load the file (because for some reason a symlink is not good enough), but being able to see it in the directory (because the symlink is, in fact, there).

Arm1stice:
Ah you're correct! `rls_analysis` doesn't seem to handle symlinks properly. I got it to work in a test environment by copying to a temporary directory. I'll go ahead and create the actual test now.

Arm1stice:
The `load_analysis` test is written and passes so we have full test coverage now

kythe/rust/indexer/tests/indexer_save_analysis.rs
```diff
@@ -0,0 +1,40 @@
...
+
+extern crate tempdir;
+use tempdir::TempDir;
+
+use std::fs;
+use std::path::PathBuf;
+
+#[test]
+fn load_analysis_works_properly() {
+    let temp_dir = TempDir::new("analysis_test").expect("Couldn't create temporary directory");
```
filmil, line 29: I think the TempDir library should be avoided when building with bazel, as it may interact poorly with restricted build environments.  Instead use the test-specific TMP_TESTDIR, which bazel will give to each test that it runs.

filmil: As an added benefit, you do not need to close the `temp_dir` once done with it, and if the test aborts, you don't leave temporary artifacts on the build machine.

kythe/rust/indexer/tests/indexer_entries.rs
```diff
@@ -0,0 +1,103 @@
...
+// limitations under the License.
+
+use crate::array_writer::ArrayWriter;
+
+extern crate kythe_rust_indexer;
+use kythe_rust_indexer::indexer::entries::EntryEmitter;
+use kythe_rust_indexer::error::KytheError;
+use storage_rust_proto::*;
+
+#[test]
```
filmil, line 22: It would be nice to leave some documentation on what these tests are supposed to do.  In case someone else needs to resume this work, it will help them orient and understand what is expected of them to do.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -0,0 +1,270 @@
...
+    ) -> Self {
+        // Create a HashMap between the file path and the VName which we can retrieve
+        // later to emit nodes
+        let mut file_vnames = HashMap::new();
+        for required_input in unit.get_required_input() {
+            let analysis_vname = required_input.get_v_name();
+            // The required input has a VName of type analysis_rust_proto::analysis::VName.
+            // We need storage_rust_proto::storage::VName to create an Entry. Therefore, we
+            // must use unsafe rust to transmute the VName.
+            let storage_vname: VName = unsafe { ::std::mem::transmute(analysis_vname.clone()) };
```
zrlk, line 64: Please pull out the VName message conversions to separate functions and if possible get rid of uses of `unsafe` (e.g., by constructing VName instances by manually copying the five fields).

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        // Emit nodes for all fact/value pairs
+        for (fact_name, fact_value) in &facts {
+            self.emitter.emit_node(def_vname, fact_name, fact_value.clone())?;
+        }
+
+        let mut anchor_vname = def_vname.clone();
+        anchor_vname.set_signature(format!("{}_anchor", def_vname.get_signature()));
+        // If the definition is a Mod, place the anchor at the first byte only,
```
zrlk, line 225: IIRC there was some business about distinguishing between the module implicitly defined by a file (for which you should anchor to the first byte of the file) and modules defined inside the file. Is this no longer necessary here, or is it being left to be implemented later? In the latter case, please leave a TODO.

shahms: I thought we'd settled on a zero-width span at the start of the file rather than a 1-byte span.

Arm1stice: Left a TODO since this will be handled in the next PR

shahms: Please also update the span in light of #4625 (properly documenting the desired semantics).

PR #4623, opened 2020-08-03
feat(textproto): prototype support for indexer plugins

justbuchanan:
~this isn't ready for a full review yet, but any high-level thoughts on the approach, etc would be appreciated~

* plugins are linked into the indexer binary ~and referenced by name~
* ~during extraction, an optional list of plugins is saved to the compilation unit's arguments under the `--kythe_textproto_plugins` flag.~
* ~at indexing time, if any plugins are specified in the compilation unit, they are initialized at the start~
* after a textproto is parsed, a list of enabled plugins is determined based on the top-level proto message name. plugins are constructed once per textproto file.
* each time a string field value is seen in the input textproto, the plugin's `AnalyzeStringField()` method is called
  * the plugin is given the text span(s) of the string value. it can decide to do nothing or emit 1+ anchors covering subsets of the span(s)
* a subset of the `TextprotoAnalyzer` class is exposed through the `PluginApi` interface, which the plugin can use to add kythe entries and get info about the textproto

Caveats:
This will only support "simple" plugins that don't need to read other textprotos as dependencies. The plugin only has access to the current textproto and any .protos it depends on. The plugin is free to link in whatever data it wants at build time, but can not dynamically load other deps at index time (since they won't be in the compilation unit). This plugin support should work for several proposed/existing textproto-based features/languages in google3, but not all.

TODO:
- [x] use a zero copy reader implementation backed by a string_view for efficiency. I'm currently using one backed by a string, which requires much more string copying than necessary.
- [x] pass the parsed textproto to the plugin constructor
- [x] refactor to share more code between `AnalyzeStringValue()` and `AnalyzeEnumValue()`. - left a TODO
- [x] consider removing `--kythe_textproto_plugins` flag in favor of determining plugins based on top-level proto message. - done.


shahms:
> * during extraction, an optional list of plugins is saved to the compilation unit's arguments under the `--kythe_textproto_plugins` flag.

Do we need to do this or is the top-level protobuf message type sufficient?

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+namespace lang_textproto {
+
+// The plugin's interface to the indexer.
+class PluginApi {
+ public:
+  virtual proto::VName CreateAndAddAnchorNode(const proto::VName& file,
+                                              int begin, int end) = 0;
+
+  virtual proto::VName CreateAndAddAnchorNode(const proto::VName& file_vname,
+                                              re2::StringPiece sp) = 0;
```
shahms, line 41: minor: prefer `absl::string_view` in public APIs

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+  virtual void EmitDiagnostic(const proto::VName& file_vname,
+                              absl::string_view signature,
+                              absl::string_view msg) = 0;
+
+  virtual absl::optional<proto::VName> VNameForRelPath(
+      absl::string_view simplified_path) const = 0;
+
+  // Create a VName for a proto descriptor.
+  template <typename SomeDescriptor>
+  StatusOr<proto::VName> VNameForDescriptor(const SomeDescriptor* descriptor) {
```
shahms, line 54: This should probably be a namespace scoped function rather than a member function.

justbuchanan:
> > * during extraction, an optional list of plugins is saved to the compilation unit's arguments under the `--kythe_textproto_plugins` flag.
> 
> Do we need to do this or is the top-level protobuf message type sufficient?

That's a good point - it would probably be sufficient to just go based on top-level protobuf message and avoid propagating a plugin flag through extraction->CU->indexer.

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+#include "kythe/cxx/common/status_or.h"
+#include "kythe/cxx/indexer/proto/vname_util.h"
+#include "kythe/proto/analysis.pb.h"
+#include "re2/re2.h"
+
+namespace kythe {
+namespace lang_textproto {
+
+// The plugin's interface to the indexer.
+class PluginApi {
```
shahms, line 35: minor: This class should have a virtual destructor and the copy ctor/assignment operator should either be protected or (public and deleted).

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+        });
+    return vname_lookup_status.ok() ? StatusOr<proto::VName>(vname)
+                                    : vname_lookup_status;
+  }
+};
+
+// All plugins must subclass this class.
+class Plugin {
+ public:
+  virtual ~Plugin() {}
```
shahms, line 74: nit: prefer `= default`.

 (See above about ctor/assignment as well)

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+  // Main entrypoint for plugins.
+  virtual absl::Status AnalyzeStringField(
+      PluginApi* api, const proto::VName& file_vname,
+      const google::protobuf::FieldDescriptor& field,
+      re2::StringPiece input) = 0;
```
shahms, line 79: absl::string_view

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+};
+
+// Callback function to instantiate a plugin by name. Returns nil if the name
+// doesn't refer to a valid plugin.
+typedef std::function<std::unique_ptr<Plugin>(absl::string_view name)>
```
shahms, line 84: nit: prefer `using` typealiases over `typedef`

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -440,6 +440,70 @@ absl::Status TextprotoAnalyzer::AnalyzeEnumValue(const proto::VName& file_vname,
...
+
+    // Use tokenizer to get range of entire string. The result contains the
+    // quotation marks.
+    std::string token_text;
+    {
+      // Note: doing input.ToString() for each field is *very* inefficient.
+      // TODO: use a ZeroCopyReader backed by a string_view, not a string, which
+      // requires copying.
+      std::istringstream iss(input.ToString());
+      google::protobuf::io::IstreamInputStream iis(&iss);
```
shahms, line 476: `google::protobuf::io::ArrayInputStream iis(input.data(), input.size());` (will take care of the TODO)

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -631,6 +701,21 @@ absl::optional<std::string> ParseProtoMessageArg(
   return absl::nullopt;
 }
 
+absl::optional<std::vector<std::string>> ParsePluginArg(
+    std::vector<std::string>* args) {
+  for (size_t i = 0; i < args->size(); i++) {
```
shahms, line 706: I would use iterators rather than indices as this will likely make the resulting code easier to read:
```
for (auto iter = args->begin(); iter != args->end(); iter++) {
  if (*iter == "--kythe_textproto_plugins") {
    if (iter + 1 < args->end()) {
      std::string value = *(iter + 1);
      args->erase(iter, iter + 2);
      return absl::StrSplit(value, ",");
    }
  }
}
```

kythe/cxx/indexer/textproto/analyzer.cc
justbuchanan: awesome, thanks!

kythe/cxx/indexer/textproto/plugin.h
justbuchanan: I would generally agree, but this relies on the `VNameForRelPath()` member function, which uses the file substitution cache for mapping paths to proto files.

kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -0,0 +1,90 @@
...
+
+  // Create a VName for a proto descriptor.
+  template <typename SomeDescriptor>
+  StatusOr<proto::VName> VNameForDescriptor(const SomeDescriptor* descriptor) {
+    absl::Status vname_lookup_status = absl::OkStatus();
+    proto::VName vname = ::kythe::lang_proto::VNameForDescriptor(
+        descriptor, [this, &vname_lookup_status](const std::string& path) {
+          auto v = VNameForRelPath(path);
+          if (!v.has_value()) {
+            vname_lookup_status = absl::UnknownError(
```
shahms, line 64: `NotFoundError`?

kythe/cxx/indexer/textproto/plugin.h
shahms: It still works fine as a non-member function, it just has to accept an instance of `PluginApi`. It could/should live alongside kythe::lang_proto::VNameForDescriptor. The only different between the two is that this one allows for failure, which should arguably be the default.

justbuchanan: Related: https://github.com/kythe/kythe/pull/4627. The *proto* indexer's vname lookup functions return an empty vname if given an invalid path. I could change those  to also return `StatusOr` alongside `kythe::lang_proto::VNameForDescriptor()`.

I think ideally in both cases (proto and textproto), the initial parse would fail if it couldn't find a dependent file and the indexing code wouldn't run. If the indexer isn't able to find a file vname, it indicates a bug in the indexer, rather than invalid inputs. Maybe we could just log a "this is a bug" error and return an empty vname?

shahms: That seems reasonable to me.

justbuchanan: I removed the StatusOr and added a log statement in case something does request a vname for an invalid path. I also removed `VNameForDescriptor()` from the plugin api.

kythe/cxx/indexer/textproto/plugins/example/testdata/plugin_test.pbtxt
```diff
@@ -0,0 +1,14 @@
+# proto-file: example.proto
+# proto-message: kythe_plugin_example.Message
+
+# The 'example' plugin adds a ref from the value to the proto descriptor for the
+# field.
+
+#- @field1 ref Field1
+#- @hello ref Field1
+field1: "hello"
```
shahms, line 9: Include an example where the field uses concatenation and might have comments between parts, e.g.
```
field3: # Comment
  "hello"
  # "world"
  "universe"
```
(where the final value should be "hellouniverse")

kythe/cxx/indexer/textproto/textproto_indexer_main.cc
```diff
@@ -111,11 +112,22 @@ int main(int argc, char* argv[]) {
       absl::GetFlag(FLAGS_flush_after_each_entry));
   KytheGraphRecorder recorder(&kythe_output);
 
+  // Plugin loader callback
+  const kythe::lang_textproto::PluginLoadCallback plugin_loader =
+      [](std::string msg_name, const google::protobuf::Message& proto) {
```
shahms, line 117: `absl::string_view` ?

kythe/cxx/indexer/textproto/analyzer.h
```diff
@@ -55,6 +55,17 @@ absl::Status AnalyzeCompilationUnit(const proto::CompilationUnit& unit,
                                     const std::vector<proto::FileData>& files,
                                     KytheGraphRecorder* recorder);
 
+// Callback function to instantiate plugins for a given proto message type.
+using PluginLoadCallback = std::function<std::vector<std::unique_ptr<Plugin>>(
+    std::string msg_name, const google::protobuf::Message& proto)>;
```
shahms, line 60: `absl::string_view`?

shahms: Also, if the function itself doesn't have to outlive the call here, I might consider using https://github.com/abseil/abseil-cpp/blob/master/absl/functional/function_ref.h instead.  It has better semantics than std::function, but it's pretty minor.

kythe/cxx/indexer/textproto/plugins/example/testdata/plugin_test.pbtxt
justbuchanan: Thanks, this is a good edge case that isn't currently handled.

justbuchanan: This also points out a potential challenge for plugin authors needing to map between the string value ("hellouniverse" here) and the bits of the string in the input. Right now the plugin is passed a string_view which spans from the 'h' in "hello" to the 'e' at the end of "universe". It would be on the plugin author to assemble the parts into the final string value and ignore the comments.

Alternatively, I could change the plugin's `AnalyzeStringValue(absl::string_view)` to something like `AnalyzeStringValue(std::string value, std::vector<absl::string_view> ranges)`. This way the analyzer handles fully parsing/assembling the string and the plugin can see what input ranges it came from.

shahms: Oof, yeah. It's obnoxious that there's no way to get total-bytes-read from the proto Tokenizer, only line and column. On the first line, the `column`/`column_end` members should map pretty closely excpet for tabs (https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.io.tokenizer#ColumnNumber.details).  It's like they were trying to make it obnoxious to map these back to the original text :-(

You could use a block_size of 1 in the ArrayInputStream constructor and use the ByteCount() accessor, but that would likely have pretty bad performance.  I think the common case (by far) will be a single token at (0, 0), but you should be able to scan the intervening range for newlines, then tabs to adjust the character offset.  It's [expletive] annoying, but doable.  I might be able to send a CL adding byte offset to the Token but I'm not super optimistic of that.

Something like:
```
absl::string_view SubStringAfter(absl::string_view input, absl::string_view token) {
  const char* start = token.data() + token.size();
  const char* end = input.data() + input.size();
  return absl::string_view(start, end - start);
}
std::vector<absl::string_view> ReadStringTokens(absl::string_view input) {
  std::vector<absl::string_view> tokens;
  ...
  tokenizer_.set_comment_style(io::Tokenizer::SH_COMMENT_STYLE);
  while (tokenizer.Next() && tokenizer.current().type == Tokenizer::TYPE_STRING) {
    absl::string_view token = tokens.empty() ? input : SubStringAfter(input, tokens.back());
    for (int line = tokenizer.current().line; line > 0; line--) {
      token.remove_prefix(token.find('\n') + 1);
    }
    // Have to ignore the token column and look for the delimiter directly :-(
    tokens.push_back(token.substr(token.find_first_of("\"'"), tokenizer.current().text.size()));
  }
  return tokens;
}
```

justbuchanan: Ugh I didn't realize it'd be so hard to get byte offsets when using the tokenizer :/. Thanks for the ideas, I'll try implementing something like your code above. Ideally I'd like to find a way to use [UTF8LineIndex](https://github.com/kythe/kythe/blob/master/kythe/cxx/common/utf8_line_index.h) here, but as-is it'd be inefficient. The span we give to the tokenizer goes from the current offset to the end of the file and we wouldn't want to re-create the line index for that whole span every time. So I'd still need to find the byte offset of the end of the last piece of string field value.

justbuchanan: I realized that the textproto indexer (including the main branch version) wasn't handling lines with tabs correctly. I broke that out into a separate smaller PR: https://github.com/kythe/kythe/pull/4656.

kythe/cxx/indexer/textproto/analyzer.h
justbuchanan: Done.

kythe/cxx/indexer/textproto/plugins/example/testdata/plugin_test.pbtxt
justbuchanan: Ok, I think I've got this working now (I merged #4656 into this one).

justbuchanan:
Thanks for the review!

PR #4624, opened 2020-08-03
feat(fuchsia): Fuchsia-specific extractor

### LINKED ISSUE #4606: Compilation extractor for Rust on Fuchsia OS

filmil:
This commit implements a compilation extractor for Fuchsia OS's Rust
source code.  Please refer to the `README.md` file for a run-down
of the internal workings of the extractor.

While the extractor is able to extract compilation into `kzip` files
as is, I expect that there may be touch-ups needed if we discover
integration issues.

Closes #4606


filmil:
@Arm1stice @shahms PTAL.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,160 @@
 # Fuchsia Rust compilation extractor
```
salguarnieri, line 1: Can we be more precise with our wording here and elsewhere in this PR? This sounds like build system support for Fuchsia rather than an extractor that takes in a rust compilation invocation and outputs kzip data.

Is there a generic binary or library that does the extraction that is called/used by the Fuchsia build system support?

filmil: > Can we be more precise with our wording here and elsewhere in this PR? This sounds like build system support for Fuchsia rather than an extractor that takes in a rust compilation invocation and outputs kzip data.

Not sure I understand the request here.  This is, indeed, a Fuchsia-specific compilation extractor for Rust. What would you rather see written here?

> Is there a generic binary or library that does the extraction that is called/used by the Fuchsia build system support?

I'm not sure what you are asking here.  I'll try to explain some background so maybe we can reformulate the question so it's easier for me to answer.

* The binary that extracts compilation is `rustc`, but invoked with a specific flag `-Zsave analysis`.  This program loads the output of `-Zsave-analysis` and converts into kzip files.
* However, since Fuchsia has a bespoke Rust toolchain, this specific program can not be applied to a piece of software other than Fuchsia. In that respect it is specifically support for Fuchsia's build system.
  * This, in turn, means that extraction can not be done "in parallel" to a regular build the way bazel extractors do it. Sigh.

Not sure if this helps.

salguarnieri: We talked about this in team chat and it sounds like rust makes the separation that we have in other supported languages difficult. For other languages we (try to) separate the extractor which is relatively generic and converts compiler invocations to kzip data from the support needed to get the build system to invoke the extractor. For example, there is build system support for gradle, bazel, cmake, etc and there are extractors for java, c++, go, etc. In some cases a light wrapper around the extractor is needed to support some build systems.

I was hoping that we could separate support for Fuchsia's build system from support for Rust in Fuchsia's build system so the build system support could be used for other languages as well and the core rust extractor could be used for other build systems.

filmil: > We talked about this in team chat and it sounds like rust makes the separation that we have in other supported languages difficult. 

I see. I too am aware of the extractor/indexer separation of concerns that Kythe uses.  Ideally, Rust compilations should be separable this way.  However, *at the moment* this is expensive to do in Rust.  IIUC, work is underway in the Rust community to turn some of Rust internals into libraries, such that it would be possible to make a purely indexing pass, at which point we could have a generic Rust indexer.  However, today, if you wanted to replay a Rust compilation, you'd need pretty much to reimplement the `cargo` build system.

But until the said libraries are available, I think this is a pragmatic way forward. I think once they are, we'd be able to evolve the extractor and indexer to use them. (And even Kythe's current bazel extractor for Rust relies on the same mechanism, so I think we're aligned there.)

> I was hoping that we could separate support for Fuchsia's build system from support for Rust in Fuchsia's build system so the build system support could be used for other languages as well and the core rust extractor could be used for other build systems.

Similarly here, GN would need to support capturing compilations in sufficient detail to make this possible, which it currently does not. In Fuchsia, there's also ongoing work to make this happen, at which point we may be able to offer a generic tool.

So, even though the current approach could be thought of as suboptimal, it will give xref support for a few hundred Fuchsia engineers which should be a significant productivity boost.

salguarnieri: This is actually different than the extractor/indexer separation. This is build system/extractor separation. That is part of why I wanted to be more precise in the language, so we clearly document what is implemented. As you say this is a fine trade off for benefit vs complexity.

filmil: Noted. I updated `README.md` to explain a bit, hopefully to your satisfaction. I'll resolve this discussion thread now, and if you want me to add specific things, please start a review and point out where you want things changed and how. Thanks!

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+
+Fuchsia code is compiled in a separate "output" directory.  However, this
+directory is created underneath `$FUCHSIA_DIR`, and each build configuration,
+for a "product-dot-board" combination such as `terminal.x64`, gets a separate
+build output directory.  The build configuration `terminal.x64` gets an output
+directory in `$FUCHSIA_DIR/out/terminal.x64`, though this last path component
+can be given a different name by the developer.  Let us call this directory
+`$FUCHSIA_OUT` for short.
+
+Multiple code corpora exist in `$FUCHSIA_DIR` and `$FUCHSIA_OUT`.  There is C
```
salguarnieri, line 50: I think the c, c++, rust, go, etc. code would all be in the same fuchsia corpus. For example, the java, c++, and proto code in chromium all belong to the same chromium corpus.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+and C++ code, Rust code, go code, test fixtures, dart and flutter code, and
+dedicated IDL generated code for languages called FIDL and Banjo.
+
+When filling out compilation unit data, we need to ensure that when we refer to
+files, their names are expressed relative to some of those directories, and
+then to correct ones.  Furthermore, sometimes relative paths are needed with
```
salguarnieri, line 56: When do files need to be relative to something other than the corpus (repository) root?

I think this has caused trouble in the past with chromium and the UI didn't know how to navigate to files that were not rooted at the repository root. If multiple "roots" are going to be used, we should make sure we handle it in the indexer in a way that works with the UI.

https://github.com/kythe/kythe/blob/master/kythe/proto/analysis.proto#L236 for reference the FileInfo message that stores these paths.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+
+;`--inputdir INPUT_DIR`: The directory that contains the `save-analysis` files.
+The compilation indexing is fully driven by these files, so the program will
+need to know where to find them.  In Fuchsia, they are all bunched up in a
+single flat directory, which is normally at `$FUCHSIA_OUT/save-analysis-temp`
+(with limited options to rename the directory).  Required.
+
+;`--inputfiles FILES`: A comma-separated list of specific `save-analysis` files
+to read, in case the `--inputdir` flag is too broad.  If this is specified together
+with `--inputdir`, the
```
salguarnieri, line 80: Unfinished comment

kythe/rust/cargo/Cargo.toml
```diff
@@ -29,11 +29,15 @@ clap = "2.33.1"
 colored = "2.0.0"
 protobuf = "=2.8.2"
 quick-error = "1.2.3"
+regex = "1.3.9"
```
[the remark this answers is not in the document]
Arm1stice: We might be able to put it in third_party and put a BUILD file in that directory. It's not on crates.io so it wouldn't work with cargo-raze

kythe/rust/fuchsia_extractor/src/bin/main.rs
```diff
@@ -0,0 +1,862 @@
...
+    /// The comma-separated revisions string, used to fill out the indexed compilation
+    /// protocol buffer field called `index`.
+    revisions: Vec<String>,
+}
+
+fn main() -> Result<()> {
+    // We'd rather use a crate that parses the program arguments directly into the `Options`
+    // struct.  However, for some weird reason our dependencies can not be so aligned to get any
+    // such crate to compile, even including a newer version of `clap`, which supports this kind of
+    // parsing.  So we parse like this.
```
Arm1stice, line 491: Which crate were you trying to use? I can try and get it to work with cargo-raze. A lot of times it just involves setting `gen_buildrs` in Cargo.toml for the dependency

kythe/rust/cargo/Cargo.toml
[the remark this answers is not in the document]
shahms: It looks like it also hasn't been updated in 7 years so may not be worth the effort.  Historically, regex libraries have offered a wide variety of syntax and bugs so standardizing on a widely available engine with specific performance, memory and syntax as been beneficial as you don't have to guess what language a particular tool is written in to know what RE syntax to use.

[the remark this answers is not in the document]
shahms: Actually, scratch that.  At least in theory the `regex` crate was "inspired by" RE2 and is likely to be adequately compatible.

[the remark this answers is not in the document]
filmil: Well, if you are worried about the use of this package, I could make do without it. I am nowhere near the problem pathological cases where the distinction between `re2` and other implementations matter.

kythe/rust/fuchsia_extractor/README.md
filmil: Well, GN kind of makes this necessary.  

If `$FUCHSIA_DIR` is the top level directory of the source repository, GN will create `$FUCHSIA_DIR/out/some-dir` for build output, and then inside it there will be multiple directories such as `obj` for objet code, but `fidlgen` for FIDL generated code, `banjoing` for banjo files etc.  And, there's still `../../` for the original source files.  All the while I think `FileInfo.path` spec requires the paths to not have `../../`, but instead be relative to a corpus root.

This to me implies that to satisfy the requirements of `CompilationUnit`one must create at least two different repo roots: one for the original source code (repo root is `$FUCHSIA_DIR/out/some-dir/../../`), one for the generated code (`$FUCHSIA_DIR/out/some-dir/fidlgen`).

kythe/rust/fuchsia_extractor/README.md
filmil: Yeah I think the correct term here should be "corpus root", not "corpus"

kythe/rust/fuchsia_extractor/src/bin/main.rs
filmil: I tried to use `structopt`, but the issue I had was that in pulls in `syn`; `syn` would not compile because it is trying to use `proc-macro`, with a combination of features that does not work.  I gave up on figuring that dependency hell out, and just used `clap` as it was already used elsewhere in the Rust code here.

kythe/rust/fuchsia_extractor/README.md
filmil: In addition, `some-dir` is not a constant name, so we can not make `$FUCHSIA_DIR` the corpus root and treat all files as rooted there.   At the same time, there is a requirement that some of the paths are expressed relative to the directory where compilation ran.  So as you see the corpus namespace is fragmented, and using multiple corpus roots seemed like a way to handle that.

filmil: To go back to the link you gave, in GN the root of the build is *not* the root of the repo.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+  --output="${_output_dir}" \
+  --corpus="fuchsia" \
+  --language="rust" \
+  "${@}"
+```
+
+The processing is sequential, and therefore fairly slow.  Expect on the order
+of 10 seconds to generate a single `kzip` for a single compilation unit. This
+sequential processing is not a fundamental limitation, but will be addressed
+only if needed. The infrastructure may.
```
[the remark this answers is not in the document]
filmil: Done.

kythe/rust/fuchsia_extractor/README.md
filmil: Done.

kythe/rust/fuchsia_extractor/README.md
filmil: Done, thanks for spotting this.

kythe/rust/fuchsia_extractor/README.md
filmil: Clarified in the text, so marking this as resolved as well.

kythe/rust/fuchsia_extractor/src/bin/main.rs
filmil: I'll resolve this since any improvements we may want to make here can be made separately.

kythe/rust/fuchsia_extractor/README.md
shahms: FileInfo.path should represent the path-as-seen-by-the-compiler (generally relative to the working_dir in the compilation unit).  The *VName* path needs to be relative to the corpus root.  For generated files, the VName path should be conceptually relative to the corpus root, with the `root` entry set to something like `out/some-dir/fidlgen`.  For example, for Bazel builds generated files tend to live in `bazel-out/<platform>/<kind>/<path>` and we typically strip out the platform, leaving a VName root of `bazel-out/<kind>`.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+`$FUCHSIA_OUT` for short.
+
+Multiple code corpora exist in `$FUCHSIA_DIR` and `$FUCHSIA_OUT`.  There is C
+and C++ code, Rust code, go code, test fixtures, dart and flutter code, and
+dedicated IDL generated code for languages called FIDL and Banjo.
+
+When filling out compilation unit data, we need to ensure that when we refer to
+files, their names are expressed relative to some of those directories, and
+then to correct ones.  Furthermore, sometimes relative paths are needed with
+respect to the corpus root, and sometimes with respect to the build directory.
```
shahms, line 61: As mentioned above, this is only true in the VNames, FileInfo.path should be relative to working_directory as seen from the invoked command's perspective.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+For example, generated code lives in `$FUCHSIA_OUT/gen`.  Object code, such as
+`.o` files, live in `$FUCHSIA_OUT/obj`, and there are about a dozen of such
+different corpus roots.
```
shahms, line 64: This is a confusing turn of phrase in Kythe terminology as files have a unique identifier consisting of 3 parts: corpus, root and path.  Corpus is a unique identifier for a particular body of code, typically 1:1 with a repository. For in-tree files, the root is empty.  For generated files the root is something which logically corresponds to the generated file root (could just be `gen` or `out` or similar). Saying "corpus root" typically indicates the root of the repository.  This part is important as clients need a way of mapping from `(corpus, root, path)` to repository file.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,168 @@
...
+
+Even further, though, even as the relationships between these directories are fixed,
+infrastructure builds may take liberties with the arrangement of these directories.
+This means that we need to keep options open to retarget these directories
+as the developer sees fit for their use.  Hence the many flags.
```
shahms, line 69: This is generally accomplished via `vnames.json` files.

kythe/rust/fuchsia_extractor/README.md
filmil: Pardon me, fixed the text.

kythe/rust/fuchsia_extractor/README.md
filmil: I used corpus root, as in "a unique `(corpus, root)`" pair.   I suppose that is too specific since IIUC, what you describe tacitly assumes that the `corpus` component is the same for the entire repo.

Also, it was not obvious to me from the specification of the field `VName.root` that the root of source is implied to be "".  I set it to "src", not knowing any better.

kythe/rust/fuchsia_extractor/README.md
filmil: Ack.

kythe/rust/fuchsia_extractor/README.md
filmil: Fixed now.  Using `""` for the root for in-tree code, and others as before.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,173 @@
...
+directory in `$FUCHSIA_DIR/out/terminal.x64`, though this last path component
+can be given a different name by the developer.  Let us call this directory
+`$FUCHSIA_OUT` for short.
+
+Multiple top-level directories exist in `$FUCHSIA_DIR` and `$FUCHSIA_OUT`.  The
+"source" corpus is at `../../` relative to the build directory.  FIDL generated
+code is at `./fidling`, Banjo generated code is at `./banjoing` and so on.
+Multiple roots are used to satisfy the spec requirement for `VName`, which
+requires each file path to be relative to its root, *and* not use any "parent"
+references.
```
shahms, line 55: This is all fairly documented at https://www.kythe.io/docs/kythe-storage.html#_a_id_termvname_a_vector_name_strong_vname_strong

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,173 @@
...
+
+When filling out compilation unit data, we need to ensure that when we refer to
+files, their names are expressed relative to some of those directories, and
+then to correct ones.  Furthermore, sometimes relative paths are needed with
+respect to the root directory of the repository tree, and sometimes with
+respect to the build directory.  For example, generated code lives in
```
shahms, line 61: The only paths which should be relative to the repository are those in the VName.  Every other path should be relative to the working_directory (or absolute if it doesn't live under the build dir), which should be the build directory.   Note that the build directory doesn't have to equal the `working_directory` but is often easier to reason about when that is the case.

kythe/rust/fuchsia_extractor/README.md
```diff
@@ -1,9 +1,173 @@
...
+libraries are available, we can update the indexer and the extractor to match.
 
 [fx]: https://www.fuchsia.dev
+[gn]: https://gn.googlesource.com/gn/
 
-Fuchsia build system is a bit specific, and is custom built on top of [GN][gn].
+Technically, this extractor produces `kzip` files as any other extractor does,
+and the `kzip` file specification is well known.  However, in practice, the
+kzip specification is open for interpretation (for example, what if you need
+to inject a specific extra file to a compilation unit?).  That in turn can
```
shahms, line 22: By putting it in the compilation unit's `required_input` list.

kythe/rust/fuchsia_extractor/README.md
filmil: I understand. 

If you have actionable feedback, please state explicitly what you want changed in this text.

kythe/rust/fuchsia_extractor/README.md
filmil: I have read that page as I was researching this mapping. 

My finding is that the text there is very much open to interpretation as it contains no normative prescriptions. So I had to use my best judgment to interpret what that page means in the context of Rust compilation extractor.

I added the link to this page into the README.md document.

If you need further changes, please formulate them so that they are actionable.

kythe/rust/fuchsia_extractor/README.md
filmil: Ack.

PR #4625, opened 2020-08-04
fix(schema): document the use of special spans for implicit modules

shahms:


kythe/docs/schema/schema.txt
```diff
@@ -2066,6 +2066,11 @@ package
 
 Brief description::
   A *package* defines a module containing declarations.
+Notes::
+  Languages in which a module is implicitly defined based on the file name
+  should use a zero-width anchor at offset 0 in that file for the
+  `defines/binding` edge and a span covering the entire file for any `defines`
```
schroederc, line 2072: The `defines/binding` edge seems suspect.  There's no binding at all.  Maybe we shouldn't emit that edge?

shahms: That was what we had settled on in the schema meeting a while ago and the current recommendation for TypeScript (but not the current implementation).  I believe this was a compromise between having the edge originate from the file node itself (which arguably is where the binding originates, but is not an anchor) and either the full file (which also is not a binding, presents challenges in contexts where the file length isn't known) or a single-byte at the start of the file (which confuses users and is "ugly").

schroederc: Why do we need both the `defines/binding` and `defines`, though?  We could just use `defines`.  If the file length is unknown, it could be the zero-width span.

shahms: I'm fine with just using the zero-width span with a `defines` edge.  (Looking over the internal bug I expressed the same skepticism around the use of `defines/binding`).  I worry that our current clients may disregard `defines` edges entirely in favor of `defines/binding` however.

shahms: Note: there was also some discussion of normalizing the `childof` edges, but I'm omitting that here as a too invasive change for now:

```
import ==[ref/imports]=> package
@(0,0) ==[defines]=>     package
file   ==[childof]=>     package
```

schroederc: 👍 Thanks for the clarification.

PR #4629, opened 2020-08-04
feat(rust): Add support to the Rust indexer for emitting module definitions and anchors

Arm1stice:
This PR adds support for emitting module definitions and anchors in the Rust indexer. Implicit module definitions have 0-length anchors at the top of the file with `defines/implicit` edges, and explicit module definitions have anchors on the module name with `defines/binding` edges.


Arm1stice:
The save_analysis files contain incorrect `byte_start` and `byte_end` values in the spans if the crate has multiple files, though the provided line and column numbers are correct. This seems to come from the offsets in the compiler source map rather than the file by itself. Because of this, I have to create my own index of byte offsets for each source file. I tried to reduce the amount of memory that gets stored by tracking if a line only contains single-byte characters

PR #4631, opened 2020-08-05
fix(cxx_indexer): avoid an assert check in Clang, silence errors

zrlk:
For some reason, in recent versions of Clang, integer template arguments (when fed into Sema as we do in MarkedSource) reach an assert that fails. To avoid tripping the assert, this PR stops trying to infer information about default template arguments if it encounters an integer template argument.

This PR also disables diagnostics while inferring which arguments are defaults. Previously, the indexer would report spurious diagnostics that were raised during inference.

### DIFF
--- a/kythe/cxx/indexer/cxx/marked_source.cc
+++ b/kythe/cxx/indexer/cxx/marked_source.cc
@@ -25,6 +25,7 @@
 #include "clang/Rewrite/Core/Rewriter.h"
 #include "clang/Sema/Template.h"
 #include "google/protobuf/stubs/common.h"
+#include "kythe/cxx/common/scope_guard.h"
 #include "kythe/cxx/indexer/cxx/clang_range_finder.h"
 #include "kythe/cxx/indexer/cxx/clang_utils.h"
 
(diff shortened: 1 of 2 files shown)


kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -588,9 +588,24 @@ std::string GetDeclName(const clang::LangOptions& lang_options,
   return "";
 }
 
+namespace {
+class : public clang::DiagnosticConsumer {
+  void HandleDiagnostic(clang::DiagnosticsEngine::Level,
+                        const clang::Diagnostic&) override {}
+} ignoring_diagnostic_consumer;
```
shahms, line 595: just use `clang::IgnoringDiagConsumer`

shahms: (And you should just construct a new one each time, rather than using a global.  They aren't entirely stateless but are very cheap to construct).

kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -588,22 +589,16 @@ std::string GetDeclName(const clang::LangOptions& lang_options,
...
     const clang::ClassTemplateSpecializationDecl* decl) {
   // While we try to figure out which template arguments are defaults, silence
   // diagnostics from the typechecker.
+  clang::IgnoringDiagConsumer consumer;
   auto* diags = &cache_->source_manager().getDiagnostics();
   bool owned = diags->ownsClient();
   auto* client = diags->getClient();
-  diags->setClient(&ignoring_diagnostic_consumer, false);
+  diags->setClient(&consumer, false);
   auto guard = MakeScopeGuard([=] { diags->setClient(client, owned); });
```
shahms, line 602: optional, what do you think of:

```
auto* diags = &cache_->source_manager().getDiagnostics();
auto guard = MakeScopeGaurd([diags=diags, client=diags->getClient(), owned=diags->ownsClient()] {
  diags->setClient(client, owned);
});
diags->setClient(&consumer, false);
```

shahms: (Also, need to depend on //kythe/cxx/common:scope_guard)

PR #4638, opened 2020-08-10
feat(rust): Support emitting remaining definition types

Arm1stice:
This PR adds support for the following types of definitions:
- Struct
- Union
- Trait
- Method
- Type
- Local
- Static
- Const
- Closures

Additionally, `tbuiltin` node types are supported with this PR. Verifier tests have been added to test this new functionality.


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -194,6 +286,101 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+                        relation
+                    ))
+                })?;
+                // Add all of the childred to the HashMap
+                for child in implementation.children.iter() {
+                    // The struct being implemented on
+                    let struct_target = relation.from;
+                    // The optional trait being implemented on
+                    let trait_target =
+                        if relation.to.krate != 4294967295 { Some(relation.to) } else { None };
```
zrlk, line 333: Is there a clearer way to represent this? Can you use -1 or ~0?

kythe/rust/indexer/testdata/variable.rs
```diff
@@ -0,0 +1,28 @@
...
+//- Static.node/kind constant
+static _TEST_STATIC: &str = "Kythe";
+
+fn main() {
+    //- @_test_variable defines/binding Local
+    //- Local.node/kind variable
+    //- Local.subkind local
+    let _test_variable = 0;
+
+    // TODO: Identify declaractions and emit an ".complete = imcomplete" fact
```
zrlk, line 17: s/imcomplete/incomplete/

kythe/rust/indexer/src/indexer/analyzers.rs
Arm1stice: This maxint is the from the data we are given in the save_analysis

zrlk: Granted that's the case, but it's a weird constant to run into when reading the code (if you're not familiar with the vagaries of 2's compliment arithmetic). Could you give it a name?

PR #4639, opened 2020-08-10
Replace `stream::iterator` with alternatives in enhanced for loops.

graememorgan:
Stream::iterator is an unsafe way to create an Iterable, and will soon be banned by an ErrorProne check.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
+++ b/kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
@@ -357,6 +357,8 @@ public final class CompilationUnitPathFileManager extends ForwardingStandardJava
               .toAbsolutePath();
       Path systemRoot = Files.createDirectory(temporaryDirectory.resolve("system"));
       try (Stream<Path> stream = Files.walk(sys.resolve("lib"))) {
+        // While this is ugly, it's safer in general than converting to a one-shot Iterable
+        // using ::iterator.
         Iterator<Path> it = stream.iterator();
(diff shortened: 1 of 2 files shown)


kythe/java/com/google/devtools/kythe/platform/java/filemanager/CompilationUnitPathFileManager.java
```diff
@@ -357,7 +357,9 @@ private void setSystemOption(String value) throws IOException {
               .toAbsolutePath();
       Path systemRoot = Files.createDirectory(temporaryDirectory.resolve("system"));
       try (Stream<Path> stream = Files.walk(sys.resolve("lib"))) {
-        for (Path path : (Iterable<Path>) stream::iterator) {
+        Iterator<Path> it = stream.iterator();
+        while (it.hasNext()) {
+          Path path = it.next();
```
shahms, line 362: As this seems objectively worse than the original, it's going to need a comment to prevent well-meaning changes from reverting this in the future.

shahms: Sorry, "objectively worse" is harsher than intended.  Mostly meant to suggest that without a corresponding comment this looks trivially like something which should be tackled with a for loop rather than an explicit iterator and that transformation likely to be re-applied.

graememorgan: Done. I think "objectively worse" is not super unfair--it's very annoying how hard it is to iterate over a stream while throwing a checked exception.

PR #4647, opened 2020-08-12
fix(rust): parallelize the extraction process

filmil:
The Fuchsia compilation extraction is easy to parallelize, so
we do so.  The rayon library farms out the extraction process,
and the extraction workers won't interfere with each other as
they only read inputs, and are guaranteed not to step over
each other.

This should speed up the extraction significantly.

### DIFF
--- a/kythe/rust/cargo/BUILD
+++ b/kythe/rust/cargo/BUILD
@@ -49,6 +49,11 @@ alias(
     tags = ["cargo-raze"],
 )
 alias(
+    name = "rayon",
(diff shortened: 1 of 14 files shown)


filmil:
@shahms PTAL?

shahms:
Ugh, it looks like I messed up the merge in the UI :-(  @filmil would you mind tackling that?

filmil:
No worries, this was going to be a hairy merge because of the changes at https://github.com/kythe/kythe/pull/4653.

PR #4650, opened 2020-08-13
fix(rust): Fixes the kzip generation

filmil:
The old code filled FileInfo.path with a path relative to the corpus
root, but should instead have filled it with a path relative to the
build directory.

Added a regression tests, and added *most* of an integration test
which runs kzip info... for additional verification of adherence
to the spec.

Howeer, that last test is currently not enabled as it is apparently
not allowed to subprocess to a binary.

### DIFF
--- a/kythe/rust/fuchsia_extractor/src/bin/main.rs
+++ b/kythe/rust/fuchsia_extractor/src/bin/main.rs
@@ -897,7 +897,6 @@ mod testing {
     }
 
     #[test]
-    #[ignore]  // Reenable once we can subprocess `kzip`.
     fn kzip_info_spec_test() {
         use std::process::Command;
 
@@ -924,7 +923,7 @@ mod testing {
         let test_runfiles = PathBuf::from(
             std::env::var("TEST_SRCDIR").expect("TEST_SRCDIR is available"));
 
-        let kzip_util_path = test_runfiles.join("io_kythe/kythe/go/platform/tools/kzip");
+        let kzip_util_path = test_runfiles.join("io_kythe/kythe/go/platform/tools/kzip/kzip");
 
         let output = Command::new(kzip_util_path)
             .arg("info")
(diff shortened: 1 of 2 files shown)


filmil:
@Arm1stice FYI

kythe/rust/fuchsia_extractor/src/bin/main.rs
```diff
@@ -875,16 +875,67 @@ mod testing {
...
+        assert_eq!(
+            "../../third_party/rust_crates/vendor/nom/lib.rs",
+            ru_info.get_path(),
+            "Expected to be relative to the build directory"
+        );
         assert_eq!("nom", cu_vname.get_signature());
     }
+
+    #[test]
+    #[ignore]  // Reenable once we can subprocess `kzip`.
```
shahms, line 900: What errors do you see trying to run the binary?

filmil: See here: https://groups.google.com/g/bazel-discuss/c/S-uSkgNcmhg
I suppose this is something about executable bits in the bazel sandbox, but I admit I've never seen this before.  The kzip binary is present in the sandbox, and I can execute it locally.

filmil: Figured it out -- my binary `kzip` was `.../kythe/go/platform/tools/kzip/kzip`, not `.../kythe/go/platform/tools/kzip`.

With this fix, things work again.

shahms: Ah, yeah, the Bazel go rules result in a lot of stuttering.

filmil: As to how I discovered this, I tried to wrap my command line `FOO` into `sh -c $FOO`, at which point bash complained `can not execute a directory` or some such.  Which was very helpful.

filmil:
Thanks! F

PR #4653, opened 2020-08-14
fix(rust): Handles different roots for Fuchsia

filmil:
Adds sorting code that bins all the Fuchsia source
files into separate roots, based on the relative
path of a file with respect to the directory the
root builds in.

There are about a dozen different roots, some of which
are "" (the "default" source, which is the "in-tree" code),
the "gen" (source code generated by build rules), "fidling_gen"
(the source code generated by the FIDL compiler) and so on.

Regression test added to confirm that the paths are set as
expected.

### DIFF
--- a/kythe/rust/cargo/BUILD
+++ b/kythe/rust/cargo/BUILD
@@ -79,6 +79,11 @@ alias(
     tags = ["cargo-raze"],
 )
 alias(
+    name = "serial_test",
+    actual = "@raze__serial_test__0_4_0//:serial_test",
+    tags = ["cargo-raze"],
+)
+alias(
     name = "tempdir",
     actual = "@raze__tempdir__0_3_7//:tempdir",
     tags = ["cargo-raze"],
--- a/kythe/rust/cargo/Cargo.toml
+++ b/kythe/rust/cargo/Cargo.toml
@@ -42,6 +42,7 @@ lazy_static = "1.4.0"
 [dev-dependencies]
 assert_cmd = "1.0.1"
 predicates = "1.0.5"
+serial_test = "0.4.0"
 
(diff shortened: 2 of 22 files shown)


shahms:
I'm not well versed in reading Rust error messages, but it appears as though something causes issues with older versions of Bazel?

filmil:
Hm, I've been seeing this even on a recently updated bazel, and there's something else at play that I don't quite understand at the moment. It seems to have to do with the use of bazel's temporary dirs on the test fixture.  

It seems that, very rarely (at least, on my machine with `--runs_per_test=1000` I observe 1 failure per 1000), the temporary directories that bazel hands the tests down through the variable `TEST_TMPDIR` are created as read-only regardless of whether they actually are supposed to be read only or not.  

I could try using a "regular" temporary directory and see if it helps bring the flakiness down.  This is not ideal, since if a test fails, the `TmpDir` implementation for Rust will remove the underlying directories, and hamper debugging.

filmil:
:/ 
I'm still seeing failures to write to temporary directory, even with the modified temp dir creation strategy.  Looking.

filmil:
@shahms  PTAL.  I think the issue is now resolved by serializing the tests.  I don't have enough data to claim for sure that `tempdir` creation in parallel by different tests somehow stepped on each other.  

But, when I ensured that tests were run serially, the flakiness went away, from 1/1000 to 0/30000.

This made the commit substantially larger as I needed to import the `serial_test` crate, and it pulled it own deps; but those files are autogenerated.

filmil:
FWIW, `tempdir` is likely not the culprit, I tried making a minimal repro with tempdir alone, and could see no issues with those tests.

PR #4655, opened 2020-08-17
fix(cxx): normalize release config, include assert stacktraces

shahms:


shahms:
This is a workaround for https://github.com/abseil/abseil-cpp/issues/769

PR #4656, opened 2020-08-17
fix(textproto indexer): handle tabs correctly

justbuchanan:
* unfortunately the proto tokenizer/parser counts tabs as 8 columns, so
  converting between column offset and byte offset is tricky. before
  this commit, the textproto analyzer failed to correctly index any lines
  of the file that included tabs
* moved proto indexer's ByteOffsetIntoLine() into a library which is
  shared with the textproto indexer
* added a test case that uses tabs


kythe/cxx/indexer/proto/offset_util.cc
```diff
@@ -0,0 +1,47 @@
...
+    LOG(ERROR) << "Error computing byte offset: expected " << column_number
+               << " columns but counted up to " << computed_column
+               << " in line \"" << line_text << "\"";
+    return -1;
+  }
+  return offset;
+}
+
+}  // namespace lang_proto
+}  // namespace kythe
```
shahms, line 47: nit: newline?

kythe/cxx/indexer/proto/offset_util.h
```diff
@@ -0,0 +1,32 @@
...
+#define KYTHE_CXX_INDEXER_PROTO_OFFSET_UTIL_H_
+
+#include "absl/strings/string_view.h"
+
+namespace kythe {
+namespace lang_proto {
+
+// Figures out just how many bytes one needs to go into `line_text` to reach
+// what the proto compiler calls column `column_number`.
+int ByteOffsetIntoLine(int column_number, absl::string_view line_text);
```
shahms, line 27: `ByteOffsetOfTabularColumn`? or something that includes specific mention of expanding tabs. I would also swap the argument order, but that's pretty minor.

justbuchanan: sgtm

PR #4663, opened 2020-08-18
feat(runextractor): Pass flags to the compdb extractor

filmil:
Adds the ability to pass arbitrary additional flags to the
extractor binary ran by `runextractor`.  This allows for example
specifying custom include library paths for `cxx_extractor`, which
allows the extractor access to include library paths for a custom
compiler if used.

Example:

```bash
runextractor compdb -extractor cxx_extractor -- -cxx-isystem=../some/dir
```

### DIFF
--- a/kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
+++ b/kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
@@ -121,7 +121,7 @@ func (c *cmakeCommand) Execute(ctx context.Context, fs *flag.FlagSet, args ...in
 		return c.Fail("Error building repository: %v", err)
 	}
 
-	if err := compdb.ExtractCompilations(ctx, extractor, filepath.Join(buildDir, "compile_commands.json")); err != nil {
+	if err := compdb.ExtractCompilations(ctx, extractor, filepath.Join(buildDir, "compile_commands.json"), nil); err != nil {
 		return c.Fail("Error extracting repository: %v", err)
 	}
 	return subcommands.ExitSuccess
--- a/kythe/go/extractors/config/runextractor/compdb/compdb_test.go
+++ b/kythe/go/extractors/config/runextractor/compdb/compdb_test.go
@@ -64,7 +64,7 @@ func TestExtractCompilationsEndToEnd(t *testing.T) {
 	if err := os.Chdir(filepath.Join(root, testPath, "testdata")); err != nil {
 		t.Fatalf("Unable to change working directory: %v", err)
 	}
-	if err := ExtractCompilations(context.Background(), extractor, "compilation_database.json"); err != nil {
+	if err := ExtractCompilations(context.Background(), extractor, "compilation_database.json", nil); err != nil {
 		t.Fatalf("Error running ExtractCompilations: %v", err)
 	}
 	err = filepath.Walk(os.Getenv("KYTHE_OUTPUT_DIRECTORY"), func(path string, info os.FileInfo, err error) error {
--- a/kythe/go/extractors/config/runextractor/compdb/compdb.go
+++ b/kythe/go/extractors/config/runextractor/compdb/compdb.go
@@ -45,7 +45,7 @@ type compileCommand struct {
 
 // ExtractCompilations runs the specified extractor over each compilation record
 // found in the compile_commands.json file at path.
-func ExtractCompilations(ctx context.Context, extractor, path string) error {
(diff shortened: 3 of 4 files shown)


kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
```diff
@@ -121,7 +121,7 @@ func (c *cmakeCommand) Execute(ctx context.Context, fs *flag.FlagSet, args ...in
 		return c.Fail("Error building repository: %v", err)
 	}
 
-	if err := compdb.ExtractCompilations(ctx, extractor, filepath.Join(buildDir, "compile_commands.json")); err != nil {
+	if err := compdb.ExtractCompilations(ctx, extractor, filepath.Join(buildDir, "compile_commands.json"), []string{}); err != nil {
```
shahms, line 124: just `nil`

kythe/go/extractors/config/runextractor/compdb/compdb_test.go
```diff
@@ -64,7 +64,7 @@ func TestExtractCompilationsEndToEnd(t *testing.T) {
 	if err := os.Chdir(filepath.Join(root, testPath, "testdata")); err != nil {
 		t.Fatalf("Unable to change working directory: %v", err)
 	}
-	if err := ExtractCompilations(context.Background(), extractor, "compilation_database.json"); err != nil {
+	if err := ExtractCompilations(context.Background(), extractor, "compilation_database.json", []string{}); err != nil {
```
shahms, line 67: ditto

kythe/go/extractors/config/runextractor/cmakecmd/cmakecmd.go
filmil: Done.

kythe/go/extractors/config/runextractor/compdb/compdb_test.go
filmil: Done.

kythe/go/extractors/config/runextractor/compdb/compdb_test.go
```diff
@@ -64,7 +64,7 @@ func TestExtractCompilationsEndToEnd(t *testing.T) {
 	if err := os.Chdir(filepath.Join(root, testPath, "testdata")); err != nil {
 		t.Fatalf("Unable to change working directory: %v", err)
 	}
-	if err := ExtractCompilations(context.Background(), extractor, "compilation_database.json"); err != nil {
+	if err := ExtractCompilations(context.Background(), extractor, "compilation_database.json", nil); err != nil {
```
creachadair, line 67: I wasn't paying attention, but I think you should reconsider this interface. Bolting extra arguments into a public interface can be a maintenance headache, especially when the argument type doesn't help the compiler protect against things like swapped arguments. It's not a huge problem in this case, but I offer the following suggestion for future consideration.

When you need to add optional arguments to a public interface, think about defining a separate type that can provide defaults, e.g.,

```go
type ExtractOptions struct {
  ExtraArguments []string // additional arguments to pass to the compiler
}

func ExtractCompilations(ctx context.Context, ext, path string, opts *ExtractOptions) error { … }
```

This not only gives you a better place to put documentation, it makes it easier to customize defaults: E.g., you can define an unexported method on the options type that the implementation uses, so the plumbing still works with a `nil` as default:

```go
func (o *ExtractOptions) extraArguments() []string {
  if o != nil && len(o.ExtraArgs) > 0 {
    return o.ExtraArgs
  }
  return nil
}
```

Relatedly: Even if you didn't want to go that far for this change (not unreasonable), I would have liked to see the doc comments on the exported names updated to reflect this API change. YMMV.

filmil: Thanks for your comment.   Indeed, I wanted to keep the scope of the changes minimal, so didn't really want to go about specifying a custom type.  In my experience, regardless of which option I pick, it's a coin toss if the reviewer will want to see the *other* option that I didn't pick. :)  So I learned to take it easy and love the process.

In this particular case, I would be glad to handle the clean-up to your liking, but I'll need to ask you to give specific actionable comments.  What I read here is you stating your preferences but leaving any follow-up work as optional.

So, if you want something changed in this code, let me know and I'll do so.  By now I know enough about the code base that I can make the change quickly.

shahms: in https://github.com/kythe/kythe/pull/4666

filmil: Thank you!

PR #4667, opened 2020-08-20
fix(bazel): align options between `query` and `build`

justbuchanan:
This flag seems to invalidate the build graph cache and causes bazel to re-fetch and re-analyze a lot of dependencies when switching between `bazel query` and `bazel build`.

I frequently use `ibazel` and this change makes the experience *significantly* better as it often runs bazel queries.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,6 +6,7 @@ build --strategy=TypeScriptCompile=sandboxed,local
 
 # Compatibility flags to avoid regressions.
 build --incompatible_disallow_empty_glob
+query --incompatible_disallow_empty_glob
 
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -6,6 +6,7 @@ build --strategy=TypeScriptCompile=sandboxed,local
 
 # Compatibility flags to avoid regressions.
 build --incompatible_disallow_empty_glob
+query --incompatible_disallow_empty_glob
```
shahms, line 9: does just putting this in `common` work?

justbuchanan: thanks, didn't see that.

PR #4669, opened 2020-08-20
feat(textproto): make example plugin more interesting

justbuchanan:
The example defines semantic nodes in a new language ("textproto_plugin_example" rather than "textproto") and adds edges that cross file boundaries, which is more in line with what a "real" plugin is likely to do.


justbuchanan:
The example defines semantic nodes in a new language ("textproto_plugin_example" rather than "textproto") and adds edges that cross file boundaries, which is more in line with what a "real" plugin is likely to do.

PR #4682, opened 2020-09-09
feat(cxx_indexer): add blame scopes for field references

zrlk:
This PR pulls out blame assignment into a separate function.
It also fixes init.cc on macOSx (and presumably other platforms
that fail to define __THROW).


zrlk:
Looks like the bazel-release failure was a timeout.

jaysachs (approved):

PR #4683, opened 2020-09-10
fix: avoid non-inclusive language

jaysachs:


salguarnieri (approved):

version.bzl
```diff
@@ -49,11 +49,11 @@ def check_version(min_required, max_supported):
     if max < _bound_size(found_version, len(max)):
         fail("Your bazel is too new. Maximum supported version {} of bazel, found {}".format(max_supported, found))
 
-def blacklist_version(version, reason):
+def prohibit_version(version, reason):
```
salguarnieri, line 52: Is this called anywhere?

jaysachs: Couldn't find any reference. Maybe it was in use pre-1.0?

salguarnieri: Maybe we can try to remove it in a followup then.

creachadair: I believe this was added in response to a version of Bazel with breaking changes, to mask out that version. Now that Kythe is tested in the Bazel release CI, this is probably less necessary.

PR #4687, opened 2020-09-16
feat(cxx_indexer): distinguish field writes from reads

zrlk:
This PR uses ref/writes edges for field writes.
Note that intermediate fields in a path (x, y in s.x.y.z)
are not considered written to in an expression like
`s.x.y.z = 3`.


shahms (approved):

creachadair:
This is a long-requested feature 🙂. Is there a corresponding schema change?

jaysachs:
> This is a long-requested feature 🙂. Is there a corresponding schema change?

Once we're past the prototyping phase, yes ... I'm still interested in what is feasible, across languages, and how it matches with what people actually want.

creachadair:
> Once we're past the prototyping phase, yes ... I'm still interested in what is feasible, across languages, and how it matches with what people actually want.

Agree. Is there a design doc or a discussion issue open for the feature? I'd like to follow along with your thoughts and results.

zrlk:
Added issue #4688 for discussion/documentation.

PR #4689, opened 2020-09-18
feat(textproto_verifier_test): accept proto_library targets instead of .proto files

justbuchanan:
The main changes are to `proto_verifier_test.bzl` and `textproto_verifier_test.bzl`. The extraction rules and the macros that use them have been changed to accept `proto_library` targets instead of a list of .proto files. This makes  it easy to use proto libraries that have many dependencies without having to explicitly list each one. The extraction rules use each `proto_library`'s `ProtoInfo` to build a list of transitive `.proto` sources.

Other changes:
* A `proto_library` has been added for each testdata proto
* Proto/textproto imports are now relative to the repo root, not the testdata directory
* The proto extractor rules now derive `--proto_path` values from the input `proto_library`s. They no longer need to be specified in the BUILD file.
* added kythe/data/vnames.proto.json, which adjusts vnames for .proto files in external repositories. The new vname pattern regex `bazel-out/[^/]+/bin/external/([^/]+)/_virtual_imports/[^/]+/(.*),` is used to assign reasonable paths and a root for external protos. See the config for specifics


kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
```diff
@@ -14,31 +14,53 @@
...
+    proto_paths = []
+    for lib in protolibs:
+        if ProtoInfo not in lib:
+            fail("Expected a proto_library, got: %s" % str(lib))
+        for src in lib[ProtoInfo].transitive_sources.to_list():
+            if src.path.endswith(".proto"):
+                srcs.append(src)
+            proto_paths += lib[ProtoInfo].transitive_proto_path.to_list()
+    srcs = collections.uniq(srcs)
+    proto_path_opts = collections.before_each("--proto_path", collections.uniq(proto_paths))
```
jaysachs, line 47: Curious: is there a possibility that identically named srcs in two different paths will now get confused? e.g. are we maintaining ordering on the proto_paths?

kythe/cxx/indexer/cxx/testdata/proto/cc_proto_verifier_test.bzl
```diff
@@ -24,8 +24,7 @@ def cc_proto_verifier_test(
         name,
         srcs,
         proto_lib,
-        proto_srcs = [],
-	proto_deps = [],
+        proto_deps = [],
```
shahms, line 27: should this not just be `deps`?

shahms: (arguably, proto_lib should be a single-element-deps, similarly to how the language specific proto rules work)

kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
```diff
@@ -14,31 +14,53 @@
...
+def get_proto_files_and_proto_path_opts(protolibs):
+    """Given a list of proto_library targets, returns a list of all transitively
+    -included .proto files and a list of options to pass to protoc (containing
+    --proto_path locations).
+    """
+    srcs = []
+    proto_paths = []
+    for lib in protolibs:
+        if ProtoInfo not in lib:
+            fail("Expected a proto_library, got: %s" % str(lib))
```
shahms, line 41: str() should be unnecessary here

kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
```diff
@@ -14,31 +14,53 @@
...
 
+def get_proto_files_and_proto_path_opts(protolibs):
+    """Given a list of proto_library targets, returns a list of all transitively
+    -included .proto files and a list of options to pass to protoc (containing
+    --proto_path locations).
+    """
+    srcs = []
+    proto_paths = []
+    for lib in protolibs:
+        if ProtoInfo not in lib:
```
shahms, line 40: just add a `providers` requirement on the attribute: https://docs.bazel.build/versions/master/skylark/lib/attr.html#label

kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
```diff
@@ -14,31 +14,53 @@
...
+            fail("Expected a proto_library, got: %s" % str(lib))
+        for src in lib[ProtoInfo].transitive_sources.to_list():
+            if src.path.endswith(".proto"):
+                srcs.append(src)
+            proto_paths += lib[ProtoInfo].transitive_proto_path.to_list()
```
shahms, line 45: You probably want to collect these into a depset(), rather than listifying them.

```
for ... :
  proto_paths.append(transitive_proto_path)
...
proto_paths = depset(transitive = proto_paths)
```

And then you can use Args rather than string options: https://docs.bazel.build/versions/master/skylark/lib/attr.html#label

kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
justbuchanan: I think that you're right that there is a possibility, although invoking `protoc` directly has the same limitations. I am maintaining the order from `ProtoInfo`, but because a `proto_library`s deps are not really ordered, I don't think it's well defined.

In the kythe repo, most `proto_library`s only have "--proto_path .", although any that depend on a `.proto` from an external dependency also have something like "bazel-out/k8-fastbuild/bin/external/com_google_protobuf/_virtual_imports/any_proto".

kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
shahms: Also, why are you adding the same proto_paths for each src in `lib[ProtoInfo].transitive_sources` ?

shahms: You *probably* only want to use the direct sources for the verifier facts.  It's not a huge deal, but verifier runtime is superlinear in the number of facts, so avoiding potentially redundantly checking them transitively might be good.

kythe/cxx/indexer/cxx/testdata/proto/cc_proto_verifier_test.bzl
justbuchanan: Removed `proto_deps` and `proto_lib` in favor of a single `proto_libs` argument. I adjusted the `proto_extract_kzip` rule to only list the top-level proto(s) of each lib as sources for the verifier, so some tests need to specify multiple proto libraries.

shahms (approved):

kythe/cxx/indexer/proto/testdata/proto_verifier_test.bzl
justbuchanan: Good point about only using direct sources for the verifier facts. I changed the helper function to now return a list of toplevel sources alongside the full set of sources. The toplevel sources are passed as sources to the extractor, while the full set of imported protos is given as `deps`.

I tried using a depset for `proto_paths`, but it later gets passed to the `extract()` rule from `verifier_test.bzl`, which accepts `opts` as a list. I could use a depset and convert it back, but not sure if it's worth the extra shuffling.

shahms: It also accepts `Args` (just not a list-of-args, which is probably a mistake).

justbuchanan: Thanks, I got it working by passing an `Args` directly (not in a list).

shahms (approved):

PR #4690, opened 2020-09-21
Use proto_library targets for skipping WKPs in C++ proto toolchains

Yannic:
This change migrates the `proto_lang_toolchain` targets to be compatible
with `--incompatible_blacklisted_protos_requires_proto_info`.

See https://github.com/bazelbuild/bazel/issues/11694


shahms:
This seems incredibly fragile to future changes in the protobuf library's list of well known protos.

kythe/extractors/BUILD
```diff
@@ -39,7 +39,18 @@ java_library(
...
+        "@com_google_protobuf//:compiler_plugin_proto",
+        "@com_google_protobuf//:descriptor_proto",
+        "@com_google_protobuf//:duration_proto",
+        "@com_google_protobuf//:empty_proto",
+        "@com_google_protobuf//:field_mask_proto",
+        "@com_google_protobuf//:source_context_proto",
+        "@com_google_protobuf//:struct_proto",
+        "@com_google_protobuf//:timestamp_proto",
+        "@com_google_protobuf//:type_proto",
+        "@com_google_protobuf//:wrappers_proto",
```
shahms, line 53: This seems fragile against future additions to the list of well-known protos.  Would it be possible to export the list of WKP from a .bzl file rather than requiring us to keep a hard-coded list in sync?

Yannic: Not sure I understand what you're asking for. Do you mean create a .bzl file in the kythe repo or in the protobuf repo. If it's the former, of course, we can do that. If it's the latter, then I'd still ask whether we can go with the first option until such a version is released so we can flip the flag without breaking you. WDYT?

shahms: Asking about the latter as otherwise it's far too easy for the list of WKP in the protobuf repository to become out of sync with the list we use here.  We're unlikely to notice a direct breakage if new types are added unless we actually *use* those types, so any such breakage is more likely to be felt downstream increasing the risk.

This is fine for now, however, but still feels pretty fragile.

Yannic: I'll send a change to protobuf!

shahms: Thanks!

shahms (approved):
(Changes to proto can be slow and I want to unblock you, so approving)

Issue #4688, opened 2020-09-16
Represent some form of dataflow in the graph


zrlk:
This is a rollup/discussion issue for experimental dataflow support, starting with C++.

blipper:
This would be super useful!

Right now dataflow analysis requires language specific tooling.  Moreover since Kythe understands cross language references the cross language dataflow opportunities are great.  Constant propagation would be an awesome first step.

PR #4696, opened 2020-09-26
feat(cxx_indexer): trace influence through function calls

zrlk:
This PR adds influence edges for function calls. In particular:

  * argument expressions influence parameter variables
  * parameter variables on incomplete functions influence the
    parameter variables on functions that complete them
  * return expressions influence their containing (complete)
    functions
  * a function call results in that function being included
    in the current influence set.

It should not be necessary to make complete functions influence
the functions they complete as they are already related through
clustering. Parameters on incomplete definitions are not clustered
with the corresponding parameters on complete definitions.

Note: it may be necessary to implement the new traverse functions
using data recursion should we start running out of stack. I have not
observed this behavior yet, but it's worth mentioning now rather than
discovering later.

Related to #4688.

### DIFF
--- a/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
+++ b/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
@@ -1514,6 +1514,9 @@ bool IndexerASTVisitor::VisitCallExpr(const clang::CallExpr* E) {
       }
     }
   }
+  if (!Job->InfluenceSets.empty() && E->getDirectCallee() != nullptr) {
+    Job->InfluenceSets.back().insert(E->getDirectCallee());
+  }
   return true;
 }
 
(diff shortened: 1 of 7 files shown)


kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1969,6 +1972,50 @@ bool IndexerASTVisitor::VisitInitListExpr(const clang::InitListExpr* ILE) {
   return true;
 }
 
+bool IndexerASTVisitor::TraverseCallExpr(clang::CallExpr* CE) {
+  auto callee = CE->getDirectCallee();
+  auto callee_exp = CE->getCallee();
+  bool valid = callee != nullptr && callee_exp != nullptr &&
+               callee->param_size() == CE->getNumArgs();
```
shahms, line 1979: So this doesn't work for functions with default arguments? or are default arguments included in getNumArgs()?

Please document when these conditions might fail, if you know.

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1969,6 +1972,50 @@ bool IndexerASTVisitor::VisitInitListExpr(const clang::InitListExpr* ILE) {
...
+  for (unsigned arg = 0; valid && arg < CE->getNumArgs(); ++arg) {
+    valid = CE->getArg(arg) != nullptr && callee->getParamDecl(arg) != nullptr;
+  }
+  if (valid) {
+    auto callee_node = BuildNodeIdForDecl(callee);
+    if (!WalkUpFromCallExpr(CE)) return false;
+    if (!TraverseStmt(callee_exp)) return false;
+    for (unsigned arg = 0; arg < CE->getNumArgs(); ++arg) {
```
shahms, line 1987: for (clang::Expr* arg : CE->arguments()) {
  ...
}

zrlk: Doesn't work with callee->getParamDecl(arg) below.

shahms (approved):

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
zrlk: I was a little too conservative with this check. I've also added tests showing that this doesn't fail (in the crashy sense) with varargs or pack expansions.

PR #4697, opened 2020-09-30
chore(cxx): increase test timeout

salguarnieri:
This test may take close to the 'small' test timeout.


shahms:
This timeout concerns me as the tests themselves aren't very large, even transitively, unlike the other slow tests (proto2).  It seems like the influences stuff may be emitting a lot of additional graph data?

shahms (approved):

PR #4701, opened 2020-10-06
feat(cxx_indexer): add user-provided template instantiation excludelist

zrlk:
Certain files, especially in the standard library (`<type_traits>`),
generate a disproportionate amount of nearly useless data from
implicit template instantiations. This PR gives users the ability
to specify that primary/partial templates defined in certain files
should not have their implicit instantiations indexed.


kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1049,11 +1049,72 @@ bool IndexerASTVisitor::TraverseCXXConstructorDecl(
          TraverseDeclarationNameInfo(DNI);
 }
 
+bool IndexerASTVisitor::ShouldIndex(const clang::Decl* Decl) {
+  if (TemplateInstanceExcludePathPattern == nullptr) {
+    return true;
+  }
```
jaysachs, line 1055: This is a lot of code, that conceptually is simple(-ish). I think it's worth a comment here describing that we return true if the decl is an explicit instantiation or a (partial) specialization, or if the regexp does not match the containing source file path.

kythe/cxx/indexer/cxx/IndexerASTHooks.h
```diff
@@ -115,7 +118,14 @@ class IndexerASTVisitor : public RecursiveTypeVisitor<IndexerASTVisitor> {
         MarkedSources(&Sema, &Observer),
         ShouldStopIndexing(std::move(ShouldStopIndexing)),
         UsrByteSize(UsrByteSize),
-        DataflowEdges(EDE) {}
+        DataflowEdges(EDE) {
+    if (!TIEPP.empty()) {
+      re2::RE2::Options options;
+      options.set_never_capture(true);
+      TemplateInstanceExcludePathPattern =
+          std::make_unique<re2::RE2>(std::string(TIEPP), options);
```
shahms, line 126: This silently ignores syntax errors in the regexp.  I think it makes more sense to accept a `std::unique_ptr<RE2>` (or `RE2*` or `re2::Set`) and CHECK that it is `ok()`.

Also, as an nit, there's no need to copy the `absl::string_view`; RE2 has a re2::StringPiece constructor so:

`absl::make_unique<re2::RE2>({TIEPP.data(), TIEPP.size()}, options);`

The above should work both internally and externally.

kythe/cxx/indexer/cxx/KytheIndexerMain.cc
```diff
@@ -57,6 +57,9 @@ ABSL_FLAG(bool, use_compilation_corpus_as_default, false,
           "Use the CompilationUnit VName corpus as the default.");
 ABSL_FLAG(bool, experimental_record_dataflow_edges, false,
           "Emit experimental dataflow edges.");
+ABSL_FLAG(std::string, template_instance_exclude_path_pattern, "",
+          "If nonempty, a regex that matches files to be excluded from "
+          "template instance indexing.");
```
shahms, line 62: Per my other comment, I would convert this to a RE2 as eagerly as possible.  Ideally at argument-parsing time:
```
struct RE2Flag {
  std::unique_ptr<re2::RE2> value;
};

bool AbslParseFlag(absl::string_view text, RE2Flag* value, std::string* error) {
  re2::RE2::Options options;
  options.set_never_capture(true);
  *value = {absl::make_unique<re2::RE2>({text.data(), text.size()}, options)};
  if (!value->value->ok()) {
    *error = value->value->error_text();
  }
  return value->value.ok();
}
std::string AbslUnparseFlag(const RE2Flag& value) {
  return value.value->pattern();
}
```

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1049,11 +1049,72 @@ bool IndexerASTVisitor::TraverseCXXConstructorDecl(
...
+            from.dyn_cast<clang::VarTemplatePartialSpecializationDecl*>()) {
+      loc = partial->getBeginLoc();
+    } else if (const auto* primary = from.dyn_cast<clang::VarTemplateDecl*>()) {
+      loc = primary->getBeginLoc();
+    } else {
+      return true;
+    }
+  } else {
+    return true;
+  }
```
shahms, line 1101: from 1056 to here should be a separate, `absl::optional<clang::SourceLocation>` returning function.

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1049,11 +1049,72 @@ bool IndexerASTVisitor::TraverseCXXConstructorDecl(
...
+  auto file = Observer.getSourceManager()->getFilename(loc);
+  if (file.data() == nullptr) {
```
shahms, line 1103: Just `empty()`? While there may technically be a semantic distinction between `empty()` and `data() == nullptr` I don't think they're relevant here (and generally checking for a specific address on `.data()` is an anti-pattern).

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1049,11 +1049,72 @@ bool IndexerASTVisitor::TraverseCXXConstructorDecl(
...
+      loc = partial->getBeginLoc();
+    } else if (const auto* primary = from.dyn_cast<clang::VarTemplateDecl*>()) {
+      loc = primary->getBeginLoc();
+    } else {
+      return true;
+    }
+  } else {
+    return true;
+  }
+  auto file = Observer.getSourceManager()->getFilename(loc);
```
shahms, line 1102: What happens if this `loc` is in a macro somewhere?

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -1049,11 +1049,72 @@ bool IndexerASTVisitor::TraverseCXXConstructorDecl(
...
+      return true;
+    }
+    auto from = rec->getInstantiatedFrom();
+    if (const auto* partial =
+            from.dyn_cast<clang::ClassTemplatePartialSpecializationDecl*>()) {
+      loc = partial->getBeginLoc();
+    } else if (const auto* primary =
+                   from.dyn_cast<clang::ClassTemplateDecl*>()) {
+      loc = primary->getBeginLoc();
+    } else {
```
shahms, line 1081: `from` is guaranteed to be one of the above or `nullptr`. I think doing the nullptr check first is more readable.

shahms (commented):
Thanks! A few comments.

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
zrlk: Let's use the spelling location in this case.

kythe/cxx/common/re2_flag.cc
```diff
@@ -0,0 +1,32 @@
...
+  *value = {std::make_shared<re2::RE2>(re2::StringPiece{text.data(), text.size()}, options)};
+  if (!value->value->ok()) {
+    *error = value->value->error();
+  }
+  return value->value->ok();
+}
+std::string AbslUnparseFlag(const RE2Flag& value) {
+  return value.value->pattern();
+}
+}
```
shahms: nit: newline?

shahms (approved):

kythe/cxx/common/re2_flag.cc
```diff
@@ -0,0 +1,33 @@
...
+  options.set_never_capture(true);
+  *value = {std::make_shared<re2::RE2>(
+      re2::StringPiece{text.data(), text.size()}, options)};
+  if (!value->value->ok()) {
+    *error = value->value->error();
+  }
+  return value->value->ok();
+}
+std::string AbslUnparseFlag(const RE2Flag& value) {
+  return value.value->pattern();
```
shahms, line 37: need to check for nullptr :-(

kythe/cxx/common/re2_flag.cc
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kythe/cxx/common/re2_flag.h"
+
+namespace kythe {
+bool AbslParseFlag(absl::string_view text, RE2Flag* value, std::string* error) {
+  re2::RE2::Options options;
```
shahms, line 24: Probably want to let the empty string mean "nullptr" as well?

PR #4704, opened 2020-10-08
fix(go_test): dynamically create files for vfs_test

### LINKED ISSUE #4703: Cloning on native Windows is not possible
It appears it's not possible to clone this repository on Windows without using WSL. This is caused by [the file path with an asterisk in its name](https://github.com/kythe/kythe/tree/master/kythe/go/platform/vfs/testdata/a/b/*/c), which is used by [vfs_test.go](https://github.com/kythe/kythe/blob/master/kythe/go/platform/vfs/vfs_test.go).

Error:
```
Failed to install: Unable to get repository: Cloning into 'redacted'...
error: invalid path 'kythe/go/platform/vfs/testdata/a/b/*/c/empty.txt'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

shahms:
Fixes #4703 (the test won't *pass* on windows, but the clone will at least be possible).

### DIFF
--- a/kythe/go/platform/vfs/BUILD
+++ b/kythe/go/platform/vfs/BUILD
@@ -11,9 +11,6 @@ go_test(
     name = "vfs_test",
     size = "small",
     srcs = ["vfs_test.go"],
-    data = glob([
-        "testdata/**",
-    ]),
     library = "vfs",
     deps = [
         "//kythe/go/test/testutil",
--- a/kythe/go/platform/vfs/vfs_test.go
+++ b/kythe/go/platform/vfs/vfs_test.go
@@ -18,23 +18,57 @@ package vfs
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


jrn:
Nice. I imagined a dependency on a genrule at https://github.com/kythe/kythe/blob/585f047bac/kythe/go/platform/vfs/BUILD#L14 generating a file with the right path; generating it in Go instead like this is much nicer.

I've read through the patch and it looks sensible to me, for what it's worth.

shahms:
> Nice. I imagined a dependency on a genrule at https://github.com/kythe/kythe/blob/585f047bac/kythe/go/platform/vfs/BUILD#L14 generating a file with the right path; generating it in Go instead like this is much nicer.
> 
> I've read through the patch and it looks sensible to me, for what it's worth.

I considered that, but it's easier to (eventually) skip the test on windows if it's self-contained.  Thanks for the feedback :-)

PR #4715, opened 2020-10-18
fix(typescript_indexer): don't emit defines/binding for computed properties

nbeloglazov:
From developer point of view computed properties themselves don't
represent a separate entity that you can click. Instead they are
just expressions and identifiers in that expression should be
xrefed.

Current behavior (which is being fixed in this commit) is the
following:

```typescript
const k = {
  [foo]: 123,
};
```

Currently `[foo]` considered a definition and it overlaps `foo`
making it impossible to click on `foo`.

Tested:
  unit tests


nbeloglazov:
@evmar

evmar (approved):

PR #4716, opened 2020-10-19
feat(typescript_indexer): don't emit references to hops in imports

nbeloglazov:
This change is a rollback of https://github.com/kythe/kythe/pull/3990. UI (CodeSearch) still doesn't support this usecase properly and current workaround introduces more confusion than it helps (we have a number of bugs about that). Even though it represents code more correctly. In particular with current approach it's impossible to easily see all usages or calls of given variable or function. Instead when you check call of function `doSomething` - you can only get refs from `import` statements from other files and then go those files one by one and search for usages of `doSomething` there.

Once UIs properly support this usecase of imported variables - we'll roll-back this change again.

Also moved import tests to separate folders so that files belonging to the same test indexed together.

Tested: unit tests


nbeloglazov:
@evmar and @ayazhafiz as we discussed it couple of months ago.

ayazhafiz (approved):

shahms:
The extant UIs should support the correct behavior, there's a flag in the Kythe pipeline which hasn't yet been enabled in live but I believe should be this quarter.

nbeloglazov:
@shahms which flag are you talking about?

shahms:
There is an internal flag on Kythe serving to enable these indirections.  It has not been enabled in prod.

nbeloglazov:
Talked to @shahms on chat. The plan is to change TS indexer to follow https://www.kythe.io/docs/schema/modules.html and guard that by a flag. That will allow us to see how it works in staging environment and later enable once the flag Shams mentioned enabled by default.

shahms:
Doing some cleanup; from your last comment it sounds like this PR isn't relevant any more? Feel free to re-open if it is.

PR #4717, opened 2020-10-21
Make the verifier library publicly visible

ivan444:
Change //kythe/cxx/verifier:lib visibility to public. That will allow Kythe users to build verification tests as simple unit tests without the need to compile-in the Kythe verification binary.

### DIFF
--- a/kythe/cxx/verifier/BUILD
+++ b/kythe/cxx/verifier/BUILD
@@ -69,6 +69,7 @@ cc_library(
         "-Wno-unused-variable",
         "-Wno-implicit-fallthrough",
     ],
+    visibility = [PUBLIC_VISIBILITY],
     deps = [
         ":lexparse",
         "//kythe/cxx/common:file_utils",


shahms:
@zrlk Thoughts? I'm not sure this is a commitment we want to make.  The "supported" interface to the verified is via the CLI binary.

zrlk:
We're not ready to commit to making this a public API (in part because we're planning on changing the solver), but we'd consider building a stable programmatic API. What use case do you have in mind?

ivan444:
Use case: tests based on the kythe verifier (more details at b/169775496).

At the moment, Kythe clients have access only to the verification binary.
That makes writing the unit tests limited -- e.g. difficult to get the unit
test code coverage right, writing custom bazel rules gets messy. The code
coverage is the main pain point. Kythe verification tests are amazingly
powerful. But it's difficult to think of all the cases -- code coverage
simplifies it significantly.

If you are strongly against making the library publicly available, does
Bazel allow setting the visibility per external project (like
GitHub.com/google/verible)?

fangism:
> b/169775496

Also b/169774483 for background on trying to measure coverage.

shahms:
It seems like this is obsolete now; please feel free to re-open if not.

PR #4719, opened 2020-10-22
chore(typescript_indexer): group related test files in subdirectories.

nbeloglazov:
Tests that verify indexing of importing/exporting symbols require indexing two or more files at the same time. Currently some of these tests live in the root testdata directory which means that each file is indexed and verified individually defeating the purpose of the test. Moving them to separate _group directories solves that.


shahms (approved):
This seems fine to me, but I'd like Evan to double check :-)

evmar (approved):
Is the _group suffix meaningful?  If not, I'd leave it out.  (Sorry to bikeshed, feel free to disregard.)

nbeloglazov:
I thought the same thing, but it is meaningful: https://github.com/kythe/kythe/blob/master/kythe/typescript/test.ts#L150

PR #4721, opened 2020-10-22
feat(textproto): expose DescriptorPool in plugin api

justbuchanan:
plugins can use this to lookup proto descriptors based on qualified names

### DIFF
--- a/kythe/cxx/indexer/textproto/plugin.h
+++ b/kythe/cxx/indexer/textproto/plugin.h
@@ -55,7 +55,7 @@ class PluginApi {
   virtual proto::VName VNameForRelPath(
       absl::string_view simplified_path) const = 0;
 
-  // Returns a reference to the descriptor pool built from the .proto files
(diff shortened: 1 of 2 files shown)


kythe/cxx/indexer/textproto/plugin.h
```diff
@@ -54,6 +54,11 @@ class PluginApi {
   // required_inputs.
   virtual proto::VName VNameForRelPath(
       absl::string_view simplified_path) const = 0;
+
+  // Returns a reference to the descriptor pool built from the .proto files
```
creachadair, line 58: A minor nit (since "reference" is a term of art in the type system):

```suggestion
  // Returns a pointer to the descriptor pool built from the .proto files
```

justbuchanan: :+1:

shahms (approved):

PR #4723, opened 2020-10-26
chore: update to new LLVM version

shahms:
This also fixes a longstanding bug in how FYI was interpreting the `Invalid` output parameter, which necessitated a change to the corresponding test. 

Luke, can you verify that `//kythe/cxx/tools/fyi/testdata:nothing` looks reasonable?


shahms:
Note: I'll also be holding off of committing this until the corresponding change is submitted internally.

salguarnieri (approved):

PR #4727, opened 2020-10-27
chore: update go dependencies

shahms:
Brotli's upstream organization is particularly awkward.  They also recently changed is to the go directory and several others are omitted from the Git archive, so I changed Kythe to use the Gazelle default imports for those, rather than a different set of more fragile shenanigans.  I also updated the proto-regeneration script to include protos outside of //kythe/proto/...


shahms:
Also, Gazelle sorts repository rules now, which caused a larger-than-necessary diff.

salguarnieri (approved):

PR #4734, opened 2020-11-10
fix(java_indexer): blame instance/static callsites on ctor/cinit methods

schroederc:
This matches the C++ indexer's existing behavior and better represents the actual callgraph.

The cinit function is newly represented in this commit as a bare function node per class.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
@@ -185,7 +185,7 @@ public class JavaEntrySets extends KytheEntrySets {
         newNode(NodeKind.FUNCTION)
             .setCorpusPath(CorpusPath.fromVName(classNode))
             .addSignatureSalt(classNode)
-            .addSignatureSalt("cinit")
+            .addSignatureSalt("clinit")
             .build());
   }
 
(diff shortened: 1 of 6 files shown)


shahms (approved):

kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
```diff
@@ -179,6 +179,16 @@ public EntrySet newLambdaAndEmit(Positions filePositions, JCTree.JCLambda lambda
             .addSignatureSalt("" + filePositions.getSpan(lambda)));
   }
 
+  /** Emits and returns a new {@link EntrySet} representing a static class initializer. */
+  public EntrySet newClassInitAndEmit(VName classNode) {
+    return emitAndReturn(
+        newNode(NodeKind.FUNCTION)
+            .setCorpusPath(CorpusPath.fromVName(classNode))
+            .addSignatureSalt(classNode)
+            .addSignatureSalt("cinit")
```
jaysachs, line 188: any reason why "cinit" and not "clinit"? is that consistency with C++ desired?

schroederc: It's just feeding the hash so it doesn't matter much but...........

kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
```diff
@@ -185,7 +185,7 @@ public EntrySet newClassInitAndEmit(VName classNode) {
         newNode(NodeKind.FUNCTION)
             .setCorpusPath(CorpusPath.fromVName(classNode))
             .addSignatureSalt(classNode)
-            .addSignatureSalt("cinit")
+            .addSignatureSalt("clinit")
```
jaysachs, line 188: Yeah, I guess it doesn't matter. There'd be a mild aesthetic improvement if we didn't just have "salt" but also e.g. a "sigPrefix" on the builder, so the sig ended up looking something like like "#clinit-a93ef8g90a9"

Issue #4748, opened 2020-11-12
Make the verifier faster


zrlk:
The verifier is insufficiently fast. It should be faster.

PR #4749, opened 2020-11-12
feat(verifier): add souffle/libffi

zrlk:
This PR adds Bazel rules to pull down and build souffle
and its libffi dependency. souffle is a high-performance
datalog implementation that supports both AOT compilation
and running in interpreter mode. We intend to add it as
a new backend for the verifier.

In service of #4748


third_party/libffi.BUILD
```diff
@@ -0,0 +1,74 @@
...
+        "-Wno-vla",
+        # libffi does not check the result of ftruncate.
+        "-Wno-unused-result",
+    ],
+    includes = [
+        "configure-bazel-gen",
+        "configure-bazel-gen/include",
+        "include",
+    ],
+    textual_hdrs = ["src/dlmalloc.c"],
```
shahms, line 72: `textual_hdrs` are exported for clients to include (which I don't think is the intent).  If this needs to be included by the library itself, but shouldn't be compiled independently (as it would be in `srcs`), but it in a separate private cc_library.

autotools.bzl
```diff
@@ -0,0 +1,80 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Originally from google/sandboxed-api; added 'bootstrap'.
+
+"""Repository rule that runs autotools' configure after extraction."""
+
+def _configure(ctx):
```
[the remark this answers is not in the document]
shahms: As Souffle itself isn't in internal third_party, you'll need to write a BUILD file for that anyway.

third_party/souffle.BUILD
```diff
@@ -0,0 +1,465 @@
...
+        "src/include/souffle/utility/EvaluatorUtil.h",
+        "src/include/souffle/utility/FunctionalUtil.h",
+        "src/include/souffle/utility/MiscUtil.h",
+        "src/include/souffle/utility/ParallelUtil.h",
+        "src/include/souffle/utility/StreamUtil.h",
+        "src/include/souffle/utility/StringUtil.h",
+        "src/include/souffle/utility/json11.h",
+        "src/include/souffle/utility/span.h",
+        "src/include/souffle/utility/tinyformat.h",
+        # other headers
```
shahms, line 210: These aren't public headers and shouldn't go in `hdrs` (they belong in `srcs`).

third_party/souffle.BUILD
```diff
@@ -0,0 +1,465 @@
...
+        "src/ram/utility/NodeMapper.h",
+        "src/ram/utility/Utils.h",
+        "src/ram/utility/Visitor.h",
+        "src/reports/DebugReport.h",
+        "src/reports/ErrorReport.h",
+        "src/synthesiser/Synthesiser.h",
+        "src/synthesiser/Relation.h",
+    ],
+    includes = [
+        "src",
```
shahms, line 459: omit this; `includes` are inherited by everything which transitively depends on a particular cc_library. Obnoxiously, if you want to change the include path while compiling the library itself, you need to do `copts = ["-Iexternal/annoying/bazel/path/to/src"]`

third_party/souffle.BUILD
shahms: It *appears* as though this is pretty close to:
```
cc_library(
  srcs = glob(["src/**", exclude = ["src/include/**"]) + [":scanner", ":parser"]),
  hdrs = glob(["src/include/**"]),
  ...
)
```

third_party/libffi.BUILD
zrlk: This is directly from google/sandboxed-api. I don't pretend to know why dlmalloc is a textual_hdr, but I do note that the import of libffi is in the same state.

third_party/souffle.BUILD
zrlk: I would prefer to avoid globs (in part to possibly remove unnecessary source files in the future).

zrlk: I also need access to the internal headers.

autotools.bzl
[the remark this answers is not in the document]
zrlk: This is another thing that's directly from sandboxed-api. Since Souffle hasn't been imported yet, I'm planning to defer massaging the build rules until I absolutely have to (or until buildbot gets upset).

third_party/souffle.BUILD
zrlk: I need access to these headers from dependent libraries.

autotools.bzl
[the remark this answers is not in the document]
shahms: As nothing in the main repository yet depends on these targets, I'm not sure buildbot will even attempt fetch these, let alone build them.

third_party/souffle.BUILD
zrlk: Which way would you find acceptable? I expect that I will need access to most of the headers (+ transitive dependencies) that main.cpp requires.

third_party/libffi.BUILD
```diff
@@ -0,0 +1,74 @@
...
+            "src/powerpc/ppc_closure.S",
+            "src/powerpc/sysv.S",
+        ],
+        "@bazel_tools//src/conditions:linux_aarch64": [
+            "src/aarch64/ffi.c",
+            "src/aarch64/internal.h",
+            "src/aarch64/sysv.S",
+        ],
+        "//conditions:default": [
+        ],
```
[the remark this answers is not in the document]
zrlk: This builds on Darwin; I tested linking it to another Kythe target and pulling in some utility code and there were no linker problems. I suspect it won't be an issue unless something tries to initialize the FFI.

third_party/souffle.BUILD
shahms: Man I hate header files. Assuming it actually works, I think moving the internal headers to srcs here and adding something like:

```
cc_library(
  name = "internal",
  hdrs = ["all/the/things"],
  strip_include_prefix = "src",
  include_prefix = "souffle",  # Or "souffle/internal" ?
)
```

so that external clients, in theory, have to use `#include <souffle/synthesiser/Relation.h>` rather than `#include "src/synthesiser/Relation.h"` would be fine?  I'm not sure whether or not you can get away with doing all of that in the same cc_library or not (but I suspect not). If that doesn't work this is fine; I'd like a little lipstick on the pig, but it's not worth losing a hand.

autotools.bzl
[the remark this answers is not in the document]
zrlk: I've manually built the target (on Darwin) and successfully linked an executable that depended on it (I instantiated an arbitrary utility class to make sure the library wasn't completely discarded).

third_party/libffi.BUILD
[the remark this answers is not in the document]
zrlk: I've added the TODO; I agree that it's worth completing.

third_party/libffi.BUILD
zrlk: Aha, thanks for looking into this. I think I fixed it in the commit I'm currently working on.

third_party/souffle.BUILD
zrlk: I messed with this for a while and eventually got to:
```
cc_library(
    name = "internal_headers",
    hdrs = glob(["src/**/*.h"], exclude = ["src/include/**"]),
    strip_include_prefix = "src",
    include_prefix = "souffle",
)

cc_library(
    name = "external_headers",
    hdrs = glob(["src/include/**/*.h"]),
    strip_include_prefix = "src/include",
    include_prefix = "souffle",
)

...
    copts = [
        "-I$(GENDIR)/external/souffle/_virtual_includes/internal_headers/souffle",
        "-I$(GENDIR)/external/souffle/_virtual_includes/external_headers/souffle",
        "-I$(GENDIR)/external/souffle/src",
    ],
...
```

I don't actually know if the library would be usable even if it builds, which it doesn't, because:
```
external/souffle/src/ast/analysis/Type.cpp:19:10: fatal error: 'ConstraintSystem.h' file not found
#include "ConstraintSystem.h"
```

ConstraintSystem.h *is* in src/ast/analysis, but it's not mapped there by Bazel inside the sandbox.

zrlk: (`external_headers` isn't necessary here because there's a src/include/souffle, but fixing that won't fix the mapping problem with the internal headers.)

third_party/souffle.BUILD
zrlk: I've added the globs.

third_party/souffle.BUILD
zrlk: Okay, it turns out that Bazel is less concerned about multiple rules claiming the same headers as I thought.

zrlk: Unfortunately this still doesn't work. If I depend on souffle from the verifier and include one of the internal headers, it's not able to find other internal headers it references:
```
bazel-out/darwin-fastbuild/bin/external/souffle/_virtual_includes/internal_headers/souffle/ast/Program.h:19:10: fatal error: 'ast/Clause.h' file not found
#include "ast/Clause.h"
```
If I add the nasty copt (`"-I$(GENDIR)/external/souffle/_virtual_includes/internal_headers/souffle",`) to the relevant verifier target it works. This is gross but maybe not as gross as dumping a bunch of souffle-internal headers all over the search path. WDYT? Maybe there's a better solution I'm missing?

shahms: Yeah, Bazel is generally fine with rules sharing inputs, it gets crabby when they share outputs.

Thanks for fighting with this. At this point, I think polluting the header space is probably the least bad.  None of the header files that I can see are named `config.h` or something equally generic.

Another option would be to use a genrule to transform the include paths used by the headers you want to re-export, but that seems like overkill here (although, I would definitely suggest it if there were something like `config.h` in use). Sorry for all of the song-and-dance.

zrlk: Okay, in the latest commit I've just gone with `includes`.

shahms (approved):
Thanks!

PR #4762, opened 2020-11-16
chore: update LLVM

shahms:
The newest sync from LLVM stops using the parallel LLVMBuild.txt files, opting instead to incorporate that information directly in the CMake files, so we can omit them as well.


justbuchanan (approved):

shahms:
FYI, I'm waiting on the internal update to be submitted before doing so here.

PR #4764, opened 2020-11-19
feat(cxx_common): extend bazel artifact selection to allow stateful selectors

shahms:
Also Aspect support and type-erased value type wrapper for them.

### DIFF
--- a/kythe/cxx/extractor/bazel_artifact.h
+++ b/kythe/cxx/extractor/bazel_artifact.h
@@ -1,0 +1,26 @@
+#ifndef KYTHE_CXX_EXTRACTOR_BAZEL_ARTIFACT_H_
+#define KYTHE_CXX_EXTRACTOR_BAZEL_ARTIFACT_H_
+
+#include <string>
+#include <vector>
+
+namespace kythe {
+
+/// \brief A pair of local path and canonical URI for a given Bazel output file.
+/// Bazel typically provides both a workspace-local path to a file as well as a
+/// canonical URI and both can be useful to clients.
+struct BazelArtifactFile {
+  std::string local_path;  ///< The workspace-relative path to the file.
+  std::string uri;  ///< The canonical URI for the file. Clients should be able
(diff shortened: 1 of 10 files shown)


shahms:
PTAL?

kythe/cxx/extractor/bazel_artifact.h
```diff
@@ -0,0 +1,59 @@
...
+#include <string>
+#include <tuple>
+#include <vector>
+
+namespace kythe {
+
+/// \brief A pair of local path and canonical URI for a given Bazel output file.
+/// Bazel typically provides both a workspace-local path to a file as well as a
+/// canonical URI and both can be useful to clients.
+struct BazelArtifactFile {
```
jaysachs, line 29: Just curious, this sort of smells like a map. Should it be?

shahms: I don't see how using a map would be beneficial.  While I wouldn't expect duplicate paths in a list of artifacts:

1) The URI isn't unique (`data:` or another content-addressable scheme may be used)
2) The unique-ness of the file paths isn't really a core part of the semantics
3) The resulting API is less readable (`std::pair` is *terrible*), so without a compelling case for #2, the decreased readability of the resulting API obtains (IMO).

kythe/cxx/extractor/bazel_artifact_selector.cc
```diff
@@ -0,0 +1,216 @@
...
+
+std::string AsUri(const build_event_stream::File& file) {
+  switch (file.file_case()) {
+    case build_event_stream::File::kUri:
+      return file.uri();
+    case build_event_stream::File::kContents:
+      // We expect inline data to be rare and small, so always base64 encode it.
+      return absl::StrCat(
+          "data:base64,",
+          // data URIs use regualr base64, not "web safe" base64.
```
jaysachs, line 37: regular

jaysachs (approved):

PR #4768, opened 2020-11-26
feat(verifier): build a souffle frontend that can read an entrystream

zrlk:
This PR builds a version of the souffle frontend that can read an
entrystream directly. For example, the program

```
.type vname = [
  signature:symbol,
  corpus:symbol,
  path:symbol,
  root:symbol,
  language:symbol
]

.decl entry(source:vname, kind:symbol, target:vname, value:symbol)
.input entry(IO=kythe_entrystream)

.decl outentry(source:vname, kind:symbol, target:vname, value:symbol)
.output outentry(IO=json)

outentry(s,k,t,v) :- entry(s,k,t,v).
```

when invoked as
```
./bazel-bin/kythe/cxx/verifier/souffle_with_kythe example.dl \
    < in.entries
```

will print the entries as JSON.

libffi still doesn't work on Darwin and the fix isn't immediately
clear (I did look at a different BUILD file, but I think their
version of libffi was different; ours seems to be missing some
Darwin headers for x86_64?). I've stubbed the relevant symbols
out to make the linker happy.

I don't know if this is the most efficient representation of the
Kythe graph or if it has other problems, so please don't rely on
it (and suggestions for improvement are welcome).

### DIFF
--- a/kythe/cxx/verifier/BUILD
+++ b/kythe/cxx/verifier/BUILD
@@ -80,6 +80,27 @@ cc_test(
 )
 
 cc_library(
+    name = "souffle_entrystream_support",
+    srcs = [
+        "souffle_entrysteam_support.cc",
+    ],
+    deps = [
+        "//kythe/proto:common_cc_proto",
+        "//kythe/proto:storage_cc_proto",
+        "@com_google_protobuf//:protobuf",
+        "@souffle",
+    ],
+)
+
+cc_binary(
+    name = "souffle_with_kythe",
+    deps = [
+        ":souffle_entrystream_support",
+        "@souffle//:main",
+    ],
+)
+
+cc_library(
     name = "lib",
     srcs = [
         "assertions.cc",
--- a/kythe/cxx/verifier/BUILD
+++ b/kythe/cxx/verifier/BUILD
@@ -90,6 +90,7 @@ cc_library(
         "@com_google_protobuf//:protobuf",
         "@souffle",
     ],
+    alwayslink = True,
 )
 
 cc_binary(
--- a/kythe/cxx/verifier/souffle_entrysteam_support.cc
+++ b/kythe/cxx/verifier/souffle_entrysteam_support.cc
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2020 The Kythe Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


kythe/cxx/verifier/BUILD
```diff
@@ -79,6 +79,27 @@ cc_test(
...
+    name = "souffle_entrystream_support",
+    srcs = [
+        "souffle_entrysteam_support.cc",
+    ],
+    deps = [
+        "//kythe/proto:common_cc_proto",
+        "//kythe/proto:storage_cc_proto",
+        "@com_google_protobuf//:protobuf",
+        "@souffle",
+    ],
```
shahms, line 92: I *think* you may need `alwayslink = True` here as the binary isn't directly using any of the symbols from this library and they may end up getting stripped as a result.

kythe/cxx/verifier/souffle_entrysteam_support.cc
```diff
@@ -0,0 +1,159 @@
...
+  /// Copies the fields from `vname` to `target`.
+  void CopyVName(const kythe::proto::VName& vname, souffle::RamDomain* target) {
+    target[kSignatureEntry] = symbolTable.unsafeLookup(vname.signature());
+    target[kCorpusEntry] = symbolTable.unsafeLookup(vname.corpus());
+    target[kPathEntry] = symbolTable.unsafeLookup(vname.path());
+    target[kRootEntry] = symbolTable.unsafeLookup(vname.root());
+    target[kLanguageEntry] = symbolTable.unsafeLookup(vname.language());
+  }
+  /// This entry is reused for each tuple.
+  kythe::proto::Entry entry_;
```
shahms, line 128: It seem like this should just be a function-local in readNextTuple? is there any benefit to re-using the same Entry proto?

kythe/cxx/verifier/souffle_entrysteam_support.cc
```diff
@@ -0,0 +1,159 @@
...
+    coded_input.SetTotalBytesLimit(INT_MAX);
+    if (!coded_input.ReadVarint32(&byte_size)) {
+      return nullptr;
+    }
+    coded_input.PushLimit(byte_size);
+    if (!entry_.ParseFromCodedStream(&coded_input)) {
+      fprintf(stderr, "error reading input stream\n");
+      return nullptr;
+    }
+    auto tuple = std::make_unique<souffle::RamDomain[]>(kEntryElements);
```
shahms, line 107: `souffle::mk` rather than absl::make_unique?

zrlk: Turns out souffle::Own is an alias for std::unique_ptr (my guess is they migrated from their own thing), but consistency is good.

zrlk: Actually, scratch that--souffle::mk doesn't work here. They use make_unique for the same purpose (e.g. https://github.com/souffle-lang/souffle/blob/e1fffe4ef91421cbf975c6898700949156cf5605/src/include/souffle/io/ReadStreamJSON.h#L105).

shahms (approved):

PR #4771, opened 2020-12-03
feat(cxx_extractor): use a stable working_directory if possible

shahms:
Particularly for remote builds, extracted compilation units can differ only in the semantically-irrelevant `working_directory` field. 
 This leads to unnecessary cache-misses and other issues.  To work around this, I've updated the C++ extractor to use the same logic as the Java indexer when normalizing the CU `working_directory` by picking the first otherwise-unused root directory from a static list.


shahms:
Follow-up in progress to account for paths in CxxCompilationUnitDetails as well.

kythe/cxx/extractor/testdata/has_include_test.cc
```diff
@@ -63,7 +63,7 @@ argument: "-I./kythe/cxx/extractor"
 argument: "./kythe/cxx/extractor/testdata/has_include.cc"
 argument: "-fsyntax-only"
 source_file: "./kythe/cxx/extractor/testdata/has_include.cc"
-working_directory: "TEST_CWD"
```
jaysachs, line 66: Are there any tests left which test the "non-stable root" path?

shahms: I'll add a test of the new logic.  None of the existing tests tested for the "non-stable root" because it was unstable, so they set it to a well-known value (`TEST_CWD`) and then used `set_working_directory` on the extracted protobuf.

So there are exactly as many tests of the previous behavior as there used to be?

shahms:
PTAL?

jaysachs (approved):

PR #4775, opened 2020-12-06
fix(cxx_indexer) the memory of the old client is released if it's own…

leonzhuLGU:
…ed by the diags, and cannot be recovered by the scopeguard

### DIFF
--- a/kythe/cxx/indexer/cxx/marked_source.cc
+++ b/kythe/cxx/indexer/cxx/marked_source.cc
@@ -596,9 +596,11 @@ void MarkedSourceGenerator::ReplaceMarkedSourceWithTemplateArgumentList(
   // diagnostics from the typechecker.
   clang::IgnoringDiagConsumer consumer;
   auto* diags = &cache_->source_manager().getDiagnostics();
-  auto guard = MakeScopeGuard(
-      [diags = diags, client = diags->getClient(),
-       owned = diags->ownsClient()] { diags->setClient(client, owned); });
+  auto guard = MakeScopeGuard([diags = diags, client = diags->getClient(),
+                               owned = diags->takeClient()]() mutable {
+    diags->setClient(client, owned.release() != nullptr);
+  });
+
   diags->setClient(&consumer, false);
   auto* template_decl = decl->getSpecializedTemplate();
   auto* template_params = template_decl->getTemplateParameters();


kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -596,9 +596,10 @@ void MarkedSourceGenerator::ReplaceMarkedSourceWithTemplateArgumentList(
   // diagnostics from the typechecker.
   clang::IgnoringDiagConsumer consumer;
   auto* diags = &cache_->source_manager().getDiagnostics();
-  auto guard = MakeScopeGuard(
-      [diags = diags, client = diags->getClient(),
-       owned = diags->ownsClient()] { diags->setClient(client, owned); });
+  
+  auto OwningPreviousClient = diags->takeClient();
+  auto PreviousClient = diags->getClient();
```
shahms, line 601: please use `lower_with_under` for local variables here

kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -697,6 +698,8 @@ void MarkedSourceGenerator::ReplaceMarkedSourceWithTemplateArgumentList(
     }
     *next_arg->mutable_pre_text() = pre_text;
   }
+
+  diags->setClient(PreviousClient, !!OwningPreviousClient.release());
```
shahms, line 702: This should still be done with a scope guard:
```
auto guard = MakeScopeGuard([diags = diags, client = diags->getClient(), owned = diags->takeClient()] {
  diags->setClient(client, owned.release() != nullptr));
});
```

shahms (commented):
Thanks! A few comments, but I appreciate the PR.

kythe/cxx/indexer/cxx/marked_source.cc
leonzhuLGU: I think the lambda expression will capture the smart pointer as const, therefore you cannot call release() inside the body without some proper cast.

shahms: Fair point, but you should just need to make the lambda mutable:
```
auto guard = MakeScopeGuard([diags = diags, client = diags->getClient(), owned = diags->takeClient()] () mutable {
  diags->setClient(client, owned.release() != nullptr));
});
```

kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -697,6 +698,8 @@ void MarkedSourceGenerator::ReplaceMarkedSourceWithTemplateArgumentList(
     }
     *next_arg->mutable_pre_text() = pre_text;
   }
+
+  diags->setClient(previous_client, !!owning_previous_client.release());
```
shahms, line 702: please use an explicit comparison to nullptr

kythe/cxx/indexer/cxx/marked_source.cc
leonzhuLGU: I see, I have changed to use scopeguard which looks more elegant

shahms (approved):
Thanks a lot!

PR #4776, opened 2020-12-08
Update to rules_go 0.25.0.

benjaminp:
This will allow Kythe to work with Bazel 4.0.

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -27,7 +27,7 @@ load("//kythe/rust/cargo:crates.bzl", "raze_fetch_remote_crates")
 
 def _rule_dependencies():
     go_rules_dependencies()
-    go_register_toolchains()
+    go_register_toolchains(version = "1.15.5")
     gazelle_dependencies()
     rules_java_dependencies()
     rules_proto_dependencies()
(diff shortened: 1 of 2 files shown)


shahms:
This appears to break our Bazel CI?

shahms (approved):
Thanks!

fangism:
Just FYI, we we've been building from top-of-tree in our project and noticed this broke our build, but I quickly worked around this in 
https://github.com/google/verible/commit/3a55e85399b478bc7544d29d7ec82457b2087511

Is there a better forward-looking fix to our WORKSPACE?

shahms:
> Just FYI, we we've been building from top-of-tree in our project and noticed this broke our build, but I quickly worked around this in
> [google/verible@3a55e85](https://github.com/google/verible/commit/3a55e85399b478bc7544d29d7ec82457b2087511)
> 
> Is there a better forward-looking fix to our WORKSPACE?

I would fix to a known good release and periodically update, or commit to forward-fixing your WORKSPACE when we update dependencies.  We have a lot of external dependencies and at least attempt to update them pretty regularly.  Given how Bazel external repos work, there is a reasonable chance that any one of these updates may not be backwards compatible.

PR #4777, opened 2020-12-10
Update rules_rust to 997a8a19a6893026a94bee5d1a775c2ba3d75a42.

benjaminp:


shahms:
Looks like some failures due to unused crates and multiple versions of the same crate?

shahms:
The lint failures are fine (I'll send out a PR adding them to the ignore list so they don't trigger in the future).

shahms (approved):

PR #4785, opened 2020-12-16
fix(cxx_common): change selector serialization to work better with protos

shahms:
The resulting design is predicated on a couple of things:

1) protos may be arena allocated, which makes output parameters preferred
2) RepeatedPtrField is awkward to work with, suggesting an overload to better accommodate it.
3) Overloads and virtual member functions don't play well.
4) So update the deserialization protocol to operate on a single Any, with non-virtual members which make use of that.

I've also included tests of the serialization/deserialization both generically and for AspectArtifactSelector.  The remaining changes are due to updates in clang-format.

### DIFF
--- a/kythe/cxx/extractor/bazel_artifact_selector.cc
+++ b/kythe/cxx/extractor/bazel_artifact_selector.cc
@@ -63,13 +63,22 @@ struct FromRange {
 template <typename T>
 FromRange(const T&) -> FromRange<T>;
 
-}  // namespace
+template <typename T>
+const T& AsConstRef(const T& value) {
+  return value;
+}
 
-absl::Status BazelArtifactSelector::Deserialize(
-    absl::Span<const google::protobuf::Any> state) {
+template <typename T>
+const T& AsConstRef(const T* value) {
+  return *value;
+}
+
+template <typename T, typename U>
+absl::Status DeserializeInternal(T& selector, const U& container) {
   absl::Status error;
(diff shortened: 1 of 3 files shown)


justbuchanan (approved):

kythe/cxx/extractor/bazel_artifact_selector.cc
```diff
@@ -61,10 +61,53 @@ struct FromRange {
...
+        return status;
+      case absl::StatusCode::kFailedPrecondition:
+        error = status;
+        continue;
+      default:
+        error = status;
+        LOG(WARNING) << "Unrecognized status code: " << status;
+    }
+  }
+  return error.ok() ? absl::NotFoundError("No state found")
```
creachadair, line 95: Are both arms of this meant to report "not found"? Given the returns above I am guessing so, but it might deserve a comment to say why there are two arms here. (Something about the presence of a value to format into the diagnostic?)

shahms: Yes, as per the function documentation.  The only difference is the inclusion of the error message (conditional string formatting is more awkward here than this formulation, IMO.

creachadair: Ah, right, I didn't look at the comment for `Deserialize`, which I should have done. 

And I agree, re: the mechanism; I was just puzzled by the sameness.

PR #4786, opened 2020-12-17
feat(go_utils): allow filling in cpp-specific information

shahms:
For C++, command line arguments aren't available to Starlark.  To work around this give `write_extra_action` some additional knowledge of `CppCompileInfo` specifically to fill out `tool` `compiler_option` and `source_file` fields.

### DIFF
--- a/kythe/go/util/tools/write_extra_action/write_extra_action.go
+++ b/kythe/go/util/tools/write_extra_action/write_extra_action.go
@@ -41,8 +41,8 @@ import (
 func main() {
 	flag.Parse()
 
-	if flag.NArg() != 2 {
-		fmt.Printf("usage: %s <text_proto_file> <output_path>\n", filepath.Base(os.Args[0]))
+	if flag.NArg() < 2 || (flag.NArg() > 2 && flag.Arg(2) != "--") {
+		fmt.Printf("usage: %s <text_proto_file> <output_path> [-- <arg>...]\n", filepath.Base(os.Args[0]))
 		os.Exit(2)
 	}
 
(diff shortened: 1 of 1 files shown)


kythe/go/util/tools/write_extra_action/write_extra_action.go
```diff
@@ -80,3 +86,17 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+func populateCppCompileInfo(cc *xapb.CppCompileInfo) {
```
schroederc, line 90: Can we add a flag to pass in this as another textproto?  Using the non-flag arguments seems too specialized for this.

shahms: We cannot. That is the entire motivation for this change.

schroederc: Why can't the caller structure the arguments into a textproto if it already has them?

shahms: This only uses the trailing non-flag arguments if there is a CppCompileInfo extension provided and only after an explicit `--` delimiter, in theory allowing for some other interpretation of those arguments for a different extension.  I don't anticipate such a need as only "legacy" rules have custom protocol buffers at all and no more are likely to be added.  Extra actions being removed entirely seems more likely than new ExtraActionInfo extensions.

shahms: The caller is Bazel. It does not provide access to the stringified command line arguments, only an opaque Args object.

schroederc (approved):

PR #4789, opened 2021-01-07
chore: make bazelrc import rooted

salguarnieri:
Without this change, when you do a build from a sub-directory, it won't be able
to find the javacopts bazelrc file.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -25,7 +25,7 @@ build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java8
 build --host_javabase=@bazel_tools//tools/jdk:remote_jdk11
 
 # Import default javacopts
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -25,7 +25,7 @@ build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java8
 build --host_javabase=@bazel_tools//tools/jdk:remote_jdk11
 
 # Import default javacopts
-import tools/javacopts.bazelrc
+import %workspace%/tools/javacopts.bazelrc
```
shahms, line 28: Should the user import at the bottom be rooted as well?

salguarnieri: Nice catch. I think yes.

shahms (approved):

PR #4791, opened 2021-01-20
feat(docs): begin to document the influence relation

zrlk:


kythe/docs/influences-relation.txt
```diff
@@ -0,0 +1,108 @@
...
+
+== Variables
+
+=== Simple assignment
+
+  v := e
+
+* Push `{}`.
+* Traverse `e`.
+* Pop onto `v`.
```
shahms, line 73: Some examples of the resulting graph shape might be useful?

zrlk: PTAL

shahms (approved):
Thanks!

PR #4796, opened 2021-01-22
chore: bazel update

shahms:


schroederc (approved):

shahms:
Sigh. I can't get the equivalent of bazel-minversion test to fail locally.

creachadair:
FWIW, I noticed our macOS install instructions are out-of-date: Bazel has moved to Homebrew core and no longer requires a tap (in fact, the tap is now obsolete, as best I can tell). I will look into updating the instructions separately, but the part that is relevant to this PR is that Homebrew Bazel is still at 3.7.2. I don't know how widespread macOS use is, but if this change requires features that won't work with 3.x it may be worth waiting for the package to update.

shahms:
> FWIW, I noticed our macOS install instructions are out-of-date: Bazel has moved to Homebrew core and no longer requires a tap (in fact, the tap is now obsolete, as best I can tell). I will look into updating the instructions separately, but the part that is relevant to this PR is that Homebrew Bazel is still at 3.7.2. I don't know how widespread macOS use is, but if this change requires features that won't work with 3.x it may be worth waiting for the package to update.
> 
> Edit: I added #4798 to address the instructions.

This change only touches the maximum version allowed and doesn't change anything else.  The macOS build is currently broken and fixing it is challenging until I can get back into the office as the macOS build box is still under my desk :-)

shahms:
Due to incompatibilities in C++ action name attributes between 4.0.0 and 3.x.y and how rbe generates the configs, I don't think it's possible for the 4.0.0 rules to live in the same tree at the moment.

PR #4817, opened 2021-02-03
feat(indexing): add synchronous FDS method

schroederc:
Add simplified synchronous file data request method for a single file to
better support clients without great streaming implementations.


kythe/proto/analysis_service.proto
```diff
@@ -57,4 +57,14 @@ service FileDataService {
...
   rpc Get(FilesRequest) returns (stream FileData) {
   }
+
+  // GetFileData returns the contents a file needed for analysis.  It is the
+  // server's responsibility to do any caching necessary to make this perform
+  // well, so that an analyzer does not need to implement its own caches unless
+  // it is doing something unusual.
+  //
+  // One or both of the path and digest fields must be nonempty, otherwise an
+  // error is returned.
```
shahms, line 67: Add a note that clients should prefer the streaming method, where feasible?

schroederc: I don't think that's necessarily the case always.

shahms (approved):

PR #4823, opened 2021-02-12
feat(java_extractor): allow specifying java sources batch size for extractor

shahms:
via the KYTHE_JAVA_SOURCE_BATCH_SIZE environment variable.

From my limited testing, this appears to work and yields indexable compilation units.


justbuchanan (approved):

creachadair:
No objection to the change, but I don't think I understand what problem this variable solves. Was there a specific bug that triggered the need for it?

shahms:
> No objection to the change, but I don't think I understand what problem this variable solves. Was there a specific bug that triggered the need for it?

There's an internal bug around many-sourced compilation units.

PR #4833, opened 2021-02-17
feat(verifier): convenience functions for building predicates and ranges

zrlk:


kythe/cxx/verifier/assertions_to_souffle_test.cc
```diff
@@ -32,6 +32,27 @@ class AstTest : public ::testing::Test {
     return new (&arena_) Identifier(yy::location{}, symbol);
   }
 
+  /// \return an `AstNode` of the form App(`head`, Tuple(`values`))
+  AstNode* Pred(AstNode* head, std::initializer_list<AstNode*> values) {
```
shahms, line 36: just use `absl::Span` ?  while std::initializer_list will work for literal initializer lists and the use as a function parameter alleviates lifetime concerns, it's a weird type that's generally best relegated to the special purposes for which it was designed.

shahms (approved):

PR #4840, opened 2021-02-26
feat(schema): add ref/id to edge kinds

schroederc:


justbuchanan (approved):

justbuchanan:
Does the generated go proto need updating?

PR #4841, opened 2021-02-27
feat(typescript): returns influence functions

zrlk:


kythe/typescript/indexer.ts
```diff
@@ -809,6 +812,32 @@ class Visitor {
...
+  /** forAllInfluencers executes fn for each influencer in the active set. */
+  forAllInfluencers(fn: (influencer: VName) => void) {
+    if (this.influencers.length != 0) {
+      this.influencers[this.influencers.length - 1].forEach(fn);
+    }
+  }
+
+  /** addInfluencer adds influencer to the active set. */
+  addInfluencer(influencer: VName) {
+    if (this.influencers.length != 0) {
```
jaysachs, line 824: Should it be an error of some sort if we try to add (or forAll()) influencers if nothing is on the stack?

kythe/typescript/indexer.ts
```diff
@@ -1598,8 +1649,9 @@ class Visitor {
     let declKw;
     if (ts.isVariableDeclaration(varDecl)) {
       declKw = initializerList.kind === ts.SyntaxKind.CatchClause ?
-          '(local var)' :
-          initializerList.flags & ts.NodeFlags.Const ? 'const' : 'let';
+                                                       '(local var)' :
+          initializerList.flags & ts.NodeFlags.Const ? 'const' :
+                                                       'let';
```
jaysachs, line 1654: It'd be nice if these pure reformattings (e.g. this and lines 1830+ below) were in a separate PR.

zrlk: https://github.com/kythe/kythe/pull/4842

kythe/typescript/indexer.ts
zrlk: I explicitly wanted this to ignore cases where we weren't currently in a context that supports influence (which is currently only inside the body of a `return` statement). Otherwise, the `if (this.influencers.length...` guard would just have to be reproduced everywhere `addInfluencer` is called.

jaysachs (approved):

PR #4859, opened 2021-03-12
Close the graph columnar db iterator.

johnedmonds:
Make sure the database iterator is closed.

This is mostly indentation changes. It's much easier to review with https://github.com/kythe/kythe/pull/4859/files?w=1


shahms:
Great catch, thanks! Would it be feasible to pull this into a separate, named, function or method rather than the generically named inline closure `loopBody`?

shahms (approved):

PR #4863, opened 2021-03-17
feat(verifier): don't spend time sorting input for the fast solver

zrlk:
The existing solver depends on sorting input facts for various
reasons. As it turns out, these sorting phases take a significant
amount of the runtime of a single verification test. This PR adds
a flag (meant to eventually invoke the fast solver) that removes
the sorting phases while preserving some invariants.

In greater detail, there are two phases. In the first, the verifier
identifies anchor nodes (for an optimization that turns anchors into
a different internal data type) and file nodes (as we support reading
goals from file nodes). Sorting is necessary to group entries
according to VName (and to provide other guarantees, like the ordering
of well-known facts) that makes this process easy to encode as a
state machine. While we no longer need the optimization, we still
need to discover goals in file nodes. Unfortunately, the "text" fact
can appear on doc nodes as well as file nodes, so we need to determine
which "text" facts are actually relevant. This PR uses a simple map
from (interned) VName to file/not-file/possibly file content symbols
to that end.

There is another mapping that needs to be preserved that's used
when goals are *not* read from file nodes. This associates file content,
provided as explicit files on the command line, with file node VNames
that are discovered from entries. This method was commonly used before
we had better Bazel support for tests and is something we could consider
dropping as reading goals from file nodes makes it redundant.

The ordering is also used after the first phase to check for duplicate
entries and for conflicting facts. It may be possible to perform the
conflicting fact check efficiently with the new solver. It is less
clear whether the duplicate entry check will need to be done before
solving (e.g., by using an interned entry set).

The second sorting phase is purely to optimize common search patterns
for the old solver and can be dropped.

Unfortunately, because we can't provide entries to the fast solver
without first giving it a full program to solve, it looks like we'll
have to (in the general case, because we need to read file nodes from
entries to get goals) (1) store all entries ourselves before giving
them to the fast solver; and (2) iterate over them at least twice.

### DIFF
--- a/kythe/cxx/verifier/verifier.cc
+++ b/kythe/cxx/verifier/verifier.cc
@@ -640,6 +640,8 @@ Verifier::Verifier(bool trace_lex, bool trace_parse)
   marked_source_code_edge_id_ =
       IdentifierFor(builtin_location_, "/kythe/edge/code");
   marked_source_false_id_ = IdentifierFor(builtin_location_, "false");
+  known_file_sym_ = symbol_table_.unique();
+  known_not_file_sym_ = symbol_table_.unique();
   SetGoalCommentPrefix("//-");
 }
 
@@ -914,6 +916,11 @@ void Verifier::DumpErrorGoal(size_t group, size_t index) {
 
 bool Verifier::VerifyAllGoals(
     std::function<bool(Verifier*, const Solver::Inspection&)> inspect) {
+  if (use_fast_solver_) {
+    highest_goal_reached_ = 0;
+    highest_group_reached_ = 0;
+    return false;
+  }
   if (!PrepareDatabase()) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


kythe/cxx/verifier/verifier.cc
```diff
@@ -1051,10 +1058,56 @@ static bool EncodedFactHasValidForm(Verifier* cxt, AstNode* a) {
   }
 }
 
+Verifier::InternedVName Verifier::InternVName(AstNode* node) {
+  auto* tuple = node->AsTuple();
+  return std::make_tuple(tuple->element(0)->AsIdentifier()->symbol(),
```
shahms, line 1063: Just use braces.  `std::make_tuple` is an attractive nuisance at this point.

shahms (approved):

PR #4867, opened 2021-03-24
fix(go_indexer): fallback to method ownership by pos

schroederc:


kythe/go/indexer/indexer.go
```diff
@@ -690,10 +692,41 @@ func (pi *PackageInfo) addOwners(pkg *types.Package) {
...
+					//    -----
+					//
+					//    package other
+					//    import "pkg"
+					//                         \/
+					//    func f(a pkg.A) { a.String() }
+					//
+					// Understanding that the above reference is to
+					// (fmt.Stringer).String() is lost due to compiler unwrapping the
+					// `type A fmt.Stringer` definition for downtream compilations.
```
shahms, line 717: "downstream" (dropped an s)

kythe/go/indexer/testdata/types.go
```diff
@@ -369,6 +371,22 @@ func main() {
 	localF(i)
 }
 
+// Test the that the String method remains identical across these different embeddings.
+type StringerI fmt.Stringer
+type StringerE interface{ fmt.Stringer }
+type StringerE2 interface{ StringerI }
+
```
shahms, line 378: would Go's type aliases also fit here? 
```
type StringerA = fmt.Stringer
```

shahms (approved):

kythe/go/indexer/testdata/types.go
schroederc: Yeah, that works.

Issue #1239, opened 2018-06-25
First pass at a Kythe indexer for the Scala Language. [phab:D924]


kythe:
_Created by **fromberger** at 2016-06-11 20:58:39_: 

Note: No instructions on running yet because it is still difficult to
self-host this with bazel.  The scala actions are skylark based, not internal
so we just get a SpawnAction that we have to parse arguments from.
It actually works if you specify it in the plugins attribute on scala_library rule, but it has no
where to write its output and it is obviously the wrong way to add a
plugin.

kythe:
_**rbraunstein** wrote at 2016-06-13 17:15:06_:

  - added path fixup code

kythe:
_**shahms** commented inline (context lost during migration) at 2016-06-14 16:23:41_:

This is likely related to the --legacy_external_runfiles change.  Can you try this with 0.3.0 after a `bazel clean --expunge`?

kythe:
_**shahms** commented inline (context lost during migration) at 2016-06-14 16:23:41_:

left over comment?

kythe:
_**shahms** commented inline (context lost during migration) at 2016-06-14 16:23:41_:

Can you put this in a `sh_binary()` and depend on it as a tool?  Arguably, you should depend on the configured `jar` binary as well, rather than system one.

kythe:
_**shahms** commented inline (context lost during migration) at 2016-06-14 16:23:41_:

nit: please fix

kythe:
_**shahms** wrote at 2016-06-14 16:23:41_:

I don't know Scala, so limiting myself to BUILD, etc.

kythe:
_**rbraunstein** wrote at 2016-06-14 20:32:18_:

  - Moved rejaring code to sh_binary

kythe:
_**rbraunstein** wrote at 2016-06-14 20:33:52_:

I think I lost my local commit message to arc, here it is:


 Moved rejaring code to sh_binary
    
    The right longterm fix is to fixup the scala.bzl rules to put the xml
    file in the right place.
    
    Getting right version of jar is harder than it should be.
    Adding deps to sh_binary says it doesn't accept filegroups, but it also
    says the dep should be a filegroup.  Adding it as data puts in in the
    runfiles directory, so we add it to the genrule that calls sh_binary for
    now.
    
    Removed comments for scala_binary rule since I'm not trying to make one
    any more.  Renamed MyMain to something better in the process.

kythe:
[...]

PR #4868, opened 2021-03-25
Add Scala indexer.

johnedmonds:
This is a continuation of rbraunstein's work in https://github.com/kythe/kythe/issues/1239.

I found a copy of the old code and added some fixes and features so it works on the current Kythe version.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -40,20 +40,26 @@ maybe(
 
 # optional: setup ScalaTest toolchain and dependencies
 load("@io_bazel_rules_scala//testing:scalatest.bzl", "scalatest_repositories", "scalatest_toolchain")
+
 scalatest_repositories()
+
 scalatest_toolchain()
+
 # gazelle:repository_macro external.bzl%_go_dependencies
 load("//:external.bzl", "kythe_dependencies")
 
 kythe_dependencies()
 
 load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
-scala_config(scala_version="2.12.12")
+
+scala_config(scala_version = "2.12.12")
 
 load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repositories")
+
 scala_repositories()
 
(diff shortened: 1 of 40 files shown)


shahms:
Thanks for reviving this! I think this has languished for so long because none of the core members use or read Scala.

Is there a specific purpose for this to be in the Kythe primary repository? We don't particularly want to have code which is unmaintained by us in the repository, as we'd be unable to provide any support, guidance or updates for it.

ron-stripe:
Is there a repo or a "contrib" subdir this can go in?
I didn't see you found a way to revive it months ago. Thank you!

I'd like to be able to use it at Stripe (where we have a lot of scala folk), but I want the provenance settled first (part of googles open source policy etc).

jsachs-google:
Thanks for the ping.

I just added a kythe-contrib repository.

johnedmonds:
@jaysachs Would I just want to move this code with the same structure over to github.com/kythe/kythe-contrib (and make publically visible anything necessary from github.com/kythe/kythe)?

jaysachs:
> @jaysachs Would I just want to move this code with the same structure over to github.com/kythe/kythe-contrib (and make publically visible anything necessary from github.com/kythe/kythe)?

Conceptually yes; I'm now curious just how much would need to be opened up in the main repo. How extensive would that visibility change be?

jaysachs:
> I'm guessing we'd get away with little to no change for now. I suppose I'm mostly thinking ahead to whether there's a way to share the logic for generating JVM signatures so we can link together the graphs generated for Java<->.jar<->Scala.

Those are definitely reasonable things to want to do, but maybe tackle the sharing visibility when you work on the Java/Scala interop.

johnedmonds:
@jaysachs https://github.com/kythe/kythe/pull/5094 should be the only visibility changes needed. Sent out the PR for kythe-contrib here: https://github.com/kythe/kythe-contrib/pull/1

PR #4872, opened 2021-03-30
fix(proto): only decorate the relative type name in fields

shahms:
Some UIs have problems decorating the multi-line spans which can result from taking the type locations directly from the proto compiler.  Split qualified names and only decorate the final unscoped type name.  This also matches the behavior from most other languages.

### DIFF
--- a/kythe/cxx/indexer/proto/file_descriptor_walker.cc
+++ b/kythe/cxx/indexer/proto/file_descriptor_walker.cc
@@ -82,6 +82,19 @@ class ScopedLookup {
   const int component_;
 };
 
+absl::optional<absl::string_view> TypeName(const FieldDescriptor& field) {
+  if (field.is_map()) {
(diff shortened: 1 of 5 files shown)


kythe/cxx/indexer/proto/file_descriptor_walker.cc
```diff
@@ -358,6 +371,17 @@ void FileDescriptorWalker::VisitField(const std::string* parent_name,
     }
     const std::vector<int>& type_span = location_map_[lookup_path];
     InitializeLocation(type_span, &type_location);
+
+    // If we're in a message or enum type, decorate only the span
+    // covering the type name itself, not the full package name.
```
salguarnieri, line 376: Can you add that we are doing this so the type name doesn't span more than 1 line so it is easier for UIs to display?

salguarnieri (approved):

PR #4874, opened 2021-03-31
fix(bazel_go_extractor): include .x files in required input

shahms:
The Go compiler produces these as well as .a archive files and they contain the minimal interface information necessary for dependencies.  I'm not entirely sure of the details (and searching for information on ".x" is predictably unfruitful), but they cause issues with xrefs when omitted.


schroederc (approved):

creachadair:
Those files used to be peculiar to the internal Go tooling, but I suppose it's made its way out into the wild. Originally, the `.x` file was just the header section of the `.a` file (the other half being the pre-linked code); I'm guessing it's mostly still the same. The header formats are slightly toolchain-version sensitive, but should work with anything recent.

PR #4881, opened 2021-04-07
feat(verifier): add stub for souffle support

zrlk:


kythe/cxx/verifier/souffle_interpreter.h
```diff
@@ -17,8 +17,24 @@
 #ifndef KYTHE_CXX_VERIFIER_SOUFFLE_INTERPRETER_H_
 #define KYTHE_CXX_VERIFIER_SOUFFLE_INTERPRETER_H_
 
+#include "kythe/cxx/verifier/assertions.h"
+
 namespace kythe::verifier {
-bool RunSouffle();
+class Verifier;
```
shahms, line 23: Is there a circular dependency? otherwise just include the header.

kythe/cxx/verifier/souffle_interpreter.h
```diff
@@ -17,8 +17,24 @@
...
+/// \param database the facts to solve against
+/// \param verifier the verifier to pass to the inspection
+/// \param inspect the inspection to use
+/// \param highest_goal_reached will be set to the highest goal reached
+/// \param highest_group_reached will be set to the highest group reached
+/// \return true if all goals could be solved
+bool RunSouffle(
+    const SymbolTable& symbol_table, const std::vector<GoalGroup>& goal_groups,
+    const Database& database, Verifier* verifier,
+    std::function<bool(Verifier*, const AssertionParser::Inspection&)> inspect,
```
shahms, line 36: Can you explain this function interface more? "The inspection to use" is pretty opaque in terms of API expectations.

kythe/cxx/verifier/souffle_interpreter.h
```diff
@@ -17,8 +17,24 @@
...
+/// \param goal_groups the goal groups to solve
+/// \param database the facts to solve against
+/// \param verifier the verifier to pass to the inspection
+/// \param inspect the inspection to use
+/// \param highest_goal_reached will be set to the highest goal reached
+/// \param highest_group_reached will be set to the highest group reached
+/// \return true if all goals could be solved
+bool RunSouffle(
+    const SymbolTable& symbol_table, const std::vector<GoalGroup>& goal_groups,
+    const Database& database, Verifier* verifier,
```
shahms, line 35: if `Verifier` is non-null, use a non-const ref

kythe/cxx/verifier/souffle_interpreter.h
```diff
@@ -17,8 +17,24 @@
...
+    std::function<bool(Verifier*, const AssertionParser::Inspection&)> inspect,
+    size_t& highest_goal_reached, size_t& highest_group_reached);
```
shahms, line 37: Prefer a return value for these, something like:

struct SouffleResult {
  bool success;
  size_t highest_goal_reached;
  size_t highest_group_reached;
};

kythe/cxx/verifier/souffle_interpreter.h
zrlk: Yes, with the :lib target.

kythe/cxx/verifier/souffle_interpreter.h
zrlk: This just gets passed to `inspect` (as happens in the current code).

shahms: If the `Verifier` argument is only going to be passed to `inspect` then it can be removed from both signatures and captured by the function if necessary. (Both still seem unused, unless I need to refresh).

kythe/cxx/verifier/souffle_interpreter.h
```diff
@@ -17,8 +17,32 @@
 #ifndef KYTHE_CXX_VERIFIER_SOUFFLE_INTERPRETER_H_
 #define KYTHE_CXX_VERIFIER_SOUFFLE_INTERPRETER_H_
 
+#include "kythe/cxx/verifier/assertions.h"
+
 namespace kythe::verifier {
-bool RunSouffle();
+class Verifier;
```
shahms, line 23: Can this be deleted now?

shahms (approved):
Thanks!

PR #4883, opened 2021-04-09
feat(common): allow named captures in vnames.json rules

shahms:
Allow the use of named captures and replacements in vnames.json.  Go already had this, but the 3 languages should be consistent.


jaysachs (approved):

shahms:
Also included a C++ test.

PR #4889, opened 2021-04-16
fix(java): properly elide type arguments for constructor calls

jaysachs:


kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/GenericMethodRef.java
```diff
@@ -26,20 +27,36 @@ private static void wildcard(Optional<?> ovar) {}
...
   }
 
-  private static void constructor() {
-    //- @Optional ref/call OptionalConstructor
-    //- @Optional ref OptionalConstructor
-    //- @Optional ref OClass
+  //- @T defines/binding AbsT
+  private static <T> void constructor() {
+    //- OptC.loc/start @^"Optional"
+    //- OptC.loc/end @$"Optional"
```
shahms, line 42: This should just be `@Optional` no?

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/GenericMethodRef.java
```diff
@@ -26,20 +27,36 @@ private static void wildcard(Optional<?> ovar) {}
...
+    //- OptC ref OptionalConstructor
+    //- OptC ref OClass
     //- @GenericMethodRef ref GClass
+    //- OptCCall.loc/start @^"new Optional<GenericMethodRef>()"
+    //- OptCCall.loc/end @$"new Optional<GenericMethodRef>()"
```
shahms, line 47: `@"new Optional<GenericMethodRef>()"`

shahms (commented):
I can't recall all of the differences between the Java and C++ graph model for generics, but for C++ there are several distinct ranges and nodes for the equivalent call:
```
@T ref Abs
@"T<A, B, C>" ref/id TApp
@"T<A, B, C>" ref TAppConstructor
@"T<A, B, C>()" ref/call TAppConstructor
new T<A, B, C>();
```

As in there is a separate node for the template and the distinct instantiation of that template being constructed along with their corresponding ranges.  I think this is largely correct as far as it goes, but it does drop the ref edge to the tapp instantiation (which, I suspect, *might* be why the wildcard tests are failing).

I also vaguely recall Cody proposing some changes to the general template model a while ago, but can't recall the details.

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/GenericMethodRef.java
jaysachs: Done

kythe/javatests/com/google/devtools/kythe/analyzers/java/testdata/pkg/GenericMethodRef.java
jaysachs: Done

jaysachs:
> I can't recall all of the differences between the Java and C++ graph model for generics, but for C++ there are several distinct ranges and nodes for the equivalent call:
> 
> ```
> @T ref Abs
> @"T<A, B, C>" ref/id TApp
> @"T<A, B, C>" ref TAppConstructor
> @"T<A, B, C>()" ref/call TAppConstructor
> new T<A, B, C>();
> ```
> 
> As in there is a separate node for the template and the distinct instantiation of that template being constructed along with their corresponding ranges. I think this is largely correct as far as it goes, but it does drop the ref edge to the tapp instantiation (which, I suspect, _might_ be why the wildcard tests are failing).

Just to be clear, I don't see where this PR drops the ref edge to the tapp -- it's just changing the anchor span for the constructor reference. We can look at addressing the missing tapp in a separate CL, especially since tapps are ignored by the UI.

C++ is likely different, since there is a different constructor for each template instantiation (which is not true semantically in Java). Is there a reference from the anchor of `T`'s constructor definition to the `Abs` node above?

> I also vaguely recall Cody proposing some changes to the general template model a while ago, but can't recall the details.

This was an addendum/fix to my previous #4888, where I didn't have the tests precise enough to notice that the identifier in "getIdentifier" is not a single symbol but could be a tapp. 

Happy to take a look at the new proposed model in subsequent changes.

shahms:
> > I can't recall all of the differences between the Java and C++ graph model for generics, but for C++ there are several distinct ranges and nodes for the equivalent call:
> > ```
> > @T ref Abs
> > @"T<A, B, C>" ref/id TApp
> > @"T<A, B, C>" ref TAppConstructor
> > @"T<A, B, C>()" ref/call TAppConstructor
> > new T<A, B, C>();
> > ```
> > 
> > 
> > As in there is a separate node for the template and the distinct instantiation of that template being constructed along with their corresponding ranges. I think this is largely correct as far as it goes, but it does drop the ref edge to the tapp instantiation (which, I suspect, _might_ be why the wildcard tests are failing).
> 
> Just to be clear, I don't see where this PR drops the ref edge to the tapp -- it's just changing the anchor span for the constructor reference. We can look at addressing the missing tapp in a separate CL, especially since tapps are ignored by the UI.

"drop" implied it was emitting the ref before (which isn't the case).  Perhaps I should've said "continues to omit" :-)

> 
> C++ is likely different, since there is a different constructor for each template instantiation (which is not true semantically in Java). Is there a reference from the anchor of `T`'s constructor definition to the `Abs` node above?

Yeah, I think the Java model is simplified in a lot of ways.  There should be a ref/id from the `T` constructor definition to the enclosing `Abs` node, IIRC.

> 
> > I also vaguely recall Cody proposing some changes to the general template model a while ago, but can't recall the details.
> 
> This was an addendum/fix to my previous #4888, where I didn't have the tests precise enough to notice that the identifier in "getIdentifier" is not a single symbol but could be a tapp.
> 
> Happy to take a look at the new proposed model in subsequent changes.

PR #4895, opened 2021-04-22
Add build_id to the XRefService Reply messages

danielmoy-google:
Explicitly return a unique identifier for the underlying dataset of the server.
This allows callers to know the state for each individual request, which may
differ from machine to machine in a distributed, productionized environment.
StatusService.Status gives good base info, but runs into eventual consistency
problems during data updates to a distributed system.


jaysachs:
> Explicitly returning a build_id indicating the state of the underlying build.
> This allows callers to know the state for each individual request, which may
> differ from machine to machine in a distributed, productionized environment.
> StatusService.Status gives good base info, but runs into eventual consistency
> problems during data updates to a distributed system.

This description doesn't seem to match with the new field. It's not the "state" of the underlying build, it's an identifier for the build (that could be used to join against some external data sources to determine the state).

danielmoy-google:
> This description doesn't seem to match with the new field. It's not the "state" of the underlying build, it's an identifier for the build (that could be used to join against some external data sources to determine the state).

Okay, updated description

kythe/proto/xref.proto
```diff
@@ -265,6 +265,9 @@ message DecorationsReply {
   // references to the list of their overrides.
   map<string, Overrides> extends_overrides = 17;
 
+  // A unique build_id for the underlying dataset serving this reply.
```
jaysachs, line 268: s/build_id/identifier/ in the doc comments here and elsewhere.

jaysachs (approved):

kythe/proto/xref.proto
danielmoy-google: done

PR #4905, opened 2021-05-04
fix: force selection of `ar` to unbreak os x build of libffi

jaysachs:
I'm not clear why a `configure` in a checkout of libffi finds `ar` properly but running `configure` inside bazel ends up choosing `libtool`, but a blunt hammer seems to work on linux as wel.


justbuchanan (approved):

zrlk:
This works on my osx machine.

PR #4907, opened 2021-05-05
fix(cxx_indexer): refactor lvalue handling for dataflow

zrlk:
Tag particular AST nodes as writes. This is cleaner than the previous
approach, which tries to scramble up the tree to figure out if a node
is in a write context. This PR also unifies lvalue interpretation
and influence target identification.

We can probably turn dataflow edges on by default, but that will be
a separate PR.

### DIFF
--- a/kythe/cxx/indexer/cxx/clang_utils.h
+++ b/kythe/cxx/indexer/cxx/clang_utils.h
@@ -35,10 +35,15 @@ const clang::Decl* FindSpecializedTemplate(const clang::Decl* decl);
 /// \return true if a reference to `decl` should be given blame context.
 bool ShouldHaveBlameContext(const clang::Decl* decl);
 
-/// \return true if `stmt` is being used in a write position according to
-/// `map`.
-bool IsUsedAsWrite(const IndexedParentMap& map, const clang::Stmt* stmt);
+/// \return the `Stmt` that is at the lvalue root position of `stmt`, or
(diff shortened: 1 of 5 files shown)


kythe/cxx/indexer/cxx/clang_utils.h
```diff
@@ -35,10 +35,15 @@ const clang::Decl* FindSpecializedTemplate(const clang::Decl* decl);
...
 bool ShouldHaveBlameContext(const clang::Decl* decl);
 
-/// \return true if `stmt` is being used in a write position according to
-/// `map`.
-bool IsUsedAsWrite(const IndexedParentMap& map, const clang::Stmt* stmt);
-
+/// \return the `Stmt` that is at the lvalue root position of `stmt`, or
+/// null otherwise. For example, in `foo[x].bar(y).z`, the member expression
+/// for `z` is in the root position.
+const clang::Stmt* FindLValueRoot(const clang::Stmt* stmt);
```
jaysachs, line 41: Hmm, are these roots or leaves? Or neither: maybe "principal" or "primary" or "proximate" or something like that?

jaysachs (approved):

kythe/cxx/indexer/cxx/clang_utils.h
zrlk: I've renamed it to `head`. Lvalues are weird in that the address you eventually get to is the most important thing. You could imagine it as the root of an inverted tree, but this is confusing.

PR #4908, opened 2021-05-05
feat(cxx_indexer): support ++ and -- for influence and r/w refs

zrlk:


kythe/cxx/indexer/cxx/testdata/df/df_unary.cc
```diff
@@ -0,0 +1,17 @@
+// Unary {post,pre}{inc,dec}rements are included in the influence graph.
+void f() {
+    int x;
+    //- @x ref VarX
+    //- @x ref/writes VarX
```
jaysachs, line 5: How about a negative test proving we don't add ref/writes to every ref?

shahms (approved):

PR #4928, opened 2021-05-14
fix(cxx_indexer): assign usr nodes to the empty corpus

justbuchanan:


shahms:
clang-format wants two spaces before the comment.

It's also unsettled whether or not we want to keep an empty corpus for USR nodes or not (my vote would be "not"). @zrlk ?

zrlk:
The way to spell a USR node's vname is part of our public interface. I think we should restore the indexer to a working state (which this PR does), *then* have a conversation about whether to change the interface.

zrlk (approved):

shahms (approved):

PR #4929, opened 2021-05-14
fix(cxx_indexer): add required parameter

justbuchanan:
this seems to have been set in google3 and was undone on import, breaking the build


kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -695,7 +695,7 @@ void MarkedSourceGenerator::ReplaceMarkedSourceWithTemplateArgumentList(
     std::string pre_text;
     {
       llvm::raw_string_ostream stream(pre_text);
-      print_arg.print(policy, stream);
```
shahms, line 698: This requires a new version of LLVM

justbuchanan: ahh makes sense. I'll close this PR and we can fix it when we upgrade LLVM

PR #4936, opened 2021-05-19
fix(tool): make "decors" an alias for "decor" in kythe tool

jaysachs:


kythe/go/services/cli/cli.go
```diff
@@ -75,7 +75,8 @@ func Execute(ctx context.Context, api API) subcommands.ExitStatus {
 	RegisterCommand(&identCommand{}, "")
 	RegisterCommand(&lsCommand{}, "")
 
-	RegisterCommand(&decorCommand{}, "xrefs")
+	RegisterCommand(&decorCommand{name: "decor"}, "xrefs")
+	RegisterCommand(&decorCommand{name: "decors"}, "xrefs")
```
shahms, line 79: Won't this list the command twice via `help`?

jaysachs: Yes. I suppose I could just tackle the TODO at

https://cs.opensource.google/kythe/kythe/+/master:kythe/go/services/cli/cli.go;l=93;drc=036f2b6793100dcc16f499b529350c08eddbff21

but I can't count how many times I type "decors" instead of "decor". I suppose I can write a shell function to deal with this for me in the short term.

PR #4940, opened 2021-05-20
fix(rust_indexer): update to a recent nightly toolchain

zrlk:


shahms (approved):
Man it's annoying that craze encodes the crate version in the repo name.

PR #4944, opened 2021-05-22
Refactor TS indexer unit tests.

nbeloglazov:
1. Introduce TestCase class. This improves code clarity. And helps to prevents issues like currently
where we for each group of files in a directory (single test case) we run tests N times, where N -
number of files in that directory instead of running test one.

2. Pass pathVNames to indexer. This is currently doesn't affect test run but it provides more
complete test setup.


nbeloglazov:
@blickly to take look for Typescript

blickly (approved):
LGTM

shahms (approved):

PR #4949, opened 2021-05-28
feat(rust_indexer): generate and check in rust protos

zrlk:


kythe/proto/generate_protobufs.py
```diff
@@ -35,47 +38,69 @@
 import stat
 import sys
 
+if len(sys.argv) > 2:
```
creachadair, line 41: I'd like to put in an appeal for using a flag-parsing library (e.g., [argparse](https://docs.python.org/3/library/argparse.html#module-argparse)) rather than bodging the arguments directly. I realize there's only one right now, but now that we've broken the seal, there will probably be more to come.

shahms: argparse is in the Python standard library and is available in both Python 3 and Python 2. It is definitely preferable to ad hoc flag parsing.

.pre-commit-config.yaml
```diff
@@ -38,7 +38,8 @@ repos:
             |(^|/)(\.[^/]*|go\.mod|go\.sum)$
             |^tools/|^third_party/|^kythe/web/site/
             |worker/info/(admin|host)$
-            |/project\.clj$|/style\.css$|bazelrc$|/Doxyfile\.kythe$|/docs/llvm-2014/Makefile$'
+            |/project\.clj$|/style\.css$|bazelrc$|/Doxyfile\.kythe$|/docs/llvm-2014/Makefile$
+            |^/kythe/proto/.*_rust_proto/'
```
shahms: I think remove the leading slash here?

`|^kythe/proto/[^/]*_rust_proto/'`

kythe/proto/generate_protobufs.py
```diff
@@ -35,47 +39,64 @@
 import stat
 import sys
 
+parser = argparse.ArgumentParser(description='Generate Go or Rust protobufs')
+parser.add_argument('--rust', dest='lang', action='store_const', const='rust',
+                    default='go', help='generate Rust instead of Go')
```
shahms, line 44: Any particular reason for this not to be a list such that it generates both Rust and Go?

The odds are good that any change to the protos will necessitate a change to both Rust and Go generated protos so being able to do so in a single command would be nice.

`parser.add_argument('--languages', default = ['go', 'rust'], etc.)` ?

zrlk: I just dropped the option entirely.

shahms (approved):

shahms:
Er, easy enough to fix the lint errors :-)

PR #4967, opened 2021-06-15
Added correct underline syntax to doc.

Bickor:
Updated ascii syntax to correct underline format so it appears of kythe site.


salguarnieri:
Nit: use [commitlint](https://github.com/conventional-changelog/commitlint) format for commit messages. For example, this could probably be `fix: Added correct underline syntax`

salguarnieri:
It looks like something happened during one of your merges that upset the CLA check. It might be best to squash to a single commit and then force push.

salguarnieri (approved):

PR #4974, opened 2021-06-17
feat(rust_indexer): Get source file contents from the FileProvider

wcalandro:
Previously, the indexer binary would extract an entire kzip to the file system, and the library would read those source files from the file system. This PR makes the indexer library request source files from a FileProvider instead.


jaysachs (commented):
Can you separate the reformatting into its own PR?

wcalandro:
@jaysachs The linter was complaining about how some of the rust code was formatted. I can try to isolate that the files it was explicitly complaining about and only include those in the PR

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -87,26 +88,29 @@ impl<'a> UnitAnalyzer<'a> {
     pub fn new(
         unit: &'a CompilationUnit,
         writer: &'a mut dyn KytheWriter,
-        root_dir: &'a Path,
+        provider: &'a mut dyn FileProvider,
     ) -> Self {
         // Create a HashMap between the file path and the VName which we can retrieve
-        // later to emit nodes
+        // later to emit nodes and create a HashMap between a file path and it's digest
```
zrlk, line 94: nit: its

zrlk (approved):

PR #4983, opened 2021-07-12
feat(proxy): Add support for transmitting protos in wire format

wcalandro:
The protobuf crate for Rust currently does not support marshalling protobufs to JSON. This PR adds two new commands to the proxy: one to request a CompilationUnit in wire format, and one to send entries back to the proxy in wire format.


kythe/go/platform/analysis/proxy/proxy.go
```diff
@@ -265,6 +325,26 @@ func decodeEntries(jsonArray json.RawMessage) ([]*spb.Entry, error) {
...
+		return nil, err
+	}
+	entries := make([]*spb.Entry, len(messages))
+	for i, msg := range messages {
+		var e spb.Entry
+		b, err := base64.StdEncoding.DecodeString(msg)
+		if err != nil {
+			return nil, err
+		}
+		if err := (proto.UnmarshalOptions{}.Unmarshal(b, &e)); err != nil {
```
shahms, line 340: outer parens are unnecessary

wcalandro:
The compiler throws an error if I don't have them, saying that there isn't a boolean on the left side.
  
  
  

  
  
>

shahms (approved):

kythe/go/platform/analysis/proxy/proxy.go
shahms: Oh, it sees the brace not as a type, but as the opening of the conditional body.

creachadair:
> The compiler throws an error if I don't have them, saying that there isn't a boolean on the left side.

This is a corner case in the grammar: When a struct literal appears at the end of the `if` initializer, the parentheses are necessary to disambiguate between the curly brackets ending the condition, and the brackets belonging to the initializer.

See the discussion about parsing ambiguity under https://golang.org/ref/spec#Composite_literals

kythe/go/platform/analysis/proxy/proxy.go
creachadair: Just so. Another way to write it here, without the parentheses, would be:

```go
var opts proto.UnmarshalOptions
if err := opts.Unmarshal(b, &e); err != nil {
```

That also has the advantage of being a little more readable if other options were added later, but of course they are equivalent in effect.

kythe/go/platform/analysis/proxy/proxy.go
```diff
@@ -110,13 +122,20 @@ type response struct {
...
+// A unit represents a compilation unit (in JSON form) to be analyzed.
 type unit struct {
 	Unit            json.RawMessage `json:"unit"`
 	Revision        string          `json:"rev,omitempty"`
 	FileDataService string          `json:"fds,omitempty"`
 }
 
+// A unit represents a compilation unit (in base64-encoded wire format) to be analyzed.
+type unit_wire struct {
+	Unit            string `json:"unit"`
```
creachadair, line 134: Rather than making this a `json.RawMessage` and explicitly encoding, I recommend making this field a plain `[]byte`. The JSON encoder will already encode byte slices as a (JSON) string encoded in Base64 (see https://pkg.go.dev/encoding/json).

wcalandro: Thanks for pointing this out! I have updated the PR based on your suggestion

kythe/go/platform/analysis/proxy/proxy.go
```diff
@@ -110,13 +122,20 @@ type response struct {
...
-// A unit represents a compilation unit to be analyzed.
+// A unit represents a compilation unit (in JSON form) to be analyzed.
 type unit struct {
 	Unit            json.RawMessage `json:"unit"`
 	Revision        string          `json:"rev,omitempty"`
 	FileDataService string          `json:"fds,omitempty"`
 }
 
+// A unit represents a compilation unit (in base64-encoded wire format) to be analyzed.
+type unit_wire struct {
```
creachadair, line 132: Names in Go should be `mixedCase`, not `snake_case`, for multiple words. See: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps (or the internal analogue).

wcalandro: Ah good catch!

PR #4984, opened 2021-07-14
feat(cxx_common): properly index generated protos

NickGonzalez:
Generated protos ending in .gen.proto are now properly indexed. Cxx indexer creates vnames
for generated protos that match the vnames created by the proto extractor. Added tests,
updated build rules and created two new entires in vnames.json.


kythe/cxx/common/protobuf_metadata_file.cc
```diff
@@ -36,6 +36,9 @@ proto::VName ProtobufMetadataSupport::VNameForAnnotation(
     out.set_corpus(context_vname.corpus());
   }
   out.set_path(annotation.source_file());
+  if (absl::EndsWith(annotation.source_file(), ".gen.proto")) {
+    out.set_root(context_vname.root());
+  }
```
shahms, line 41: I'm hesitant to hard-code more file extension-specific behavior in the indexer.  Regardless, this should come in the same block as setting the corpus (as in, it should only happen if looking up the annotation-provided source file VName directly fails).

shahms: Notably, because `context_vname` is generated from the name of the metadata file, which will always be "generated", it's possible (though unlikely, currently) that the root of the generated proto will differ from the root used here.

NickGonzalez: Moved set_root to the same block as setting the corpus.

NickGonzalez:
This change allows for generated proto files with the .gen.proto extension to be properly indexed.  There is a followup pull request coming that adds metadata support for generated protos. 

If you prefer, I can combine these two changes into one pull request. Just let me know.

Also, I can't get the presubmit checks to pass when pushing to origin.  The bazel tests are failing with "no such package '@debian_java_deps//': ..  Is this expected? I did a git push --no-verify.

shahms:
> This change allows for generated proto files with the .gen.proto extension to be properly indexed. There is a followup pull request coming that adds metadata support for generated protos.
> 
> If you prefer, I can combine these two changes into one pull request. Just let me know.
> 
> Also, I can't get the presubmit checks to pass when pushing to origin. The bazel tests are failing with "no such package '@debian_java_deps//': .. Is this expected? I did a git push --no-verify.

No, the tests should pass.  This won't be mergeable if the tests fail in the presubmit.

shahms (approved):

tools/build_rules/verifier_test/cc_indexer_test.bzl
```diff
@@ -180,22 +180,29 @@ def _transitive_entries(deps):
             compressed += dep[KytheEntries].compressed
     return KytheEntries(compressed = depset(transitive = compressed), files = depset(transitive = files))
 
+def _fix_path_for_generated_file(path):
```
zrlk, line 183: Could you explain why you need to do this? My concern is that we will also need to change the "real" build implementation to do something similar.

NickGonzalez: The initial problem is that when building the command line parameters for the protoc compiler, the call to paths._relativize fails with the following error:

Path 'bazel-out/k8-fastbuild/bin/kythe/testdata/indexers/proto/_virtual_imports/testdata5_proto/kythe/testdata/indexers/proto/gen/testdata5.gen.proto' is not beneath 'kythe/testdata/indexers/proto'.

If we manage to get past this error, the next problem we run into is in the protoc compiler itself.  When constructing the GeneratedSourceInfo message used to generate the metadata, the protoc compiler uses  path() in the FileDescriptor for the proto being compiled, in this case testdata5.gen.proto in the _virtual_imports directory, which results in metadata that references a file with _virtual_imports in the path. This becomes a problem downstream because the proto paths in the cxx index don't match the paths in the proto index, so we lose the "generates" relationship. 

By stripping everything up to "kythe/testdata/..." we get past the relativize issue, and more importantly the paths in the proto index and cxx index match, all we have to do now is make sure the vnames that are constructed have the correct root that resolve to the generated files directory.  For that we use the .gen.proto file extension.

A similar issue with _virtual_imports was identified and fixed in https://github.com/bazelbuild/bazel/commit/f65cc726655eb81c6e0aeb04d05b42831640dce9.

zrlk (approved):

PR #4985, opened 2021-07-15
feat(rust_indexer): Add support for communicating with the analyzer driver

wcalandro:
This PR adds support to the Rust indexer to communicate with the analyzer driver. This includes support within the indexer library and a new binary to continuously process compilation units.


shahms:
You probably want to setup the git commit hooks (from https://www.kythe.io/contributing/) so you can get the lint warnings earlier.

jaysachs (commented):
So bear with me, I've only spent 30 minutes reading RustByExample.

kythe/rust/indexer/src/bin/proxy/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    let mut kythe_writer = ProxyWriter::new();
+    let mut indexer = KytheIndexer::new(&mut kythe_writer);
+
+    // Request and process
+    loop {
+        let unit = request_compilation_unit()?;
+        // Retrieve the save_analysis file
+        let temp_dir = TempDir::new("rust_indexer").context("Couldn't create temporary directory")?;
+        let temp_path = PathBuf::new().join(temp_dir.path());
+        write_analysis_to_directory(&unit, &temp_path, &mut file_provider)?;
```
jaysachs, line 38: So if this fails & returns (or the `TempDir::new`) we don't write "done" back to the proxy, and so the conversation will be out of sync.

wcalandro:
Since this is a fatal error, shouldn't the indexer simply die, which will be detected by the proxy?
  
  

  
  
>

kythe/rust/indexer/src/bin/proxy/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    io::stdin()
+        .read_line(&mut response_string)
+        .context("Failed to read response from proxy")?;
+
+    // Convert to json and extract information
+    let response: Value = serde_json::from_str(&response_string)
+        .context("Failed to parse response as JSON")?;
+    if response["rsp"].as_str().unwrap() == "ok" {
+        let args = response["args"].clone();
+        let unit_base64 = args["unit"].as_str().unwrap();
```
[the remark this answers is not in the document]
jaysachs: > Since this is a fatal error, shouldn't the indexer simply die, which will be detected by the proxy?

Yes, that's true. I'm just confirming that if there's no "unit" field, then the indexer is right to decide "the proxy is misbehaving so I should panic". That does seem to make sense.

kythe/rust/indexer/src/writer.rs
```diff
@@ -79,6 +80,76 @@ impl<'a> KytheWriter for CodedOutputStreamWriter<'a> {
...
+        Ok(())
+    }
+
+    fn flush(&mut self) -> Result<(), KytheError> {
+        // Combine all of the entries
+        let buffer_json_array = self.buffer.join(", ");
+        self.buffer.clear();
+
+        // Write the proxy command to output
+        println!(r#"{{"req":"output_wire","args":[{}]}}"#, buffer_json_array);
```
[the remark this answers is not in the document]
wcalandro: It won't automatically flush and the error isn't swallowed, it panics. I can change this to call the function directly so we can catch the error ourselves and do what we want with it.

kythe/rust/indexer/src/providers.rs
```diff
@@ -139,3 +140,72 @@ impl FileProvider for KzipFileProvider {
...
+        // Read the response from the proxy
+        let mut response_string = String::new();
+        io::stdin()
+            .read_line(&mut response_string)
+            .map_err(|err| KytheError::IndexerError(format!("Failed to read proxy file response from stdin: {:?}", err)))?;
+
+        // Convert to json and extract information
+        let response: Value = serde_json::from_str(&response_string)
+            .map_err(|err| KytheError::IndexerError(format!("Failed to read proxy file response from stdin: {:?}", err)))?;
+        if response["rsp"].as_str().unwrap() == "error" {
```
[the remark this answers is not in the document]
wcalandro: I'd be fine with caching the existence data. Maybe limited in size?

kythe/rust/indexer/src/providers.rs
```diff
@@ -139,3 +140,72 @@ impl FileProvider for KzipFileProvider {
...
+            // The file couldn't be found
```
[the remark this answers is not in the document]
wcalandro: Is it wise to cache file data in RAM? I'm worried about the possibility of OOM'ing

kythe/rust/indexer/src/bin/proxy/main.rs
wcalandro: Gotcha, I'll change the structure of how this all works

kythe/rust/indexer/src/bin/proxy/main.rs
```diff
@@ -0,0 +1,122 @@
...
+    let mut indexer = KytheIndexer::new(&mut kythe_writer);
+
+    // Request and process
+    loop {
+        let unit = request_compilation_unit()?;
+        // Retrieve the save_analysis file
+        let temp_dir = TempDir::new("rust_indexer").context("Couldn't create temporary directory")?;
+        let temp_path = PathBuf::new().join(temp_dir.path());
+        if let Err(err) = write_analysis_to_directory(&unit, &temp_path, &mut file_provider) {
+            println!(r#"{{"req":"done", "args"{{"ok":"false","msg":"{:?}"}}}}"#, err);
```
jaysachs, line 39: Don't we need to get the proxy response (i.e. lines 54-57 below)?

wcalandro: That's correct. Fixed

kythe/rust/indexer/src/bin/proxy/main.rs
```diff
@@ -0,0 +1,126 @@
...
+            let mut response_string = String::new();
```
jaysachs, line 40: So rather than duplicating that here, what about factoring out a `finish_unit` function, to which you pass a `Result<()>`, and then calling it here on error, and then also below via `let res = indexer.index_cu(...); finish_unit(res);` ?

wcalandro: Done

kythe/rust/indexer/src/bin/proxy/main.rs
```diff
@@ -0,0 +1,147 @@
...
+        let temp_dir = TempDir::new("rust_indexer").context("Couldn't create temporary directory")?;
+        let temp_path = PathBuf::new().join(temp_dir.path());
+        match finish_unit(write_analysis_to_directory(&unit, &temp_path, &mut file_provider), true) {
+            Ok(true) => {},
+            Ok(false) => { continue; },
+            Err(err) => { return Err(err); },
+        };
+
+        // Index the CompilationUnit and let the proxy know we are done
+        finish_unit(indexer.index_cu(&unit, &temp_path, &mut file_provider), false)?;
```
jaysachs, line 50: Does it make sense to pull this into the match and get rid of the continue there?

jaysachs (approved):

kythe/rust/indexer/src/bin/proxy/main.rs
wcalandro: I refactored the function to clean up this logic

PR #4986, opened 2021-07-20
feat(extraction): accept "arguments" in addition to "command" in compdb

afq984:
[some](https://github.com/rizsotto/Bear) tools generate "arguments" instead of "command",
both are valid according to https://clang.llvm.org/docs/JSONCompilationDatabase.html

### DIFF
--- a/kythe/go/extractors/config/runextractor/compdb/compdb.go
+++ b/kythe/go/extractors/config/runextractor/compdb/compdb.go
@@ -39,8 +39,23 @@ import (
 // A compileCommand holds the decoded arguments of a LLVM compilation database
 // JSON command spec.
 type compileCommand struct {
+	Arguments []string
 	Command   string
 	Directory string
+}
+
+func (cc *compileCommand) asCommand() string {
+	if len(cc.Arguments) > 0 {
+		return shell.Join(cc.Arguments)
+	}
+	return cc.Command
+}
+
+func (cc *compileCommand) asArguments() ([]string, bool) {
+	if len(cc.Arguments) > 0 {
+		return cc.Arguments, true
+	}
+	return shell.Split(cc.Command)
 }
 
 // ExtractOptions holds additional options related to compilation DB extraction.
@@ -77,7 +92,7 @@ func ExtractCompilations(ctx context.Context, extractor, path string, opts *Extr
 			if err := extractOne(ctx, extractor, entry, env, opts); err != nil {
 				// Log error, but continue processing other compilations.
 				atomic.AddUint64(&failCount, 1)
-				log.Printf("Error extracting compilation with command '%s': %v", entry.Command, err)
+				log.Printf("Error extracting compilation with command '%s': %v", entry.asCommand(), err)
 			}
 		}(entry)
 	}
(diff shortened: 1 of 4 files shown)


kythe/go/extractors/config/runextractor/compdb/compdb_test.go
```diff
@@ -38,10 +38,7 @@ const (
 func setupEnvironment(t *testing.T) string {
 	t.Helper()
 	// TODO(shahms): ExtractCompilations should take an output path.
-	output := os.Getenv("TEST_TMPDIR")
-	if output == "" {
-		t.Skip("Skipping test due to incompatible environment (missing TEST_TMPDIR)")
-	}
+	output := t.TempDir()
```
afq984, line 41: Changing this because otherwise the second run of testExtractCompilationsEndToEndWithDatabase will try to overwrite a generated file and fail.

shahms: This works (and is documented to do so by Bazel as well).

kythe/go/extractors/config/runextractor/compdb/compdb.go
```diff
@@ -39,10 +39,25 @@ import (
...
 }
 
+func (cc *compileCommand) asCommand() string {
+	if len(cc.Arguments) > 0 {
+		return shell.Join(cc.Arguments)
+	}
+	return cc.Command
+}
+
+func (cc *compileCommand) asArguments() (args []string, ok bool) {
```
shahms, line 54: The return values are obvious from context without the names and should likely be omitted.

afq984: Thanks for reviewing, removed the names.

shahms (approved):
/gcbrun

afq984:
rebased from master

shahms:
It looks like this causes `go test -mod=readonly ./kythe/...` to fail:

```
--- FAIL: TestExtractCompilationsEndToEnd (0.00s)
    --- FAIL: TestExtractCompilationsEndToEnd/command (0.00s)
        compdb_test.go:62: Unable to change working directory: chdir /workspace/kythe/go/extractors/config/runextractor/compdb/kythe/go/extractors/config/runextractor/compdb/testdata: no such file or directory
    --- FAIL: TestExtractCompilationsEndToEnd/arguments (0.00s)
        compdb_test.go:62: Unable to change working directory: chdir /workspace/kythe/go/extractors/config/runextractor/compdb/kythe/go/extractors/config/runextractor/compdb/testdata: no such file or directory
FAIL
```

afq984:
Looks like when tested by go test `testPath` and `extractorPath` becomes invalid as go [runs the test in the package directory](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/cmd/go/internal/test/test.go;l=365-368;drc=953d1feca9b21af075ad5fc8a3dad096d3ccc3a0).

If we fix the paths, it'd still depend on cxx_executor being available somewhere.
Maybe we can continue to skip it in `go test` based on the presence of TEST_TMPDIR or other conditions?

shahms:
> Looks like when tested by go test `testPath` and `extractorPath` becomes invalid as go [runs the test in the package directory](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/cmd/go/internal/test/test.go;l=365-368;drc=953d1feca9b21af075ad5fc8a3dad096d3ccc3a0).
> 
> If we fix the paths, it'd still depend on cxx_executor being available somewhere.
> Maybe we can continue to skip it in `go test` based on the presence of TEST_TMPDIR or other conditions?

Skipping based on `TEST_TMPDIR` works for me.

PR #4990, opened 2021-07-21
feat(rust_extractor): remove dependency on rustc_driver to avoid dylink

wcalandro:
Remove dependency on a dynamic link to rustc_driver

### DIFF
--- a/kythe/rust/extractor/src/lib.rs
+++ b/kythe/rust/extractor/src/lib.rs
@@ -13,14 +13,21 @@
 // limitations under the License.
 #![feature(rustc_private)]
 
-extern crate rustc_driver;
+#[macro_use]
+extern crate anyhow;
+
+extern crate rustc_error_codes;
+extern crate rustc_errors;
 extern crate rustc_interface;
 extern crate rustc_save_analysis;
 extern crate rustc_session;
 
-use rustc_driver::{Callbacks, Compilation, RunCompiler};
-use rustc_interface::{interface, Queries};
+use anyhow::{Context, Result};
+use rustc_errors::registry::Registry;
+use rustc_interface::interface;
 use rustc_save_analysis::DumpHandler;
+use rustc_session::config::{self, Input};
+use rustc_session::{getopts, DiagnosticOutput};
 use std::path::PathBuf;
 
 /// Generate a save_analysis in `output_dir`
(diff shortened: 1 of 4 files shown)


wcalandro:
Looks like still might have a dynamic link dependency on `libLLVM-12-rust-1.54.0-nightly.so`

kythe/rust/extractor/src/lib.rs
```diff
@@ -29,75 +36,81 @@ use std::path::PathBuf;
...
+        return Err(anyhow!("The first argument must be an empty string"));
     }
 
-    let mut callback_shim = CallbackShim::new(output_dir);
+    // Generate the configuration to parse the options
+    let mut options = getopts::Options::new();
+    for option in config::rustc_optgroups() {
+        (option.apply)(&mut options);
+    }
+    let matches = options.parse(&rustc_arguments[1..]).context("Failed to parse options")?;
```
jaysachs, line 51: Are we going to need this same or very similar logic for this creating of the `config` and running the compiler for the indexer as well?

wcalandro: No, because the indexer does not run the compiler

kythe/rust/extractor/src/lib.rs
```diff
@@ -29,75 +36,81 @@ use std::path::PathBuf;
 /// first element must be an empty string.
```
jaysachs, line 36: Simply stating this doesn't feel like enough explanation for what seems like a strange arbitrary constraint. What's the "why" for this?

kythe/rust/extractor/src/lib.rs
```diff
@@ -29,75 +36,81 @@ use std::path::PathBuf;
...
-        // falsely claims that there is a mismatch between rustc_save_analysis's
-        // `rls_data::config::Config` and ours, even though we use the same version.
-        // This forces us to use the environment variable method of configuration
-        // instead.
-        std::env::set_var(
-            "RUST_SAVE_ANALYSIS_CONFIG",
-            r#"{"output_file":null,"full_docs":true,"pub_only":false,"reachable_only":false,"distro_crate":false,"signatures":false,"borrow_data":false}"#,
-        );
+            std::env::set_var(
+                "RUST_SAVE_ANALYSIS_CONFIG",
```
jaysachs, line 93: I'm not fully following the flow here, but this doesn't need to happen *before* `run_compiler` is invoked?

kythe/rust/extractor/src/lib.rs
wcalandro: The first argument always gets ignored, which is why it should be an empty string. But I agree, this comment should be better.

kythe/rust/extractor/src/lib.rs
wcalandro: This environment variable isn't read until we actually process the crate with the save analysis crate, so we set it right before we do the analysis.

jaysachs (approved):

wcalandro:
This change is no longer the best way forward as it introduces other dynamically linked dependencies. Closing for now

PR #4991, opened 2021-07-21
chore: fix cloudbuild rustup installation

shahms:


wcalandro (approved):

wcalandro:
Ah good find! I didn't realize that would be an issue

PR #4996, opened 2021-07-28
fix(rust_extractor): change argument processing due to rules_rust change

wcalandro:
Fixes a bug in the Rust extractor caused by changes in the arguments passed to the rules_rust `process_wrapper` binary

### DIFF
--- a/kythe/extractors/BUILD
+++ b/kythe/extractors/BUILD
@@ -198,6 +198,7 @@ extractor_action(
     data = ["//external:vnames_config"],
     env = {
         "LD_LIBRARY_PATH": "external/rust_linux_x86_64/lib:external/rust_darwin_x86_64/lib",
+        "KYTHE_CORPUS": "kythe.io",
     },
     extractor = "@io_kythe//kythe/rust/extractor",
     mnemonics = [
--- a/kythe/extractors/BUILD
+++ b/kythe/extractors/BUILD
@@ -198,6 +198,7 @@ extractor_action(
     data = ["//external:vnames_config"],
     env = {
         "LD_LIBRARY_PATH": "external/rust_linux_x86_64/lib:external/rust_darwin_x86_64/lib",
+        # TODO(wcalandro): Remove this once the extractor has support for VName configuration
         "KYTHE_CORPUS": "kythe.io",
     },
     extractor = "@io_kythe//kythe/rust/extractor",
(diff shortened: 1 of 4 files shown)


kythe/extractors/BUILD
```diff
@@ -198,6 +198,7 @@ extractor_action(
     data = ["//external:vnames_config"],
     env = {
         "LD_LIBRARY_PATH": "external/rust_linux_x86_64/lib:external/rust_darwin_x86_64/lib",
+        "KYTHE_CORPUS": "kythe.io",
```
jaysachs, line 201: Why is this needed?

kythe/rust/extractor/src/bin/save_analysis.rs
```diff
@@ -38,10 +38,12 @@ fn generate_arguments(arguments: Vec<String>, output_dir: &Path) -> Result<Vec<S
         .position(|arg| arg == "--")
         .ok_or_else(|| anyhow!("Could not find the start of the rustc arguments"))?;
 
-    // Keep the "--" argument and replace it with an empty string because
+    // The rust compiler executable path is the argument directly after "--"
+    // We keep the path argument and  replace it with an empty string because
     // `kythe_rust_extractor::generate_analysis` requires the first argument in
     // `arguments` to be an empty string
-    let mut rustc_arguments = arguments.split_at(argument_position).1.to_vec();
+    let mut rustc_arguments = arguments.split_at(argument_position + 1).1.to_vec();
```
jaysachs, line 45: So it's not clear to me where this logic belongs. kythe_rust_extractor::generate_analysis could be changed to do this, moving the oddity closer to where it's needed. Why do it at the extractor?

kythe/extractors/BUILD
wcalandro: The extractor requires the "KYTHE_CORPUS" environment variable to be set. I don't think this was caught previously because the extractor hasn't been run on the Kythe codebase in a while. Another option is to simply set a default corpus in the extractor.

kythe/rust/extractor/src/bin/save_analysis.rs
wcalandro: The main reason this was broken out was to keep the library build-system agnostic agnostic, with binary clients performing the pre-processing of the Rust compiler arguments. Initially this was done to make it easier for Fuchsia's extractor, but they seemingly don't pull in the extractor library anyways. I can definitely move this logic into the library if you believe that's the best course of action.

jaysachs: It's not clear that either place is better, since they're both in the extractor.

jaysachs (approved):

PR #5001, opened 2021-07-30
Load inline metadata

sophie-google:
Modify the Kythe Golang Indexer to add support for inline base64 encoded GeneratedCodeInfo protos specified in a generated Golang file.

These changes allow generated Go files with inline metadata to have Kythe edges to their source files.


shahms:
Please run `gofmt` and `buildifier` on the modified files.

kythe/proto/BUILD
```diff
@@ -681,6 +681,11 @@ proto_library(
     deps = [":storage_proto"],
 )
 
+go_kythe_proto(
+	proto = ":metadata_proto",
```
shahms, line 685: This will require running `kythe/proto/generate_protobufs.py`

kythe/go/indexer/indexer.go
```diff
@@ -57,6 +58,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/tools/go/gcexportdata"
 
+	mpb "kythe.io/kythe/proto/metadata_go_proto"
```
shahms, line 61: This is likely going to require running `go test -mod=mod ./kythe/...` to update the go.mod file.

kythe/go/indexer/indexer.go
```diff
@@ -853,6 +899,20 @@ func matchesBuildTags(fpath string, data []byte, bc *build.Context) bool {
 	return err == nil && match
 }
 
+// unmarshalProtoBase64 parses a base64 encoded proto string into the supplied
+// message, mutating msg. If the string cannot be unmarshalled, the function
```
schroederc, line 903: The "If the string cannot be unmarshalled, the function returns an error" bit is extraneous

kythe/go/indexer/indexer.go
```diff
@@ -853,6 +899,20 @@ func matchesBuildTags(fpath string, data []byte, bc *build.Context) bool {
...
+//  returns an error.
+func unmarshalProtoBase64(base64Str string, msg proto.Message) error {
+	b, err := base64.StdEncoding.DecodeString(base64Str)
+	if err != nil {
+		return err
+	}
+	if err := proto.Unmarshal(b, msg); err != nil {
```
schroederc, line 910: Just `return proto.Umarshal(b, msg)`

kythe/go/indexer/testdata/basic/inline.go
```diff
@@ -0,0 +1,25 @@
+package subject
+
+var Primary = struct {
```
schroederc, line 3: Unless I'm missing something, you can do this instead of the manual `loc/{start,end}` goals below:

`//- @Primary defines/binding Primary`

and

`//- @MyParam defines/binding MyParam` above the relevant MyParam line below

sophie-google: The reason I did not use @Primary is because @Primary would have to be defined one or more lines above var Primary. This file is testing that the changes to the indexer have successfully loaded metadata which contains specific byte location information. Since my goals are dependent on specific byte locations, adding a goal containing @Primary above var Primary would shift the byte offsets of where Primary itself occurs, hence why we used the manual loc/{start, end} syntax.

kythe/proto/BUILD
sophie-google: I ran it and received this error:

Traceback (most recent call last):
  File "***/kythe/proto/generate_protobufs.py", line 101, in <module>
    do_lang(lang='go', ext='pb.go')
  File "***/kythe/proto/generate_protobufs.py", line 62, in do_lang
    rule_dir, proto = rule.lstrip('/').rsplit(':', 1)
TypeError: a bytes-like object is required, not 'str'

kythe/go/indexer/indexer.go
sophie-google: go: finding module for package kythe.io/kythe/proto/metadata_go_proto
kythe/go/indexer/indexer.go:63:2: no matching versions for query "latest"

kythe/proto/BUILD
wcalandro: Ensure that you are running with Python 2, as you will get that error if you run with Python 3

sophie-google: Thanks! It worked with Python2.

kythe/go/indexer/indexer.go
sophie-google: I used Python2 to run generate_protobufs.py. However, even so, subsequently running go test -mod=mod ./kythe/... still results in no modified or added files. I get lengthy output in my terminal with some errors and failures.

shahms:
This looks good to me, once Cody is satisfied :-)

kythe/go/indexer/indexer.go
```diff
@@ -44,21 +45,22 @@ import (
 	"go/types"
 	"io"
 	"io/ioutil"
+	"kythe.io/kythe/go/extractors/govname"
```
schroederc, line 48: These imports should be a separate group below the std lib imports and above the third_party ones.

sophie-google: Done

kythe/go/indexer/indexer.go
```diff
@@ -48,17 +49,19 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-
```
schroederc, line 51: Remove the leading whitespace from this line to make the linter happy

sophie-google: Done - is there a way to run the linter before committing so the checks don't unnecessarily fail?

schroederc:
You can setup pre-commit (https://kythe.io/contributing/)

schroederc (approved):

PR #5002, opened 2021-07-30
fix(rust_extractor): make extractor tests more resilient

wcalandro:


kythe/rust/extractor/extractor_test.bzl
```diff
@@ -61,13 +63,13 @@ rust_extractor_test = rule(
         "src": attr.label(
             mandatory = True,
             executable = True,
-            cfg = "host",
+            cfg = "target",
```
shahms, line 66: These should both be `host` no? They are going to be run on the host, not built for the target platform.

wcalandro:
```
Otherwise, executables that are used at runtime (e.g. as part of a test) should be built for the target configuration. In this case, specify cfg="target" in the attribute.
```
https://docs.bazel.build/versions/main/skylark/rules.html#configurations

shahms (approved):

PR #5003, opened 2021-07-30
proto: update generator script to support Python3

creachadair:
Convert Bazel output to a string rather than bytes, so that string operations
in Python 3 will work correctly. This change assumes the output is UTF-8, which
should be safe in the Kythe repository.


shahms (approved):
Thanks!

creachadair:
Oops. I misread that as a merge.

PR #5008, opened 2021-08-02
feat(cxx_indexer): use default corpus for meta, builtin nodes

justbuchanan:
This should be a no-op unless --use_compilation_corpus_as_default is
enabled, in which case these nodes will be assigned the compilation
unit's corpus.

KytheGraphObserver's constructor emits nodes for builtins and meta
nodes, so setting the default corpus later means it won't apply to these
nodes. The constructor has been modified to accept a default_corpus
parameter, which is set to the compilation unit's corpus when
--use_compilation_corpus_as_default is enabled.

A new test was added that reads the indexer output for a test example
and checks that each vname has a non-empty corpus.


justbuchanan:
I'm having trouble getting `//kythe/cxx/indexer/cxx/testdata:vname_corpus_empty_test` to work because the expansion of `"$(location :vname_corpus_entries)"` gives `kythe/cxx/indexer/cxx/testdata/vname_corpus_entries.entries.gz`, but the actual path should be something like `bazel-out/k8-fastbuild/bin/kythe/cxx/indexer/cxx/testdata/vname_corpus_entries.entries.gz`. Any tips on getting that working?

kythe/go/test/tools/empty_corpus_checker/empty_corpus_checker.go
```diff
@@ -0,0 +1,58 @@
+// Binary empty_corpus_checker reads an entrystream (in delimited binary proto
```
shahms, line 1: This file is missing it's copyright header.

shahms:
> I'm having trouble getting `//kythe/cxx/indexer/cxx/testdata:vname_corpus_empty_test` to work because the expansion of `"$(location :vname_corpus_entries)"` gives `kythe/cxx/indexer/cxx/testdata/vname_corpus_entries.entries.gz`, but the actual path should be something like `bazel-out/k8-fastbuild/bin/kythe/cxx/indexer/cxx/testdata/vname_corpus_entries.entries.gz`.
> 
> Any tips on getting that working?



> I'm having trouble getting `//kythe/cxx/indexer/cxx/testdata:vname_corpus_empty_test` to work because the expansion of `"$(location :vname_corpus_entries)"` gives `kythe/cxx/indexer/cxx/testdata/vname_corpus_entries.entries.gz`, but the actual path should be something like `bazel-out/k8-fastbuild/bin/kythe/cxx/indexer/cxx/testdata/vname_corpus_entries.entries.gz`.
> 
> Any tips on getting that working?

Use `$(execpath ...)` https://docs.bazel.build/versions/main/be/make-variables.html#predefined_label_variables

(Or match the VName path and use vnames.json to map it appropriately when extracted)

justbuchanan:
Thanks! I tried that and realized the issue was that I needed to use `:vname_corpus_entries.entries.gz` instead of `:vname_corpus_entries`. After that change, the file shows up at the path provided by `$(location ...)`

kythe/go/test/tools/empty_corpus_checker/empty_corpus_checker.go
shahms: This file is still lacking a copyright header.

kythe/cxx/indexer/cxx/KytheGraphObserver.h
```diff
@@ -140,13 +140,15 @@ class KytheGraphObserver : public GraphObserver {
                               const MetadataSupports* meta_supports,
                               const llvm::IntrusiveRefCntPtr<IndexVFS>& vfs,
                               ProfilingCallback ReportProfileEventCallback,
-                              std::string build_config = "")
+                              std::string build_config = "",
+                              std::string default_corpus = "")
```
shahms, line 144: This is awkward; repeated function parameters of the same type (particularly when defaulted) are error-prone as it's often hard to tell at the call site which argument belongs to which parameter.

What was wrong with the previous formulation with the external call to `set_default_corpus`?

justbuchanan: `RegisterBuiltins()` and `EmitMetaNodes()` are called from the constructor, so the external call to `set_default_corpus()` was happening too late.

I could revert back to externally calling `set_default_corpus()` and also move `EmitMetaNodes()` outside the constructor.

shahms: Seems like as good a time as any to pull the optional arguments into a separate struct, then.

1) remove set_default_corpus if it can't have the intended effect 
2) change to:
```
struct KytheGraphObserverOptions {
  // The custom build_config, if any, to apply to anchor nodes.
  std::string build_config = "";
  // The default corpus to use for nodes which would otherwise have an empty corpus.
  std::string default_corpus = "";
};
class KytheGraphObserver ... {
 public:
  // Nested classes in C++ are obnoxious, so Clang won't like the default construction below
  // if `Options` is a nested class.
  using Options = KytheGraphObserverOptions;

  explicit KytheGraphObserver(..., const Options& options = {}) ...
};
```

kythe/cxx/indexer/cxx/KytheGraphObserver.h
```diff
@@ -131,24 +131,35 @@ class KytheClaimToken : public GraphObserver::ClaimToken {
...
+  // corpus.
+  std::string default_corpus = "";
+};
+
 /// \brief Records details in the form of Kythe nodes and edges about elements
 /// discovered during indexing to the provided `KytheGraphRecorder`.
 class KytheGraphObserver : public GraphObserver {
  public:
+  // Nested classes in C++ are obnoxious, so Clang won't like the default
+  // construction below if `Options` is a nested class.
```
shahms, line 147: Sorry, I just meant this comment for you.  No need to include it in the code :-)

justbuchanan: sounds good :)

kythe/cxx/indexer/cxx/KytheGraphObserver.h
justbuchanan: :+1:

shahms (approved):

PR #5014, opened 2021-08-02
fix(rust_extractor): process env vars and compiler args passed to wrapper

### LINKED ISSUE #5011: Rust extractor doesn't support pulling in compiler wrapper arguments
The Bazel Rust rules utilize a wrapper binary to configure the build environment before calling the Rust compiler. This brings in environment variables and other information output by build scripts. Currently, the Rust extractor does not replicate this behavior, which can cause extraction to fail when a compilation target requires some of those environment variables or linking arguments to compile.

wcalandro:
Fixes #5011

### DIFF
--- a/kythe/rust/extractor/src/bin/save_analysis.rs
+++ b/kythe/rust/extractor/src/bin/save_analysis.rs
@@ -86,7 +86,7 @@ fn process_env_file(path: &Path) -> Result<()> {
     let lines = io::BufReader::new(file).lines();
     let pwd = std::env::current_dir().context("Couldn't determine pwd")?;
     for line in lines {
-        let line_string = line.context("Failed to read line of arg file")?;
+        let line_string = line.context("Failed to read line of env file")?;
         let split: Vec<&str> = line_string.split('=').collect();
         let value = split[1].replace("${pwd}", pwd.to_str().unwrap());
         std::env::set_var(split[0], value);
--- a/kythe/rust/extractor/src/bin/save_analysis.rs
+++ b/kythe/rust/extractor/src/bin/save_analysis.rs
@@ -96,10 +96,8 @@ fn process_env_file(path: &Path) -> Result<()> {
 ///
 /// * `path` - The path of the compiler argument file
 fn process_arg_file(path: &Path) -> Result<Vec<String>> {
-    let mut arguments = Vec::new();
-    for line in io::BufReader::new(File::open(path)?).lines() {
-        let line_string = line.context("Failed to read line of arg file")?;
-        arguments.push(line_string);
-    }
-    Ok(arguments)
+    io::BufReader::new(File::open(path)?)
+        .lines()
+        .map(|line| line.context("Failed to read line of arg file"))
+        .collect()
 }
--- a/kythe/rust/extractor/src/bin/save_analysis.rs
+++ b/kythe/rust/extractor/src/bin/save_analysis.rs
@@ -58,7 +58,7 @@ fn generate_arguments(arguments: Vec<String>, output_dir: &Path) -> Result<Vec<S
     rustc_arguments[outdir_position] = format!("--out-dir={}", outdir_str);
 
     // Check if there is a file containing environment variables
-    let env_file_position = arguments.iter().position(|arg| arg.contains("--env-file"));
+    let env_file_position = arguments.iter().position(|arg| arg == "--env-file");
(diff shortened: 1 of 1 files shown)


kythe/rust/extractor/src/bin/save_analysis.rs
```diff
@@ -55,5 +57,53 @@ fn generate_arguments(arguments: Vec<String>, output_dir: &Path) -> Result<Vec<S
         output_dir.to_str().ok_or_else(|| anyhow!("Couldn't convert temporary path to string"))?;
     rustc_arguments[outdir_position] = format!("--out-dir={}", outdir_str);
 
+    // Check if there is a file containing environment variables
+    let env_file_position = arguments.iter().position(|arg| arg.contains("--env-file"));
```
shahms, line 61: `arg.starts_with` or `==` ?

kythe/rust/extractor/src/bin/save_analysis.rs
```diff
@@ -55,5 +57,53 @@ fn generate_arguments(arguments: Vec<String>, output_dir: &Path) -> Result<Vec<S
...
+    if let Some(position) = env_file_position {
+        let env_file = &arguments[position + 1];
+        let path = Path::new(env_file);
+        process_env_file(path).context("Failed to process wrapper environment file")?;
+    }
+
+    // Check if there is a file containing extra compiler arguments
+    let arg_file_position = arguments.iter().position(|arg| arg.contains("--arg-file"));
```
shahms, line 69: ditto

kythe/rust/extractor/src/bin/save_analysis.rs
```diff
@@ -55,5 +57,53 @@ fn generate_arguments(arguments: Vec<String>, output_dir: &Path) -> Result<Vec<S
...
+/// * `path` - The path of the compiler argument file
+fn process_arg_file(path: &Path) -> Result<Vec<String>> {
+    let file = File::open(path)?;
+    let lines = io::BufReader::new(file).lines();
+    let mut arguments = Vec::new();
+    for line in lines {
+        let line_string = line.context("Failed to read line of arg file")?;
+        arguments.push(line_string);
+    }
+    Ok(arguments)
```
shahms, line 108: FWIW, this is also just:
```
    io::BufReader::new(File::open(path)?)
        .lines()
        .map(|line| line.context("..."))
        .collect()
```

kythe/rust/extractor/src/bin/save_analysis.rs
wcalandro: Oh yes this should be ==

kythe/rust/extractor/src/bin/save_analysis.rs
wcalandro: Oh I actually never knew that a question mark could be used inline. However, the iterator returns an Option so I don't believe that map would work correctly.

wcalandro: I stand corrected

kythe/rust/extractor/src/bin/save_analysis.rs
```diff
@@ -55,5 +57,47 @@ fn generate_arguments(arguments: Vec<String>, output_dir: &Path) -> Result<Vec<S
         output_dir.to_str().ok_or_else(|| anyhow!("Couldn't convert temporary path to string"))?;
     rustc_arguments[outdir_position] = format!("--out-dir={}", outdir_str);
 
+    // Check if there is a file containing environment variables
+    let env_file_position = arguments.iter().position(|arg| arg == "--env-file");
+    if let Some(position) = env_file_position {
+        let env_file = &arguments[position + 1];
```
shahms, line 63: This (and below) assume the arguments are well-formed such that there is an argument value at `position + 1`.  I suspect using `arguments.windows(2)` may be more robust?

```
for window in arguments.windows(2) {
  if let [arg, value] = window {
    if arg == "--env-file" {
      process_env_file(Path::new(value)).context("....")?;
      break;
    }
  }
}
```

wcalandro: This operation is controlled by the Rust rules, and these arguments are being passed before the arguments that are passed to the Rust compiler, so there is a guarantee that there will at least be an argument after the flag, so we will not have an out-of-bounds read

shahms (approved):

PR #5022, opened 2021-08-04
feat(kzip info): check for invalid absolute paths

justbuchanan:
vname paths for required inputs should be relative to the root of the
repo and not contain a leading '/'.


shahms:
I think you need to regenerate the go protos

shahms (approved):

PR #5023, opened 2021-08-04
fix(rust_indexer): allow extractor target to be set manually

wcalandro:
This allows the extractor target in the rust_indexer_test macro to be set manually


tools/build_rules/verifier_test/rust_indexer_test.bzl
```diff
@@ -88,8 +88,9 @@ rust_extract = rule(
         "crate_name": attr.string(
             default = "test_crate",
         ),
-        "_extractor": attr.label(
+        "extractor": attr.label(
```
shahms, line 91: Having an executable that is core to the rule's function be user-specified is often unwise; what's the use case you have envisioned here?

PR #5028, opened 2021-08-09
fix(go_indexer): check whether embedded type is Named

schroederc:


zrlk (approved):

shahms (approved):

creachadair:
This seems like a reasonable change, but I'm curious to know what motivated it? Was something being indexed incorrectly without this check?

schroederc:
Go 1.17 tickles this long-standing bug (Embedded is deprecated and may return nil) and causes the indexer to crash.

shallowclouds:
> This seems like a reasonable change, but I'm curious to know what motivated it? Was something being indexed incorrectly without this check?

When indexing code like:

```Go
type SampleInterface = interface {
        ReadSizeProvider
}

type ReadSizeProvider = interface {
	RemainingBytes() (numBytes uint64)
}
```
It got crashed.

I met this problem recently, and I'm happy to see this issue has been fixed, thanks.

PR #5032, opened 2021-08-11
fix(rust_indexer): Fix proxy indexer

wcalandro:


jaysachs (commented):
I sorta wish this was 3 separate PRs:

1. `hex` -> `base64` dependency
2. removal of path from `FileProvider` methods
3. the actual fix (`"true"` -> `true`)

shahms:
Why remove path when it exists on the underlying RPC?

wcalandro:
> I sorta wish this was 3 separate PRs:
> 
> 1. `hex` -> `base64` dependency
> 2. removal of path from `FileProvider` methods
> 3. the actual fix (`"true"` -> `true`)

I agree that the FileProvider change could probably be broken out, but the "true" -> true change and the hex -> base64 change are both part of the fix

wcalandro:
> Why remove path when it exists on the underlying RPC?

Based on looking at other indexers that utilize the proxy interface, the precedent is to completely ignore the path argument.

shahms:
> > Why remove path when it exists on the underlying RPC?
> 
> Based on looking at other indexers that utilize the proxy interface, the precedent is to completely ignore the path argument.

Right, but it exists to allow for implementations to ignore either and provide better diagnostics in either case.

PR #5033, opened 2021-08-11
feat(corpus checker): allow specifying an allowlist of corpora

justbuchanan:


kythe/go/test/tools/empty_corpus_checker/empty_corpus_checker.go
```diff
@@ -77,5 +81,14 @@ func main() {
 	if emptyCorpusCount != 0 {
 		log.Fatalf("FAILURE: found %d sources with empty corpus", emptyCorpusCount)
 	}
+
+	if len(allowedCorpora) > 0 {
+		allowedCorporaSet := stringset.New(allowedCorpora...)
```
shahms, line 86: Just use flagutil.StringSet rather than converting here?

justbuchanan: Thanks! I forgot we had that.

shahms (approved):

Issue #5037, opened 2021-08-18
Support marking (proto) declarations with semantics


zrlk:
While we support using metadata to link generated code to generator source (as with protobuf), we have no way to specify the kind of semantics a particular generated function has. For example, it is useful to know whether a protobuf function mutates a particular field. This issue concerns adding `GraphObserver::UseKind` values to metadata.

An example metadata file that marks a particular function as writing to a remote node is:
```
{
  "type": "kythe0",
  "meta": [
    {
      "type": "anchor_defines",
      "begin": 122,
      "end": 135,
      "vname": {
        "signature": "gsig",
        "corpus": "gcorp",
        "path": "gpath",
        "language": "glang",
        "root": "groot"
      },
      "semantic": "SEMA_WRITE"
    }
  ]
}
```

The indexer should emit an additional `ref/writes` edge that points to the VName in the metadata record when it encounters a call to the function:

```
// Checks that we can apply semantics to functions.
#pragma kythe_metadata "semantic.meta"
//- @foo defines/binding FnFoo
void foo();

void bar() {
  //- // @"foo()" ref/writes vname(gsig, gcorp, groot, gpath, glang)
  //- @foo ref FnFoo
  //- @"foo()" ref/call FnFoo
  foo();
}
```

PR #5045, opened 2021-08-20
Apply semantic metadata rules for function calls for #5037

zrlk:
This PR adds a cache of alternate semantics that maps from
source locations to alternate semantics as specified in metadata
files. When a relevant declaration is discovered at one
of these source locations, its canonical representative is
added to a separate cache so it can be looked up directly.
This way, we don't have to complicate IndexedParentMap any more
(or worse, add another prepass over the AST).


kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -1070,6 +1071,10 @@ class GraphObserver {
   /// Note that the `VNameRef` should not outlive `id`.
   VNameRef DecodeMintedVName(const NodeId& id) const;
 
+  /// \brief Return a map of file IDs to metadata.
+  virtual const std::multimap<clang::FileID, std::shared_ptr<MetadataFile>>&
```
shahms, line 1075: minor: prefer `absl::flat_hash_map<clang::FileID, std::vector<...>>` unless this needs order (which doesn't appear to be the case)

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -3065,6 +3072,7 @@ bool IndexerASTVisitor::VisitFunctionDecl(clang::FunctionDecl* Decl) {
   if (SkipAliasedDecl(Decl)) {
     return true;
   }
+  (void)AlternateSemanticForDecl(Decl);
```
shahms, line 3075: This needs a comment.

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -5697,4 +5705,55 @@ void IndexerASTVisitor::LogErrorWithASTDump(absl::string_view msg,
...
+  const auto* named = llvm::dyn_cast<clang::NamedDecl>(decl);
+  // No sense in annotating an unnamed declaration. We still deal in `Decl`s
+  // in this function, though, because bare `Decl`s appear in many contexts
+  // (like the callee of a call expression).
+  if (named == nullptr) return nullptr;
+  auto range = RangeForNameOfDeclaration(named);
+  if (auto cached =
+          alternate_semantic_cache_.find(range.getBegin().getRawEncoding());
+      cached != alternate_semantic_cache_.end() &&
+      cached->second.end == range.getEnd()) {
```
shahms, line 5752: It looks like you're co-mingling the key and the value here, rather than just using a pair for the key?

kythe/cxx/indexer/cxx/IndexerASTHooks.h
```diff
@@ -1036,6 +1040,23 @@ class IndexerASTVisitor : public RecursiveTypeVisitor<IndexerASTVisitor> {
 
   /// \brief AST nodes we know are used in specific ways.
   absl::flat_hash_map<const clang::Stmt*, GraphObserver::UseKind> use_kinds_;
+
+  struct AlternateSemantic {
+    clang::SourceLocation end;
```
shahms, line 1045: right, just move this to the key

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -1174,12 +1179,19 @@ class NullGraphObserver : public GraphObserver {
 
   const ClaimToken* getVNameClaimToken() const override { return &VnameToken; }
 
+  const std::multimap<clang::FileID, std::shared_ptr<MetadataFile>>& meta()
```
shahms, line 1182: It seems like this interface wants to be merely:

std::shared_ptr<MetadataFile> GetMetadataFile(clang::SourceRange);

(or CharSourceRange)

shahms (changes requested):

kythe/cxx/indexer/cxx/GraphObserver.h
zrlk: There can be more than one metadata file associated with a given FileID.

kythe/cxx/indexer/cxx/GraphObserver.h
zrlk: See below re: duplicate keys.

shahms: So return a pair of iterators; exposing the entire map is unnecessary, especially since std::multimap is bad. Note: the possibility of duplicate keys is why I suggested `flat_hash_map<..., std::vector<...>>`

shahms: Or even just a `std::vector<MetadataFile*>`; there's no need to provide callers with a strong reference or expose that to callers.

zrlk: Oh, I was so focused on the `flat_hash_map` I missed the `vector`.

kythe/cxx/indexer/cxx/GraphObserver.h
zrlk: Also, `PrepareAlternateSemanticCache` scans through all loaded metadata at once, not just the metadata for a particular source range.

shahms (approved):
Thanks!

PR #5061, opened 2021-09-01
fix(rust_indexer): don't try to canonicalize tmp_directory

wcalandro:


jaysachs (approved):
So is the issue that `fs::canonicalize` doesn't resolve symlinks?

PR #5065, opened 2021-09-03
fix(rust_indexer): don't throw error when parent vname can't be found

wcalandro:
Sometimes, definitions may have parents that are not preset in the save_analysis. We simply want to skip emitting childof edges. This may have to do with rustc_save_analysis not supporting macros yet, but also may be related to certain relations to the standard library.


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -397,32 +397,33 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            if let Some(krate_id) = self.krate_ids.get(&def.id.krate) {
+                let krate_signature = format!(
+                    "{}_{}_{}",
+                    krate_id.disambiguator.0, krate_id.disambiguator.1, krate_id.name
+                );
+
+                // Check for "./" at the beginning and remove it
+                let file_vname = self.file_vnames.get(def.span.file_name.to_str().unwrap());
+
+                // save_analysis sometimes references files that we have as file nodes
```
jaysachs, line 409: "that we don't have as"?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -503,14 +504,18 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
-                    let parent_vname = self.definition_vnames.get(&parent_id).ok_or_else(|| {
-                        KytheError::IndexerError(format!(
-                            "Failed to get vname for parent of definition {:?}",
+                    if let Some(parent_vname) = self.definition_vnames.get(&parent_id) {
+                        // Emit the childof edge between this node and the parent
+                        self.emitter.emit_edge(def_vname, parent_vname, "/kythe/edge/childof")?;
+                    } else {
+                        // TODO(wcalandro): This error should no longer happen once we support
+                        // macros
+                        eprintln!(
```
jaysachs, line 513: Emit a diagnostic node instead of going to stderr?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -503,14 +504,29 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+                        // Emit the childof edge between this node and the parent
+                        self.emitter.emit_edge(def_vname, parent_vname, "/kythe/edge/childof")?;
+                    } else {
+                        // Generate a diagnostic node indicating that we couldn't find the parent
+                        let mut diagnostic_vname = def_vname.clone();
+                        let mut anchor_vname = def_vname.clone();
+                        let def_signature = def_vname.get_signature();
+                        diagnostic_vname.set_signature(format!("{}_diagnostic", def_signature));
+                        anchor_vname.set_signature(format!("{}_anchor", def_signature));
+                        self.emitter.emit_node(
```
jaysachs, line 517: I suggest have a separate `emit_diagnostic` function ... it will come in handy.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -397,32 +397,33 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
-                continue;
+                // Generate node based on definition type
+                let mut def_vname = self.krate_vname.clone();
+                let def_signature = format!("{}_def_{}", krate_signature, def.id.index);
+                def_vname.set_signature(def_signature.clone());
+                def_vname.set_language("rust".to_string());
+                def_vname.clear_path();
+                self.emit_definition_node(&def_vname, def, file_vname.unwrap())?;
+            } else {
+                eprintln!(
```
jaysachs, line 422: Diagnostic node here as well?

wcalandro: Not possible, we have no signature to create an anchor nor diagnostic node with

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -503,14 +513,24 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
-                    self.emitter.emit_edge(def_vname, parent_vname, "/kythe/edge/childof")?;
+                    if let Some(parent_vname) = self.definition_vnames.get(&parent_id) {
+                        // Emit the childof edge between this node and the parent
+                        self.emitter.emit_edge(def_vname, parent_vname, "/kythe/edge/childof")?;
+                    } else {
+                        // Generate a diagnostic node indicating that we couldn't find the parent
+                        let mut diagnostic_vname = def_vname.clone();
+                        let mut anchor_vname = def_vname.clone();
+                        let def_signature = def_vname.get_signature();
+                        diagnostic_vname.set_signature(format!("{}_diagnostic", def_signature));
```
jaysachs, line 524: Might there be other diagnostic nodes for that def?

kythe/rust/indexer/src/indexer/entries.rs
```diff
@@ -108,4 +108,37 @@ impl<'a> EntryEmitter<'a> {
...
     }
+
+    /// Creates a diagnostic node with a tagged edge from an anchor or file
+    /// vname to the node Accepts a message, optional details, and an
+    /// optional url
+    ///
+    /// # Errors
+    /// If an error occurs while writing the entry, an error is returned.
+    pub fn emit_diagnostic<S: Into<String>>(
+        &mut self,
```
jaysachs, line 119: What if you pulled the vname generation in here? Use the compilation unit hash plus some (in process) sequence to make sure the sigs are unique?

kythe/rust/indexer/src/indexer/entries.rs
```diff
@@ -115,30 +116,43 @@ impl<'a> EntryEmitter<'a> {
...
+        let bytes = sha256.finalize();
+        let sha256sum = hex::encode(bytes);
+
+        // Use source_vname signature and sha256 sum to generate diagnostic signature
+        let mut diagnostic_vname = VName::new();
+        let source_signature = source_vname.get_signature();
+        diagnostic_vname.set_signature(format!("{}_{}", source_signature, sha256sum));
+
+        // Emit diagnostic node
         self.emit_node(&diagnostic_vname, "/kythe/node/kind", b"diagnostic".to_vec())?;
```
jaysachs, line 140: This had me confused for a minute - it reads like you're emitting the node multiple times. `emit_node` is really `emit_fact`.

kythe/rust/indexer/src/indexer/entries.rs
```diff
@@ -115,30 +116,43 @@ impl<'a> EntryEmitter<'a> {
...
-        diagnostic_vname: &VName,
         message: &str,
-        details: Option<S>,
-        url: Option<S>,
+        details: Option<&str>,
+        url: Option<&str>,
     ) -> Result<(), KytheError> {
+        // Combine diagnostic message fields and create sha256 sum
+        let combined_description =
+            format!("{}||{}||{}", message, details.unwrap_or("NONE"), url.unwrap_or("NONE"));
```
jaysachs, line 128: Maybe use empty string instead of "NONE"?`

kythe/rust/indexer/src/indexer/entries.rs
```diff
@@ -115,30 +116,43 @@ impl<'a> EntryEmitter<'a> {
...
+        let mut sha256 = Sha256::new();
+        sha256.update(combined_description.as_str().as_bytes());
+        let bytes = sha256.finalize();
+        let sha256sum = hex::encode(bytes);
```
jaysachs, line 132: sha256 works; maybe a bit heavyweight but we shouldn't be going down this path frequently.

jaysachs (commented):
I'd like shahms' take on this as well.

kythe/rust/indexer/src/indexer/entries.rs
wcalandro: Another option is to base64 but that would lead to long signatures.

kythe/rust/indexer/src/indexer/entries.rs
wcalandro: I agree. Not sure why I originally named it that. Changed

kythe/rust/indexer/src/indexer/entries.rs
shahms: The C++ indexer uses the base64 encoding of a SHA256 hash for all signatures longer than 43 bytes. This is fine here :-)

shahms (approved):

PR #5066, opened 2021-09-08
fix(rust_indexer): handle ending offsets for files without ending newline

wcalandro:
If a file being indexed does not have an ending newline (such as generated Rust protobuf files), the ending column of a definition span might exceed the number of columns in that line. Therefore if a byte offset is requested for a column that is longer than the line, we just return the offset for the last column in the line.


jaysachs (commented):
Can you add a test for this?

wcalandro:
@jaysachs The easiest way to create the test was to remove the newline at the bottom of the anchors.rs test file. Before this PR, removing that newline would cause the indexer to panic, now it successfully indexes the test file and passes the verifier

jaysachs (approved):

PR #5067, opened 2021-09-08
fix(rust_extractor): use post-processed file paths for source file field

wcalandro:
Previously, the raw path of a source file was used for the source files field in the CompilationUnit. However, if the vname rules modified the path, there would be a mismatch between the path in the required inputs and the source files. Now, we use the post-processed path in both locations.

### DIFF
--- a/kythe/rust/extractor/src/bin/extractor.rs
+++ b/kythe/rust/extractor/src/bin/extractor.rs
@@ -73,10 +73,12 @@ fn main() -> Result<()> {
         .filter(|file| file.ends_with(".rs"))
         .map(String::from) // Map the &String to a new String
         .collect();
+    let mut source_file_processed_paths = Vec::new();
     let mut required_input: Vec<CompilationUnit_FileInput> = Vec::new();
     for file_path in &rust_source_files {
         // Generate the file vname based on the vname rules
         let file_vname: VName = create_vname(&mut vname_rules, file_path, &default_corpus);
+        source_file_processed_paths.push(file_vname.get_path().to_string());
         kzip_add_required_input(file_path, file_vname, &mut kzip, &mut required_input)?;
     }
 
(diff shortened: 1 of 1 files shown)


shahms:
That mismatch is to be expected; the source files should correspond to the `FileInfo` path, not the `VName`

wcalandro:
> That mismatch is to be expected; the source files should correspond to the `FileInfo` path, not the `VName`

Is this to ensure that no references are emitted to generated files?

shahms:
> > That mismatch is to be expected; the source files should correspond to the `FileInfo` path, not the `VName`
> 
> Is this to ensure that no references are emitted to generated files?

No, it's because required_input is a lookup map from compiler paths (`FileInfo`) to VName paths and the source_file field is the former.  We want to emit references to generated files, but they need to have the right VName.

wcalandro:
> No, it's because required_input is a lookup map from compiler paths (`FileInfo`) to VName paths and the source_file field is the former. We want to emit references to generated files, but they need to have the right VName.

What is the use of required_inputs then, if we use the source_files vnames when emitting references?

shahms:
The paths from source_files should be looked up (by the indexer) via `required_input.info.path` and use the corresponding VName, but the extractor should (mostly) preserve the compiler paths.

creachadair:
> The paths from source_files should be looked up (by the indexer) via `required_input.info.path` and use the corresponding VName, but the extractor should (mostly) preserve the compiler paths.

The main reason we preserve the paths as the compiler saw them (as much as possible) is that some compilers make decisions (e.g., configuration, where to find imports, etc.) based on where input files are physically located in the original filesystem. The original culprit was `gcc`, but we've seen other examples too (e.g., python resolving relative imports). I keep thinking newer languages won't repeat the same mistakes, but I've been repeatedly surprised.

Often—usually, in fact—virtualizing paths relative to the corpus is a no-op (except for generated files). But we try to avoid language-specific corner cases in the compilation record, except where we can't help it (hence the language details).

PR #5068, opened 2021-09-09
fix(rust_indexer): use processed file paths

wcalandro:
Previously, the indexer would get file paths from both the required inputs and the source files, which don't always match up. Because the Rust indexer does not have to do any work using the original compiler paths, we only use the paths from the required inputs.

### DIFF
--- a/kythe/rust/indexer/src/indexer/analyzers.rs
+++ b/kythe/rust/indexer/src/indexer/analyzers.rs
@@ -97,9 +97,11 @@ impl<'a> UnitAnalyzer<'a> {
         for required_input in unit.get_required_input() {
             let analysis_vname = required_input.get_v_name();
             let storage_vname: VName = analysis_to_storage_vname(analysis_vname);
-            let path = storage_vname.get_path().to_owned();
+            let info = required_input.get_info();
+            let path = info.get_path().to_owned();
+            let digest = info.get_digest().to_owned();
             file_vnames.insert(path.clone(), storage_vname);
-            file_digests.insert(path.clone(), required_input.get_info().get_digest().to_string());
+            file_digests.insert(path.clone(), digest);
         }
 
         let unit_storage_vname: VName = analysis_to_storage_vname(unit.get_v_name());
--- a/kythe/rust/indexer/src/indexer/analyzers.rs
+++ b/kythe/rust/indexer/src/indexer/analyzers.rs
@@ -38,6 +38,9 @@ pub struct UnitAnalyzer<'a> {
     file_vnames: HashMap<String, VName>,
     // A map between a file name and its sha256 digest
     file_digests: HashMap<String, String>,
+    // A map between the compiler path of generated files and the path shown in
+    // the save_analysis file
+    generated_file_names: HashMap<String, String>,
     // An index for computing byte offsets in files based on line and column number
     offset_index: OffsetIndex,
     // A file provider
@@ -94,14 +97,19 @@ impl<'a> UnitAnalyzer<'a> {
         // later to emit nodes and create a HashMap between a file path and its digest
         let mut file_vnames = HashMap::new();
         let mut file_digests = HashMap::new();
+        let mut generated_file_names = HashMap::new();
         for required_input in unit.get_required_input() {
             let analysis_vname = required_input.get_v_name();
             let storage_vname: VName = analysis_to_storage_vname(analysis_vname);
             let info = required_input.get_info();
-            let path = info.get_path().to_owned();
+            let source_path = info.get_path().to_owned();
+            let processed_path = storage_vname.get_path().to_owned();
+            if source_path != processed_path {
+                generated_file_names.insert(processed_path, source_path.clone());
+            }
             let digest = info.get_digest().to_owned();
-            file_vnames.insert(path.clone(), storage_vname);
-            file_digests.insert(path.clone(), digest);
+            file_vnames.insert(source_path.clone(), storage_vname);
+            file_digests.insert(source_path.clone(), digest);
         }
 
         let unit_storage_vname: VName = analysis_to_storage_vname(unit.get_v_name());
@@ -111,6 +119,7 @@ impl<'a> UnitAnalyzer<'a> {
             emitter: EntryEmitter::new(writer),
             file_vnames,
             file_digests,
+            generated_file_names,
             offset_index: OffsetIndex::default(),
             provider,
         }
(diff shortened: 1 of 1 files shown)


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -118,35 +115,33 @@ impl<'a> UnitAnalyzer<'a> {
     /// generate the OffsetIndex
     pub fn handle_files(&mut self) -> Result<(), KytheError> {
         // https://kythe.io/docs/schema/#file
-        for source_file in self.unit.get_source_file() {
-            let vname = self.get_file_vname(source_file)?;
-
+        for (file_path, vname) in &self.file_vnames {
```
shahms, line 118: For languages-not-C++ the indexer should generally restrict itself to indexing the files listed in the source_file attribute rather than all of the required_input.

Do you have an example of where the path from `source_file` isn't present in `required_input`?  It's likely that the extractor should be normalizing and relativizing paths.

wcalandro: If a generated file is used as compiler input, the path after passing through the vname rules (in required_input) is different than the path that is present in path used by the compiler and in the save_analysis file.

wcalandro: Currently, only rust source files are added to `required_input` in the extractor, so `source_file` === `required_input`, just with different paths. Whether the extractor should be changed to add dependency libraries may be an additional conversation to be had.

shahms: Right, that's as expected.  There are two paths in required_input: `required_input.info.path` and `required_input.v_name.path`. The former should be used for finding compiler paths and should match the `source_file` (and save_analysis) path for generated files.  The latter should rarely be used directly as it is largely meaningless without the corresponding root and corpus.

Does the Rust extractor not include information about dependencies in `required_input`? Using a hypothetical BUILD rules for an example:

```
rust_binary(name = "bin", srcs = ["main.rs"], deps = [":lib"])
rust_library(name = "lib", srcs = ["lib.rs"])
```

The compilation unit for `bin` should only include `main.rs` in the `source_file` but would include whatever files were necessary to reference entities arising from `lib.rs` in `required_input`.

wcalandro: Ah I see what you are saying regarding `required_input.info.path`, I forgot that field existed. No information about dependencies are kept in `required_input`, as it would require saving compiled binary `.rlib` files to the kzip. The extractor knows nothing about the source files that are used to build the dependencies.

wcalandro:
I think this smaller overall change to the indexer better reflects the patterns from the other indexers that you described

kythe/rust/indexer/src/indexer/analyzers.rs
shahms: If you have file paths referenced from the save_analysis file which have no entries in `required_input`, the extractor needs to put that information somewhere. I think a language-specific message in the `details` field that is a simple `map<string, VName>` would suffice.  It could also be limited to the cases where `(corpus, root, path)` differ from the simple case.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -97,9 +97,11 @@ impl<'a> UnitAnalyzer<'a> {
         for required_input in unit.get_required_input() {
             let analysis_vname = required_input.get_v_name();
             let storage_vname: VName = analysis_to_storage_vname(analysis_vname);
-            let path = storage_vname.get_path().to_owned();
+            let info = required_input.get_info();
+            let path = info.get_path().to_owned();
+            let digest = info.get_digest().to_owned();
```
shahms, line 102: This still won't work for generated files?

wcalandro: Okay I fixed a small bug and now I have confirmed that this code works for generated files. The last question is whether the path in the vnames for the references should have the compiler file path or the processed file path?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -38,6 +38,9 @@ pub struct UnitAnalyzer<'a> {
     file_vnames: HashMap<String, VName>,
     // A map between a file name and its sha256 digest
     file_digests: HashMap<String, String>,
+    // A map between the compiler path of generated files and the path shown in
+    // the save_analysis file
```
shahms, line 42: The compile path is the save_analysis path, no?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -94,14 +97,19 @@ impl<'a> UnitAnalyzer<'a> {
...
         let mut file_vnames = HashMap::new();
         let mut file_digests = HashMap::new();
+        let mut generated_file_names = HashMap::new();
         for required_input in unit.get_required_input() {
             let analysis_vname = required_input.get_v_name();
             let storage_vname: VName = analysis_to_storage_vname(analysis_vname);
             let info = required_input.get_info();
-            let path = info.get_path().to_owned();
+            let source_path = info.get_path().to_owned();
+            let processed_path = storage_vname.get_path().to_owned();
```
shahms, line 106: It's possible for more than one file to have the same "processed_path" even without generated files.  The "source_path" should be unique, however.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -94,14 +97,19 @@ impl<'a> UnitAnalyzer<'a> {
...
+            if source_path != processed_path {
+                generated_file_names.insert(processed_path, source_path.clone());
```
shahms, line 108: Based on the lookup in handle_files I think this is backwards?

shahms (commented):
VNames should always contain VName paths ("processed path"), never compiler paths.  Think of VName paths like URL paths.  They are meaningless without the host (root) and scheme (corpus) which are provided by the extractor.

The extractor maps between compiler/physical paths and VNames, taking into account how the build system lays out the file system, etc.  The indexer reads compiler paths, translating them into VNames using the extractor-provided mapping only when it needs to produce graph data and then only in the context of a full VName.

kythe/rust/indexer/src/indexer/analyzers.rs
wcalandro: This refers to the path that the compiler uses to reference files in the save_analysis file

kythe/rust/indexer/src/indexer/analyzers.rs
wcalandro: The processed path is based on the source path, and the processed path is equal to the path inside the save_analysis file. The main reason for this PR overall is that the paths in the save_analysis file for generated files does not match up with the source_file path.

kythe/rust/indexer/src/indexer/analyzers.rs
wcalandro: I don't believe so. This worked as intended on my test case

shahms: Please include a rust_indexer_test of this behavior.

PR #5071, opened 2021-09-13
fix(rust_indexer): serialize proxy communication with serde_json

wcalandro:
Previously we constructed the proxystub commands with format strings, which could lead to issues if there were stray quotation marks. With this PR we let serde_json handle the JSON serialization for us.

### DIFF
--- a/kythe/rust/indexer/src/error.rs
+++ b/kythe/rust/indexer/src/error.rs
@@ -41,6 +41,11 @@ quick_error! {
             from()
             display("Failed to parse Protobuf: {}", err)
         }
+        /// There was an issue serializing a type to JSON
+        JsonSerializationError(err: serde_json::Error) {
+            from()
+            display("JSON Serialization Error: {}", err)
+        }
         /// The KytheWriter encounters an error
         WriterError(err: ProtobufError) {
             display("Writer encountered an error: {}", err)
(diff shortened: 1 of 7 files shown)


kythe/rust/indexer/src/error.rs
```diff
@@ -41,6 +41,11 @@ quick_error! {
             from()
             display("Failed to parse Protobuf: {}", err)
         }
+        /// There was an issue serializing a type to JSON
+        JsonSerializationError(err: serde_json::Error) {
+            from()
```
jaysachs, line 46: Where is this used?

kythe/rust/indexer/src/writer.rs
```diff
@@ -103,23 +104,19 @@ impl KytheWriter for ProxyWriter {
     /// An error is returned if the write fails.
     fn write_entry(&mut self, entry: Entry) -> Result<(), KytheError> {
         let bytes = entry.write_to_bytes().map_err(KytheError::WriterError)?;
-        // Convert the entry bytes to base64 and surround with quotes to prepare for
-        // inserting into a JSON array
-        let b64 = format!(r#""{}""#, base64::encode(&bytes));
-        self.buffer.push(b64);
+        // Convert the entry bytes to base64
```
jaysachs, line 107: I think the comment is vacuous now.

kythe/rust/indexer/src/error.rs
wcalandro: When the requests are generated, the serde_json serialization returns a `serde_json::Result` which may include a `serde_json::Error`. This code converts any `serde_json::Error` into a `KytheError::JsonSerializationError`.

jaysachs (approved):
Wonder if it's worth a future cleanup to convert the responses in the same way for symmetry?

wcalandro:
> Wonder if it's worth a future cleanup to convert the responses in the same way for symmetry?

This is doable. Would just need a generic response type that would allow us to detect the response argument type based on the response command, then we could serialize the response based on that. Can be done in a future PR

PR #5077, opened 2021-09-17
fix(rust_indexer): don't fail indexing if generated file couldn't be emitted

wcalandro:


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -119,7 +119,12 @@ impl<'a> UnitAnalyzer<'a> {
     pub fn handle_files(&mut self) -> Result<(), KytheError> {
         // https://kythe.io/docs/schema/#file
         for source_file in self.unit.get_source_file() {
-            let vname = self.get_file_vname(source_file)?;
+            let vname_result = self.get_file_vname(source_file);
+            // Generated files won't have a file vname returned
+            if vname_result.is_err() {
+                continue;
```
jaysachs, line 125: Can you emit a diagnostic here?

wcalandro: What would the diagnostic be linked to? We don't have an anchor or file node

shahms: We should log an error, at least

jaysachs: Ah, good question. Presumably if this avoids the error, there's actually no references to this file from anything in  required_input to this file, in which case it's fine to not be able to resolve this. This is only interesting if there's a reference to that file that we encounter during analysis, and during analysis, we'll have something (the required_input) to anchor to.

jaysachs (approved):

Issue #5084, opened 2021-10-04
Don't package the `.rlib` files into kzips for `fuchsia_extractor`


filmil:
As agreed with @wcalandro, the fuchsia extractor should not be packaging `rlib` files into the `kzip` archive when extracting.
This departs from the kythe recommendations for extractors, but is compatible with the currently present rust indexer.

filmil:
@wcalandro please assign me to this bug, since I have no permission to do that myself.

PR #5085, opened 2021-10-04
[fuchsia_extractor] Don't pack rlib files by default

filmil:
This change adds a flag --withrlibs, false by default, which when set
causes the extractor to package rlibs files for compilation replay.

However, since the current rust indexer does not use that analysis method,
the default behavior is to *not* package these files.

Fixes: #5084

### DIFF
--- a/kythe/rust/fuchsia_extractor/src/bin/main.rs
+++ b/kythe/rust/fuchsia_extractor/src/bin/main.rs
@@ -218,7 +218,10 @@ lazy_static! {
 }
 
 /// Examines a set of rustc arguments and extracts the "rlib" files from it.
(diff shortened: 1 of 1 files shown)


filmil:
@wcalandro PTAL.
I tested this change locally (in addition to the tests) and it seems to behave as advertised.

wcalandro (approved):

shahms:
Minor rust-format changes:
```
Diff in /workspace/kythe/rust/fuchsia_extractor/src/bin/main.rs at line 613:
     quiet: bool,
     /// If set, the `rlib` files are packaged into kzip files too.
     with_rlibs: bool,
-
 }
 
 fn main() -> Result<()> {
```

wcalandro:
Changes look good, please run `./tools/rust/rustfmt.sh` to format the code.

PR #5088, opened 2021-10-05
fix(rust_indexer): fix path confusion in vname and offset lookups

wcalandro:
Currently, the Rust indexer can run into errors if a file's VName path and FileInfo path are different. This is because some functions of the indexer, such as cached vname lookups, are done based on file name. This change uses the FileInfo path in all locations to do the lookup, because this path is relative to the build directory and will be unique for each file.

### DIFF
--- a/kythe/rust/indexer/src/indexer/analyzers.rs
+++ b/kythe/rust/indexer/src/indexer/analyzers.rs
@@ -96,8 +96,8 @@ impl<'a> UnitAnalyzer<'a> {
         let mut file_digests = HashMap::new();
         for required_input in unit.get_required_input() {
             let analysis_vname = required_input.get_v_name();
+            let path = required_input.get_info().get_path().to_owned();
             let storage_vname: VName = analysis_to_storage_vname(analysis_vname);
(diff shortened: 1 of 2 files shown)


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -118,14 +118,22 @@ impl<'a> UnitAnalyzer<'a> {
...
             let vname = vname_result.unwrap();
 
+            // Not all required inputs are Rust source files, so ignore if not
+            // a source file
+            if !info_path.ends_with(".rs") {
+                continue;
+            }
+
             // Create the file node fact
             self.emitter.emit_fact(&vname, "/kythe/node/kind", b"file".to_vec())?;
```
shahms, line 138: The indexer should only emit file nodes (among other things) for the files listed in the `source_file` field, not all of `required_input`.

shahms (commented):
On the whole this looks good, but the indexer shouldn't generally do anything for all of required_input other than look up a source path.

shahms (approved):

PR #5097, opened 2021-10-11
fix: adjust regex in simple_vnames.json

wcalandro:
This change adjusts one of the regex patterns so that it works properly with Rust

### DIFF
--- a/kythe/data/simple_vnames.json
+++ b/kythe/data/simple_vnames.json
@@ -40,7 +40,7 @@
     }
   },
   {
-    "pattern": "/(.*)",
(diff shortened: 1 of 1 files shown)


kythe/data/simple_vnames.json
```diff
@@ -40,7 +40,7 @@
     }
   },
   {
-    "pattern": "/(.*)",
```
shahms, line 43: The semantics of these patterns are equivalent to RE2::FullMatch patterns; adding an anchor here is working around a bug in the Rust indexer.

PR #5101, opened 2021-10-13
fix(rust_indexer): don't use the span file name in reference vnames

wcalandro:


wcalandro:
> Would it be possible to add some tests to guard this behavior? It seems important.

I'll try and put together a new test today

PR #5103, opened 2021-10-14
fix(rust_indexer): remove language ans signature fields from vname for files

wcalandro:


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -129,9 +129,6 @@ impl<'a> UnitAnalyzer<'a> {
             // Create the file node fact
             self.emitter.emit_fact(&vname, "/kythe/node/kind", b"file".to_vec())?;
 
-            // Create language fact
-            self.emitter.emit_fact(&vname, "/kythe/language", b"rust".to_vec())?;
```
shahms, line 133: The fact is fine, but the file VName shouldn't have a language.

jaysachs (commented):
Update the PR description please.

PR #5107, opened 2021-10-18
chore(rust_common): remove Rust runfiles helper library

wcalandro:


shahms:
Sigh. It's probably worth sending a PR to check `RUNFILES_DIR` in the upstream rules_rust repo, but until then just continue using the internal implementation.  It looks like even the Bazel Java Runfiles class (mostly) just checks the environment variable(s): https://github.com/bazelbuild/bazel/blob/aebad8133b20eee2776c9afb3814b0b6cde17a8a/tools/java/runfiles/Runfiles.java
Same for C++:
https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/tools/cpp/runfiles/runfiles_src.h#L119

PR #5109, opened 2021-10-19
feat(rust_indexer): add support for Rust import xrefs

wcalandro:
This PR adds support for emitting cross references for imported objects in `use` statements.

### DIFF
--- a/kythe/rust/indexer/src/indexer/analyzers.rs
+++ b/kythe/rust/indexer/src/indexer/analyzers.rs
@@ -265,7 +265,7 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
             format!("{}_{}_{}", krate_id.disambiguator.0, krate_id.disambiguator.1, krate_id.name);
         let mut krate_vname = self.unit_vname.clone();
         krate_vname.set_signature(signature);
-        krate_vname.set_language("rust".to_string());
+        krate_vname.set_language("rust".to_owned());
         krate_vname.clear_path();
         krate_vname
     }
@@ -558,9 +558,8 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
                 if let Some(method_impl) = self.method_index.remove(&def.id) {
                     // Get the vname of the parent struct
                     // Need to have the option variable to avoid multiple borrows on "self".
-                    let parent_vname_option =
-                        self.definition_vnames.get(&method_impl.struct_target);
-                    let parent_vname = if let Some(vname) = parent_vname_option {
+                    let parent_vname = self.definition_vnames.get(&method_impl.struct_target);
+                    let parent_vname = if let Some(vname) = parent_vname {
                         vname.clone()
                     } else {
                         // Usually this condition occurs if the save_analysis
@@ -719,8 +718,8 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
 
             // Get the CrateId for the referenced crate
             let ref_id = reference.ref_id.unwrap();
-            let krate_id_option = self.krate_ids.get(&ref_id.krate);
-            if krate_id_option.is_none() {
+            let krate_id = self.krate_ids.get(&ref_id.krate);
+            if krate_id.is_none() {
                 // This is a little bit of chicken and egg here. We could try
                 // and find the file vname first before the crate id, but then
                 // we wouldn't have a definition anchor to emit the diagnostic
@@ -738,7 +737,7 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
                 )?;
                 continue;
             }
-            let krate_id = krate_id_option.unwrap();
+            let krate_id = krate_id.unwrap();
 
             // Create VName for target of reference
             let definition_vname = self.generate_def_vname(krate_id, ref_id.index);
@@ -759,11 +758,11 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
             }
             reference_vname.set_path(file_vname.unwrap().get_path().to_string());
 
-            let byte_span_option = self.get_byte_span(&ref_id, span)?;
-            if byte_span_option.is_none() {
+            let byte_span = self.get_byte_span(&ref_id, span)?;
+            if byte_span.is_none() {
                 continue;
             }
-            let byte_span = byte_span_option.unwrap();
+            let byte_span = byte_span.unwrap();
 
             // Create signature based on span
             reference_vname.set_signature(self.create_ref_signature(krate_signature, &byte_span));
@@ -789,8 +788,8 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
             let ref_id = reference.ref_id;
 
             // Get the CrateId for the referenced crate
-            let krate_id_option = self.krate_ids.get(&ref_id.krate);
-            if krate_id_option.is_none() {
+            let krate_id = self.krate_ids.get(&ref_id.krate);
+            if krate_id.is_none() {
                 // This is a little bit of chicken and egg here. We could try
                 // and find the file vname first before the crate id, but then
                 // we wouldn't have a definition anchor to emit the diagnostic
@@ -807,7 +806,7 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
                 )?;
                 continue;
             }
-            let krate_id = krate_id_option.unwrap();
+            let krate_id = krate_id.unwrap();
 
             // Create VName for target of reference
             let target_vname = self.generate_def_vname(krate_id, ref_id.index);
@@ -826,11 +825,11 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
             }
             reference_vname.set_path(file_vname.unwrap().get_path().to_string());
 
-            let byte_span_option = self.get_byte_span(&reference.ref_id, span)?;
-            if byte_span_option.is_none() {
+            let byte_span = self.get_byte_span(&reference.ref_id, span)?;
+            if byte_span.is_none() {
                 continue;
             }
-            let byte_span = byte_span_option.unwrap();
+            let byte_span = byte_span.unwrap();
 
             // Create signature based on span
             reference_vname.set_signature(self.create_ref_signature(krate_signature, &byte_span));
@@ -861,16 +860,16 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
     ) -> Result<Option<ByteSpan>, KytheError> {
         let file_name = span.file_name.to_str().unwrap();
         // Get byte span
-        let start_byte_option =
+        let start_byte =
             self.offset_index.get_byte_offset(file_name, span.line_start.0, span.column_start.0);
 
         // If the start byte is none, then the save_analysis is giving information about
         // standard library files and we should skip
-        if start_byte_option.is_none() {
+        if start_byte.is_none() {
             return Ok(None);
         }
 
-        let start_byte = start_byte_option.unwrap();
+        let start_byte = start_byte.unwrap();
         let end_byte = self
             .offset_index
             .get_byte_offset(file_name, span.line_end.0, span.column_end.0)


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -81,6 +81,12 @@ pub struct MethodImpl {
     pub trait_target: Option<rls_data::Id>,
 }
 
+/// A struct containing byte offsets for xrefs
+struct ByteOffsets {
```
jaysachs, line 85: ByteRange ? ByteSpan? Span?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +726,73 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            }
+            let byte_offsets = byte_offsets_option.unwrap();
+
+            // Create signature based on span
+            reference_vname.set_signature(format!(
+                "{}_ref_{}_{}",
+                krate_signature, byte_offsets.start_byte, byte_offsets.end_byte
+            ));
+
+            self.emitter.emit_reference(
```
jaysachs, line 784: If you're going to introduce the `ByteSpan` (or whatever), maybe the emitter should be modified to accept it instead of (or in addition to) the two separate arguments?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +726,73 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+                    &target_vname,
+                    "Failed to get file VName for reference",
+                    Some(format!("The Rust indexer was unable to locate the file VName for the reference in the file \"{}\"", span.file_name.to_str().unwrap()).as_ref()),
+                    None,
+                )?;
+                continue;
+            }
+            reference_vname.set_path(file_vname.unwrap().get_path().to_string());
+
+            let byte_offsets_option = self.get_byte_offsets(&ref_id, span)?;
```
jaysachs, line 772: get_byte_offsets should probably return a `ByteSpan` struct (and be renamed appropriately)

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +726,73 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            }
+
+            let ref_id = reference.ref_id.unwrap();
+
+            let mut reference_vname = template_vname.clone();
+            let span = &reference.span;
+
+            // Create VName for target of reference
+            let mut target_vname = template_vname.clone();
+            target_vname.set_signature(format!("{}_def_{}", krate_signature, ref_id.index));
```
jaysachs, line 756: I think this is the 4th such occurrence.  Maybe factor out a function (named ... "signatureForFoo"? not sure -- is this a signature given a crate signature and an index, so maybe "signatureForCrateAndIndex"?) and use it everywhere

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +726,73 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            target_vname.set_language("rust".to_string());
+
+            // Create VName for the reference node
+            let file_vname = self.file_vnames.get(span.file_name.to_str().unwrap());
```
jaysachs, line 760: maybe a helper function to do all the to_str and unwrapping? this is a pretty common stanza.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +726,73 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            if file_vname.is_none() {
+                self.emitter.emit_diagnostic(
+                    &target_vname,
+                    "Failed to get file VName for reference",
+                    Some(format!("The Rust indexer was unable to locate the file VName for the reference in the file \"{}\"", span.file_name.to_str().unwrap()).as_ref()),
+                    None,
+                )?;
+                continue;
+            }
+            reference_vname.set_path(file_vname.unwrap().get_path().to_string());
```
jaysachs, line 770: Not sure if a helper to get the file_vname path from a file_name would be useful or not.

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -751,42 +825,60 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+        // Get byte span
+        let start_byte_option = self.offset_index.get_byte_offset(
+            span.file_name.to_str().unwrap(),
+            span.line_start.0,
+            span.column_start.0,
+        );
+
+        // If the start byte is none, then the save_analysis is giving information about
+        // standard library files and we should skip
+        if start_byte_option.is_none() {
```
jaysachs, line 865: Is there a verifier test for this case? Can you add one?

kythe/rust/indexer/src/indexer/analyzers.rs
wcalandro: A result with an option has to be returned because of line 865 where a check is made for a reference to a standard library.

kythe/rust/indexer/src/indexer/analyzers.rs
wcalandro: Changed to ByteSpan

kythe/rust/indexer/src/indexer/analyzers.rs
jaysachs: https://kythe.io/docs/kythe-verifier.html section "Checking that rules are unsatisfiable"

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +709,89 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            }
+
+            // Currently this is only known to support `Use` ImportKinds
+            if reference.kind != rls_data::ImportKind::Use {
+                continue;
+            }
+
+            let ref_id = reference.ref_id.unwrap();
+
+            let mut reference_vname = template_vname.clone();
```
jaysachs, line 734: Move this declaration closer to its first use?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,6 +709,89 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+
+            let byte_span_option = self.get_byte_span(&ref_id, span)?;
+            if byte_span_option.is_none() {
+                continue;
+            }
+            let byte_span = byte_span_option.unwrap();
+
+            // Create signature based on span
+            reference_vname.set_signature(format!(
+                "{}_ref_{}_{}",
```
jaysachs, line 784: You do this twice; pull it out to a function that takes the krate_signature and the ByteSpan

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -81,6 +81,12 @@ pub struct MethodImpl {
     pub trait_target: Option<rls_data::Id>,
 }
 
+/// A struct containing byte offsets for xrefs
```
jaysachs, line 84: "Represents a span within a file based on byte offets."

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -280,21 +289,15 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
 
         // First emit the node for our own crate and add it to the hashmap
         let krate_id = &krate_prelude.crate_id;
-        let krate_signature =
-            format!("{}_{}_{}", krate_id.disambiguator.0, krate_id.disambiguator.1, krate_id.name);
-        let krate_vname = self.generate_crate_vname(&krate_signature);
+        let krate_vname = self.generate_crate_vname(krate_id);
```
jaysachs, line 292: Be consistent in var naming, either "foo_v_name" or "foo_vname"

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -259,9 +266,11 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
 
     /// Given a signature, generates a VName for a crate based on the VName of
     /// the CompilationUnit
-    fn generate_crate_vname(&self, signature: &str) -> VName {
+    fn generate_crate_vname(&self, krate_id: &rls_analysis::CrateId) -> VName {
+        let signature =
+            format!("{}_{}_{}", krate_id.disambiguator.0, krate_id.disambiguator.1, krate_id.name);
         let mut crate_v_name = self.unit_vname.clone();
-        crate_v_name.set_signature(signature.to_string());
+        crate_v_name.set_signature(signature);
```
jaysachs, line 273: Why use "crate_v_name" here and "krate_vname" below?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -413,16 +416,9 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
-                    krate_id.disambiguator.0, krate_id.disambiguator.1, krate_id.name
-                );
-
                 // Generate node based on definition type
                 let mut def_vname = self.krate_vname.clone();
-                let def_signature = format!("{}_def_{}", krate_signature, def.id.index);
-                def_vname.set_signature(def_signature.clone());
-                def_vname.set_language("rust".to_string());
+                def_vname.set_signature(self.create_def_signature(krate_id, def.id.index));
                 def_vname.clear_path();
```
jaysachs, line 422: The krate_vname (from which this was cloned) shouldn't have a path. I get this is defensive. Maybe instead of `create_def_signature` have a `create_def_v_name(base_vname, krate_id, index)` ?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -731,62 +797,103 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+                // of a crate id not being in our HashMap is very low, so if this
+                // fails we just emit a diagnostic node to the first source file
+                // of the crate we are indexing
+                let file_vname = self.file_vnames.values().next().unwrap();
+                self.emitter.emit_diagnostic(
+                    file_vname,
+                    "Cross reference could not be generated",
+                    Some(&format!("Failed to generate cross reference for \"{:?}\" because the referenced crate could not be found", ref_id)),
+                    None
+                )?;
```
jaysachs, line 818: Isn't this block identical to the similar block above? Should there be a `continue` here as well?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -731,62 +797,103 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
             let mut target_vname = template_vname.clone();
-            target_vname
-                .set_signature(format!("{}_def_{}", krate_signature, reference.ref_id.index));
+            target_vname.set_signature(self.create_def_signature(krate_id, ref_id.index));
             target_vname.set_language("rust".to_string());
 
             // Create VName for the reference node
-            let file_vname = self.file_vnames.get(span.file_name.to_str().unwrap());
+            let file_name = span.file_name.to_str().unwrap();
+            let file_vname = self.file_vnames.get(file_name);
```
jaysachs, line 829: It looks like every use of `self.file_vnames.get()` is based on a (rust analysis) `span`. Should we have  a method
`file_vname_from_span(span)` ?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,74 +701,191 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+            let krate_id_option = self.krate_ids.get(&ref_id.krate);
+            if krate_id_option.is_none() {
+                // This is a little bit of chicken and egg here. We could try
+                // and find the file vname first before the crate id, but then
+                // we wouldn't have a definition anchor to emit the diagnostic
+                // node to because we wouldn't have the crate id yet. The chance
+                // of a crate id not being in our HashMap is very low, so if this
+                // fails we just emit a diagnostic node to the first source file
+                // of the crate we are indexing
+                let file_vname = self.file_vnames.values().next().unwrap();
```
jaysachs, line 737: In the (degenerate?) case that we have no file vnames, this will panic, right?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,74 +701,191 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+
+            // Create VName for target of reference
+            let target_vname = self.create_def_vname(krate_id, ref_id.index);
+
+            // Create VName for the reference node
+            let file_name = span.file_name.to_str().unwrap();
+            let file_vname = self.file_vnames.get(file_name);
+            if file_vname.is_none() {
+                self.emitter.emit_diagnostic(
+                    &target_vname,
```
jaysachs, line 756: Diagnostics should be pointing from a file or anchor node; isn't target_vname the semantic node? (does this question apply elsewhere?) Maybe these variables just need better names to make that clear, i.e. rename `target_node` to `anchor_node`?

kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -719,74 +701,191 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
...
+                continue;
+            }
+            let krate_id = krate_id_option.unwrap();
+
+            // Create VName for target of reference
+            let target_vname = self.create_def_vname(krate_id, ref_id.index);
+
+            // Create VName for the reference node
+            let file_name = span.file_name.to_str().unwrap();
+            let file_vname = self.file_vnames.get(file_name);
```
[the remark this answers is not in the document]
wcalandro: Ah that function seems to be obsolete, it has been removed

kythe/rust/indexer/src/indexer/analyzers.rs
wcalandro: Renamed the variable and made sure to convert to an anchor vname before emitting the diagnostic

PR #5123, opened 2021-11-03
feat(schema): add tvar/tparam entries

schroederc:
This node/edge pair is expected to replace and simplify the existing `abs` structure in the Kythe graph.


shahms (approved):

kythe/docs/schema/schema.txt
```diff
@@ -176,23 +176,23 @@ bounded/upper or bounded/lower
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Brief description::
-  <<absvar>> A is *bounded/upper* by B when A is constrained to be a subtype
-  of B. <<absvar>> A is *bounded/lower* by B when A is constrained to be a
-  supertype of B.
+  <<absvar>>/<<tvar>> A is *bounded/upper* by B when A is constrained to be a
```
jaysachs, line 179: Should we be stronger and indicate that absvar is deprecated and will be replaced?

zrlk (approved):

kythe/docs/schema/schema.txt
creachadair: I'm not sure I understand the difference. Is the idea that we're bypassing an indirection through the `abs` to get to the parameters from the type?

schroederc: Yeah, we're removing `abs` nodes altogether.  What was their child node becomes the binder for the `tvar` (nee `absvar`) nodes.

PR #5127, opened 2021-11-10
feat(schema): add completedby and its implementation in C++

zrlk:
We're deprecating completes and completes/uniquely in favor of
a simpler, completely semantic edge. This new edge also has the
advantage of reducing fanout.


shahms (approved):

creachadair:
I like this change. I have a vague memory that we talked about something similar to this approach earlier on, but I can no longer remember why we decided not to do it.

Or possibly my memory is failing me. Either way, though, 👍🏼

PR #5130, opened 2021-11-11
feat(serving): offer a callback to rewrite edge labels

zrlk:
This will help with turnup of new labels.

### DIFF
--- a/kythe/go/serving/xrefs/xrefs.go
+++ b/kythe/go/serving/xrefs/xrefs.go
@@ -93,22 +93,59 @@ type SplitTable struct {
 
 	// Documentation is a table of srvpb.Documents keyed by their node ticket.
 	Documentation table.Proto
+
+	// EdgeFilter is an optional callback to rewrite edge labels.
+	EdgeFilter func(context.Context, *string)
+}
+
+func (s *SplitTable) filterFileDecorations(ctx context.Context, fd *srvpb.FileDecorations, err error) (*srvpb.FileDecorations, error) {
+	if fd == nil || err != nil || s.EdgeFilter == nil || fd.Decoration == nil {
+		return fd, err
+	}
+	for _, d := range fd.Decoration {
+		s.EdgeFilter(ctx, &d.Kind)
+	}
+	return fd, err
+}
+
+func (s *SplitTable) filterCrossReferencesGroup(ctx context.Context, g *srvpb.PagedCrossReferences_Group) {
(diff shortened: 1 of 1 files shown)


kythe/go/serving/xrefs/xrefs.go
```diff
@@ -93,22 +93,67 @@ type SplitTable struct {
 
 	// Documentation is a table of srvpb.Documents keyed by their node ticket.
 	Documentation table.Proto
+
+	// EdgeFilter is an optional callback to rewrite edge labels.
+	// It will be called once per request; the function it returns will then be
+	// called once per edge.
+	EdgeFilter func(context.Context) func(*string)
```
salguarnieri, line 100: nit: this is more of a map than a filter

creachadair: Could just call it "RewriteEdgeLabel", since that's what it does. 🙂

kythe/go/serving/xrefs/xrefs.go
```diff
@@ -93,22 +93,67 @@ type SplitTable struct {
 
 	// Documentation is a table of srvpb.Documents keyed by their node ticket.
 	Documentation table.Proto
+
+	// RewriteEdgeLabel is an optional callback to rewrite edge labels.
+	// It will be called once per request; the function it returns will then be
+	// called once per edge.
+	RewriteEdgeLabel func(context.Context) func(*string)
```
salguarnieri, line 100: Why is the callback func(*string) instead of func(string) string?

zrlk: The rewrite* functions all rewrite messages and fields in place.

salguarnieri: Where are the existing rewrite functions?

zrlk: They're added in this PR. If something needs to be changed please let me know.

salguarnieri: Sorry, your answer to my question made it sound like you didn't have control over the rewrite functions or were following existing behavior.

Why are you rewriting in place instead of assigning a new value to the field? It seems easier for clients to write a function that goes from string to string than change the contents of an existing string.

zrlk: No worries. I've changed the type in the most recent push.

salguarnieri (approved):

Issue #5133, opened 2021-11-23
How can support for macos?


Qchrlis:

zrlk:
This is blocking on https://github.com/abseil/abseil-cpp/issues/1067

wcalandro:
Closing due to the previously mentioned commit

PR #5134, opened 2021-11-24
fix: the mac build

zrlk:
Macs are (mostly) case-insensitive. A change to LLVM resulted in
a file called CXXABI.h being visible in the system search path
ahead of the actual cxxabi.h. This caused certain Abseil libraries
to fail to build as they depend on the content of that header.

This PR pushes Abseil to head (why not?), LLVM to head (as they
fixed this problem upstream), and rules_cc to head (because the
LLVM fix requires a more recent version).

Addresses #5133

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -160,8 +160,8 @@ def _cc_dependencies():
         github_archive,
         name = "com_google_absl",
         repo_name = "abseil/abseil-cpp",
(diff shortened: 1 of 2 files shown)


zrlk:
This is the fix to the non-Abseil issue https://github.com/abseil/abseil-cpp/issues/1067 (mentioning in case someone gets there via search)

shahms (approved):

PR #5136, opened 2021-12-02
feat(typescript_indexer): emit tvar and tparam.N

zrlk:
This PR is part of the schema migration away from abs/absvars.
The TypeScript indexer will emit both the (incomplete, as it never
used abs nodes) old artifacts as well as the new artifacts.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -128,11 +128,14 @@ function stripExtension(path: string): string {
  *   https://www.typescriptlang.org/docs/handbook/declaration-merging.html
  * for a listing of namespace groups for various declaration types and further
  * discussion.
+ *
+ * TYPE_MIGRATION is a temporary namespace to be used during the tvar migration.
  */
 export enum TSNamespace {
   TYPE,
   VALUE,
   NAMESPACE,
+  TYPE_MIGRATION,
 }
 
 /**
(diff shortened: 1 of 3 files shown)


kythe/typescript/indexer.ts
```diff
@@ -782,9 +787,11 @@ class Visitor {
   }
 
   /** newAnchor emits a new anchor entry that covers a TypeScript node. */
-  newAnchor(node: ts.Node, start = node.getStart(), end = node.end): VName {
+  newAnchor(node: ts.Node, start = node.getStart(), end = node.end, tag = ''):
+      VName {
     const name = Object.assign(
-        {...this.kFile}, {signature: `@${start}:${end}`, language: LANGUAGE});
+        {...this.kFile},
+        {signature: `@${tag}${start}:${end}`, language: LANGUAGE});
```
schroederc, line 794: What's the purpose of distinguishing the anchor defining the two type param nodes?

schroederc (approved):

kythe/typescript/indexer.ts
zrlk: I wanted to keep the two subgraphs entirely parallel. Does this make any difference after denormalization? ISTR exactly overlapping anchors get special treatment somewhere.

schroederc: It'll be fine in post-processing.  Only `imputes` edges matter when it comes to overlapping anchor spans.

PR #5140, opened 2021-12-06
Update BUILD since $(PLUGIN_OUT) is removed and $(OUT) placeholder replaced.

buildbreaker2021:


meteorcloudy:
Thanks! Can you look into the presubmit failure?

shahms:
This file is part of our release; in which released version of Bazel is this change live?

meteorcloudy:
https://github.com/bazelbuild/bazel/commit/d69a55cf7b64df1f8a86a31e35a8a0b10cabeabc is only available at HEAD right now, it's not cut into Bazel 5.0. So probably in later rolling releases and 6.0.

shahms:
I don't see how I can merge this, then. While building Kythe at Bazel HEAD will fail without this change, doing so with an older version will silently do the wrong thing.  While getting rid of `$(PLUGIN_OUT)` seems feasible, changing `$(OUT)` to "%s" is a breaking change with no available workaround.

kythe/extractors/BUILD
```diff
@@ -51,7 +51,7 @@ proto_lang_toolchain(
         "@com_google_protobuf//:type_proto",
         "@com_google_protobuf//:wrappers_proto",
     ],
-    command_line = "--$(PLUGIN_OUT)=:$(OUT)",
+    command_line = "--PLUGIN_java_stubby_out=:%s",
```
shahms, line 54: AFAIK, this should be `--PLUGIN_C++_out` (although actually *determining* that, without the aid of the `$(PLUGIN_OUT)` replacement) is incredibly challenging.  I had to run `bazel -s` to do so.

meteorcloudy:
@buildbreaker2021 Is there any easy way to migrate this in a way that works with both newer and older Bazel? Otherwise, I think we probably should rollback bazelbuild/bazel@d69a55c first, and reintroduce it in a way that's easier to migrate for.

buildbreaker2021:
Will try something first.

shahms:
I believe this works, thanks!

shahms (approved):

PR #5142, opened 2021-12-07
chore(cxx_indexer): fork (most) template/generic tests into abs/tvar

zrlk:
cp -R'd the relevant directories.
tvar_template/template_file_call.cc had to be updated
because it matches on a vname path.


shahms (approved):
Man this is painful without copy-detection enabled.

PR #5143, opened 2021-12-08
feat(cxx_indexer): emit tvar/tparam edges

zrlk:


kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -578,6 +578,13 @@ class GraphObserver {
   virtual void recordAbsVarNode(
       const NodeId& Node, const absl::optional<MarkedSource>& MarkedSource) {}
 
+  /// \brief Records a node representing a variable in a dependent type
```
shahms, line 581: nit: C++ allows abstract variables, e.g.
```
template <typename T>
T pi = 3.14f;
```

So this won't necessarily be a "type" abstraction nor itself a type (for non-type template parameters).

kythe/cxx/indexer/cxx/testdata/objc/tvar_generics/generic_bound.m
```diff
@@ -19,10 +19,10 @@ @implementation O2
 // No source range defines BoxDecl since this is a generic type.
 //- @Type defines/binding TypeVar
 //- @Box defines/binding BoxAbs
-//- TypeVar.node/kind absvar
+//- TypeVar.node/kind tvar
 //- BoxDecl childof BoxAbs
 //- BoxAbs.node/kind abs
```
shahms, line 24: Shouldn't we no longer emit the `abs` nodes in this case?

zrlk: In the fullness of time, yes. As work proceeds on implementing the new model I'll continue to update the forked tests. Right now only tvar/tparam are implemented.

kythe/cxx/indexer/cxx/GraphObserver.h
zrlk: In `template <unsigned U> auto v = U;`, you can imagine the integer `u` substituted for `U` as `u` reified to a singleton type; C<1> and C<2> are not miscible because their types are different. I think it's valid to refer to tvars as variables at the type level.

shahms (approved):

PR #5167, opened 2022-01-06
feat(tooling): Support filtering of kzips by language.

nbeloglazov:
Adds new param to `kzip filter` command that allows to filter compilation
units by list of languages. Also adds a build rule `filter_kzip` similar
to `merge_kzips` to allow filtering on build level. This is useful when
kzips come from targets that produce multiple compilation units for
2 or more languages and we need only one of those languages.


kythe/go/platform/tools/kzip/filtercmd/filtercmd.go
```diff
@@ -37,9 +38,10 @@ import (
 type filterCommand struct {
 	cmdutil.Info
 
-	input    string
-	output   string
-	encoding flags.EncodingFlag
+	input     string
+	output    string
+	encoding  flags.EncodingFlag
+	languages flagutil.StringList
```
shahms, line 44: I think you want `flagutil.StringSet`. Obviates the subsequent conversion to `stringset.Set`.

nbeloglazov: I tried but I'm getting these errors:
```
kythe/go/platform/tools/kzip/filtercmd/filtercmd.go:91:34: languages.Empty undefined (type flagutil.StringSet has no field or method Empty)
kythe/go/platform/tools/kzip/filtercmd/filtercmd.go:97:23: languages.Empty undefined (type flagutil.StringSet has no field or method Empty)
kythe/go/platform/tools/kzip/filtercmd/filtercmd.go:99:21: languages.Contains undefined (type flagutil.StringSet has no field or method Contains)
```

And when I'm trying to use StringSet as Set (just to see if that works) I get this:

```
var languages stringset.Set = c.languages
kythe/go/platform/tools/kzip/filtercmd/filtercmd.go:89:32: cannot use c.languages (variable of type flagutil.StringSet) as type stringset.Set in variable declaration
````

Am I missing some basic golang understanding?

Comment (shahms) [reply kythe/go/platform/tools/kzip/filtercmd/filtercmd.go:44]: Right, you'll need to cast it to a `stringset.Set` for example: https://github.com/kythe/kythe/blob/145859480b64849c1911cebfc09c42be69835476/kythe/go/test/tools/empty_corpus_checker/empty_corpus_checker.go#L86

The flag type doesn't have an `Empty` method, but does have `Len`.  It may be reasonable to add `Empty` though.

**Comment by shahms** [comment]
The lint failures are buildifier related; should just be able to run buildifier to resolve them.

**Review by shahms — APPROVED** [review]

**Comment by shahms** [comment]
Looks like one more buildifier nit:
```
--- a/kythe/go/platform/tools/kzip/filtercmd/BUILD
+++ b/kythe/go/platform/tools/kzip/filtercmd/BUILD
@@ -9,8 +9,8 @@ go_library(
         "//kythe/go/platform/kzip",
         "//kythe/go/platform/tools/kzip/flags",
         "//kythe/go/platform/vfs",
-        "//kythe/go/util/flagutil",
         "//kythe/go/util/cmdutil",
+        "//kythe/go/util/flagutil",
         "@com_github_google_subcommands//:go_default_library",
         "@org_bitbucket_creachadair_stringset//:go_default_library",
     ],
```

PR #5168, opened 2022-01-07
chore: add context to kzip error message

salguarnieri:


kythe/go/platform/kzip/kzip.go
```diff
@@ -191,7 +191,7 @@ func NewReader(r io.ReaderAt, size int64) (*Reader, error) {
 	if len(archive.File) == 0 {
 		return nil, errors.New("archive is empty")
 	} else if fi := archive.File[0].FileInfo(); !fi.IsDir() {
-		return nil, errors.New("archive root is not a directory")
+		return nil, fmt.Errorf("archive root directory missing: expected a 'root' directory but got %v", archive.File[0].Name)
```
shahms, line 194: There is no specific requirement that the root be called `root`, just that there is a single root directory and that it be the first entry in the zip.

salguarnieri: I updated the message, WDYT?

This is coming from a bug where they didn't explicitly include the directory in the zip. The files inside the zip had the right path, but there wasn't an entry for the root directory.

shahms (approved):

PR #5169, opened 2022-01-07
feat(go extractor): add flag to put deps in default corpus

justbuchanan:
This will be used in our extraction of repos for the oss pipeline, where we want all extracted items to be assigned to the main/default corpus.


jaysachs (commented):
Is it possible to test this behavior?

justbuchanan:
I added a test to `govname_test.go`.

jaysachs (approved):

justbuchanan:
This will be used in our extraction of repos for the oss pipeline, where we want all extracted items to be assigned to the main/default corpus.

PR #5170, opened 2022-01-10
docs: clarify that "value" is the `value` field and not the message's value

yannack:
This created confusion for our team. Moving the comment closer to the definition of the `value` field, and using backticks makes it somewhat clearer what the comment is referring to.

### DIFF
--- a/kythe/proto/analysis.proto
+++ b/kythe/proto/analysis.proto
@@ -18,13 +18,13 @@ syntax = "proto3";
 
 package kythe.proto;
 
-option go_package = "analysis_go_proto";
-option java_package = "com.google.devtools.kythe.proto";
-option cc_enable_arenas = true;
-
 import "google/protobuf/any.proto";
 import "google/protobuf/timestamp.proto";
 import "kythe/proto/storage.proto";
+
+option go_package = "analysis_go_proto";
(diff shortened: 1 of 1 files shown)


shahms (approved):

kythe/proto/analysis.proto
```diff
@@ -48,10 +48,9 @@ message AnalysisRequest {
 // place.  A given analysis may not produce any outputs.  It is okay for an
 // indexer to send an empty AnalysisOutput message if needed to keep the RPC
 // channel alive; the driver must correctly handle this.
-//
-// The format of value is determined by the analyzer. Kythe language indexers
-// emit wire-format kythe.proto.Entry messages.
 message AnalysisOutput {
+  // The format of `value` is determined by the analyzer. Kythe language
```
creachadair, line 52: FWIW Go style eschews use of markup in comments.

yannack: This isn't Go code, but proto though, so is the Go style relevant?

creachadair: > This isn't Go code, but proto though, so is the Go style relevant?

No, except perhaps as a note for me to read more closely. 😆

PR #5179, opened 2022-01-13
chore: Allow duplicate CUs in kzip merge command

salguarnieri:


kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -62,6 +63,7 @@ func (c *mergeCommand) SetFlags(fs *flag.FlagSet) {
 	fs.Var(&c.encoding, "encoding", "Encoding to use on output, one of JSON, PROTO, or ALL")
 	fs.BoolVar(&c.recursive, "recursive", false, "Recurisvely merge .kzip files from directories")
 	fs.Var(&c.rules, "rules", "VName rules to apply while merging (optional)")
+	fs.BoolVar(&c.ignoreDuplicateCUs, "ignore_duplicate_cus", false, "Do not fail if we try to add the same CU twice")
```
jaysachs, line 66: Is it ambiguous if we leave off the "_cus" part of the flag?

jaysachs (approved):

kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
salguarnieri: We are merging kzips, so it might be possible to think we ignore duplicate kzips. But the end results would be the same.

PR #5195, opened 2022-01-25
fix(go_indexer): fix refs to anonymous members across packages

schroederc:


kythe/go/indexer/indexer.go
```diff
@@ -638,7 +638,17 @@ func (pi *PackageInfo) newSignature(obj types.Object) (tag, base string) {
...
-	return topLevelTag, fmt.Sprintf("[%p].%s", obj, obj.Name())
+	return topLevelTag, pi.anonSignature(obj)
+}
+
+func (pi *PackageInfo) anonSignature(obj types.Object) string {
+	// Use the object's line number and file basename to differentiate the
+	// node while allowing for cross-package references (other parts of the
+	// Position may differ).  This may collide if a source file isn't gofmt'd
+	// and defines multiple anonymous fields with the same name on the same
+	// line, but that's unlikely to happen in practice.
```
shahms, line 649: It seems like it would be possible to track and use the number of anonymous fields defined on the same line and incorporate the index if it comes to that. I hope it doesn't come to that.

shahms (approved):

kythe/go/indexer/indexer.go
schroederc: I don't think a dependent package would be able to do the same count.

shahms: Ah, yeah, good point.

PR #5197, opened 2022-01-26
feat(rust_indexer): add flag to disable emitting xrefs for stdlib

wcalandro:
Open Source Code Search has a strange bug where, occasionally, anchors for cross-references can be clickable even when the definition node (including an anchor node) for the target of the cross reference does not exist. Clicking the cross reference anchor brings up the References panel which shows an error because the definition node does not exist (we don't index the Rust standard library, so we can't emit definition nodes and anchors for it). An example of this is here: https://cs.opensource.google/fuchsia/fuchsia/+/main:src/connectivity/network/dns/src/main.rs;l=30;bpv=0;bpt=1 (hover over the "std" and try clicking on it)

This PR adds an optional flag to the Rust indexer that disables emitting any cross references to the standard library, which should mitigate this issue.

### DIFF
--- a/kythe/rust/indexer/BUILD
+++ b/kythe/rust/indexer/BUILD
@@ -36,6 +36,7 @@ rust_binary(
     deps = [
         ":kythe_rust_indexer",
         "//kythe/rust/cargo:anyhow",
+        "//kythe/rust/cargo:clap",
         "@rules_rust//proto/raze:protobuf",
     ],
 )
@@ -52,6 +53,7 @@ rust_binary(
         "//kythe/proto:analysis_rust_proto",
         "//kythe/rust/cargo:anyhow",
         "//kythe/rust/cargo:base64",
+        "//kythe/rust/cargo:clap",
         "//kythe/rust/cargo:serde_json",
         "@rules_rust//proto/raze:protobuf",
     ],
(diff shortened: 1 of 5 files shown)


jaysachs (approved):

creachadair:
This is a reasonable change, but it's worth noting for posterity that we originally considered this behaviour to be a feature, not a bug: Even if a definition site wasn't indexed, it can be useful for the user to see xrefs for other uses of the same thing.

(Internal CodeSearch didn't cope with this super gracefully, as it was oriented toward definition sites—but the open-source UI seems to provide an option)

jaysachs:
> This is a reasonable change, but it's worth noting for posterity that we originally considered this behaviour to be a feature, not a bug: Even if a definition site wasn't indexed, it can be useful for the user to see xrefs for other uses of the same thing.
> 
> (Internal CodeSearch didn't cope with this super gracefully, as it was oriented toward definition sites—but the open-source UI seems to provide an option)

That's true in general. For reasons we haven't determined yet, we're not seeing cross references for that symbol. Also note that due to constraints on the current Rust indexer implementation, we don't get all `std` references during indexing.

Arguably, the utility of finding "uses of the std crate" is low, so in this specific case there seems to be no huge loss.

Ideally, in the fullness of time, we'll index the Rust stdlib and then this flag will just be off permanently. (And also figure out the bug in the UI pathways.)

PR #5202, opened 2022-02-01
fix(rust_indexer): emit built-in types in the same corpus as the CU

wcalandro:
Previously, built-in types in Rust were emitted under the "std" corpus. This PR changes this behavior so that the nodes are emitted with a corpus identical to the parent Compilation Unit


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -240,7 +240,8 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
         ];
         let mut type_vnames: HashMap<String, VName> = HashMap::new();
         let mut vname_template = VName::new();
-        vname_template.set_corpus("std".to_string());
+        let unit_corpus = unit_vname.get_corpus().to_string();
+        vname_template.set_corpus(unit_corpus);
```
shahms, line 244: Please make this flag controlled.

shahms (approved):

PR #5208, opened 2022-02-04
fix(rust_common): update to new rules_rust and fix references to deprecated code

wcalandro:
This PR migrates Kythe to the new implementation of initializing the Rust toolchain and changes references to the internals of the toolchain.


shahms:
Seems like rustc can't find libc?

wcalandro:
Only seems to happen with this test target in fastbuild. It's most likely to do with the fact that, without "-c opt", the C++ libraries are built with "-fPIC" which seems to be confusing the Rust linker inside RBE only. Something must have changed in rules_rust between our old version and HEAD.
  
  
  

  
  
>

shahms (approved):

PR #5229, opened 2022-03-01
fix(doc): update description and examples for extends edge kind

jaysachs:


shahms (approved):

shahms (approved):

kythe/docs/schema/schema.txt
```diff
@@ -631,7 +631,8 @@ extends
 ~~~~~~~
 
 Brief description::
-  A *extends* B if A is a direct nominal subtype of B.
+  A *extends* B if A explicitly derives from B. It neither implies nor excludes
+  any type relationship between A and B.
```
creachadair, line 635: I feel like these two sentences contradict each other: If A explicitly derives from B, isn't that a type relationship?

(Was the intent, maybe, "any _other_ type relationship"? I'm trying to think of an example to illustrate the case, but I can't think of a good one)

shahms: This is primarily intended to address cross-language extensions from generated code.  A Java interface, which is used to generate JavaScript which then implements that interface.  Java and JavaScript don't share a type domain, but the one does logically extend the other (via the generated intermediary).

jaysachs: *extends* already specifically includes C++ private inheritance, which does not imply a type relationship. We also want to leave this open to include cross-lingual relationships, where "subtyping" isn't applicable. (Philosophically there may be some type-like relationship, but not formally.)

PR #5247, opened 2022-03-29
fix(cxx_indexer): remove another source of corpusless nodes

justbuchanan:
In some instances, a c++ symbol/node is not associated with a file and ends up with an empty corpus. This seems to happen when a macro is defined on the command line (i.e. "-DMY_MACRO=1"), then later referred to in the source code. This change sets the corpus to the default value (from the compilation unit) instead.

TODO: test this - it works in google3 localrun, but getting the issue to reproduce in the oss project is tricky


shahms:
You can use the `copts` attribute of `cc_indexer_test` to pass arbitrary compiler flags, which should include `-D` and that should enable you to write a normal verifier test for this.

justbuchanan:
Ok, I finally got the test working to a point that it actually fails when I remove the fix in `KytheGraphObserver.cc`. Interestingly it's not enough to just define the macro on the command line. You have to define it, undefine it, then define it again... Maybe this test case can be simplified and it's likely there's a deeper underlying bug here.

PR #5259, opened 2022-04-19
chore(javascript): run `npm audit fix --force`

justbuchanan:
This updates packages to fix some known known vulnerabilities. Should clean up a couple of our dependabot alerts.

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -18,6 +18,7 @@ load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_regi
 load("@rules_rust//tools/rust_analyzer:deps.bzl", "rust_analyzer_deps")
 load("@rules_rust//proto:repositories.bzl", "rust_proto_repositories")
 load("@build_bazel_rules_nodejs//:index.bzl", "npm_install")
+load("@build_bazel_rules_nodejs//:repositories.bzl", "build_bazel_rules_nodejs_dependencies")
 load(
     "@bazelruby_rules_ruby//ruby:deps.bzl",
     "rules_ruby_dependencies",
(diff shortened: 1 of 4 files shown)


shahms:
Error: Expected package major version to equal @build_bazel_rules_nodejs major version
    @bazel/jasmine - 5.4.0  
    @build_bazel_rules_nodejs - 3.2.3
  See https://github.com/bazelbuild/rules_nodejs/wiki/Avoiding-version-skew

justbuchanan:
I tried updating @build_bazel_rules_nodejs and now there's a dependency cycle somewhere.

```
ERROR: Failed to load Starlark extension '@rules_nodejs//nodejs:providers.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @rules_nodejs
This could either mean you have to add the '@rules_nodejs' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
INFO: Elapsed time: 0.243s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

I tried grepping for "@rules_nodejs" and I'm not seeing it anywhere.

shahms:
It looks like upstream substantially changed the installation procedure (or at least the Starlark file interdependencies) such that we'd need a 3-phase setup routine to keep working.  Right now, we want to do all of the setup in setup.bzl then use the repositories defined there in external.bzl, but the nodejs rules require defining a bunch of their own repositories via the `build_bazel_rules_nodejs_dependencies` macro. Bazel external dependency management remains terrible.

shahms:
Updating rules_nodejs remains a huge PITA. It's likely you'll either need to do so incrementally (to 3.8.0 I think?) or do a larger removal and migration to the current documented practices.

PR #5260, opened 2022-04-19
fix(kzip_info): don't count c++ stdlib includes as invalid absolute paths

justbuchanan:
In general, paths in the kzip should be relative (to the root of the repo), so we count absolute paths as errors. in some cases like c++ stdlib includes, these are expected and should not count as errors. There are likely other exceptions to the rule that we'll want to add in the future.

### DIFF
--- a/kythe/go/platform/kzip/info/info.go
+++ b/kythe/go/platform/kzip/info/info.go
@@ -86,8 +86,10 @@ func (a *Accumulator) Accumulate(u *kzip.Unit) {
 	var srcCorpora stringset.Set
 	srcsWithRI := stringset.New()
 	for _, ri := range u.Proto.RequiredInput {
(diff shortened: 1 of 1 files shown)


shahms:
Even the C++ standard library includes should be relative; we need to update the various vnames.json to ensure this is the case. IIRC we put them in the `/` root (or could invent something like `<system>`), but they should still

1) be relative
2) not have an empty root

shahms:
(Pretty sure we should do something similar for the `kythe_builtins` as well.)

justbuchanan:
Sounds good, thanks. A number of kzips that are built by other teams currently contain a lot of files in /usr/include/c++, so I guess we'll need to ask them to update their vnames.json configs.

shahms:
Right, I think the default vnames.json we ship don't always do the right thing here.

PR #5262, opened 2022-04-24
feat(typescript_indexer): emit edges from object literals to corresponding types

nbeloglazov:
This change emits `ref` edges from properties on object literals and destructuring to fields of the corresponding types, if type is known. Example:

```typescript
interface Person {
  name: string;
}

const p: Person = {name: 'Alice'};
const {name} = p;
```
With this change there will be two new `ref` edges from `name` properties in object literal and destructuring to `Person.name` field. 

This change supports object literals returned from functions, passed as function arguments and type casting. See examples in `testdata/interface.ts` test. It doesn't support more tricky cases that require deeper type checking, for example the following case won't have refs:

```typescript
const p: Person = isAlice ? {name: 'Alice'} : {name: 'Bob'};
```

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -759,6 +759,9 @@ class Visitor {
   /** influencers is a stack of influencer VNames. */
   private readonly influencers: Set<VName>[] = [];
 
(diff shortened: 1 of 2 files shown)


jaysachs:
go/gcbrun

PR #5263, opened 2022-04-27
fix(rust_indexer): set corpus+lang on diagnostic

wcalandro:
Previously, the Rust indexer didn't set the source corpus and language on the vname for the diagnostic. Fixed this and added tests.


kythe/rust/indexer/BUILD
```diff
@@ -75,6 +75,8 @@ rust_test(
         ":kythe_rust_indexer",
         "//kythe/proto:analysis_rust_proto",
         "//kythe/proto:storage_rust_proto",
+        "//kythe/rust/cargo:sha2",
+        "//kythe/rust/cargo:hex",
```
shahms, line 79: Run buildifier?

kythe/rust/indexer/src/indexer/entries.rs
```diff
@@ -139,6 +139,9 @@ impl<'a> EntryEmitter<'a> {
         let mut diagnostic_vname = VName::new();
         let source_signature = source_vname.get_signature();
         diagnostic_vname.set_signature(format!("{}_{}", source_signature, sha256sum));
+        // Use the corpus and language from the source_vname
+        diagnostic_vname.set_corpus(source_vname.get_corpus().to_string());
+        diagnostic_vname.set_language(source_vname.get_language().to_string());
```
shahms, line 144: It seems like this should be a copy of the source_vname, with a modified signature rather than a new vname with individually copied fields?

shahms (approved):

PR #5269, opened 2022-04-30
feat(verifier): add a flag to elide nodes not bound as goals

shahms:
For even simple code the resulting graph can be pretty daunting.  While there are definitely circumstances which call for viewing that graph in all its glorious complexity, it makes producing minimal representations quite challenging.  These minimal representations aid in understanding the specific corner of the graph being investigated without all of the associated clutter.

This PR adds a `--minimal_graphviz` which includes only the nodes bound to goals and the edges between them.

### DIFF
--- a/kythe/cxx/verifier/verifier.h
+++ b/kythe/cxx/verifier/verifier.h
@@ -183,7 +183,7 @@ class Verifier {
   void ShowAnchors() { show_anchors_ = true; }
 
   /// \brief Elide unlabeled nodes from graph dumps.
-  void ElideUnlabeled() { show_unlabled_ = false; }
+  void ElideUnlabeled() { show_unlabeled_ = false; }
 
   /// \brief Check for singleton EVars.
   /// \return true if there were singletons.
--- a/kythe/cxx/verifier/verifier.h
+++ b/kythe/cxx/verifier/verifier.h
@@ -183,7 +183,7 @@ class Verifier {
   void ShowAnchors() { show_anchors_ = true; }
 
   /// \brief Elide unlabeled nodes from graph dumps.
-  void ElideUnlabled() { show_unlabled_ = false; }
+  void ElideUnlabeled() { show_unlabled_ = false; }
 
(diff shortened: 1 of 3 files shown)


kythe/cxx/verifier/verifier.h
```diff
@@ -336,6 +339,9 @@ class Verifier {
   /// If true, show anchor locations in graph dumps (instead of @).
   bool show_anchors_ = false;
 
+  /// If true, show unlabeled nodes in graph dumps.
+  bool show_unlabled_ = true;
```
creachadair, line 343:
```suggestion
  bool show_unlabeled_ = true;
```

kythe/cxx/verifier/verifier_main.cc
```diff
@@ -91,6 +93,11 @@ invocation and rule syntax.
     v.SaveEVarAssignments();
   }
 
+  if (absl::GetFlag(FLAGS_minimal_graphviz)) {
+    v.SaveEVarAssignments();
+    v.ElideUnlabled();
```
creachadair, line 98:
```suggestion
    v.ElideUnlabeled();
```

kythe/cxx/verifier/verifier.h
```diff
@@ -182,6 +182,9 @@ class Verifier {
   /// \brief Show anchor locations in graph dumps (instead of @).
   void ShowAnchors() { show_anchors_ = true; }
 
+  /// \brief Elide unlabeled nodes from graph dumps.
+  void ElideUnlabled() { show_unlabled_ = false; }
```
creachadair, line 186:
```suggestion
  void ElideUnlabeled() { show_unlabled_ = false; }
```

creachadair (commented):
Just s drive-by. I like this change.

kythe/cxx/verifier/verifier.h
shahms: That's what I get for coding on a weekend, thanks!

wcalandro (approved):

PR #5277, opened 2022-05-14
feat(typescript_indexer): merge nodes where define and ref belong to same anchor

nbeloglazov:
Adds additional step in TS indexer. Edges and facts are not immediately emitted,
instead they are kept in memory and at the end of the indexing we run post processing
which might merge, remove and otherwise modify output data.

Also this commit implements specific post processing - merging of nodes which
are defined at the same location where some other node is referenced. Example:

```
AnchorA ==ref==> NodeA
AnchorB ==defines/binding==> NodeA
AnchorB ==ref==> NodeB
```

It will remove NodeA and all edges that point to NodeA will point to NodeB.
So in this case final data will be:

```
AnchorA ==ref==> NodeB
AnchorB ==ref==> NodeB
```


shahms:
Do you have an example of where this occurs in TS? The transformation isn't generally correct, but may be fine here. For example, in C++ this pattern occurs fairly regularly with classes and their constructors where replacing a reference to the constructor with a reference to the class is incorrect (same with Java and a few others). This is what the `ref/id` edge exists for.

nbeloglazov:
Ignore this PR for now. if we can surface ref/id refs in UI (currently they are not visible) - then I'll revert this PR. And there is a internal thread about it.

PR #5278, opened 2022-05-14
Update metadata.proto

good-nife:
Add comment to kythe.proto.MappingRule.end about it being exclusive rather than inclusive.


schroederc:
Closing extremely stale PR.

PR #5279, opened 2022-05-18
fix(rust_indexer): clean relative paths in indexer to match vname paths

wcalandro:
Modules can be defined with a custom path that can be relative to the current directory. This causes the save-analysis to include file names that have `..` in them, which causes the indexer to fail to find the file VName. This PR adds functionality to clean the file paths from the spans in the save-analysis so that the paths match file VNames.

A dependency on the path-clean crate has been added to provide this functionality.


kythe/rust/indexer/src/indexer/analyzers.rs
```diff
@@ -417,7 +418,8 @@ impl<'a, 'b> CrateAnalyzer<'a, 'b> {
         let defs = self.analysis.defs.clone();
 
         for def in &defs {
-            let file_vname = self.file_vnames.get(def.span.file_name.to_str().unwrap());
+            let file_name = clean(def.span.file_name.to_str().unwrap());
```
shahms, line 421: As `def.span.file_name` should be a `PathBuf` I believe you can:

```
use path_clean::PathClean;
...
let file_name = def.span.file_name.clean();
```
directly

kythe/rust/indexer/testdata/relative/relative_module/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+//- RelativeMod.node/kind record
+//- RelativeMod.subkind module
+//- RelativeMod.complete definition
+
+//- @print_hello_world defines/binding FnPrint
+//- FnPrint.node/kind function
+//- FnPrint.complete definition
+pub fn print_hello_world() {
+    println!("Hello, world!");
+}
```
shahms, line 14: nit: add a newline at the end of the file

kythe/rust/indexer/testdata/relative/src/main.rs
```diff
@@ -0,0 +1,8 @@
+#[path = "../relative_module/mod.rs"]
+//- @relative_module ref RelativeMod
+mod relative_module;
+
+fn main() {
+    //- @print_hello_world ref FnPrint
+    relative_module::print_hello_world();
+}
```
shahms, line 8: ditto

kythe/rust/indexer/src/indexer/analyzers.rs
shahms: There also appears to be a recent fork of `path-clean` (called `clean-path`) which may be a better choice? I'm not good at evaluating Rust library quality, though.

wcalandro: I feel that `clean()` is a better use case here because it returns a String rather than another PathBuf.

kythe/rust/indexer/testdata/relative/relative_module/mod.rs
wcalandro: Thanks, I'm surprised pre-commit didn't catch this.

kythe/rust/indexer/src/indexer/analyzers.rs
shahms: That's fine. Arguably, all of this should be using Path or PathBuf (as they are paths) rather than string, but that's a bigger change.

shahms (approved):

PR #5281, opened 2022-05-20
feat(java_indexer): use ref/id in annotation utterance contexts

zrlk:
This PR makes it easier to distinguish actual Java annotation use sites from other places those annotations are used.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
@@ -148,6 +148,8 @@ public class KytheTreeScanner extends JCTreeScanner<JavaNode, TreeContext> {
 
   private KytheDocTreeScanner docScanner;
 
+  private boolean inRefIdContext = false;
+
   private KytheTreeScanner(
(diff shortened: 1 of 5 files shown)


kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
```diff
@@ -961,7 +963,11 @@ public JavaNode visitTypeArray(JCArrayTypeTree arrayType, TreeContext owner) {
   public JavaNode visitAnnotation(JCAnnotation annotation, TreeContext owner) {
     TreeContext ctx = owner.down(annotation);
     scanList(annotation.getArguments(), ctx);
-    return scan(annotation.getAnnotationType(), ctx);
+    boolean savedInRefIdContext = inRefIdContext;
+    inRefIdContext = true;
+    JavaNode node = scan(annotation.getAnnotationType(), ctx);
```
jaysachs, line 968: I don't think it's possible to have an annotation scoped "inside" an annotation? I guess the saving/restoring is the canonical pattern to use, though.

jaysachs (approved):

PR #5286, opened 2022-05-24
feat(verifier): allow simplifying graph output data

shahms:
This adds two additional verifier flags which control removing the `/kythe/(edge/)?` prefix and hiding node VNames when dumping graphviz graphs.  This results in much more compact and readable graphs (to say nothing of the dot file itself).

This also incidentally updates the PrinterPrinter interface to use absl::string_view instead of `const std::string&`

### DIFF
--- a/kythe/cxx/verifier/verifier_main.cc
+++ b/kythe/cxx/verifier/verifier_main.cc
@@ -53,6 +53,10 @@ ABSL_FLAG(std::string, goal_regex, "",
 ABSL_FLAG(bool, convert_marked_source, false,
           "Convert MarkedSource-valued facts to subgraphs.");
 ABSL_FLAG(bool, show_anchors, false, "Show anchor locations instead of @s");
+ABSL_FLAG(bool, show_vnames, true,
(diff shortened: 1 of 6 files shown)


zrlk (approved):

kythe/cxx/verifier/verifier_main.cc
```diff
@@ -53,6 +53,10 @@ ABSL_FLAG(std::string, goal_regex, "",
 ABSL_FLAG(bool, convert_marked_source, false,
           "Convert MarkedSource-valued facts to subgraphs.");
 ABSL_FLAG(bool, show_anchors, false, "Show anchor locations instead of @s");
+ABSL_FLAG(bool, show_vnames, true,
```
jaysachs, line 56: I wonder if these flag names (as well as other graphviz-related flags) ought to have a `dot_` or `graphviz_` (or `gv_`) prefix

shahms: Possibly? In theory they could apply to the JSON dumps as well (but don't currently).

PR #5290, opened 2022-05-26
fix(cxx_indexer): emit non-implicit refs to explicit template specializations

shahms:
Obnoxiously, we don't want *all* of the nodes visited during a traversal of a class template specialization to be within it's context.  For explicit instantiations, we want the test of the instantiation itself to remain explicit.

### DIFF
--- a/kythe/cxx/indexer/cxx/testdata/template/template_class_inst_explicit.cc
+++ b/kythe/cxx/indexer/cxx/testdata/template/template_class_inst_explicit.cc
@@ -4,6 +4,8 @@ using ExternalDef = int;
 template <typename T> struct C {
   using X = ExternalDef;
 };
+//- @C defines/binding CInstantiation
+//- @C ref CInstantiation
 template struct C<int>;
 //- ExplicitC.node/kind record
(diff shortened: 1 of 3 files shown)


kythe/cxx/indexer/cxx/testdata/template/template_class_inst_explicit.cc
```diff
@@ -4,6 +4,8 @@ using ExternalDef = int;
 template <typename T> struct C {
   using X = ExternalDef;
 };
+//- @C defines/binding CInstantiation
+//- @C ref CInstantiation
 template struct C<int>;
 //- ExplicitC.node/kind record
```
jaysachs, line 10: Any relationship (or non-relationship) to assert between `ExplicitC` and `CInstantiation` ? What about to `TemplateC`?

shahms: Fixed (they're the same nodes).  The relationship to `TemplateC` is covered in other tests.

zrlk (approved):

PR #5292, opened 2022-05-27
feat(typescript_indexer): use ref/id for destructuring variables

nbeloglazov:
It also affects object literal properties that reference inteface fields.
Those should use `override` instead but it requires extra work and will
be done as follow up.


jaysachs (approved):

jaysachs:
go/gcbrun

PR #5293, opened 2022-05-27
feat(typescript_indexer): emit `field` subkind fact for class/interface/object literal properties

nbeloglazov:


nbeloglazov:
@zrlk can you take a look?

zrlk (approved):

PR #5305, opened 2022-06-09
feat(java extractor): use default value if corpus is ambiguous

justbuchanan:
Before this change, java CUs don't receive a corpus if the required_inputs are not all in the same corpus. With this change, instead of assigning the empty corpus, the CU is assigned a corpus from the KYTHE_CORPUS environment variable.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
+++ b/kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
@@ -304,10 +304,13 @@ public class JavaCompilationUnitExtractor {
       unit.addSourceFile(sourceFile);
       sourceFileCorpora.add(inputCorpus.getOrDefault(sourceFile, ""));
     }
-    if (sourceFileCorpora.size() == 1) {
-      // Attribute the source files' corpus to the CompilationUnit if it is unambiguous.
-      unit.getVNameBuilder().setCorpus(Iterables.getOnlyElement(sourceFileCorpora));
-    }
+    // Attribute the source files' corpus to the CompilationUnit if it is unambiguous. Otherwise use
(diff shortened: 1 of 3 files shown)


kythe/java/com/google/devtools/kythe/extractors/java/JavaCompilationUnitExtractor.java
```diff
@@ -304,10 +304,13 @@ private CompilationUnit buildCompilationUnit(
...
-      // Attribute the source files' corpus to the CompilationUnit if it is unambiguous.
-      unit.getVNameBuilder().setCorpus(Iterables.getOnlyElement(sourceFileCorpora));
-    }
+    // Attribute the source files' corpus to the CompilationUnit if it is unambiguous. Otherwise use
+    // the default corpus.
+    String cuCorpus =
+        sourceFileCorpora.size() == 1
+            ? Iterables.getOnlyElement(sourceFileCorpora)
+            : fileVNames.getDefaultCorpus();
+    unit.getVNameBuilder().setCorpus(cuCorpus);
```
jaysachs, line 313: Maybe we should log these events somehow?

(ideally set something in the CU itself to say what happened, but I'm not sure we have any place for that.)

justbuchanan: I'm not opposed to logging this, but it does seem to be a fairly common occurrence in google3. Many j2cl compilation units include "google3" java files as well as several "j2cl" java files that are assigned different corpora based on file path.

jaysachs: Ah, ok, I guess. (I wonder about the extra corpora, but that's different.)

PR #5314, opened 2022-06-22
feat(go_indexer): add --override_stdlib_corpus flag

justbuchanan:
Problem: we index the go std library with the "google3/golib" corpus and the rest of go code with "google3". We want the cross-references to work between these, so std library elements in both need to have the same corpus.

For many projects we index, we want stdlib nodes to be assigned the compilation unit's corpus (enabled with --use_compilation_corpus_as_default), however in some cases we want to give stdlib items an explicit corpus so we can get cross-references to a shared instance of the std library.

### DIFF
--- a/kythe/go/extractors/govname/govname.go
+++ b/kythe/go/extractors/govname/govname.go
@@ -34,7 +34,7 @@ import (
 const (
 	pathTail     = `(?:/(?P<path>.+))?$`
 	packageSig   = "package"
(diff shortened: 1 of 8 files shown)


shahms:
Is it feasible to add tests for the overwriting behavior?

justbuchanan:
I added a new verifier test showing the main corpus being used for functions defined in the file and the override corpus used for imported stdlib items.

PR #5315, opened 2022-06-22
Generate compilation database for workspace

mai93:
Since [.ycm_extra_conf.py](https://github.com/kythe/kythe/blob/master/.ycm_extra_conf.py#L35) is calling `bazel print_action` directly, [generate_compilation_database.sh](https://github.com/kythe/kythe/blob/master/tools/cpp/generate_compilation_database.sh) is only called by [clang_tidy.sh](https://github.com/kythe/kythe/blob/master/tools/clang_tidy/clang_tidy.sh#L9) and this  call is broken as `clang_tidy` does not pass a certain file path to `generate_compilation_database.sh`. 

This PR fixes this by making `generate_compilation_database.sh` generate the compilation database for the project workspace and not require a file as input.

### DIFF
--- a/tools/cpp/generate_compilation_database.sh
+++ b/tools/cpp/generate_compilation_database.sh
@@ -5,7 +5,7 @@
 
 set -e
 
-TARGETS=($(bazel query 'kind(cc_.*, //...) - attr(tags, manual, //...)'))
+mapfile -t TARGETS < <(bazel query 'kind(cc_.*, //...) - attr(tags, manual, //...)')
 bazel build \
   --experimental_action_listener=//kythe/cxx/tools/generate_compile_commands:extract_json \
   --noshow_progress \
--- a/tools/cpp/generate_compilation_database.sh
+++ b/tools/cpp/generate_compilation_database.sh
@@ -5,12 +5,13 @@
 
 set -e
 
+TARGETS=($(bazel query 'kind(cc_.*, //...) - attr(tags, manual, //...)'))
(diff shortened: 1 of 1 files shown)


mai93:
Thanks @justbuchanan, I see that one of the checks is failing but I do not have permission to see the details of the failure.

justbuchanan:
Here's the relevant snippet from the logs:

```
Step #1: shellcheck...............................................................Failed
Step #1: - hook id: shellcheck
Step #1: - exit code: 1
Step #1: 
Step #1: In tools/cpp/generate_compilation_database.sh line 13:
Step #1:   $(bazel query 'kind(cc_.*, //...) - attr(tags, manual, //...)') > /dev/null
Step #1:   ^-- SC2046: Quote this to prevent word splitting.
Step #1: 
Step #1: For more information:
Step #1:   https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...
```

shahms:
Since you actually want word splitting to happen here, my preferred approach for that is to use a bash array:

```
TARGETS=($(bazel query ....))
bazel build ... "${TARGETS[@]}"
```

shahms:
Sigh; shellcheck doesn't like that and suggests using `mapfile`, which would look like:

```
mapfile -t TARGETS < <(bazel query ...)
```

Hopefully I remember this the next time it comes up. Thanks!

shahms (approved):

PR #5319, opened 2022-06-30
chore(java_indexer): delete deprecated getMethodVName() method

justbuchanan:
We should be using the other version that accepts a `CorpusPath` parameter in all places.


schroederc (changes requested):
Looks like this might be used in some indexers

justbuchanan:
The only usage I saw was in the soy indexer and I sent a cl to change it to use the other version of this method.

schroederc (approved):

PR #5323, opened 2022-07-19
feat(cxx_indexer): record full signatures in a separate file

zrlk:


shahms:
What's the motivation behind this change?

zrlk:
This makes it easier to debug production issues. Because signatures are based on structure, being able to go from a hash back to the original string turns out to be very useful:
```
f_pqs_p7393XtVZLv2ZOKKphe7ZcTsqPLZ3V6vfaXcU	talias(__builtin_ms_va_list#n,ptr#builtin(char#builtin))
rRZmIXgKRsITbsW6J15_H1fUu1JfaiZm4FpCfEGvCQg	talias(__NSConstantString#n,__NSConstantString_tag#c#bjwedRx4FtI)
```

As it turns out, simply turning off hashing makes it intractable to denormalize a graph for a "real" compilation.

shahms:
> This makes it easier to debug production issues. Because signatures are based on structure, being able to go from a hash back to the original string turns out to be very useful:
> 
> ```
> f_pqs_p7393XtVZLv2ZOKKphe7ZcTsqPLZ3V6vfaXcU	talias(__builtin_ms_va_list#n,ptr#builtin(char#builtin))
> rRZmIXgKRsITbsW6J15_H1fUu1JfaiZm4FpCfEGvCQg	talias(__NSConstantString#n,__NSConstantString_tag#c#bjwedRx4FtI)
> ```
> 
> As it turns out, simply turning off hashing makes it intractable to denormalize a graph for a "real" compilation.

Then we'll definitely need to go with the callback approach to use the necessary internal abstractions.

zrlk:
> Then we'll definitely need to go with the callback approach to use the necessary internal abstractions.

This meant to run locally (ie not in actual production). Do you still think it's useful to stub out the file stuff?

shahms:
> 

Yes; it keeps the file I/O and hash mapping appropriately isolated, rather than adding "dead" code to every GraphObserver implementation and complicating the resulting API.

shahms:
I would just stub out the hash stuff.  Where you currently have the `RecordHashesFile` option, I'd make that a `CompressSignature = std::function<std::string(absl::string_view)>` callback.

The one place `Force` is currently specified is (according to the comment) to protect against "certain characters" so it's likely better to replace that with just an explicit allowlist such that if the total length of the string is below the cuttoff and contains only allowed characters, don't compress.

Similarly, I'm not entirely sure why the one place which specifies `DontRecord` does so. It seems like recording that hash, if any, would be harmless?

zrlk:
Realized that there's no reason we need to store the original strings after the hash is recorded, we just need to make sure we only record a value for a given hash once.

There's no reason to support different hash functions; I've pulled out an interface for recording though.

I'm actually uncertain what the issue was with `Force` (I *think* it was invalid utf8 getting in through docstrings). This should probably be a different PR.

`DontRecord` is asserted for anchor nodes, the signatures of which are uninteresting. IIRC not recording these will cut the output size by half.

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -72,6 +70,25 @@ class ProfileBlock {
   const char* SectionName;
 };
 
+class HashRecorder {
+ public:
+  virtual void RecordHash(const std::string& hash, absl::string_view web64hash,
```
shahms, line 75: The `hash` argument should also be an `absl::string_view`, no?

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -72,6 +70,25 @@ class ProfileBlock {
...
+                          absl::string_view original) {}
+  virtual ~HashRecorder() {}
```
shahms, line 77: nit: prefer = default

https://abseil.io/tips/131

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -72,6 +70,25 @@ class ProfileBlock {
...
+};
+
+class FileHashRecorder : public HashRecorder {
+ public:
+  FileHashRecorder(const std::string& path);
```
shahms, line 82: This constructor definitely wants to be `explicit` and probably wants to take `absl::string_view` (even if it will need to be converted to a null-terminated C-string to pass to fopen)

https://abseil.io/tips/1

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -72,6 +70,25 @@ class ProfileBlock {
...
+  void RecordHash(const std::string& hash, absl::string_view web64hash,
+                  absl::string_view original) override;
+  ~FileHashRecorder();
```
shahms, line 85: nit: `override`

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -163,10 +184,19 @@ class GraphObserver {
     const ClaimToken* getToken() const { return Token; }
 
    private:
-    const ClaimToken* Token;
+    friend class GraphObserver;
+    NodeId(const ClaimToken* Token, const std::string& Identity)
```
shahms, line 188: nit: `explicit`

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -1093,7 +1123,7 @@ class GraphObserver {
     return {};
   }
 
-  virtual ~GraphObserver() = 0;
+  virtual ~GraphObserver() {}
```
shahms, line 1126: nit: =default (or keep the =0)

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -1119,10 +1149,9 @@ class GraphObserver {
   clang::LangOptions* LangOptions = nullptr;
   clang::Preprocessor* Preprocessor = nullptr;
   ProfilingCallback ReportProfileEvent = [](const char*, ProfilingEvent) {};
+  std::unique_ptr<HashRecorder> hash_recorder_;
```
shahms, line 1152: This seems more naturally and unowned pointer if it is always being passed in externally.

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -72,6 +70,25 @@ class ProfileBlock {
   const char* SectionName;
 };
 
+class HashRecorder {
```
shahms, line 73: Any particularly reason to go with a single-function interface over just `absl::AnyInvocable` or `std::function`?

kythe/cxx/indexer/cxx/IndexerASTHooks.h
```diff
@@ -143,6 +143,8 @@ struct IndexerOptions {
       CreateWorklist = [](IndexerASTVisitor* indexer) {
         return IndexerWorklist::CreateDefaultWorklist(indexer);
       };
+  /// \brief Notified each time a semantic signature is hashed.
+  std::unique_ptr<HashRecorder> HashRecorder;
```
shahms, line 147: This makes this struct no longer copyable (which might be fine), but this also seems more natural as an unowned pointer.

kythe/cxx/indexer/cxx/IndexerFrontendAction.cc
```diff
@@ -216,6 +216,7 @@ std::string IndexCompilationUnit(
   options.build_config = ExtractBuildConfig(Unit);
   options.default_corpus =
       Options.UseCompilationCorpusAsDefault ? Unit.v_name().corpus() : "";
+  options.hash_recorder = std::move(Options.HashRecorder);
```
shahms, line 219: The caller didn't transfer ownership of the options struct, stealing from its guts is inappropriate.

This should either be passed by value or the HashRecorder ownership be retained by the caller.

kythe/cxx/indexer/cxx/KytheGraphObserver.h
```diff
@@ -163,6 +165,7 @@ class KytheGraphObserver : public GraphObserver {
     ReportProfileEvent = std::move(ReportProfileEventCallback);
     RegisterBuiltins();
     EmitMetaNodes();
+    hash_recorder_ = std::move(options.hash_recorder);
```
shahms, line 168: ditto

kythe/cxx/indexer/cxx/GraphObserver.h
zrlk: I find it easier to reason about the lifetime and of the internal state of the class.

kythe/cxx/indexer/cxx/KytheIndexerMain.cc
```diff
@@ -120,9 +120,11 @@ int main(int argc, char* argv[]) {
   options.CreateWorklist = [](IndexerASTVisitor* indexer) {
     return IndexerWorklist::CreateDefaultWorklist(indexer);
   };
+  std::unique_ptr<FileHashRecorder> recorder;
```
shahms, line 123: minor: just use std::optional (or absl::optional); there's no need for heap allocation here.

https://abseil.io/tips/123
https://abseil.io/tips/187

shahms (approved):

kythe/cxx/indexer/cxx/KytheIndexerMain.cc
```diff
@@ -119,6 +120,11 @@ int main(int argc, char* argv[]) {
   options.CreateWorklist = [](IndexerASTVisitor* indexer) {
     return IndexerWorklist::CreateDefaultWorklist(indexer);
   };
+  std::optional<FileHashRecorder> recorder;
+  if (!absl::GetFlag(FLAGS_record_hashes_file).empty()) {
+    recorder.emplace(absl::GetFlag(FLAGS_record_hashes_file));
+    options.HashRecorder = &*recorder;
```
shahms, line 126: FYI (no change needed):  optional::emplace returns a reference to the emplaced object, so this can just be:
```
options.HashRecorder = &recorder.emplace(...);
```

PR #5324, opened 2022-07-19
feat(rust_common): migrate to crate_universe from cargo-raze

wcalandro:


shahms (approved):
Any particular reason? is cargo-raze deprecated or dispreferred?

 (This looks better to me, just curious).

wcalandro:
While crate_universe is still more experimental, there is less file spam and it's maintained by rules_rust rather than a separate group. Also, updating dependencies was leading to PRs with a lot of file changes and I was planning on doing some work on our dependencies soon anyways so I thought I'd just switch us over.

PR #5325, opened 2022-07-20
feat(cxx_indexer): implement directory traversal in indexer VFS

shahms:
Additionally, add some rudimentary tests and clean up IndexVFS a bit.

This addresses issues with `/winsysroot`-style header paths being able to find their SDK headers without providing `HeaderSearchInfo`.


kythe/cxx/indexer/cxx/KytheVFS.cc
```diff
@@ -121,6 +119,12 @@ llvm::ErrorOr<std::unique_ptr<llvm::vfs::File>> IndexVFS::openFileForRead(
 
 llvm::vfs::directory_iterator IndexVFS::dir_begin(const llvm::Twine& dir,
                                                   std::error_code& error_code) {
+  std::string path = dir.str();
+  if (auto* record =
+          FileRecordForPath(path, BehaviorOnMissing::kReturnError, 0)) {
+    return llvm::vfs::directory_iterator(
+        std::make_shared<DirectoryIteratorImpl>(std::move(path), record));
+  }
```
jaysachs, line 127: Worth at least logging an error, or is that done in FileRecordForPath?

jaysachs (approved):

kythe/cxx/indexer/cxx/KytheVFS.cc
shahms: I don't believe it's worth logging an error at this level as it would be very noisy (file not found is a generally expected and unsurprising condition), but it's definitely worth properly reporting the error :-)

Updated and tested.

PR #5330, opened 2022-07-27
chore(rust_common): manually configure the Rust protobuf toolchain

wcalandro:
`rules_rust` comes with a toolchain to generate Rust protobuf source files with `protobuf` crate version `2.8.2`. However, we will soon need to build with version `2.27.1` because of internal version updates. The `protobuf` crate requires that the version used to generate the sources be the same as the version used as the dependency of crates referencing those sources. This PR configures the Rust protobuf toolchain manually using the rule provided in `rules_rust`. This will allow us to upgrade to future versions by simply updating the `protobuf` and `protobuf-codegen` dependencies inside `WORKSPACE`.


BUILD
```diff
@@ -49,3 +50,17 @@ bzl_library(
...
+    edition = "2021",
+    proto_compile_deps = ["@crate_index//:protobuf"],
+    proto_plugin = "@crate_index//:protobuf-codegen__protoc-gen-rust",
+)
+
+toolchain(
+    name = "rust_proto_toolchain",
+    toolchain = ":rust_proto_toolchain_impl",
+    toolchain_type = "@rules_rust//proto:toolchain",
+)
```
shahms, line 66: nit: please add a newline here (this is also the lint complain)

shahms (approved):

PR #5337, opened 2022-08-02
fix(java_indexer): assign all diagnostics a corpus

justbuchanan:
when use_compilation_corpus_as_default is enabled

### DIFF
--- a/kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java
@@ -202,7 +202,11 @@ public class KytheEntrySets {
 
   /** Emits and returns a DIAGNOSTIC node attached to no file. */
   public EntrySet emitDiagnostic(Diagnostic d) {
-    return emitDiagnostic(null, d);
+    VName fileVName =
+        useCompilationCorpusAsDefault
+            ? VName.newBuilder().setCorpus(compilationVName.getCorpus()).build()
+            : null;
+    return emitDiagnostic(fileVName, d);
   }
 
   /**
--- a/kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java
+++ b/kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java
@@ -222,14 +222,17 @@ public class KytheEntrySets {
     if (!d.getContextUrl().isEmpty()) {
       builder.setProperty("context/url", d.getContextUrl());
     }
-    EntrySet dn = emitAndReturn(builder);
     if (fileVName == null) {
       if (getUseCompilationCorpusAsDefault()) {
         builder.setCorpusPath(defaultCorpusPath());
       }
-      return dn;
     } else {
       builder.setCorpusPath(CorpusPath.fromVName(fileVName));
+    }
+    EntrySet dn = emitAndReturn(builder);
+    if (fileVName == null) {
+      return dn;
+    } else {
       if (d.hasSpan()) {
         Span s =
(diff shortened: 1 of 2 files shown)


jaysachs (approved):

kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java
```diff
@@ -202,7 +202,11 @@ public EntrySet newAnchorAndEmit(VName fileVName, Span loc, Span snippet) {
 
   /** Emits and returns a DIAGNOSTIC node attached to no file. */
   public EntrySet emitDiagnostic(Diagnostic d) {
-    return emitDiagnostic(null, d);
+    VName fileVName =
+        useCompilationCorpusAsDefault
+            ? VName.newBuilder().setCorpus(compilationVName.getCorpus()).build()
+            : null;
+    return emitDiagnostic(fileVName, d);
```
schroederc, line 209: This incorrectly attaches the diagnostic to a file VName with an empty path.  The corpus needs to be set in the diagnostics builder

justbuchanan: Done. PTAL.  The diagnostics that are ending up without a corpus are those where the indexer is unable to find a symbol, so there's no file to attach the diagnostic to.

kythe/java/com/google/devtools/kythe/analyzers/java/JavaEntrySets.java
```diff
@@ -159,8 +159,12 @@ public VName getNode(
...
         logger.atWarning().log("%s", msg);
       }
-      Diagnostic.Builder d = Diagnostic.newBuilder().setMessage(msg);
-      return emitDiagnostic(d.build()).getVName();
+      Diagnostic d = Diagnostic.newBuilder().setMessage(msg).build();
+      VName dVName =
+          getUseCompilationCorpusAsDefault()
+              ? VName.newBuilder().setCorpus(getCompilationVName().getCorpus()).build()
+              : null;
+      return emitDiagnostic(dVName, d).getVName();
```
schroederc, line 167: This is still calling the API that tags a file with a diagnostic.  I think this change needs to be addressed in KytheEntrySets.

schroederc (changes requested):

kythe/java/com/google/devtools/kythe/analyzers/base/KytheEntrySets.java
schroederc: Sorry for any confusion.

`emitDiagnostic(VName, Diagnostic)` will emit a `tagged` edge if the `VName` is non-null.  That's why this change is incorrect.  We will be adding a new `tagged` edge from a `VName` with only a corpus.

I believe this change is meant to only change the corpus on the `Diagnostic` itself which means the change should be located in `emitDiagnostic(VName, Diagnostic)` on both sides of the `fileVName == null` condition.

justbuchanan: Hmm, when you say to only change the corpus on the `Diagnostic` itself, where is it stored? The `Diagnostic` proto doesn't have any fields to store that AFAIK, only the `ResolvedDiagnostic` does.

justbuchanan: Can I just set the `CorpusPath` on the `NodeBuilder` in `emitDiagnostic(VName fileVName, Diagnostic d)`?

schroederc: Yeah, LGTM.

schroederc (approved):

PR #5338, opened 2022-08-02
feat(go_indexer): rework --use_compilation_corpus_as_default

justbuchanan:
Most of this change is a flag/option rename. The most important bit is in `rewrittenCorpusForVName()` in emit.go.

The `--override_stdlib_corpus` wasn't working as intended in google3 because it identifies stdlib nodes by the 'golang.org' corpus, but in google3, these are assigned the 'google3/golib' corpus. I renamed `--use_compilation_corpus_as_default` to `--use_compilation_corpus_for_all` to better indicate it's function, which is to change the corpus of *all* VNames (including stdlib and builtin items) at emit time to that of the compilation corpus, which is what we want for oss repos. `rewrittenCorpusForVName()` has been changed so that when `--use_compilation_corpus_for_all` is disabled, the compilation unit's corpus is only used for items that don't already have a corpus set, which is what we want for google3 and google3/golib. Builtin types such as `string` or `bool` will still be assigned 'golang.org', which is what we were already doing before I started messing with corpus rewriting. We could use `--override_stdlib_corpus` to change these to 'golang.org', but I'm inclined to remove the override flag and just continue using 'golang.org' for builtins.

### DIFF
--- a/kythe/go/indexer/emit.go
+++ b/kythe/go/indexer/emit.go
@@ -1003,22 +1003,30 @@ func (e *emitter) writeSatisfies(src, tgt types.Object) {
 }
 
 func (e *emitter) writeFact(src *spb.VName, name, value string) {
-	src = proto.Clone(src).(*spb.VName)
-	src.Corpus = e.rewrittenCorpusForVName(src)
+	if corpus := e.rewrittenCorpusForVName(src); corpus != src.GetCorpus() {
+		src = proto.Clone(src).(*spb.VName)
+		src.Corpus = corpus
+	}
 	e.check(e.sink.writeFact(e.ctx, src, name, value))
 }
 
 func (e *emitter) writeEdge(src, tgt *spb.VName, kind string) {
-	src = proto.Clone(src).(*spb.VName)
-	src.Corpus = e.rewrittenCorpusForVName(src)
-	tgt = proto.Clone(tgt).(*spb.VName)
-	tgt.Corpus = e.rewrittenCorpusForVName(tgt)
+	if corpus := e.rewrittenCorpusForVName(src); corpus != src.GetCorpus() {
+		src = proto.Clone(src).(*spb.VName)
+		src.Corpus = corpus
+	}
+	if corpus := e.rewrittenCorpusForVName(tgt); corpus != tgt.GetCorpus() {
+		tgt = proto.Clone(tgt).(*spb.VName)
+		tgt.Corpus = corpus
+	}
 	e.check(e.sink.writeEdge(e.ctx, src, tgt, kind))
 }
 
 func (e *emitter) writeAnchor(node ast.Node, src *spb.VName, start, end int) {
-	src = proto.Clone(src).(*spb.VName)
-	src.Corpus = e.rewrittenCorpusForVName(src)
+	if corpus := e.rewrittenCorpusForVName(src); corpus != src.GetCorpus() {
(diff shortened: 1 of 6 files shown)


kythe/go/indexer/emit.go
```diff
@@ -1000,28 +1003,22 @@ func (e *emitter) writeSatisfies(src, tgt types.Object) {
 }
 
 func (e *emitter) writeFact(src *spb.VName, name, value string) {
-	if e.opts.UseCompilationCorpusAsDefault || e.opts.OverrideStdlibCorpus != "" {
-		src = proto.Clone(src).(*spb.VName)
-		src.Corpus = e.rewrittenCorpusForVName(src)
-	}
+	src = proto.Clone(src).(*spb.VName)
```
schroederc, line 1006: if corpus := e.rewrittenCorpusForVName(src); corpus != src.GetCorpus() {
  src = proto.Clone(src).(*spb.VName)
  src.SetCorpus(corpus)
}

kythe/go/indexer/emit.go
```diff
@@ -1000,28 +1003,22 @@ func (e *emitter) writeSatisfies(src, tgt types.Object) {
...
 }
 
 func (e *emitter) writeEdge(src, tgt *spb.VName, kind string) {
-	if e.opts.UseCompilationCorpusAsDefault || e.opts.OverrideStdlibCorpus != "" {
-		src = proto.Clone(src).(*spb.VName)
-		src.Corpus = e.rewrittenCorpusForVName(src)
-		tgt = proto.Clone(tgt).(*spb.VName)
-		tgt.Corpus = e.rewrittenCorpusForVName(tgt)
-	}
+	src = proto.Clone(src).(*spb.VName)
```
schroederc, line 1012: Do the same checks as above to reduce unnecessary cloning.

kythe/go/indexer/emit.go
```diff
@@ -1000,28 +1003,22 @@ func (e *emitter) writeSatisfies(src, tgt types.Object) {
...
+	tgt.Corpus = e.rewrittenCorpusForVName(tgt)
 	e.check(e.sink.writeEdge(e.ctx, src, tgt, kind))
 }
 
 func (e *emitter) writeAnchor(node ast.Node, src *spb.VName, start, end int) {
-	if e.opts.UseCompilationCorpusAsDefault || e.opts.OverrideStdlibCorpus != "" {
-		src = proto.Clone(src).(*spb.VName)
-		src.Corpus = e.rewrittenCorpusForVName(src)
-	}
+	src = proto.Clone(src).(*spb.VName)
```
schroederc, line 1020: ditto

kythe/go/indexer/emit.go
```diff
@@ -1030,10 +1027,8 @@ func (e *emitter) writeAnchor(node ast.Node, src *spb.VName, start, end int) {
 }
 
 func (e *emitter) writeDiagnostic(src *spb.VName, d diagnostic) {
-	if e.opts.UseCompilationCorpusAsDefault || e.opts.OverrideStdlibCorpus != "" {
-		src = proto.Clone(src).(*spb.VName)
-		src.Corpus = e.rewrittenCorpusForVName(src)
-	}
+	src = proto.Clone(src).(*spb.VName)
```
schroederc, line 1030: ditto

schroederc (approved):

PR #5351, opened 2022-08-11
fix(go): re-add Compilation struct and associated methods

justbuchanan:
These were deleted from kindex.go in https://github.com/kythe/kythe/pull/5348, but are still in use in google3


shahms:
It looks like the `go test` tests fail

justbuchanan:
Oops, missed a few go imports.

jaysachs (approved):

PR #5354, opened 2022-08-15
fix(rust_extractor): improve logic for finding the save analysis file

wcalandro:
This PR uses the output file name of a target for the name of the save analysis file rather than using the crate name from the compiler context. This makes the path of the save analysis file deterministic and future-proofs the extractor against potential changes to crate naming for targets that break the path determination, which has happened in the past. The PR also introduces an indexer test for inline test code which have previously been affected by issues related to save analysis file naming.


kythe/rust/extractor/src/bin/extractor.rs
```diff
@@ -67,13 +67,29 @@ fn main() -> Result<()> {
...
+    let build_output_path = spawn_info
+        .get_output_file()
+        .get(0)
+        .ok_or_else(|| anyhow!("Failed to get output file from spawn info"))?;
+
+    // Get the name of the output file (without extension)
+    let output_file_name = PathBuf::from(build_output_path)
+        .with_extension("")
+        .file_name()
+        .unwrap()
```
shahms, line 80: Shouldn't this also be an `or_else` or equivalent rather than panicking on error?

wcalandro: The `file_name` only returns `None` if the path ends with `..`, but I have gone ahead and added the `ok_or_else`

shahms (approved):

PR #5356, opened 2022-08-16
chore: perform some cleanup on the Rust indexer

schroederc:


wcalandro:
Please run `./tools/rust/rustfmt.sh` (requires rust to be installed locally). I'm going to be changing the repo to use the rustfmt aspect soon to alleviate thisl

wcalandro:
I'm slightly concerned with the addition of `assert!`s (since they would panic), but they realistically should never throw so it's probably fine=.

schroederc:
> I'm slightly concerned with the addition of `assert!`s (since they would panic), but they realistically should never throw so it's probably fine=.

The asserts are maintaining the invariant already expressed here: https://github.com/kythe/kythe/pull/5356/files#diff-cb12149740fdf097c5ef11d74c7ec519d615b38d7a13e0bb12fd3c81166852f7R293

wcalandro:
Ah that's fair

wcalandro:
rustfmt is still unhappy. I would double-check that your IDE isn't trying to format some of the files itself as Google uses a specific rustfmt.toml (present in tools/rust/rustfmt.toml)

shahms (approved):

wcalandro (approved):

PR #5357, opened 2022-08-17
chore(rust_common): rustfmt improvements

wcalandro:
The `rustfmt` tool changes over time and our repo should always aim to use the binary related to our current version of Rust. This PR updates `tools/rust/rustfmt.sh` to use the binary included with `rules_rust`, falling back to the local version if necessary. Additionally, `.bazelrc` now configures the `rustfmt` aspect from `rules_rust` so that the checks occur automatically during builds.

The `Dockerfile` for our pre-commit cloudbuilder has been updated to ensure that the installed version of Rust matches our current toolchain, and this will need to be updated periodically. Other changes have been made to the `Dockerfile` related to our Go version, as `staticcheck` could not be installed with the version of Go available in Focal Fossa.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -95,11 +95,10 @@ build:clang-tidy --copt=-Wundefined-bool-conversion --copt=-Wunreachable-code --
 build:clang-tidy --copt=-Wunused-function --copt=-Wunused-lambda-capture --copt=-Wunused-local-typedef
 build:clang-tidy --copt=-Wunused-private-field --copt=-Wuser-defined-warnings
 
-# Support user-provided user.bazelrc
-try-import %workspace%/user.bazelrc
-    .
-
 # Enable rustfmt tests on all Rust targets
 build --@rules_rust//:rustfmt.toml=//tools/rust:rustfmt.toml
 build --aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect
 build --output_groups=+rustfmt_checks
+
+# Support user-provided user.bazelrc
+try-import %workspace%/user.bazelrc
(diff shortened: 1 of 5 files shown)


shahms (approved):

.bazelrc
```diff
@@ -98,3 +98,8 @@ build:clang-tidy --copt=-Wunused-private-field --copt=-Wuser-defined-warnings
 # Support user-provided user.bazelrc
 try-import %workspace%/user.bazelrc
     .
+
+# Enable rustfmt tests on all Rust targets
+build --@rules_rust//:rustfmt.toml=//tools/rust:rustfmt.toml
+build --aspects=@rules_rust//rust:defs.bzl%rustfmt_aspect
+build --output_groups=+rustfmt_checks
```
shahms, line 105: This should probably be limited to the prepush config, no?

shahms: And certainly come before the try-import for user.bazelrc (which should always come last)

wcalandro: IMO this is better because it will come up when running `bazel build` during local development

shahms: I'm dubious that it should be part of the default configuration (rather than its own), but am not particularly bothered by it.

But it should definitely come before the `try-import user.bazelrc`.

.bazelrc
```diff
@@ -98,3 +98,8 @@ build:clang-tidy --copt=-Wunused-private-field --copt=-Wuser-defined-warnings
 # Support user-provided user.bazelrc
 try-import %workspace%/user.bazelrc
     .
```
shahms, line 100: As long as you're modifying this file, this errant period should be removed.  It appears to have arrived in a2568e3099 but I'm not really sure where it came from other than a typo.

shahms (approved):
Thanks!

PR #5367, opened 2022-08-29
build: update max version to Bazel v5.3.0

schroederc:


shahms (approved):

.bazelversion
```diff
@@ -1 +1 @@
-5.0.0
```
hzeller, line 1: Wouldn't we now be tied to exactly 5.3.0, leaving out the flexibility of anything between 5.0..5.3 ?
Could we just remove this file and have version.bzl take care of the range ?

schroederc: That file only tells bazelisk which version to use which is important for building past versions.  The other file gives the range for any Bazel version that can be used.  Both are important.

PR #5377, opened 2022-09-14
fix(cxx_indexer): make logic less tricky and more functional

zrlk:


kythe/cxx/indexer/cxx/IndexerASTHooks.cc
```diff
@@ -4251,10 +4251,10 @@ IndexerASTVisitor::BuildNodeIdForTemplateName(const clang::TemplateName& Name) {
             // templates.
             if (const auto* FTD = dyn_cast<clang::FunctionTemplateDecl>(
                     Name.getAsTemplateDecl())) {
-              for (const auto* ID = FTD;
-                   (ID = ID->getInstantiatedFromMemberTemplate()); FTD = ID)
+              while ((FTD = FTD->getInstantiatedFromMemberTemplate())) {
```
justbuchanan, line 4254: I think you can drop one set of parens here

justbuchanan (approved):

kythe/cxx/indexer/cxx/IndexerASTHooks.cc
zrlk: This results in a compiler warning about using the result of an assignment in a condition (which is intentional here).

justbuchanan: fair enough :)

PR #5378, opened 2022-09-14
feat(cxx_indexer): distinguish direct function calls

zrlk:


zrlk:
In partial service of #5379 

There are other AST nodes that need supported (like constructor calls).

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -738,6 +738,14 @@ class GraphObserver {
     Yes
   };
 
+  /// \brief Determines whether a feature is direct or admits indirection.
+  enum class Directness {
+    /// This feature is indirect (e.g., performs dynamic dispatch).
+    Indirect,
```
shahms, line 744: Thoughts about:
```
enum class CallDispatch {
  kDefault,
  kDirect,
};
```
As the distinction?

shahms (commented):
This looks good, just a minor question on naming.

kythe/cxx/indexer/cxx/GraphObserver.h
zrlk: SGTM

PR #5380, opened 2022-09-15
rust: remove some needless borrows

krasimirgg:
No functional changes intended.

New versions of Clippy warn about these.


shahms (approved):
Thanks!

PR #5381, opened 2022-09-16
fix(cxx_indexer): dereference member templates

zrlk:
This PR also cleans up BuildNodeIdForTemplateName, since it no longer has to dereference member templates itself.

### DIFF
--- a/kythe/cxx/indexer/cxx/clang_utils.cc
+++ b/kythe/cxx/indexer/cxx/clang_utils.cc
@@ -17,6 +17,7 @@
 #include "kythe/cxx/indexer/cxx/clang_utils.h"
 
 #include "clang/AST/DeclTemplate.h"
+#include "clang/AST/DeclVisitor.h"
 #include "clang/Basic/CharInfo.h"
 #include "clang/Lex/Lexer.h"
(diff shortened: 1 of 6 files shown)


kythe/cxx/indexer/cxx/clang_utils.cc
```diff
@@ -45,12 +45,48 @@ clang::SourceLocation GetLocForEndOfToken(
                                            source_manager, lang_options);
 }
 
-// TODO(zarko): Update this to handle member specializations.
-const clang::Decl* FindSpecializedTemplate(const clang::Decl* decl) {
+const clang::Decl* DereferenceMemberTemplates(const clang::Decl* decl,
+                                              bool use_mts) {
+  if (!use_mts) return decl;
+  if (const auto* ft = clang::dyn_cast<clang::FunctionTemplateDecl>(decl)) {
```
shahms, line 51: Use a clang::ConstDeclVisitor for this rather than the else if chains? It will be (marginally) faster, but generally keep the code better structured.

zrlk: Done.

shahms (approved):

PR #5396, opened 2022-10-18
chore: remove last //third_party/javac dependency

shahms:


shahms:
(The target no longer exists)

justbuchanan (approved):

PR #5398, opened 2022-10-19
feat(go_indexer): implement ref/writes edges

schroederc:


schroederc:
This is for simple variables.  Member writes will be handled in a future change.

shahms (approved):

PR #5408, opened 2022-10-25
chore: include build:prepush config; bump minor version

shahms:
`test` inherits from `build` but if there isn't a `prepush` defined for `build` it can cause problems.

Bump the Bazel release version.

Don't re-run a full query for the tests, reuse the results of the previous query.


tools/git/test-affected-targets.sh
```diff
@@ -42,7 +42,7 @@ fi
 readarray -t TESTS < <(bazel query \
   --keep_going \
   --noshow_progress \
-  "tests(rdeps(//..., set($*))) except attr('tags', '$TAG_RE', //...)")
+  "tests(set(${TARGETS[*]}))")
```
schroederc, line 45: This removes the filtering of OMIT_TAGS

schroederc (changes requested):

tools/git/test-affected-targets.sh
schroederc: Oh, I see.

schroederc (approved):

tools/git/test-affected-targets.sh
shahms: That filtering is done on `TARGETS` already. But fair that it wouldn't filter expanded `test_suite` rules.

PR #5409, opened 2022-10-25
feat(tooling): add a flag to ignore conflicting /kythe/code facts

BrianWieder:
Adding a flag to the verifier to ignore conflicting /kythe/code facts.


d4n1elchen:
Some more context for this PR: there are some conflicting /kythe/code facts generated because of C++'s template instantiation, which in my case is inevitable because they are used in the dependency of the tested file.

shahms (approved):

jaysachs:
I would be very interested in a pointer to the failures you're observing that motivated this CL.

d4n1elchen:
I realized that this PR doesn't prevent `is_ok` from flipping. We might need additional PR to fix that.

PR #5421, opened 2022-11-01
chore: remove most remaining mentions of abs/absvar

jaysachs:
Just a few mentions left in comments in the C++ indexer, and a small section of the schema.


kythe/typescript/kythe.ts
```diff
@@ -110,7 +110,6 @@ export function makeOrdinalEdge(edge: EdgeKind, ordinal: number): OrdinalEdge {
  *   https://github.com/kythe/kythe/tree/master/kythe/data/schema_index.textproto#L64
  */
 export enum NodeKind {
-  ABSVAR = 'absvar',
```
jaysachs, line 113: This was already out of sync with the missing ABS enum. Maybe there should be two "UNUSED_1" and "UNUSED_2" in there?

schroederc: Do these actually have to match the proto enum values?  It seems more like the C++ enum where the integers don't matter.

schroederc (approved):

kythe/typescript/kythe.ts
jaysachs: Everything seems to work, and it already didn't match, so I'm assuming it doesn't matter.

PR #5422, opened 2022-11-02
perf(serving): reduce patching to offsets; drop text

schroederc:


jaysachs (commented):
Can you add a test? While I can follow the code, it's difficult to understand the intent of the change.

schroederc:
The existing tests already cover this.  We just don't need the text held in the Patcher.

jaysachs (approved):

PR #5426, opened 2022-11-04
chore: add compatibility shims to work with Java 17+

shahms:
Also remove the legacy platform restrictions in a vain attempt to get `cquery` working (but objc_library remains stubbornly incompatible at this time).

Rather than using reflection, this sticks various compatibility routines into a separate interface (where possible) which is loaded via service-provider.  I'm not sure that latter is necessary, but it's at a least a token towards "runtime support" for now.


jaysachs (approved):
That doesn't seem anywhere as terrible as I was afraid it would be. Nicely done.

PR #5436, opened 2022-11-09
feat(cxx_indexer): support importing protobuf metadata

zrlk:


kythe/cxx/common/protobuf_metadata_file.cc
```diff
@@ -102,6 +102,17 @@ std::unique_ptr<kythe::MetadataFile> ProtobufMetadataSupport::ParseFile(
       } else if (absl::StartsWith(token, "mutable_")) {
         rule.semantic = kythe::MetadataFile::Semantic::kReadWrite;
       }
+    } else {
+      switch (annotation.semantic()) {
+        case proto2::GeneratedCodeInfo_Annotation_Semantic_SET:
```
shahms, line 107: wrong namespace here (also, prefer the nested names over the underscore variety).

`google::protobuf::GeneratedCodeInfo::Annotation::Semantic::SET`

zrlk: Haha, it's actually much more of a mess than I expected. The new Semantic enum isn't being exported, so we need to upgrade protobuf (to a version past the current one at HEAD, since the one at HEAD still doesn't export Semantic). Upgrading protobuf appears to be nontrivial.

Also, for some reason, the nested names don't work (at least in a version of protobuf with Semantic).

shahms: Ah, fair. I've a branch updating to HEAD to deal with enum-annotations, but it's non-trivial and they're in the process of harmonizing the internal/external targets, so figured it best to wait.

(It's also blocked by a variety of inconsistencies between the git-exposed targets and the release-exposed ones which need upstream changes).

PR #5449, opened 2022-11-22
chore(java_indexer): downgrade doc reference failure logging

shahms:


kythe/java/com/google/devtools/kythe/analyzers/java/KytheDocTreeScanner.java
```diff
@@ -109,7 +109,8 @@ public Void visitReference(ReferenceTree tree, DCDocComment doc) {
     try {
       sym = (Symbol) trees.getElement(getCurrentPath());
     } catch (Throwable e) {
-      logger.atWarning().withCause(e).log("Failed to resolve documentation reference: %s", tree);
+      logger.atInfo().log("Failed to resolve documentation reference: %s\n%s", tree, e);
```
jaysachs, line 112: Do we even need this? How is it useful outside of the case where we turn on fine logging to get the full trace?

shahms: I think it's still useful to know about; it's something I expect people will file bugs about so is good to have a trace in the logs.

jaysachs: Should we add a diagnostic instead?

shahms: Yeah, good call. Just a bit more involved.

jaysachs (approved):

PR #5469, opened 2022-12-06
feat(serving): add option to read xref pages concurrently

schroederc:


shahms:
```
go: updates to go.mod needed, disabled by -mod=readonly; to update it:
	go mod tidy
```

kythe/go/serving/xrefs/xrefs.go
```diff
@@ -624,6 +627,53 @@ func decorationToReference(norm *span.Normalizer, d *srvpb.FileDecorations_Decor
...
+			reply.Total.Declarations += int64(idx.Count)
+		} else {
+			reply.Filtered.Declarations += int64(idx.Count)
+		}
+	case xrefCategoryRef:
+		if pageSet.Contains(idx) {
+			reply.Total.References += int64(idx.Count)
+		} else {
+			reply.Filtered.References += int64(idx.Count)
+		}
```
shahms, line 661: should this be?
```
case xrefCategoryDef, xrefCategoryDecl, xrefCategoryRef:
		if pageSet.Contains(idx) {
			reply.Total.Definitions += int64(idx.Count)
		} else {
			reply.Filtered.Definitions += int64(idx.Count)
		}
```

kythe/go/serving/xrefs/xrefs.go
```diff
@@ -916,12 +1031,7 @@ readLoop:
 					reply.Filtered.Declarations += int64(filtered)
 					stats.addAnchors(&crs.Declaration, p.Group)
 				}
-			case xrefs.IsRefKind(req.ReferenceKind, idx.Kind):
-				if !pageSet.Contains(idx) {
-					reply.Filtered.References += int64(idx.Count)
-					continue
-				}
-				reply.Total.References += int64(idx.Count)
+			case xrefCategoryRef:
```
shahms, line 1034: same for these three. The case-code for xrefCategory{Def,Decl,Ref} appears identical; should they be a single case.

kythe/go/serving/xrefs/xrefs.go
schroederc: I don't understand.  Why would decls and refs both be counted as defs?

kythe/go/serving/xrefs/xrefs.go
schroederc: They affect different categories in the reply.

shahms:
Apparently my eyes don't work today.

PR #5471, opened 2022-12-07
fix(rust_indexer): set root when creating vname for reference

wcalandro:


wcalandro:
> if not too onerous, a test of this would be great

Is it possible to the vname of an edge?

wcalandro:
> if not too onerous, a test of this would be great

I'm not sure of a way to properly test this since the `defines/binding` anchor has the right root, it's just that the vname for the target anchor of the edge doesn't.

shahms:
> > if not too onerous, a test of this would be great
> 
> I'm not sure of a way to properly test this since the `defines/binding` anchor has the right root, it's just that the vname for the target anchor of the edge doesn't.

https://www.kythe.io/docs/kythe-verifier.html#_explicit_unification

Something like:
```
//- @SomeAnchor ref AnotherNode=vname(_, _, "root", _, _)
....
```

Issue #5490, opened 2023-01-11
migrate TS rules from rules_nodejs to rules_ts


jaysachs:
Specifically, migrate from https://bazelbuild.github.io/rules_nodejs/ to https://github.com/aspect-build/rules_ts

This will unblock moving to modern versions of the tscompiler dependency.

jaysachs:
Plan of action:
1. Introduce a `codejson` (or similar) fact for MarkedSource use.
2. Update proxystub to convert `codejson` to `code` and the payload to binary encoded proto
3. Update the TS indexer to emit this `codejson` fact instead of `code` and remove its dependency on proto
4. Update the TS build rules to `rules_ts`

shahms:
The rules still need to be migrated :-)

jaysachs:
Why the assumption of one PR to fix an issue?

creachadair:
> > Why the assumption of one PR to fix an issue?
> 
> I think it looks for some magic words in the description/title, but can't recall the details.
> 

If you write "Updates #xx" it will tag the issue without closing it; the words "Fix" and "Fixes" (and maybe some others) cause it to close the issue when the PR merges. 

For better or worse.

PR #5491, opened 2023-01-11
feat(indexing): support rewriting JSON-encoded code facts in proxy

schroederc:


shahms (approved):

shahms:
Addresses point 2 from #5490

jaysachs (commented):
Do we want to update schema.proto as well?

schroederc:
> Do we want to update schema.proto as well?

It's not really a general-purpose fact so I'm against putting it in the schema.  It's only really for the proxy.

shahms:
It's probably a good idea to document it somewhere, though. Maybe a note in the schema?

shahms (approved):

PR #5502, opened 2023-01-17
Adjust test for file objects in the system module path

cushon:
The paths for these file objects changes from e.g. `/modules/java.base/java/io/Serializable.class` to `/java.base/java/io/Serializable.class` in an upcoming version, breaking the existing regex.

This change instead tests for a location whose name starts with `SYSTEM_MODULES`. The location instance for these file objects is not one of the `StandardLocations` constants, it's a custom subclass, but the name reliably starts with `SYSTEM_MODULES`.


shahms:
(buildkite is currently broken due to an unsigned repository, that failure is non-blocking)

shahms (approved):
Thanks!

PR #5508, opened 2023-01-20
chore(c++): use string_view in more places

justbuchanan:
this is from a recent LSC


kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -68,13 +68,13 @@ constexpr int kNonRepeatedFieldIndex = -1;
...
-  void AddError(const std::string& filename, int line, int column,
-                const std::string& message) override {
+  void AddError(absl::string_view filename, int line, int column,
+                absl::string_view message) override {
     LOG(ERROR) << filename << "@" << line << ":" << column << ": " << message;
   }
 
-  void AddWarning(const std::string& filename, int line, int column,
-                  const std::string& message) override {
+  void AddWarning(absl::string_view filename, int line, int column,
```
creachadair, line 76: Why by value here vs. by reference above?

(I don't know what's correct, I just noticed they're different, although both are replacing `const std::string&`).

shahms:
This will require a newer version of protobuf and/or and member function name change.  I have a question out on one of the LSC CLs to this effect.

kythe/cxx/indexer/textproto/analyzer.cc
shahms: By value. std::string_view is a lightweight "reference type" intended to be passed by value.

justbuchanan: @creachadair thanks, that was a typo. Fixed

justbuchanan:
> This will require a newer version of protobuf and/or and member function name change. I have a question out on one of the LSC CLs to this effect.

Sounds good. I'll hold off on this until then.

justbuchanan:
Looks like this was just updated in the proto repo yesterday: https://github.com/protocolbuffers/protobuf/commit/d538808032d66366364e410200246c236e275e0f

shahms:
> Looks like this was just updated in the proto repo yesterday: [protocolbuffers/protobuf@d538808](https://github.com/protocolbuffers/protobuf/commit/d538808032d66366364e410200246c236e275e0f)

Excellent, then we'll just need to wait for a new release. While they are in the process of making the at-HEAD BUILD file better, actually using it at the moment remains quite an undertaking.

PR #5510, opened 2023-01-23
chore: enable building kythe extractor docker image via GCB

jaysachs:
Haven't tried this out yet, looking for feedback on the approach

### DIFF
--- a/kythe/extractors/bazel/gcb.yaml
+++ b/kythe/extractors/bazel/gcb.yaml
@@ -1,0 +1,49 @@
+# Cloud Build configuration for mandatory presubmit tests.
+steps:
+  - name: gcr.io/cloud-builders/git
+    id: clone
+    args:
+      - clone
+      - '--recurse-submodules'
+      - 'https://github.com/kythe/kythe'
+      - .
+    dir: code,
+  - name: gcr.io/cloud-builders/git,
+    id: checkout,
+    args:
+      - checkout
+      - '--recurse-submodules'
+      - ${_COMMIT}
+    dir: code
+    waitFor:
+      - clone
+  - name: gcr.io/kythe-repo/bazelisk-builder-client
+    id: bazel-release
+    args:
+      - bazel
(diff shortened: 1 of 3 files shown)


kythe/extractors/bazel/gcb.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - '--stamp'
+      - '//kythe/extractors/bazel:stage'
+  - name: gcr.io/cloud-builders/docker
+    id: docker-build
+    args:
+      - build
+      - '-t'
+      - gcr.io/kythe-public/bazel-extractor:latest
+      - .
+    dir: code/bazel-out/k8-opt/bin/kythe/extractors/bazel/stage.done
```
jaysachs, line 37: Felt dirty hardcoding this.

shahms: I don't know of any other way, unfortunately.

justbuchanan:
Overall looks good. A couple questions:

What do we gain by running the `docker build` step outside of bazel? Is this related to the glibc version issue?

Does the image built in the docker-build step still exist when the docker-push step runs? Data persistence across build steps is fairly limited (mostly to /workspace or custom volumes), but maybe there's something special for docker images. See also https://cloud.google.com/build/docs/configuring-builds/pass-data-between-steps#passing_data_using_workspaces

justbuchanan:
> What do we gain by running the `docker build` step outside of bazel? Is this related to the glibc version issue?

Following up on my own question - is the issue that the docker binary isn't present in the bazel image, so we're using a separate image with docker installed to do the actual `docker build` part?

jaysachs:
With the latest commit ([cdee542](https://github.com/kythe/kythe/pull/5510/commits/cdee542c0a46a95bdf5d003d3094a623ca22ce0d))  and some GCP IAM fiddling this successfully builds and pushes.

shahms (approved):

PR #5520, opened 2023-02-11
chore(typescript_indexer): consolidate indexer options in a single object

nbeloglazov:
This is necessary to add more options in future. Currently each options requires new method argument and that hurts readability.

Also using this opportunity modify a bit API of index() function and plugins. Now we pass files to index as explicit `paths` array rather than it being part of IndexerHost. That way index methods are a bit easier to understand of *what* exactly they need to index.

Note: this change will require some google3 changes as it's changing arguments of index() method and there are few users. Changes should be trivial: converting list of arguments to an object.

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -37,6 +37,8 @@ enum RefType {
  * code duplication.
  */
 export interface IndexerHost {
+  options: IndexingOptions;
+
   /**
    * Gets the offset table for a file path.
    * These are used to lookup UTF-8 offsets (used by Kythe) from UTF-16 offsets
@@ -76,14 +78,18 @@ export interface IndexerHost {
   moduleName(path: string): string;
   /**
    * Paths to index.
+   * TODO: Remove completely and instead pass paths explicitly
+   * as inputs to methods.
    */
   paths: string[];
   /**
    * TypeScript program.
+   * TODO: migrate usages to options.program
    */
   program: ts.Program;
   /**
    * Strategy to emit Kythe entries by.
+   * TODO: migrate usages to options.emit
    */
   emit(obj: JSONFact|JSONEdge): void;
 }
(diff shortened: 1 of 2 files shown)


jaysachs:
go /gcbrun

kythe/typescript/indexer.ts
```diff
@@ -31,12 +31,44 @@ enum RefType {
...
 
+/**
+ * Various options that are required in order to perform indexing.
+ */
+export interface IndexingOptions {
+  /** A VName for the entire compilation, containing e.g. corpus name. */
+  compilationUnit: VName;
+
+  /** A map of file path to path-specific VName. */
+  pathVNames: Map<string, VName>;
```
jaysachs, line 42: What do you of breaking this into two components:

`compilationUnit`: has VName, pathVNames, and srcfiles (the latter  are currently the paths passed to `index`)

`IndexingOptions` (or `IndexingEnvironment`): has the `emit` and  `plugins`

So what about `program` and `readFile` .... can the indexer itself create the program object? Pass in more stuff in `options` if needed for that. 

`readFile`. Hmm. Maybe let's make a full/complete `CompilationUnit` object? Put sources, the vname map, the root vname, and the readFile function on that?

nbeloglazov: Good suggestion. I kept readFile() in IndexingOptions though. For me CompilationUnit is "dynamic" part, that changes from indexing to indexing, while IndexingOptions normally static and depends on environment and command line arguments. I think even CompilerOptions while currently coming from compilation units, eventually should become static and we should apply same CompilerOptions for all units. 

Note that this made `index()` method signature to diverge more so it will require more work on merging to google3. Though the merging is still relatively trivial.

jaysachs: I know the implementation of readFile has to depend on the compilation unit, so it seems a little odd that it's passed separately from the CU, and just once at creation. But conceptually it doesn't belong on the CU, so I think this works.

jaysachs (approved):

PR #5523, opened 2023-02-17
feat(typescript): add option to pass only source files to program initialization

nbeloglazov:
Passing all files to `createProgram` should not be necessary as TS compiler can load necessary dependencies lazily. It is done through CompilerHost.


nbeloglazov:
Please hol on reviewing just yet. It might not be working as I think it's working.

jaysachs (approved):

nbeloglazov:
I won't proceed with this PR. Instead I'll open another one to update names and comments of the variables to make it clearer how TS indexer works.

Issue #4021, opened 2019-08-23
Emit references and definitions for the same anchor


ayazhafiz:
See a much longer discussion at #3934.

The Python indexer, and since #3990 the TypeScript indexer, deal with imports defining a local import declaration _and_ referencing a remote definition by creating a unique anchor for the import definitions and a unique anchor for the import reference.

```typescript
//- @import defines/binding LocalFooDefintion
//- @foo ref RemoteFooDefinition
import { foo } from './bar'
```

This is done because the most popular Kythe UI currently chooses an anchor as referencing other semantic nodes _or_ defining a semantic node, but not both.

If the UI is extended to support both definitions and references on one anchor, this can be closed and we can emit definitions and references on the same anchor.

```typescript
//- @foo defines/binding LocalFooDefintion
//- @foo ref RemoteFooDefinition
import { foo } from './bar'
```

kamahen:
Using `import` for the `ref` confuses users (it's been tried, plus a few variants).

shahms:
IIRC from the outcome of that discussion, the test should also include:

```
//- @foo defines/binding LocalFooDefintion
//- @foo ref RemoteFooDefinition
//- LocalFooDefinition aliases RemoteFooDefinition
import { foo } from './bar'
```

kamahen:
A generalization of issue #3815 -- it could be desirable for the indexer to specify the ordering of edges that the UI displays. (Issue #3815 refers to a single kind of edge from a node; but it might be desirable to specify an ordering for all edges from a node.)

robinp:
As discussed in context of Python's `from foo import *` syntax in https://github.com/TreeTide/underhood/issues/9#issuecomment-538885438, is it right to say that blanket imports happening at the `*` should have the same reference behavior, but actually emitting the for all the imported names *that are actually used in the module*? See that discussion for the rationale.

Example:

```
// foo exports a, b, c, ..., z
//- @"*" defines/binding LocalB
//- @"*" ref RemoteB
//- LocalB aliases RemoteB
//- @"*" defines/binding LocalC
//- @"*" ref RemoteC
//- LocalC aliases RemoteC
//- !{ @"*" defines/binding LocalA }    // smoke check, among other exported names
from foo import *

//- @b ref LocalB
//- @c ref LocalC
bar = b + c
```

ayazhafiz:
@robinp. I think yes. And as others have mentioned, there is an issue of practicality here for the user. In something like

```python
from foo import * # imports a, b

[...]

PR #5527, opened 2023-02-24
feat(typescript_indexer): implement ref inlining for imported symbols

nbeloglazov:
The goal of this PR is to connect usages of imported symbol to the its declaration from original file. Today indexer produces the following graph:

```ts
// foo.ts
//- @ANSWER defines/binding AnswerOrig
export const ANSWER = 42;


// bar.ts
//- @ANSWER ref/imports AnswerOrig
//- @ANSWER defines/binding AnswerLocal
import {ANSWER} from './foo';

//- @ANSWER ref AnswerLocal
console.log(ANSWER);
```

After this PR the graph will look like this:

```ts
// foo.ts
//- @ANSWER defines/binding AnswerOrig
export const ANSWER = 42;


// bar.ts
//- @ANSWER ref/imports AnswerOrig
import {ANSWER} from './foo';

//- @ANSWER ref AnswerOrig
console.log(ANSWER);
```

Note that `AnswerLocal` is gone. 

The change is guarded by `enableImportsEdgeReassignment` flag. This will allow us to easier roll it out at Google. 

This is related to #4021.


kythe/typescript/post_processing.ts
```diff
@@ -0,0 +1,140 @@
...
+ * and removes `@ANSWER defines/binding AnswerLocal` definition as it doesn't
+ * have any refs to it now.
+ *
+ *
+ * Notice that ANSWER anchor in console.log call now points at the AnswerOrig node.
+ * This is what users expect: that usages of imported symbols will point to the
+ * original definition and not local aliases.
+ */
+function reassignEdgesForImports(data: Readonly<KytheData>): KytheData {
+    const result: KytheData = [];
```
jaysachs, line 103: Maybe premature optimization, but we've seen memory issues in the TS indexer.

So we're not just buffering all the emitted data, but buffering it 2x here. Is there a way to avoid this? Can we at least drop references to the AST and file data before doing this? Is there a way to do this my mutating the original data in-place?

nbeloglazov: Actually it wouldn't be 2x here. We create a copy of an array but we are not copying entries themselves. We mutating them in-place. `Readonly<KytheData>` is a bit misleading: it means we can't modify array itself but can and do modify elements of the array.

As for dropping AST or file data - I don't think it can be controlled in TS/JS. I believe that NodeJS should be able to clear those objects as they are unused by that point. I added enablePostProcessing flag so it should help us to see impact of just adding buffering. 

Mutating array fully in-place a bit clunky as we need to remove some entries from the array. So we would need to make array elements nullable and then carefully handle nulls. Which is a bit annoying. But if it turns out that enabling post processing has bad impact - I'll do that.

creachadair: For historical context, this specific issue (memory overload) was an important motivation for some of Kythe's design. We had Python indexer shards falling over because they needed to retain the whole indexer output for a module in-memory, and since the dependencies are all source or source-equivalent, it didn't fit in some cases.

As a rule, this is why we designed the graph schema not to do denormalizations "directly" in the indexer, but instead to handle them as postprocessing steps. It's safer in languages with separate compilation, but TS is not one of those 🙂.

One can't necessarily avoid having to have all the source in-memory, but ideally the indexer should try hard to not keep the graph artifacts in-memory. In some cases one can get away with a cache of smaller pieces, though.

nbeloglazov: Actually I probably can do that inlining without buffering, by complicating TS indexer itself.

nbeloglazov: Rewrote implementation by having local per-source file vname => vname mapping. Similar like we do in JS indexer.

nbeloglazov:
@jaysachs bump

jaysachs (approved):

Issue #5528, opened 2023-03-06
Add recordio text format support to TextProto indexer.


chuthan20:
Recordio format specifies multiple record in a single file using a separator. Currently text proto indexer fails for recordio text-formatted file. 

By default, separator is empty line. Users can override that to be any set of characters.
Common separators using one or more repeated characters of:
`***`, `===`, `---`, `###`, `\n`, `\n\n`, `// ----`

Currently, text proto indexer uses `proto2::TextFormat::Parser` to parse the text format which also produces `proto2::TextFormat::ParseInfoTree` to keep track of source locations. Since proto2::TextFormat::Parser does not support parsing multiple records from single file, we need to extract each chunk out and also keep track of line-offset to get the correct location.

PR #5529, opened 2023-03-07
chore(textproto_indexer): add recordio textformat support (#5528)

chuthan20:
Parse each recordio of textformat using the specified record separator and index it. Recordio textformat parser function also uses proto2::TextFormat::Parser but keeps track of line offset where that record begins in the file.

By default, empty line is the record separator, although user can specify custom separator.

### DIFF
--- a/kythe/cxx/extractor/textproto/textproto_extractor.cc
+++ b/kythe/cxx/extractor/textproto/textproto_extractor.cc
@@ -47,6 +47,9 @@ ABSL_FLAG(std::string, proto_message, "",
 ABSL_FLAG(std::vector<std::string>, proto_files, {},
           "A comma-separated list of proto files needed to fully define "
           "the textproto's schema.");
+ABSL_FLAG(std::string, record_separator, "",
+          "Delimitates each record within a recordio file. Presence of this"
+          "flag indicates this is a recordio formatted file.");
 
 namespace kythe {
 namespace lang_textproto {
@@ -80,7 +83,8 @@ Examples:
   export KYTHE_OUTPUT_FILE=foo.kzip
   textproto_extractor foo.pbtxt
   textproto_extractor foo.pbtxt --proto_message MyMessage --proto_files foo.proto,bar.proto
-  textproto_extractor foo.pbtxt --proto_message MyMessage --proto_files foo.proto -- --proto_path dir/with/my/deps")");
+  textproto_extractor foo.pbtxt --proto_message MyMessage --proto_files foo.proto -- --proto_path dir/with/my/deps
+  textproto_extractor foo.recordiotxt --proto_message MyMessage --proto_files foo.proto --record_separator @@@ -- --proto_path dir/with/my/deps")");
   std::vector<char*> remain = absl::ParseCommandLine(argc, argv);
   std::vector<std::string> final_args(remain.begin() + 1, remain.end());
 
@@ -150,6 +154,11 @@ Examples:
   compilation.mutable_unit()->add_argument(textproto_filename);
   compilation.mutable_unit()->add_argument("--proto_message");
   compilation.mutable_unit()->add_argument(std::string(schema.proto_message));
+  std::string record_separator = absl::GetFlag(FLAGS_record_separator);
+  if (!record_separator.empty()) {
+    compilation.mutable_unit()->add_argument("--record_separator");
+    compilation.mutable_unit()->add_argument(record_separator);
+  }
   // Add protoc args.
   if (!proto_extractor.path_substitutions.empty()) {
     compilation.mutable_unit()->add_argument("--");
--- a/kythe/cxx/indexer/textproto/analyzer.cc
+++ b/kythe/cxx/indexer/textproto/analyzer.cc
@@ -44,6 +44,7 @@
 #include "kythe/cxx/indexer/proto/source_tree.h"
 #include "kythe/cxx/indexer/proto/vname_util.h"
 #include "kythe/cxx/indexer/textproto/plugin.h"
+#include "kythe/cxx/indexer/textproto/recordio_textparser.h"
 #include "kythe/proto/analysis.pb.h"
 #include "re2/re2.h"
 
@@ -95,6 +96,13 @@ proto::VName LookupVNameForFullPath(absl::string_view full_path,
   return proto::VName{};
 }
 
+// The TreeInfo contains the ParseInfoTree from proto2 textformat parser
+// and line offset of the textproto within a file.
+struct TreeInfo {
+  const TextFormat::ParseInfoTree* parse_tree;
+  int line_offset = 0;
+};
+
 // The TextprotoAnalyzer maintains state needed across indexing operations and
 // provides some relevant helper methods.
 class TextprotoAnalyzer : public PluginApi {
@@ -122,14 +130,14 @@ class TextprotoAnalyzer : public PluginApi {
   absl::Status AnalyzeMessage(const proto::VName& file_vname,
                               const Message& proto,
                               const Descriptor& descriptor,
-                              const TextFormat::ParseInfoTree& parse_tree);
+                              const TreeInfo& tree_info);
 
   // Analyzes the message contained inside a google.protobuf.Any field. The
   // parse location of the field (if nonzero) is used to add an anchor for the
   // Any's type specifier (i.e. [some.url/mypackage.MyMessage]).
   absl::Status AnalyzeAny(const proto::VName& file_vname, const Message& proto,
                           const Descriptor& descriptor,
-                          const TextFormat::ParseInfoTree& parse_tree,
+                          const TreeInfo& tree_info,
                           TextFormat::ParseLocation field_loc);
 
   absl::StatusOr<proto::VName> AnalyzeAnyTypeUrl(
@@ -178,8 +186,7 @@ class TextprotoAnalyzer : public PluginApi {
 
  private:
   absl::Status AnalyzeField(const proto::VName& file_vname,
-                            const Message& proto,
-                            const TextFormat::ParseInfoTree& parse_tree,
+                            const Message& proto, const TreeInfo& parse_tree,
                             const FieldDescriptor& field, int field_index);
 
   std::vector<StringToken> ReadStringTokens(absl::string_view input);
@@ -228,9 +235,10 @@ proto::VName TextprotoAnalyzer::VNameForRelPath(
   return LookupVNameForFullPath(full_path, *unit_);
 }
 
-absl::Status TextprotoAnalyzer::AnalyzeMessage(
-    const proto::VName& file_vname, const Message& proto,
-    const Descriptor& descriptor, const TextFormat::ParseInfoTree& parse_tree) {
+absl::Status TextprotoAnalyzer::AnalyzeMessage(const proto::VName& file_vname,
+                                               const Message& proto,
+                                               const Descriptor& descriptor,
+                                               const TreeInfo& tree_info) {
   const Reflection* reflection = proto.GetReflection();
 
   // Iterate across all fields in the message. For proto1 and 2, each field has
@@ -249,11 +257,11 @@ absl::Status TextprotoAnalyzer::AnalyzeMessage(
 
       // Add a ref for each instance of the repeated field.
       for (int i = 0; i < count; i++) {
-        auto s = AnalyzeField(file_vname, proto, parse_tree, field, i);
+        auto s = AnalyzeField(file_vname, proto, tree_info, field, i);
         if (!s.ok()) return s;
       }
     } else {
-      auto s = AnalyzeField(file_vname, proto, parse_tree, field,
+      auto s = AnalyzeField(file_vname, proto, tree_info, field,
                             kNonRepeatedFieldIndex);
       if (!s.ok()) return s;
     }
@@ -271,11 +279,11 @@ absl::Status TextprotoAnalyzer::AnalyzeMessage(
     if (field->is_repeated()) {
       const size_t count = reflection->FieldSize(proto, field);
       for (size_t i = 0; i < count; i++) {
-        auto s = AnalyzeField(file_vname, proto, parse_tree, *field, i);
+        auto s = AnalyzeField(file_vname, proto, tree_info, *field, i);
         if (!s.ok()) return s;
       }
     } else {
-      auto s = AnalyzeField(file_vname, proto, parse_tree, *field,
+      auto s = AnalyzeField(file_vname, proto, tree_info, *field,
                             kNonRepeatedFieldIndex);
       if (!s.ok()) return s;
     }
@@ -342,7 +350,7 @@ absl::StatusOr<proto::VName> TextprotoAnalyzer::AnalyzeAnyTypeUrl(
 // matches fields up with the ParseInfoTree.
 absl::Status TextprotoAnalyzer::AnalyzeAny(
     const proto::VName& file_vname, const Message& proto,
-    const Descriptor& descriptor, const TextFormat::ParseInfoTree& parse_tree,
+    const Descriptor& descriptor, const TreeInfo& tree_info,
     TextFormat::ParseLocation field_loc) {
   CHECK(descriptor.full_name() == "google.protobuf.Any");
 
@@ -353,7 +361,7 @@ absl::Status TextprotoAnalyzer::AnalyzeAny(
   // assume it's a directly-specified Any and defer to AnalyzeMessage.
   auto s = AnalyzeAnyTypeUrl(file_vname, field_loc);
   if (!s.ok()) {
-    return AnalyzeMessage(file_vname, proto, descriptor, parse_tree);
+    return AnalyzeMessage(file_vname, proto, descriptor, tree_info);
   }
   const proto::VName type_url_anchor = *s;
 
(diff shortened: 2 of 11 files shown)


kythe/cxx/extractor/textproto/textproto_extractor.cc
```diff
@@ -47,6 +47,9 @@ ABSL_FLAG(std::string, proto_message, "",
 ABSL_FLAG(std::vector<std::string>, proto_files, {},
           "A comma-separated list of proto files needed to fully define "
           "the textproto's schema.");
+ABSL_FLAG(std::optional<std::string>, multiline_separator, std::nullopt,
+          "Delimitates each record within a recordio file. Presence of this"
+          "flag indicates this is a recordio formatted file.");
```
jaysachs, line 52: "indicates this expects recordio formatted files

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -927,34 +941,39 @@ absl::Status AnalyzeCompilationUnit(PluginLoadCallback plugin_loader,
...
+  for (auto& [filepath, filecontent] : file_data_by_path) {
     // Use reflection to create an instance of the top-level proto message.
     // note: msg_factory must outlive any protos created from it.
     google::protobuf::DynamicMessageFactory msg_factory;
     std::unique_ptr<Message> proto(msg_factory.GetPrototype(descriptor)->New());
 
     // Parse textproto into @proto, recording input locations to @parse_tree.
     TextFormat::ParseInfoTree parse_tree;
-    {
+    if (!is_recordio) {
```
jaysachs, line 957: Can this block be moved into the `else` of the conditional below?

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,99 @@
...
+    google::protobuf::DynamicMessageFactory& factory,
+    const google::protobuf::Descriptor* descriptor,
+    std::function<void(google::protobuf::Message*,
+                       google::protobuf::TextFormat::ParseInfoTree&, int)>
+        callback) {
+  std::string delimiter = "";
+  if (record_delimiter.has_value()) {
+    delimiter = absl::StripAsciiWhitespace(record_delimiter.value());
+  }
+  // TODO(chuthan): Accept parser as input instead of creating one here.
```
jaysachs, line 42: Please unify this with the parser setup for the non-recordio case (e.g. handling this TODO).

kythe/cxx/indexer/textproto/recordio_textparser.h
```diff
@@ -0,0 +1,55 @@
...
+#include "google/protobuf/text_format.h"
+
+namespace kythe {
+namespace lang_textproto {
+
+// Parses recordio formatted content separated by record delimiter. Callback
+// is invoked for each chunk that is parsed into proto message along with the
+// line offset of where that chunk begins in the file.
+//
+// Empty lines can be record delimiter and they can also separate two chunks
```
jaysachs, line 32: Really, that's a case we need to handle?

jaysachs (commented):
Is there a canonical public reference for "recordio" format?

shahms:
> Is there a canonical public reference for "recordio" format?

The closest thing is https://mesos.apache.org/documentation/latest/recordio/

But regardless, bother internal and external "recordio" formats are binary formats, not just delimited ones.

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,99 @@
...
+  parser.AllowPartialMessage(true);
+  parser.AllowUnknownExtension(true);
+
+  std::istringstream stream(content);
+  std::ostringstream chunk;
+  std::string line;
+  int num_of_lines = 0;
+  int chunk_begin_line = 0;
+  bool textprto_chunk_read = false;
+  while (std::getline(stream, line)) {
```
shahms, line 58: This is largely just `absl::StrSplit(content, absl::StrCat("\n", delimiter, "\n"))` with unnecessary copies, but I agree with Jay that this should be pulled into the main loop.

shahms (changes requested):
In general, the way to accomplish this is to create an outer message into which the "records" can be embedded, while still being a single protocol buffer.

kythe/cxx/indexer/textproto/recordio_textparser.cc
chuthan20: When the separator is new_line, I still need to handle the case where comments are separated by new-line. So I need to check if each chunk only contains comments/whitespaces and then make it part of next chunk.

Also, one of the recordio-textformatted files within google3 has 70K and next few are around 30K lines. 
Created some references to files within this internal document go/github-kythe-pr5529

What do you mean this should be pulled into the main loop?

shahms: By "Pulled into the main loop" I mean that the multi-chunk and extant "single-chunk" logic should not be duplicated as they are here.

Yes, this is why I say "largely". As written, the use of `std::getline` could be replaced by `absl::StrSplit(contents, ByChar('\n'))` for "free".

kythe/cxx/indexer/textproto/recordio_textparser.h
chuthan20: I've seen in google3... and when I just pass a chunk with just comment to TextFormat::Parser, it fails.

kythe/cxx/extractor/textproto/textproto_extractor.cc
chuthan20: Done.

kythe/cxx/indexer/textproto/analyzer.cc
chuthan20: Done.

kythe/cxx/indexer/textproto/recordio_textparser.cc
chuthan20: I've made this function to split and return the chunk only.. instead of also trying to parse.

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -95,6 +96,13 @@ proto::VName LookupVNameForFullPath(absl::string_view full_path,
   return proto::VName{};
 }
 
+// The TreeInfo contains the ParseInfoTree from proto2 textformat parser
+// and line offset of the textproto within a file.
+struct TreeInfo {
+  const TextFormat::ParseInfoTree& parse_tree;
```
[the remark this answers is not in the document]
chuthan20: Done.

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -95,6 +96,13 @@ proto::VName LookupVNameForFullPath(absl::string_view full_path,
...
+  int line_offset;
```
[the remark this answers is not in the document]
chuthan20: I didn't make it member, since ParseInfoTree was being passed around and also, passing this offset along makes it easy to do threading when there is multiple chunk with big recordio files.

kythe/cxx/indexer/textproto/recordio_textparser.cc
chuthan20: Thanks, I've tried to make changes to reflect what you suggested. I've made the function split into chunks and return that.. rather than also trying to convert it to proto object.

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,72 @@
...
+  std::vector<std::string> lines = absl::StrSplit(content, absl::ByChar('\n'));
+
+  std::ostringstream chunk;
+  int currentline = 0, chunk_begin_line = 0;
+  // Keep track of whether any lines in the chunk contains textproto data
+  // and not all the lines in the chunk are only comment/empty line.
+  bool seen_proto_chunk = false;
+  for (const auto& line : lines) {
+    currentline++;
+    const auto& trimmed = absl::StripAsciiWhitespace(line);
```
shahms, line 46: absl:string_view trimmed = ...

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,72 @@
...
+
+void ParseRecordioTextChunks(
+    std::string content, std::string record_delimiter,
+    std::function<void(absl::string_view chunk, int chunk_start_line)>
+        callback) {
+  absl::string_view trimmed_delimiter =
+      absl::StripAsciiWhitespace(record_delimiter);
+  std::vector<std::string> lines = absl::StrSplit(content, absl::ByChar('\n'));
+
+  std::ostringstream chunk;
```
shahms, line 39: Don't copy data to `chunk`; use the StrSplit-emitted string_views directly to calculate the "chunk". As long as you're not copying them around, they will refer to the underlying string_view from the argument and can be used directly.

Any time you would be appending to the stream you should be "extending" the current chunk instead, e.g.

void ExtendBy(absl::string_view& data, std::size_t count) {
  data = absl::string_view(data.data(), data.size() + count);  
}

But you need to ensure that `data` never exceeds the size of the underlying data, this is likely best accomplished by writing your own delimiter for StrSplit which includes the delimiter itself, if found, in the returned line, e.g.
```
struct WithChar {
  explicit WithChar(char ch) : delimiter_(ch) {}
  absl::string_view Find(absl::string_view text, size_t pos) const {
    absl::string_view sep = delimiter_.Find(text, size_t);
    // Always return a zero-width span after the delimiter, so that it's included if present.
    sep.remove_prefix(sep.size()); 
    return sep.
  }
  private:
    absl::ByChar delimiter_;
};
```

shahms (commented):
Thanks! A few more suggestions on efficient implementation.

kythe/cxx/indexer/textproto/analyzer.cc
```diff
@@ -970,10 +974,43 @@ absl::Status AnalyzeCompilationUnit(PluginLoadCallback plugin_loader,
...
+    parser.AllowPartialMessage(true);
+    parser.AllowUnknownExtension(true);
+
+    auto analyzeMessage = [&](absl::string_view chunk, int start_line) {
+      LOG(INFO) << "Analyze chunk at line: " << start_line;
+      // Parse textproto into @proto, recording input locations to @parse_tree.
+      TextFormat::ParseInfoTree parse_tree;
+      parser.WriteLocationsTo(&parse_tree);
+
+      if (!parser.ParseFromString(std::string(chunk), proto.get())) {
```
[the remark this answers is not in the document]
shahms: Sigh. This requires updating to a newer version of protobuf which accepts absl::string_view and that's another whole thing for a variety of reasons.

To work around that, I would write:
```
google::protobuf::io::ArrayInputStream stream(chunk.data(), chunk.size());
if (!parser.Parse(&stream, proto.get())) {
...
```

kythe/cxx/indexer/textproto/recordio_textparser.cc
chuthan20: I am bit confused about the extending "chunk" part. 

Do I need to keep a reference to first line in this case, and keep extending that until I find a record delimiter? Something like this? This is not working as expected for me.
```
absl::string_view chunk;
for (absl::string_view line : absl::StrSplit(content, WithChar('\n'))) {
    if (chunk.empty()) {
      chunk = line;
    }
    if (IsRecordDelimiter(line) {
       callback(...);
       continue;
    }
    if (IsEmptyOrAComment(line)) {
       ExtendBy(chunk, line.size());
       continue;
    }
   ...
}
```

shahms: Yes, you should be able to transform what you have now into roughly:
```
    if (seen_proto_chunk && (trimmed == trimmed_delimiter)) {
      LOG(INFO) << "Read chunk at " << chunk_begin_line;
      callback(chunk, chunk_begin_line);

      chunk.remove_prefix(chunk.size());
      seen_proto_chunk = false;
      chunk_begin_line = currentline;
      continue;
    }
    if (absl::StartsWith(trimmed, "#") || trimmed.empty()) {
      ExtendBy(chunk, line.size());
      continue;
    }
    // Textproto data, so append it to chunk.
    seen_proto_chunk = true;
    ExtendBy(chunk, line.size());
  }
  // Last line might contains textproto data.
  if (seen_proto_chunk) {
    callback(chunk, chunk_begin_line);
  }
```

chuthan20: Thanks! it works now!

shahms (commented):
Thanks! Very close, just a couple more minor comments.

kythe/cxx/indexer/textproto/recordio_textparser.h
```diff
@@ -0,0 +1,56 @@
...
+// with the line offset of where that chunk begins in the file.
+//
+// Empty lines can be record delimiter and they can also separate two chunks
+// of comments. Always gather chunk of comments into one big chunk along
+// with next message. For example:
+// 1.  // FooBar
+// 2.  // Bar
+// 3.
+// 4.  // BarBar
+// 5.  name: "hello"
```
shahms, line 41: The comment delimiters are wrong here, they should be `#`

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,107 @@
...
+    absl::FunctionRef<void(absl::string_view chunk, int chunk_start_line)>
+        callback) {
+  int currentline = 0, chunk_begin_line = 0;
+  // Keep track of whether any lines in the chunk contains textproto data
+  // and not all the lines in the chunk are only comment/empty line.
+  bool seen_proto_chunk = false;
+  absl::string_view trimmed_delimiter =
+      absl::StripAsciiWhitespace(record_delimiter);
+  absl::string_view chunk;
+  for (absl::string_view line : absl::StrSplit(content, WithChar('\n'))) {
```
shahms, line 68: Optional: I suspect most of this function can be replaced with StrSplit and a delimiter class and be simpler, e.g.

```
class ProtoLineDelimiter {
 public:
  explicit ProtoLineDelimiter(absl::string_view delimiter, int* line_count = nullptr)
      : delimiter_(delimiter), line_count_(line_count) {}

  /// \brief Finds the next occurrence of the configured delimiter
  /// on a line by itself, after the first non-comment, non-empty line.
  absl::string_view Find(absl::string_view text, size_t pos) {
    for (absl::string_view line :
         absl::StrSplit(text.substr(pos), WithChar('\n'))) {
      if (line_count_) {
        ++*line_count_;
      }
      // Don't look for the delimiter until we've seen our first non-empty,
      // non-comment line.
      data_seen_ = data_seen_ || !(absl::StartsWith(line, "#") ||
                                              absl::StripPrefix(line, "\n").empty());
      if (!data_seen_) continue;

      if (absl::string_view stripped = absl::StripPrefix(line, delimiter_);
          // The line consists entirely of the delimiter.
          data_seen_ && absl::StripPrefix(stripped, "\n").empty()) {
        return line;
      }
    }
    return text.substr(text.size());
  }

 private:
  std::string delimiter_;
  int* line_count_;

  bool data_seen_ = false;
};
...
int line_count = 0;
for (absl::string_view chunk : absl::StrSplit(data, ProtoLineDelimiter(record_delimiter, &line_count)) {
  callback(chunk, line_count);
}
```

This lets StrSplit itself keep track of all of the chunk-information and the loop just has to count lines and find the next delimiter.

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,107 @@
...
+
+void ParseRecordTextChunks(
+    absl::string_view content, absl::string_view record_delimiter,
+    absl::FunctionRef<void(absl::string_view chunk, int chunk_start_line)>
+        callback) {
+  int currentline = 0, chunk_begin_line = 0;
+  // Keep track of whether any lines in the chunk contains textproto data
+  // and not all the lines in the chunk are only comment/empty line.
+  bool seen_proto_chunk = false;
+  absl::string_view trimmed_delimiter =
```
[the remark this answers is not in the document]
chuthan20: This is what the actual tool within google3 that converts recordio-text to proto does. But, if you feel strongly, I can check without trimming and maybe send CLs to fix up the text files that is violating this.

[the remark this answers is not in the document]
shahms: Yes, I'm generally not a fan of ignoring/modifying the user-provided delimiter.

kythe/cxx/indexer/textproto/recordio_textparser.cc
chuthan20: I am using this code now.. but changed:
- record delimiter could start with a comment.. for example: `# ---`
- each chunk needs the start line of the chunk so made changes to reflect this.

kythe/cxx/indexer/textproto/recordio_textparser.cc
```diff
@@ -0,0 +1,107 @@
...
+
+    bool initial_chunk = false;
+    if (chunk.empty()) {
+      chunk = line;
+      initial_chunk = true;
+    }
+    absl::string_view trimmed = absl::StripAsciiWhitespace(line);
+    // So far, chunk has textproto data (not just comments or empty lines)
+    // and current line is the delimiter.
+    if (seen_proto_chunk && (trimmed == trimmed_delimiter)) {
```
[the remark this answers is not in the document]
chuthan20: not trimming the delimiter anymore.

kythe/cxx/indexer/textproto/recordio_textparser.cc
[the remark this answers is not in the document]
chuthan20: not trimming the delimiter anymore.

chuthan20:
Oops, didn't mean to merge, just wanted to pull the latest changes from master.

PR #5530, opened 2023-03-09
feat(serving): support scoped cross-references

schroederc:
This change only adds support to the ScopedReferences in the serving data, and the actual server will unwrap them into unscoped references. A future change will support ScopedReferences in the CrossReferences API.


justbuchanan (approved):

kythe/proto/serving.proto
```diff
@@ -276,6 +276,20 @@ message PagedCrossReferences {
     int32 ordinal = 2;
   }
 
+  // References to the source node with an associated semantic scope.
+  message ScopedReference {
```
creachadair, line 280: This seems like an interesting change! Will this require any schema changes for the indexers to generate the scope data?

schroederc: Not really.  It requires the `childof` edges that are already part of the callsite relation.  A couple of indexers already emit this data: https://github.com/kythe/kythe/pull/4187 and https://github.com/kythe/kythe/commit/0a3afc9bffba3e09c638ae28575605a1441ebf5e

creachadair: Very cool. 🥰

PR #5537, opened 2023-03-14
fix(cxx_extractor): the main source file should be marked always_process

shahms:
There is some argument to be made that this logic should (also?) go in the indexer as the indexer should always process the primary source file, regardless of claiming, but we already have a facility for this in the extractor and just need to change one place so that the primary source file is always processed.

Background: when using dynamic claiming, prospective claims are issued with a timeout, but if a given compilation unit abandoned midway through indexing it can sometimes be retried before the previous prospective claims have expired, thus skipping files which it should index.  While this is problematic for header files as well, it is most troublesome for source files which are unlikely to be indexed at all as a result. Header files may potentially be included by another compilation unit and therefore indexed separately.

Pros of this approach: very small change, uses pre-existing facility.
Cons: 
 * This arguably belongs in the indexer itself
 * can't be flag-controlled in extractor
 * need to update all of the tests due to the change in the resulting proto files

### DIFF
--- a/kythe/cxx/extractor/cxx_extractor.cc
+++ b/kythe/cxx/extractor/cxx_extractor.cc
@@ -548,7 +548,7 @@ void ExtractorPPCallbacks::FileChanged(
   if (Reason == EnterFile) {
     if (last_inclusion_directive_path_.empty()) {
       current_files_.push(FileState{std::string(GetMainFile()->getName()),
-                                    ClaimDirective::NoDirectivesFound});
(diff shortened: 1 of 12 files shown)


zrlk (approved):

jaysachs (approved):
I'd agree it belongs in the indexer but this is terribly convenient.

creachadair:
This seems like a nice workaround. Our original idea for static claiming was too expensive to be practical on google3, but this is a decent tradeoff.

PR #5538, opened 2023-03-14
feat(cxx_indexer): support simple aliases to declrefexprs

zrlk:
This PR marks the ref to DeclRefExpr E in *&E = ... as a write.


kythe/cxx/indexer/cxx/clang_utils.cc
```diff
@@ -159,6 +159,12 @@ bool ShouldHaveBlameContext(const clang::Decl* decl) {
 
 const clang::Stmt* FindLValueHead(const clang::Stmt* stmt) {
   if (stmt == nullptr) return nullptr;
+  stmt = Unparen(stmt);
```
shahms, line 162: This is just `clang::Expr::IgnoreParens()`

`FindLValueHead` is called exclusively on `clang::Expr`, and `GetInfluencedDeclFromLValueHead` would be as well if `FindLValueHead` and `UsedAsWrite` were changed to `Expr` (which is what both operate on exclusively).

zrlk: Is IgnoreParens safe to use if an expr has embedded nulls?

shahms: It looks like it will return nullptr in that case:
https://clang.llvm.org/doxygen/IgnoreExpr_8h_source.html#l00034
https://clang.llvm.org/doxygen/IgnoreExpr_8h_source.html#l00150

However, `getSubExpr` cannot return nullptr:
https://clang.llvm.org/doxygen/Expr_8h_source.html#l02138

 `cast<Expr>(Val)` will neither accept nor return a null pointer:
https://llvm.org/doxygen/Casting_8h_source.html#l00558

shahms (approved):

PR #5542, opened 2023-03-20
fix(cxx_indexer): emit proper marked source for macro-renamed functions

shahms:
This also changes the behavior of macro-expanded ranges slightly: if the range would expand to the entirety of the macro body, use the full expansion range rather than the zero-width span.  This could cause some knock-on range changes but should be more intuitive for those affected cases than the current behavior.

Prior to this, the marked source for functions renamed in this fashion would include the *entire* function body, not just the signature.


zrlk (approved):

kythe/cxx/indexer/cxx/clang_range_finder.cc
```diff
@@ -50,35 +44,53 @@ CharSourceRange GetImmediateFileRange(
...
+  // For macro body ranges we want to return a zero-width range
+  // pointing to the beginning of the macro expansion so that any
+  // edges originating from within the macro body are suppressed.
+  // TODO(shahms): Emit the properly expanded range in all cases once UI
+  // issues are addressed.
+  if (
+      // Only if both the immediately prior range is within a macro body for
+      // both locations.
+      (begin_in_macro_body && end_in_macro_body) &&
+      // And only if both ends would not cover the entire macro expansion.
```
creachadair, line 79: Maybe I'm misunderstanding, but wouldn't this rule out, e.g., name decorating macros, like

```c
#define F(X) curse_ ## X

F(you)
```

where the resulting identifier _is_ the whole expansion?

shahms: I don't know what you mean by "rule out", but this changes the `F(you)` from being undecorated to decorating the entire `F(you)` range.

PR #5546, opened 2023-03-22
chore: update Bazel java-related rules

shahms:
As part of updating the protobuf dependency, update to newer rules_java and rules_jvm_external.  The newer rules_jvm_external required re-pinning the Maven dependencies, which also re-pinned Cargo for some reason.

The new toolchains threw some ErrorProne issues, which were fixed.

Finally, update the maximum Bazel version along the way.


zrlk:
Are the presubmit errors expected?

shahms:
> Are the presubmit errors expected?

No, Bazel 5.3.0 is crabby. Looking to see if the fix is easy or if I need to bump the minversion.

shahms:
It looks like newer versions of the remote_java_tools repos removed the targets Bazel 5.3.0 was relying on, so that's not an easy fix.

jaysachs (approved):

Issue #5543, opened 2023-03-21
Support C++ modules


zrlk:
Support implicit and explicit C++ modules in extraction and indexing.

PR #5554, opened 2023-03-28
fix(cxx_extractor): don't crash when modules are on

zrlk:
In service of #5543.

This PR should only affect the behavior of the extractor where it would have either crashed because of a null dereference or due to a CHECK-fail. The conditions leading to these cases occur when modules are turned on.


shahms (commented):
Please add a test for this.

zrlk:
It appears that the OSS Bazel rules don't support header-only modules. An internal test would probably work; otherwise I think we're looking at a sh test with the cxx_extractor binary (ugh) or a hand-edited xa (ugh).  I would be happy to be proven incorrect.

shahms:
> 

None of the existing extractor integration tests use the Bazel rules or shell, see, for example: https://github.com/kythe/kythe/blob/master/kythe/cxx/extractor/testdata/root_directory_test.cc

Is there some reason that approach won't work here?

zrlk:
OK, the new test hits the new code path. I expect to refine the set of flags as I dig deeper into this. (also ugh `-Xclang`)

kythe/cxx/extractor/cxx_extractor.cc
```diff
@@ -591,8 +619,10 @@ PreprocessorTranscript ExtractorPPCallbacks::PopFile() {
 }
 
 void ExtractorPPCallbacks::EndOfMainFile() {
-  AddFile(GetMainFile(), std::string(GetMainFile()->getName()));
-  *main_source_file_transcript_ = PopFile();
+  if (auto* mfile = GetMainFile()) {
+    AddFile(GetMainFile(), std::string(GetMainFile()->getName()));
```
shahms, line 623: `s/GetMainFile()/mfile/g` on this line?

shahms (approved):
Thanks!

PR #5568, opened 2023-04-04
fix(go_indexer): don't strip extensions from import paths

schroederc:


kythe/go/extractors/govname/govname.go
```diff
@@ -227,7 +227,7 @@ func ImportPath(v *spb.VName, goRoot string) string {
 		// that nonstandard install locations will work sensibly.
 		return tail
 	}
-	return filepath.Join(v.Corpus, trimmed)
+	return filepath.Join(v.Corpus, v.Path)
```
shahms, line 230: I think the original is correct, but there is some unfortunate logic in the extractor which already trims the file extension, e.g.

https://github.com/kythe/kythe/blob/acdc44d2ab1c3f2fd1dfa301e8f621b6750c0b9e/kythe/data/vnames.go.json#L32

schroederc: That's trimming .x and .a which is correct.  We shouldn't trim version extensions; this fixes that.

shahms: I do not believe the extractor should be trimming the file extension, as it leads to issues like this.

Clearly not all paths have their extensions trimmed during extraction, which just leads to more ambiguity as to whether a path is a path or Something Else. The failing test suggests that this isn't the only place that mistake has happened.

schroederc: The "path" here is the import/package path, not a file path.  It definitely should not have those extensions.

shahms: If that were consistently true, this wouldn't be an issue (or would be less of an issue).  Looking at all of the places ImportPath is called, it's clearly not always a package path:

## Indexer
https://github.com/kythe/kythe/blob/72849483b91ad3264db7493395524d0620728d5b/kythe/go/indexer/indexer.go#L349 (file VName, file path)
https://github.com/kythe/kythe/blob/72849483b91ad3264db7493395524d0620728d5b/kythe/go/indexer/indexer.go#L390 (CU VName, ??? path)
https://github.com/kythe/kythe/blob/72849483b91ad3264db7493395524d0620728d5b/kythe/go/indexer/indexer.go#L872 (package VName, package path)

## Go-tool Extractor
https://github.com/kythe/kythe/blob/7c9b65bb652856bd9fb3337017fd222000c87f77/kythe/go/extractors/golang/golang.go#L279 (CU VName, ??? path)
https://github.com/kythe/kythe/blob/7c9b65bb652856bd9fb3337017fd222000c87f77/kythe/go/extractors/golang/golang.go#L408 (file VName?, file path)
https://github.com/kythe/kythe/blob/7c9b65bb652856bd9fb3337017fd222000c87f77/kythe/go/extractors/golang/golang.go#L468 (file VName?, file path)

shahms (approved):
This seems like a reasonable compromise, at least until folks start routinely suffixing packages with `.a` and `.x`

PR #5569, opened 2023-04-07
feat(typescript): add ref/id edges for shorthands

nbeloglazov:
Also refactors code a bit:
*  Exit early when vname is undefined as lot's of logic depends on vname being present. 
* Consistently use `ts.is*` functions.


nbeloglazov:
@shahms could you approve?

shahms (approved):

PR #5571, opened 2023-04-13
fix(cxx_extractor): don't validate pchs

zrlk:
The extractor will refuse to load pch/modules if they are built from a different libclang than the extractor binary. This is not ideal. Of course, there may be breaking changes in the pch format, but it's not unreasonable to assume that these changes happen less than daily. (Initial impressions hold that we still need to get around this problem to read even simple file content from a pch.)

This PR disables pch validation. Since we don't officially support builds that require pch/modules yet, the danger here is that we crash on a bad or incompatible (or actually corrupt) pch instead of quitting nicely with a diagnostic.

In service of #5543.

### DIFF
--- a/kythe/cxx/extractor/cxx_extractor.cc
+++ b/kythe/cxx/extractor/cxx_extractor.cc
@@ -41,6 +41,7 @@
 #include "clang/Lex/MacroArgs.h"
 #include "clang/Lex/PPCallbacks.h"
 #include "clang/Lex/Preprocessor.h"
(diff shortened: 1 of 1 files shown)


shahms:
```
kythe/cxx/extractor/cxx_extractor.cc:999:29: error: member access into incomplete type 'clang::PreprocessorOptions'
    CI.getPreprocessorOpts().DisablePCHOrModuleValidation =
```

PR #5576, opened 2023-04-14
serving: add fields for missing text facts

schroederc:


kythe/proto/serving.proto
```diff
@@ -144,6 +147,10 @@ message File {
   string encoding = 3;
 
   FileInfo info = 4;
+
+  // Whether the File did not have a text fact.  Necessary to differentiate
+  // between an empty text and missing text.
```
shahms, line 152: proto3 supports `optional` now which can differentiate between missing and empty; is that a viable option instead?

schroederc: That works

shahms (approved):

kythe/proto/serving.proto
creachadair: Ah, I see proto3 is working its way back toward being proto2 😂

PR #5577, opened 2023-04-14
chore(cxx_extractor): clarify differently rooted paths and clean up relative path names

shahms:
Consistently omit the leading "./" from relative paths and make root-vs-working path mistakes harder to make in the future.


shahms:
Actually, looking at the IndexVFS code, the indexer already handles `..` and `.` components lexically; I can just implement the TODO..

kythe/cxx/extractor/cxx_extractor.cc
```diff
@@ -621,14 +654,14 @@ PreprocessorTranscript ExtractorPPCallbacks::PopFile() {
...
+    *main_source_file_ = AddFile(mfile, mfile->getName());
     *main_source_file_transcript_ = PopFile();
   }
 }
 
 std::string ExtractorPPCallbacks::FixStdinPath(const clang::FileEntry* file,
-                                               const std::string& in_path) {
-  if (in_path == "-" || in_path == "<stdin>") {
+                                               llvm::StringRef path) {
+  if (path == "-" || path == "<stdin>") {
```
zrlk, line 664: Any reason why `IsStdinPath` shouldn't be used here?

zrlk (approved):

kythe/cxx/extractor/cxx_extractor.cc
shahms: Not really, no.

PR #5580, opened 2023-04-17
feat(api): guard returning file entries with missing text

schroederc:


kythe/proto/filetree.proto
```diff
@@ -77,6 +80,9 @@ message DirectoryReply {
 
     // True if the entry is generated.
     bool generated = 4;
+
+    // Whether the FILE entry is missing text.
+    bool missing_text = 5;
```
jaysachs, line 85: Might be more natural in the positive sense. Though does `has_text` conflict with proto3 hazzers?

schroederc: The picked the default that would be more common so we don't just populate `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true`, `true` in every response.

jaysachs (approved):

PR #5583, opened 2023-04-17
feat(java): Add java20 shim for java indexer

salguarnieri:


buildenv/java/BUILD
```diff
@@ -13,6 +13,14 @@ default_java_toolchain(
     target_version = "19",
 )
 
+default_java_toolchain(
+    name = "toolchain_jdk_20",
+    configuration = dict(),
+    java_runtime = "@remotejdk20_linux//:jdk",
```
shahms, line 19: You'll have to add this repository as well. It's annoying.  You can probably follow the format from https://github.com/bazelbuild/rules_java/blob/master/java/repositories.bzl but I'm not sure how to get a good URL for the JDK itself.

salguarnieri: Thanks for the link. I got some URLs from https://github.com/bazelbuild/bazel/blob/master/distdir_deps.bzl and the presubmit seems to be happy. Not sure if there are other checks I can do.

shahms (approved):

buildenv/java/BUILD
salguarnieri: I ran bazel test --java_language_version=20 ... and it didn't fail.

PR #5604, opened 2023-04-26
chore(java_indexer): log in addition to emitDiagnostic

salguarnieri:


kythe/java/com/google/devtools/kythe/analyzers/java/KytheTreeScanner.java
```diff
@@ -484,7 +484,10 @@ public JavaNode visitClassDef(JCClassDecl classDef, TreeContext owner) {
       if (member instanceof JCMethodDecl) {
         JCMethodDecl method = (JCMethodDecl) member;
         if (method.sym == null) {
-          emitDiagnostic(ctx, "missing method symbol", null, null);
+          String name = method.name == null ? "" : method.name.toString();
+          logger.atWarning().log(
+              "Missing method symbol: %s at %s:%s", name, ctx.getSourcePath(), ctx.getTreeSpan());
+          var unused = emitDiagnostic(ctx, "missing method symbol: " + name, null, null);
```
shahms, line 490: why the unused var?

salguarnieri: I don't think we really want to exit early here and return the diagnostic node, but since emitDiagnostic returns a value linters require we assign it to `unused`.

shahms: That seems like the wrong default for such a linter.

shahms (approved):

PR #5612, opened 2023-05-01
fix(cxx_indexer): find vnames for pch/pcm files

zrlk:
Files loaded from pch/pcms don't trigger the same preprocessor callbacks as files encountered through normal includes. This causes us to assign them the default claim token/default file vname, which in turn breaks references from dependent modules. This PR ascribes vnames to pch/pcm-sourced files in the standard way.

It may be possible to preload claim tokens for pch/pcm files. This is left to a future PR.

In service of #5543

### DIFF
--- a/kythe/cxx/indexer/cxx/GraphObserver.h
+++ b/kythe/cxx/indexer/cxx/GraphObserver.h
@@ -351,15 +351,14 @@ class GraphObserver {
 
   /// \brief Returns a claim token for namespaces declared at `Loc`.
   /// \param Loc The declaration site of the namespace.
-  virtual const ClaimToken* getNamespaceClaimToken(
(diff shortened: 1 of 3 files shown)


kythe/cxx/indexer/cxx/GraphObserver.h
shahms, line 1: These are all still logically `const` and should remain so. If this implementation needs to be non-const, use `mutable` and a mutex.

schroederc (approved):

shahms (changes requested):

kythe/cxx/indexer/cxx/GraphObserver.h
shahms: Then just `mutable` and a thread-hostile class comment should suffice.

PR #5614, opened 2023-05-01
fix(cxx_indexer): find vnames for pch/pcm files (const edition)

zrlk:
Files loaded from pch/pcms don't trigger the same preprocessor
callbacks as files encountered through normal includes. This
causes us to assign them the default claim token/default file
vname, which in turn breaks references from dependent modules.
This PR ascribes vnames to pch/pcm-sourced files in the
standard way.

It may be possible to preload claim tokens for pch/pcm files.
This is left to a future PR.

In service of #5543


shahms (approved):
Thanks!

shahms (approved):

PR #5626, opened 2023-05-10
chore(java_indexer): prep for multijar

salguarnieri:
Put shim implementations in different directories under the same class name so they can be included in a multijar.


shahms (approved):
Do we want to do this with the extractor shims as well?

salguarnieri:
https://github.com/kythe/kythe/pull/5627 for the extractor shims

PR #5629, opened 2023-05-12
feat(java): add semantic/generated=set facts for annotated proto accessors

justbuchanan:
I have some proto verifier tests to exercise this, but they require the internal changes to `protoc` to be pushed to github, then included in a release, then updated in our dependencies. Merging this earlier will allow us to try this out in internally sooner.

### DIFF
--- a/kythe/java/com/google/devtools/kythe/platform/shared/Metadata.java
+++ b/kythe/java/com/google/devtools/kythe/platform/shared/Metadata.java
@@ -34,8 +34,10 @@ import java.util.List;
 public class Metadata {
   /** An additional semantic to apply to the given (java) node. */
   public enum Semantic {
-    NONE, /// < No special semantics.
-    WRITE, /// < Write semantics.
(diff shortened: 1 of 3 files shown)


kythe/java/com/google/devtools/kythe/platform/shared/Metadata.java
```diff
@@ -32,6 +32,14 @@
  * path to a loadable metadata file.
  */
 public class Metadata {
+  /** An additional semantic to apply to the given (java) node. */
+  public enum Semantic {
```
justbuchanan, line 36: @zrlk I added a separate enum here rather than re-using the proto-defined one since the c++ implementation did the same. Is this necessary? The only advantage I think for java is that we're not using ALIAS, so that's omitted from the java-defined enum.

justbuchanan: Sounds good, done.

PR #5630, opened 2023-05-12
feat(java_indexer): add Semantic field to protobuf Metadata

nbeloglazov:
The field will be used by java-based indexers to emit `semantic/generated` facts when processing proto-generated code.


justbuchanan:
This has some overlap with https://github.com/kythe/kythe/pull/5629. I'm not sure why I added a separate enum for the semantic type though - I think I copied it from the c++ implementation, but it shouldn't be necessary. Let's go with your PR and I'll rebase mine afterwards.

This LGTM, but let's have @zrlk take a look.

nbeloglazov:
Ok, sg! I'll wait for your PR to get to google3 and then use it for JS indexer.

nbeloglazov:
Nice. You literally opened your PR like 20 min ago.

PR #5638, opened 2023-05-18
fix(cxx_common): attempt to reduce memory usage of bazel artifact selector

shahms:
This should reduce memory usage somewhat, although there are likely more gains to be had particularly during serialization.  This should incidentally reduce the size of the serialized state by removing the unused fields and using a slightly more compact format for the fields which *are* used.

### DIFF
--- a/kythe/cxx/extractor/bazel_artifact_selector.cc
+++ b/kythe/cxx/extractor/bazel_artifact_selector.cc
@@ -127,10 +127,20 @@ absl::optional<BazelArtifact> AspectArtifactSelector::Select(
 bool AspectArtifactSelector::SerializeInto(google::protobuf::Any& state) const {
   kythe::proto::BazelAspectArtifactSelectorState raw;
   *raw.mutable_disposed() = FromRange{state_.disposed};
-  *raw.mutable_filesets() = FromRange{state_.filesets};
+  for (const auto& [key, fileset] : state_.filesets) {
+    auto& entry = (*raw.mutable_filesets())[key];
+    for (const auto& file : fileset.files) {
+      auto* file_entry = entry.add_files();
+      file_entry->set_name(file.local_path);
+      file_entry->set_uri(file.uri);
+    }
+    for (const auto& id : fileset.children) {
+      entry.add_file_sets()->set_id(id);
+    }
+  }
   *raw.mutable_pending() = FromRange{state_.pending};
 
-  state.PackFrom(raw);
+  state.PackFrom(std::move(raw));
   return true;
 }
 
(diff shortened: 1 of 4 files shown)


kythe/cxx/extractor/bazel_artifact_selector.h
```diff
@@ -168,13 +174,19 @@ class AspectArtifactSelector final : public BazelArtifactSelector {
   absl::Status DeserializeFrom(const google::protobuf::Any& state) final;
 
  private:
+  struct FileSet {
+    std::vector<const BazelArtifactFile*> files;
```
jaysachs, line 178: The `files` hash_map owns these objects, right?

kythe/cxx/extractor/bazel_artifact_selector.cc
```diff
@@ -214,19 +238,22 @@ void AspectArtifactSelector::ReadFilesInto(
...
-    for (const auto& file : filesets.files()) {
-      files.push_back({
-          .local_path = AsLocalPath(file),
-          .uri = AsUri(file),
-      });
+    const FileSet& fileset = node.mapped();
+    files.reserve(files.size() + fileset.files.size());
+    for (const auto* file : fileset.files) {
+      auto iter = state_.files.find(*file);
+      CHECK(iter != state_.files.end()) << "Attempt to remove a missing file!";
```
jaysachs, line 247: So if we get bad data we crash? i.e. this is an unrecoverable state?

kythe/cxx/extractor/bazel_artifact_selector.cc
```diff
@@ -214,19 +238,22 @@ void AspectArtifactSelector::ReadFilesInto(
...
     state_.disposed.insert(std::string(id));
-    auto node = state_.filesets.extract(iter);
-    const build_event_stream::NamedSetOfFiles& filesets = node.mapped();
-
-    for (const auto& file : filesets.files()) {
-      files.push_back({
-          .local_path = AsLocalPath(file),
-          .uri = AsUri(file),
-      });
+    const FileSet& fileset = node.mapped();
```
jaysachs, line 243: I'll just note that with everything named "file" or "iter" it's a little challenging to follow what's going on.

jaysachs (commented):
Are there any tests that could be updated or written?

kythe/cxx/extractor/bazel_artifact_selector.h
shahms: Yeah, that's right. I'll add a comment to that effect in a separate PR.

kythe/cxx/extractor/bazel_artifact_selector.cc
shahms: This is unrecoverable. The `files` set owns the data. If we fail to lookup the record from a pointer to that record, we're well in UB territory and thoroughly screwed.

shahms:
> Are there any tests that could be updated or written?

The user-visible behavior here is intentionally identical so there's not really anything to test at the moment.

PR #5658, opened 2023-05-24
deprecation(schema): remove completes edge kind

schroederc:


tools/schema/edges
```diff
@@ -5,9 +5,6 @@ upper
 lower
 childof
 context
-completedby
```
shahms, line 8: retain `completedby`?

shahms (approved):

PR #5664, opened 2023-05-26
chore(tooling): record -source values from java CUs for kzip info

salguarnieri:


kythe/go/platform/kzip/info/info.go
```diff
@@ -124,6 +132,26 @@ func (a *Accumulator) Accumulate(u *kzip.Unit) {
 	a.KzipInfo.AbsolutePaths = absPaths.Elements()
 }
 
+// javaSourceVersion checks if the CU args include setting source to a language level and if it does
+// it returns that version. If no source flag is set or it can't be read, it returns 0.
+func javaSourceVersion(args []string) int {
+	for i, arg := range args {
+		if arg == "-source" || arg == "--source" {
```
justbuchanan, line 139: Is it possible to get something like `-source=1.8`? Or will it always be separate [`-source`, `1.8`]

salguarnieri: A simple test says it doesn't like `=`

```
$ javac -source=9 A.java
error: invalid flag: -source=9
Usage: javac <options> <source files>
use --help for a list of possible options
```

justbuchanan: :+1:

justbuchanan (approved):

PR #5670, opened 2023-05-30
fix(verifier): depend on absl optional library

zrlk:


shahms (approved):
We can also just switch everything over to `std::optional`

PR #5673, opened 2023-05-31
fix(go): avoid formatting protos via %v for non-debug uses

jaysachs:


salguarnieri (approved):

schroederc:
You can also use `%s` to use the standard proto String representation.

PR #5681, opened 2023-06-01
test: deterministically format VNames

schroederc:


jaysachs (approved):
So my original PR was closer to correct. Vindicated!

PR #5728, opened 2023-06-30
chore(cxx_common): migrate to Abseil logging

shahms:
Migrate off of `google/glog` as it is no longer maintained and onto `absl/log` as it is.  This unblocks a few other cleanups and re-enables logging-related flags.  This PR includes some hacks to deal with the fact that Abseil logging does not yet have `VLOG` support.  The pattern used is intended to be trivial to replace on import, but is still ugly.


shahms:
Ugh, missed the commit fixing that because commitlint didn't like `//kythe/cxx/...`

jaysachs (approved):

PR #5730, opened 2023-06-30
feat(cxx_indexer): add GraphObserver.recordDiagnostic()

justbuchanan:


kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -970,6 +970,13 @@ class GraphObserver {
   virtual void recordDeprecated(const NodeId& NodeId,
                                 const llvm::StringRef& Advice) {}
 
+  /// \brief Records a diagnostic at the given source range
+  /// \param Range The source range
+  /// \param Signature The signature to use for the diagnostic VName
+  /// \param Message The diagnostic message
+  virtual void recordDiagnostic(const Range& Range, std::string_view Signature,
+                                std::string_view Message) {}
```
shahms, line 978: minor: This should probably use `llvm::StringRef` (or we should switch every method to use `std::string_view`)

kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -974,8 +974,9 @@ class GraphObserver {
   /// \param Range The source range
   /// \param Signature The signature to use for the diagnostic VName
   /// \param Message The diagnostic message
-  virtual void recordDiagnostic(const Range& Range, std::string_view Signature,
-                                std::string_view Message) {}
+  virtual void recordDiagnostic(const Range& Range,
+                                const llvm::StringRef& Signature,
+                                const llvm::StringRef& Message) {}
```
shahms, line 979: `llvm::StringRef` is like `absl::string_view` in that both should be passed by value rather than `const-&`

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
```diff
@@ -1257,21 +1257,21 @@ void KytheGraphObserver::recordDeprecated(const NodeId& NodeId,
...
-                                          std::string_view Signature,
-                                          std::string_view Message) {
+                                          const llvm::StringRef& Signature,
+                                          const llvm::StringRef& Message) {
   proto::VName anchor_vname = VNameFromRange(Range);
 
   proto::VName dn_vname;
-  dn_vname.set_signature(
-      std::string(absl::StrCat(anchor_vname.signature(), "-", Signature)));
+  dn_vname.set_signature(std::string(
```
shahms, line 1265: `absl::StrCat` returns `std::string`; the explicit conversion here is unnecessary

shahms (commented):
Thanks! Just a couple minor follow-ups.

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
justbuchanan: Done.

kythe/cxx/indexer/cxx/GraphObserver.h
justbuchanan: Makes sense. I did it this way because I saw several other `StringRef&` parameters elsewhere in the file. I can send a followup to remove the `const &` from those.

shahms (approved):

PR #5748, opened 2023-07-20
chore(cxx): s/absl::string_view/std::string_view/g

justbuchanan:
The only change I reverted was in parsetextproto.h, which defines its own mock of absl::string_view

### DIFF
--- a/kythe/cxx/common/file_vname_generator.cc
+++ b/kythe/cxx/common/file_vname_generator.cc
@@ -34,7 +34,7 @@ std::string EscapeBackslashes(std::string_view value) {
 }
 
 absl::optional<std::string_view> FindMatch(std::string_view text,
-                                            const RE2& pattern) {
+                                           const RE2& pattern) {
   re2::StringPiece match;
   if (pattern.Match(text, 0, text.size(), RE2::UNANCHORED, &match, 1)) {
     return std::string_view(match.data(), match.size());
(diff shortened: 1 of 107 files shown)


shahms:
You'll need to re-run clang-format on the changed files, change the `#include "absl/strings/string_view.h" => "#include <string_view>` and update the BUILD file dependencies as well.

One possible complication here is that internally `absl::string_view` is *mostly* equivalent to `std::string_view` except in some more exotic (for google3) environments. I don't *think* we care about those, but IIRC one of them is macOS, so we may.

justbuchanan:
I updated `#include`s and reran clang-format. Is there a good automated way to prune unused BUILD deps in the open source world?

shahms:
Not for C++, sadly. Gazelle works for more languages than Go these days, but we don't even use it for that at the moment.

shahms:
Actually, I think you can leave the build dependency. ISTR that string_view is included in the "strings" umbrella library which has a bunch of other stuff. It's probably not worth disentangling that at the moment.

shahms:
The current diff is potentially a difference between clang-format-11 (from the GCB worker) and a more recent version of clang-format on your machine. We may need to update the clang-format version on GCB as 11 no longer installs on our workstations when last I checked.

justbuchanan:
I'll take a look at updating the GCB worker

justbuchanan:
The debian package repository shows clang-format-N for versions 11 to 17. Not sure which one we want.

On my machine, `clang-format --version` says clang-format "google3-trunk" and a commit hash.

shahms:
Yes, the default internal version is compiled from trunk (which is why we hard-code a specific version in the pre-commit config).  14 is probably reasonable as it was released  25 March 2022

shahms:
Thanks!

PR #5752, opened 2023-07-20
chore(cxx): s/absl::optional/std::optional/g

justbuchanan:


shahms:
Nice, thanks! I'd definitely update the BUILD rules here; it should be just a matter of deleting the `"@com_google_absl//absl/types:optional"`  dep everywhere.

justbuchanan:
> Nice, thanks! I'd definitely update the BUILD rules here; it should be just a matter of deleting the `"@com_google_absl//absl/types:optional"` dep everywhere.

Done

shahms (approved):

justbuchanan:
looks like I need to change over `nullopt` as well

shahms (approved):

justbuchanan:
std::string_view didn't go well, so I'll drop this one too for now

shahms:
This doesn't have any of the problems that string_view does, AFAIK.  string_view is special for reasons which don't apply to optional.

justbuchanan:
Ok, I'll give it a shot :)

Issue #5705, opened 2023-06-15
Clean up experimental_ flags in the C++ indexer


zrlk:
We've found useful settings for many of the experimental_ flags. These no longer need to be configurable.

PR #5756, opened 2023-07-21
chore(cxx_indexer): remove experimental_guess_proto_semantics

zrlk:
Our version of protobuf now contains the semantic field in GeneratedCodeInfo, so experimental_guess_proto_semantics is now redundant.

Our version of protobuf does *not* fill out all semantics. The ones we expect to come in a future import are commented out in the test. IIRC it was determined that arena maintenance functions should not have semantics set, so these have been removed entirely.

In service of #5705


shahms:
Is there a bug/issue to re-enable the assertions when we have a new enough version of protobuf?

shahms (approved):

PR #5762, opened 2023-07-26
chore: format proto and go build rules with custom gazelle

shahms:
To facilitate ease-of-import, our Go build rules aren't entirely compatible with the default gazelle-generated rules and so I've made a few modifications.

1) I've wrapped the default go language generation rules to suppress emitting a separate go_library which is embedded in go_binary targets.
2) I've patched gazelle to allow generating per-_test.go-file go_test targets rather than just a single go_test per package. I plan on attempting to upstream this patch, but for now it lives here.
3) The go_proto_library generation expects the go_package option to be the full import path, not the relative package (despite the name of the option, obnoxiously) so I've updated that.

This also adds a `bazel run //:gazelle` step to pre-commit to keep them from getting out of date.


.pre-commit-config.yaml
```diff
@@ -93,6 +93,7 @@ repos:
         description: Clean build dependencies with gazelle
         language: system
         entry: bazel run //:gazelle
+        stages: [push]
```
schroederc, line 96: Running on push seems like an odd choice since a commit has already been made.  Does this block the push if there are differences?

shahms: Yeah, it's like the other formatters. I mostly just changed it because the lint-runner is much more lightweight and doesn't have bazel[isk] installed. I can change it back and add bazel to that image, though.

kythe/go/util/vnameutil/BUILD
```diff
@@ -1,3 +1,4 @@
+# gazelle:go_test file
```
schroederc, line 1: What are these directives?

shahms: These are the directives added by the Gazelle patch to support per-file test generation, rather than per-package.

schroederc (approved):

PR #5765, opened 2023-07-27
chore(cxx_indexer): remove unused marked_source flags

zrlk:
In service of #5705


shahms:
This seems good, but shouldn't we be generally using `PARAMETER_LOOKUP_BY_PARAM` for C++ function marked source?

shahms (approved):

PR #5784, opened 2023-08-10
feat(serving): return estimates for all ref edge subkinds

salguarnieri:
Instead of returning a total count of ref edges, return a map containing the count of each ref subkind. With ref/writes, clients want to have more visibility into the counts of different ref kinds.


kythe/go/serving/xrefs/xrefs_test.go
```diff
@@ -1330,11 +1330,18 @@ func TestCrossReferences(t *testing.T) {
 
 	if err := testutil.DeepEqual(&xpb.CrossReferencesReply_Total{
 		Definitions: 1,
-		References:  2,
+		RefEdgeToCount: map[string]int64{
+			"%/kythe/edge/ref": 2,
```
schroederc, line 1334: Get rid of the leading `%` in the counts

kythe/go/serving/xrefs/xrefs.go
```diff
@@ -758,8 +758,11 @@ func (t *Table) CrossReferences(ctx context.Context, req *xpb.CrossReferencesReq
 		CrossReferences: make(map[string]*xpb.CrossReferencesReply_CrossReferenceSet, len(req.Ticket)),
 		Nodes:           make(map[string]*cpb.NodeInfo, len(req.Ticket)),
 
-		Total: &xpb.CrossReferencesReply_Total{},
+		Total: &xpb.CrossReferencesReply_Total{
+			RefEdgeToCount: make(map[string]int64),
```
schroederc, line 762: Add these together for the existing References field until it is removed.

schroederc (changes requested):

kythe/go/serving/xrefs/xrefs_test.go
salguarnieri: done

kythe/go/serving/xrefs/xrefs.go
salguarnieri: I added the old code back in until the field can be deleted

kythe/go/serving/xrefs/xrefs_test.go
```diff
@@ -1331,10 +1331,18 @@ func TestCrossReferences(t *testing.T) {
...
+		RefEdgeToCount: map[string]int64{
+			"/kythe/edge/ref": 2,
+		},
 	}, reply.Total); err != nil {
 		t.Error(err)
 	}
-	if err := testutil.DeepEqual(&xpb.CrossReferencesReply_Total{}, reply.Filtered); err != nil {
+	wantFiltered := &xpb.CrossReferencesReply_Total{
+		RefEdgeToCount: map[string]int64{
+			"/kythe/edge/ref": 0,
```
schroederc, line 1342: Do we want these superfluous zero counts?

salguarnieri: I added code to remove them.

PR #5794, opened 2023-08-18
fix(verifier): support inspections using the new solver

zrlk:
This PR adds support for inspections using the new solver. Currently, inspections get access to the internal AST; this PR begins to move things over to just giving them access to string representations (as the new solver doesn't use the verifier AST during execution).

### DIFF
--- a/kythe/cxx/verifier/assertion_ast.h
+++ b/kythe/cxx/verifier/assertion_ast.h
@@ -375,17 +375,17 @@ using Database = std::vector<AstNode*>;
 using AnchorMap = std::multimap<std::pair<size_t, size_t>, AstNode*>;
 
 /// An EVar whose assignment is interesting to display.
-class Inspection {
+struct Inspection {
  public:
   enum class Kind {
     EXPLICIT,  ///< The user requested this inspection (with "?").
     IMPLICIT   ///< This inspection was added by default.
   };
   std::string label;  ///< A label for user reference.
-  Kind kind;          ///< Whether this inspection was added by default.
   EVar* evar;         ///< The EVar to inspect.
+  Kind kind;          ///< Whether this inspection was added by default.
   Inspection(const std::string& label, EVar* evar, Kind kind)
-      : label(label), kind(kind), evar(evar) {}
+      : label(label), evar(evar), kind(kind) {}
 };
 
 }  // namespace verifier
--- a/kythe/cxx/verifier/souffle_interpreter.h
+++ b/kythe/cxx/verifier/souffle_interpreter.h
@@ -39,13 +39,16 @@ struct SouffleResult {
 /// \param inspect the inspection callback that will be used against the
 /// provided list of inspections; a false return value stops iterating through
 /// inspection results and fails the solution, while a true result continues.
+/// If set, `value` is a string representation of the assignment of the EVar to
+/// which the inspection refers.
 /// \param get_symbol returns a string representation of the given symbol
 /// \return a `SouffleResult` describing how the run went.
 SouffleResult RunSouffle(
     const SymbolTable& symbol_table, const std::vector<GoalGroup>& goal_groups,
     const Database& database, const AnchorMap& anchors,
     const std::vector<Inspection>& inspections,
-    std::function<bool(const Inspection&, std::optional<std::string_view>)>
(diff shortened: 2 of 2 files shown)


kythe/cxx/verifier/assertion_ast.h
```diff
@@ -375,16 +375,22 @@ using Database = std::vector<AstNode*>;
...
   };
   std::string label;  ///< A label for user reference.
-  EVar* evar;         ///< The EVar to inspect.
   Kind kind;          ///< Whether this inspection was added by default.
   Inspection(const std::string& label, EVar* evar, Kind kind)
-      : label(label), evar(evar), kind(kind) {}
+      : label(label), kind(kind), evar(evar) {}
+
+ private:
+  friend class Verifier;
```
shahms, line 390: friendship is only necessary in the face of private members

kythe/cxx/verifier/assertions_to_souffle.cc
```diff
@@ -158,17 +181,47 @@ bool SouffleProgram::LowerGoal(const SymbolTable& symbol_table, AstNode* goal) {
 }
 
 bool SouffleProgram::Lower(const SymbolTable& symbol_table,
-                           const std::vector<GoalGroup>& goal_groups) {
+                           const std::vector<GoalGroup>& goal_groups,
+                           const std::vector<Inspection>& inspections) {
   code_.clear();
   if (emit_prelude_) {
+  }
```
shahms, line 188: just remove the empty conditional?

kythe/cxx/verifier/souffle_interpreter.cc
```diff
@@ -141,28 +141,68 @@ class KytheReadStreamFactory : public souffle::ReadStreamFactory {
...
+
 class KytheWriteStream : public souffle::WriteStream {
  public:
   explicit KytheWriteStream(
       const std::map<std::string, std::string>& rw_operation,
       const souffle::SymbolTable& symbol_table,
-      const souffle::RecordTable& record_table, size_t* output)
+      const souffle::RecordTable& record_table,
+      const std::vector<EVarType>& output_types,
+      const std::function<std::string(size_t)>& get_symbol, KytheOutput* output)
```
shahms, line 155: It's generally advisable to avoid retaining pointers to parameters which are passed by const-ref as they can easily bind to temporaries and result in use-after-free.

It's likely these should be passed by pointer, copied to the members or annotated as `ABSL_ATTRIBUTE_LIFETIME_BOUND`

kythe/cxx/verifier/souffle_interpreter.cc
```diff
@@ -182,28 +223,40 @@ class KytheWriteStreamFactory : public souffle::WriteStreamFactory {
...
   }
 
-  size_t GetOutput(size_t id) const { return outputs_[id]; }
+  const KytheOutput& GetOutput(size_t id) const { return outputs_[id]; }
 
  private:
-  std::vector<size_t> outputs_;
+  std::vector<KytheOutput> outputs_;
+  std::vector<EVarType> output_types_;
+  const std::function<std::string(size_t)>& get_symbol_;
```
shahms, line 235: you probably just want to store this member by value

kythe/cxx/verifier/souffle_interpreter.h
```diff
@@ -36,12 +39,15 @@ struct SouffleResult {
...
-SouffleResult RunSouffle(const SymbolTable& symbol_table,
-                         const std::vector<GoalGroup>& goal_groups,
-                         const Database& database, const AnchorMap& anchors,
-                         const std::vector<Inspection>& inspections,
-                         std::function<bool(const Inspection&)> inspect);
+SouffleResult RunSouffle(
+    const SymbolTable& symbol_table, const std::vector<GoalGroup>& goal_groups,
+    const Database& database, const AnchorMap& anchors,
+    const std::vector<Inspection>& inspections,
+    std::function<bool(const Inspection&, std::optional<std::string_view>)>
```
shahms, line 48: What is the second optional param for? it's not documented that I can see an all uses seem to use the unhelpful parameter name `o`

Is this the output?

shahms (approved):

PR #5795, opened 2023-08-19
fix(typescript_indexer): improve handling of anonymous functions

nbeloglazov:
This PR improves function that returns TS Symbol for Node. Previously it handled only named Nodes such as functions with names. Now we can get Symbols of anonymous functions as well. This fixes certain aspects of anonymous functions such as adds missing `childof` edges.  And additionally it brings close handling of named and anonymous entities. 

Unfortunately it relies on using private Node.symbol property. I considered alternative options such as:

1. Support two different branches: named nodes with Symbol and anonymous nodes without Symbols. That would require re-implementing a [SymbolVNameStore](https://github.com/kythe/kythe/blob/a4e2b7822d4857d26a87341b5b4c3a014b74cc30/kythe/typescript/indexer.ts#L112C5-L112C5) cache to deduplicate vnames for anonymous nodes. Keeping both in sync would be challenging and confusing.

2. For anonymous nodes, where we can't get Symbols using public API, create our own Symbol instance. Symbol is public interface so we can create objects that implement it. Would also need to introduce Node=>FakeSymbol cache. 

Both approaches seem more complicated. So I decided to use private `Node.symbol` API. If future TS versions remove it - we'll have to switch to one of the alternative approaches.


kythe/typescript/indexer.ts
```diff
@@ -216,11 +216,32 @@ class StandardIndexerContext implements IndexerHost {
...
   getSymbolAtLocation(node: ts.Node): ts.Symbol|undefined {
-    return this.typeChecker.getSymbolAtLocation(node);
+    // Practically any interesting node has a Symbol: variables, classes, functions.
+    // Both named and anonymous have Symbols. We tie Symbols to Vnames so its
+    // important to get Symbol object for as many nodes as possible. Unfortunately
+    // Typescript doesn't provide good API for extracting Symbol from Nodes.
+    // It is supported well for named nodes, probably logic being that if you can't
+    // refer to a node then no need to have Symbol. But for Kythe we need to handle
+    // anonymous nodes as well. So we do hacks here.
+    // See similar bugs that hasn't been resolved though propely:
```
shahms, line 226: `propely` => `properly`

"See similar bugs that haven't been resolved properly"

shahms (commented):
Just a grammar/spelling nit.

While I worry about depending on an unstable API, it's at least more isolated than the internal Java APIs used by the Java indexer.

shahms (approved):

kythe/typescript/indexer.ts

PR #5823, opened 2023-08-31
fix(cxx_indexer): use operator symbols for markedsource; add flag

zrlk:
This PR addresses a longstanding issue where we used names (operator PlusPlus) instead of tokens (operator++) for overloaded operators in C++. It also adds a flag to turn on current/future cleanup work in signature generation.

### DIFF
--- a/kythe/cxx/indexer/cxx/marked_source.cc
+++ b/kythe/cxx/indexer/cxx/marked_source.cc
@@ -502,6 +502,39 @@ class DeclAnnotator : public clang::DeclVisitor<DeclAnnotator> {
   std::vector<Annotation> annotations_;
 };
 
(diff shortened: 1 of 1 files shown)


kythe/cxx/indexer/cxx/marked_source.cc
```diff
@@ -551,6 +566,23 @@ std::string GetDeclName(const clang::LangOptions& lang_options,
   return "";
 }
 
+void CleanMarkedSource(MarkedSource* to_clean) {
```
shahms, line 569: This should be in the anonymous namespace above (or marked `static`). So should `GetDeclName`, but that's pre-existing.

shahms (approved):

PR #5827, opened 2023-09-01
feat(proto_indexer): add kinds to signatures

schroederc:


kythe/cxx/indexer/proto/marked_source.cc
```diff
@@ -46,6 +47,31 @@ bool GenerateMarkedSourceForDottedName(absl::string_view name,
   return true;
 }
 
+template <typename T>
+std::optional<MarkedSource> GenerateMarkedSourceForDescriptor(
```
shahms, line 51: this should be marked `static` or in an internal namespace

kythe/cxx/indexer/proto/marked_source.cc
```diff
@@ -17,6 +17,7 @@
 #include "kythe/cxx/indexer/proto/marked_source.h"
 
 #include "absl/strings/str_split.h"
+#include "marked_source.h"
```
shahms, line 20: This is a redundant include (please use repo-relative include paths)

PR #5881, opened 2023-10-06
build: undo option not supported by version

schroederc:


shahms (approved):
But does it work with 6.3.2?

wcalandro (approved):

PR #5887, opened 2023-10-11
test(cxx): demonstrate bug in anchor spans for lambdas

justbuchanan:
looking to actually fix this in a followup commit, but if you have any ideas offhand for where this span is generated, pointers would be appreciated :)


kythe/cxx/indexer/cxx/testdata/basic/scopes.cc
```diff
@@ -0,0 +1,4 @@
+// TODO(justinbuchanan): fix indexer to include leading '(' in lambda anchor
+//- @") {}" defines Lambda
```
shahms, line 2: Since  lambda is syntactically introduced by the capture list (`[]`), the definition should probably span the entire thing.  The parens are entirely optional here; `[]{}` is a valid lambda definition.

kythe/cxx/indexer/cxx/testdata/BUILD
```diff
@@ -959,6 +959,12 @@ cc_indexer_test(
     tags = ["basic"],
 )
 
+cc_indexer_test(
+    name = "scopes_test",
```
shahms, line 963: This should probably be named something including the word "lambda".

kythe/cxx/indexer/cxx/testdata/basic/scopes.cc
justbuchanan: Done.

shahms (approved):

PR #5902, opened 2023-10-17
Allow missing file VNames with --file_vnames on.

zrlk:
The old solver is a little too forgiving with --file_vnames=false (which is not the default); it will match against more than the empty token for the corpus/path/root of an anchor. There is one test in our suite that mixes expected behaviors (in that it tries to match against anchors with a corpus and path set while also using a script-style input with no VName). Rather than plumb through the loose matching to the new solver, instead provide a flag that allows (here) zero-cost migration to the stricter matching rules.

### DIFF
--- a/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
+++ b/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
@@ -3058,7 +3058,8 @@ bool IndexerASTVisitor::TraverseFunctionDecl(clang::FunctionDecl* FD) {
     //   };
     // TODO(shahms): Fix this upstream by getting TraverseFunctionHelper to
     // do the right thing.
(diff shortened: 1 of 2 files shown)


kythe/cxx/verifier/verifier.h
```diff
@@ -201,6 +201,9 @@ class Verifier {
   /// \brief Look for assertions in file node text.
   void UseFileNodes() { assertions_from_file_nodes_ = true; }
 
+  /// \brief Only raise a warning if a file VName is missing.
+  void WarnOnMissingFileVNames() { warn_on_missing_file_vnames_ = true; }
```
shahms, line 205: `AllowMissingFileVNames` ? something that indicates these will no longer be an error

shahms (approved):

PR #5916, opened 2023-11-03
fix(verifier): do not allow evars created in negative contexts to be inspected

zrlk:
Absent special logic (see the TODO), an attempt to inspect an evar that first appears in a negative context will result in an invalid Datalog program (because it's grounded in an aggregate and leaks out of that aggregate's scope as an output to the relation, roughly:

result(vN) :- 0 = count:{entry(vN,_,_,_,_)}.

)

### DIFF
--- a/kythe/cxx/verifier/assertions_to_souffle.cc
+++ b/kythe/cxx/verifier/assertions_to_souffle.cc
@@ -288,7 +288,8 @@ bool SouffleProgram::Lower(const SymbolTable& symbol_table,
         // witness for *why* that negative goal group failed), but this will
         // complicate error recovery. This message is still better than getting
         // a diagnostic from Souffle about a leaky witness.
-        LOG(ERROR) << "can't inspect a negated evar";
+        LOG(ERROR) << i.evar->location() << ": " << i.label
+                   << ": can't inspect a negated evar";
         return false;
       }
       auto id = FindEVar(i.evar);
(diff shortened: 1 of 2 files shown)


kythe/cxx/verifier/assertions_to_souffle.cc
```diff
@@ -270,6 +280,17 @@ bool SouffleProgram::Lower(const SymbolTable& symbol_table,
...
       }
+      if (negated_evars_.contains(i.evar)) {
+        // TODO(zarko): If we intend to preserve this restriction, it would be
+        // better to catch it earlier (possibly during goal parsing). It's
+        // possible to support it (e.g., an evar in a negative context with an
+        // inspection will be inspected if the negative context fails, giving a
+        // witness for *why* that negative goal group failed), but this will
+        // complicate error recovery. This message is still better than getting
+        // a diagnostic from Souffle about a leaky witness.
+        LOG(ERROR) << "can't inspect a negated evar";
```
shahms, line 291: Can we log some additional context, such as the evar name or line or something?

kythe/cxx/verifier/assertions_to_souffle.h
```diff
@@ -121,9 +126,27 @@ class SouffleProgram {
     return evars_.try_emplace(evar, evars_.size()).first->second;
   }
 
+  /// An association between an EVar and a Datalog variable.
+  struct FoundEVar {
+    /// The Datalog id of the EVar.
+    size_t id;
+    /// Whether this was the first time this EVar was seen.
+    bool is_fresh;
```
shahms, line 134: minor: initialize these to a reasonable default to avoid accidental uninitialized access?

shahms (approved):
Thanks!

PR #5922, opened 2023-11-08
Fix builds on macOS with arm64 CPUs

amorde:
Fixes a few issues encountered when trying to compile locally on macOS with apple silicon / arm CPUs.

1. RBE config

The RBE config is missing a CC toolchain for `darwin_arm64`, causing `bazel build //...` to fail with a message like

```
tools/platforms/configs/rbe_default/cc/BUILD:47:19: in cc_toolchain_suite rule //tools/platforms/configs/rbe_default/cc:toolchain: cc_toolchain_suite '//tools/platforms/configs/rbe_default/cc:toolchain' does not contain a toolchain for cpu 'darwin_arm64'
```

Edit: The fix I initially proposed will not work, so I've removed it from this PR for now. See discussion [here](https://github.com/kythe/kythe/pull/5922#discussion_r1387206594)

2. JDK default toolchain selection

The configuration for the default JDK toolchain is hard-coded to only expose linux variants. Upon analyzing a target that requires the JDK it fails with a message like below:

```
bazel build //...                  
ERROR: /Users/eamorde/Development/kythe/kythe/javatests/com/google/devtools/kythe/analyzers/base/BUILD:3:10: While resolving toolchains for target //kythe/javatests/com/google/devtools/kythe/analyzers/base:stream_fact_emitter_test: No matching toolchains found for types @bazel_tools//tools/jdk:runtime_toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@bazel_tools//tools/jdk:runtime_toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
ERROR: Analysis of target '//kythe/javatests/com/google/devtools/kythe/analyzers/base:stream_fact_emitter_test'
```

3. JDK CPU

After fixing the default toolchain issue we encounter another issue where we attempt to run a binary compiled for linux x86.

`_remote_jdk_repository` only provides the `x86` CPU mapping and appears to be the cause of this issue.

```
ERROR: /Users/eamorde/Development/kythe/kythe/java/com/google/devtools/kythe/platform/shared/BUILD:5:13: Building kythe/java/com/google/devtools/kythe/platform/shared/libanalysis_exception.jar (1 source file) [for tool] failed: Worker process did not return a WorkResponse:

---8<---8<--- Start of log, file at /private/var/tmp/_bazel_eamorde/7396ce93e4a02c32055c37ee57a28b26/bazel-workers/multiplex-worker-8-Javac.log ---8<---8<---
/private/var/tmp/_bazel_eamorde/7396ce93e4a02c32055c37ee57a28b26/execroot/io_kythe/external/remotejdk11_linux/bin/java: /private/var/tmp/_bazel_eamorde/7396ce93e4a02c32055c37ee57a28b26/execroot/io_kythe/external/remotejdk11_linux/bin/java: cannot execute binary file
```


tools/platforms/configs/rbe_default/cc/BUILD
```diff
@@ -51,6 +51,7 @@ cc_toolchain_suite(
         "k8": ":cc-compiler-k8",
         "armeabi-v7a|compiler": ":cc-compiler-armeabi-v7a",
         "armeabi-v7a": ":cc-compiler-armeabi-v7a",
+        "darwin_arm64": ":cc-compiler-darwin-arm64",
```
amorde, line 54: even though the rbe platform is intended for linux, trying to `bazel build //...` hits this and will fail on macOS + arm with an error like below:

```
 bazel build //...
ERROR: /Users/eamorde/Development/kythe/tools/platforms/configs/rbe_default/cc/BUILD:47:19: in cc_toolchain_suite rule //tools/platforms/configs/rbe_default/cc:toolchain: cc_toolchain_suite '//tools/platforms/configs/rbe_default/cc:toolchain' does not contain a toolchain for cpu 'darwin_arm64'
ERROR: /Users/eamorde/Development/kythe/tools/platforms/configs/rbe_default/cc/BUILD:47:19: Analysis of target '//tools/platforms/configs/rbe_default/cc:toolchain' failed
```

buildenv/java/BUILD
```diff
@@ -12,7 +12,10 @@ package(default_visibility = ["//visibility:public"])
     default_java_toolchain(
         name = "toolchain_jdk_{version}".format(version = version),
```
amorde, line 13: because this name is not unique by platform we can't call this multiple times for each platform. I wasn't sure what changing the name of this would entail, so using a platform-controlled config setting seemed like decent workaround

tools/platforms/configs/rbe_default/cc/BUILD
shahms: This is automatically generated and should not be manually updated any changes to it will be overwritten.

shahms (commented):
Thanks! In order for me to approve this PR, you'll need to sign the Google CLA.

shahms:
The lint failure:
```
diff --git a/buildenv/java/BUILD b/buildenv/java/BUILD
index 073565780..4eb317f2a 100644
--- a/buildenv/java/BUILD
+++ b/buildenv/java/BUILD
@@ -111,4 +111,3 @@ config_setting(
         "@platforms//os:linux",
     ],
 )
-
diff --git a/external.bzl b/external.bzl
index 11bfdea12..e5e3a796c 100644
--- a/external.bzl
+++ b/external.bzl
@@ -50,7 +50,7 @@ def _rule_dependencies():
                 version = version,
                 platform = platform,
             ))
-            if platform == 'macos':
+            if platform == "macos":
                 native.register_toolchains("@remotejdk{version}_{platform}_aarch64_toolchain_config_repo//:toolchain".format(
                     version = version,
                     platform = platform,
```

tools/platforms/configs/rbe_default/cc/BUILD
amorde: Ah ok thanks. How is this generated?

amorde: Found it, looks like it's generated from a tool provided by bazel-toolchains.

Their README as a [section](https://github.com/bazelbuild/bazel-toolchains#custom-execution-properties) on abstracting the RBE platform, perhaps that would solve the issue I was having here. will poke around some more but let me know if you have any suggestions for other ways to fix this

amorde: Hm I don't see a way to address this here. I'll remove this change for now.

macOS users will need to run:

```
bazel build -- //... -//tools/platforms/configs/rbe_default/...
```

instead of `bazel build //...`

amorde:
@shahms could you trigger another run of the lint job?

jaysachs:
go/gcbrun

shahms (approved):
Thanks!

PR #5933, opened 2023-11-16
feat(verifier): warn on and allow specifying default file corpora

zrlk:
This PR adds a flag to specify default file vname corpus values for anchor goals. These are chiefly useful when providing files as additional input to the verifier with --file_vnames on (the default). It's meant to help migrate indexers and test harnesses that set file corpora in anchors but do not also provide them for the corresponding file nodes (which happens surprisingly often), causing tests to fail when anchor corpora are checked (as happens in the fast solver, but has been "relaxed" in the old solver for a very long time).

### DIFF
--- a/kythe/cxx/verifier/verifier.cc
+++ b/kythe/cxx/verifier/verifier.cc
@@ -1140,7 +1140,7 @@ Verifier::InternedVName Verifier::InternVName(AstNode* node) {
           tuple->element(4)->AsIdentifier()->symbol()};
 }
 
-AstNode* Verifier::CorrectFileVName(AstNode* node) {
+AstNode* Verifier::FixFileVName(AstNode* node) {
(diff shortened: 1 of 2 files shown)


kythe/cxx/verifier/verifier.cc
```diff
@@ -1137,6 +1140,32 @@ Verifier::InternedVName Verifier::InternVName(AstNode* node) {
           tuple->element(4)->AsIdentifier()->symbol()};
 }
 
+AstNode* Verifier::CorrectFileVName(AstNode* node) {
```
jaysachs, line 1143: FixFileVName? EnsureFileVNameHasCorpus ? Somehow I don't immediately parse "Correct" as a verbe.

jaysachs (approved):

PR #5939, opened 2023-11-16
feat(verifier): avoid the arity ceiling in outputs

zrlk:
The new solver/Souffle has a hard limit on the arity of relations. This is tunable but requires a change to the library. Before this PR, that limit (20 as I write this) was easy to reach through inspections, especially when `--annotated_graphviz` is on (which adds an inspection to *every* evar). Luckily, this limit does not apply to records. This PR changes lowering to use a unary output relation when inspections are on. The single element is a record containing the values of the inspected evars.

### DIFF
--- a/kythe/cxx/verifier/assertions_to_souffle.cc
+++ b/kythe/cxx/verifier/assertions_to_souffle.cc
@@ -312,7 +312,10 @@ bool SouffleProgram::Lower(const SymbolTable& symbol_table,
         inspections_.push_back(i.evar);
       }
     }
-    if (!result_tyspec.empty()) {
+    if (!result_spec.empty()) {
+      // result_ty has been defined and will always be a record with at least
+      // one element; we need to terminate both the type definition and the
+      // record expression.
       absl::StrAppend(&result_tyspec, "]");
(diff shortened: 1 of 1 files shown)


kythe/cxx/verifier/assertions_to_souffle.cc
```diff
@@ -294,19 +298,31 @@ bool SouffleProgram::Lower(const SymbolTable& symbol_table,
...
+        absl::StrAppend(&result_tyspec,
+                        result_tyspec.empty() ? ".type result_ty = [" : ", ",
+                        "rv", id, ":",
                         type->second == EVarType::kVName ? "vname" : "number");
+
         inspections_.push_back(i.evar);
       }
     }
+    if (!result_tyspec.empty()) {
+      absl::StrAppend(&result_tyspec, "]");
```
jaysachs, line 316: So it looks like the corresponding `[` do not _both_ depend on `result_tyspec`; one does, but the other on `result_clause`. Maybe the logic lines up; but it would be nice if that were clearer.

jaysachs (approved):

PR #5946, opened 2023-11-23
flag --short is no longer defined, remove it from docs

cstroe:
When I run `kythe help --short`, I get the following:

```
flag provided but not defined: -short
help [<subcommand>]:
	With an argument, prints detailed information on the use of
	the specified subcommand. With no argument, print a list of
	all commands and a brief description of each.
```

And the command returns status code `2`.

# What should happen

We should not try using the `--short` flag in our docs, it seems it was removed.

### DIFF
--- a/kythe/docs/kythes-command-line-tool.txt
+++ b/kythe/docs/kythes-command-line-tool.txt
@@ -45,34 +45,32 @@ $ kythe help
 
 .Output
 ----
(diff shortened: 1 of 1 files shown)


cstroe:
Looks like the command output was out of date also, updated that too.

jaysachs:
go/gcbrun

shahms (approved):

PR #5951, opened 2023-12-05
chore: update absl/protobuf, use NoDestructor

zrlk:


external.bzl
```diff
@@ -195,7 +195,7 @@ def _cc_dependencies():
         github_archive,
         name = "com_google_absl",
         repo_name = "abseil/abseil-cpp",
-        commit = "276f88cb77dd543ae9cc4ed55c08fb5f74f405ea",
+        commit = "71d553b12397ef81e9111b4fa21c68af3c0bf8b9",
```
shahms, line 198: This is likely to fail due to missing `<filesystem>`. I'm pushing a PR now to update the RBE base image to something more recent which should at least facilitate fixing that problem, if it doesn't do so directly.

shahms:
Oh, you're also going to need to regen all-the-protos.

shahms:
The dangly protos under //third_party/bazel and //kythe/go/util/riegeli are annoying :-(

shahms (approved):
Unsure what's going on with the bazel build

shahms:
https://en.cppreference.com/w/cpp/header/compare is a new-to-C++20-header which is being picked up by the `-std=c++20` flag, but I don't know where that's coming from.

PR #5970, opened 2023-12-14
feat(kzip): add option for providing input list via file

justbuchanan:
The tool is often used like:

$ kzip merge -output out.kzip $(find build/ -name "*.kzip")

This works in most cases, but can fail if there are too many files and the maximum number of commandline arguments is exceeded. This new flag is a workaround that allows you to do instead:

$ find build/ -name "*kzip" > filelist.txt
$ kzip merge -output out.kzip -input_file_list filelist.txt


kythe/go/platform/tools/kzip/mergecmd/mergecmd.go
```diff
@@ -90,7 +93,15 @@ func (c *mergeCommand) Execute(ctx context.Context, fs *flag.FlagSet, _ ...any)
 			vfs.Remove(ctx, tmpName)
 		}
 	}()
-	archives := fs.Args()
+	var archives []string
+	if c.inputFileList != "" {
```
jaysachs, line 97: May allow `"-"` to mean stdin, so the following would work?

```kzip merge -output out.kzip -input_file_list -```

Comment (justbuchanan) [reply kythe/go/platform/tools/kzip/mergecmd/mergecmd.go:101]: :+1: done

File: kythe/go/platform/tools/kzip/mergecmd/mergecmd.go [hunk 2]
```diff
@@ -90,7 +93,15 @@ func (c *mergeCommand) Execute(ctx context.Context, fs *flag.FlagSet, _ ...any)
...
+		archives, err = fileListFromTextFile(c.inputFileList)
+		if err != nil {
+			return c.Fail("Error reading input file list: %v", err)
+		}
+	} else {
+		archives = fs.Args()
```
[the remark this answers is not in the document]
justbuchanan: Done. I added a check that prints an error and exits if you try to use both.

PR #5971, opened 2023-12-15
fix(verifier): give more useful diagnostics during recovery

zrlk:
We might want to consider continuing recovery after lowering fails. In practice this happens when a program with inspections is partially lowered (because the full program can't be solved) and some inspected evars never appear, so are never given types. Note that we'll also need to fix the output record/expression as those datalog variables will be unconstrained.

This PR does a simpler thing and cleans up errors during recovery (to clarify that the 'no type found for X' messages are expected); it also selects a more relevant error goal.

### DIFF
--- a/kythe/cxx/verifier/testdata/singleton_inspect_expected_error_fs.txt
+++ b/kythe/cxx/verifier/testdata/singleton_inspect_expected_error_fs.txt
@@ -1,3 +1,4 @@
+ (during error recovery) evar typing missing for v1
 Could not verify all goals. The furthest we reached was:
   singleton_inspect_input_fs.txt:1:5-1:34 Nonsingleton ref Nonsingleton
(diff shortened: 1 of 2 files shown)


kythe/cxx/verifier/testdata/singleton_inspect_expected_error_fs.txt
```diff
@@ -0,0 +1,3 @@
+Could not verify all goals. The furthest we reached was:
+  singleton_inspect_input_fs.txt:1:5-1:34 Nonsingleton ref Nonsingleton
+  Goal: fact(<null>, /kythe/edge/ref, <null>, /, "")
```
jaysachs, line 3: Is it possible (and worth it) to add a test for the message during error recovery ?

jaysachs (approved):

PR #5976, opened 2023-12-18
chore: update LLVM and fix deprecation warnings

shahms:


shahms:
Heh, the schema doc had a friend-template-with-definition as well. Fixing.

PR #5999, opened 2024-01-24
fix(build): explicitly specify some deps to fix the build

justbuchanan:
without these, `bazel --bazelrc=/dev/null build --config=release //kythe/release` fails on my machine


salguarnieri (commented):
Do you know how the build ever worked without those deps?

justbuchanan:
> Do you know how the build ever worked without those deps?

I'm not sure - it's either something to do with clang (c++ modules) or bazel's `layering_check` or a combination of the two. The added dependencies are correct, but for whatever reason, transitively depending them worked fine before. Building now (on my work laptop), is requiring them to be direct dependencies.

salguarnieri (approved):

PR #6002, opened 2024-01-26
fix(java_indexer): add kind=variable subkind=import to enum

danielmoy-google:
This is documented here: https://kythe.io/docs/schema/#variable


shahms:
Thanks! I tackled that along with a missing include in #6001

PR #6003, opened 2024-01-26
chore: Update default gotool encoding to proto

salguarnieri:
Don't default to json encoded kzips for gotool.


schroederc (approved):

jingyuanliang:
This is equivalent to setting KYTHE_KZIP_ENCODING to empty by default then it's `--encoding ""` below. This would fail execution; tested with:
```
$ docker run -v /tmp/workspace:/workspace:z -w /workspace/out --entrypoint kzip gcr.io/kythe-public/golang-extractor:latest merge --encoding "" --output x.kzip buildmetadata.kzip compilations.kzip
invalid value "" for flag -encoding: unknown encoding 
--output path kzip-file*

Options:
  -append
    	Whether to additionally merge the contents of the existing output file, if it exists
  -encoding value
    	Encoding to use on output, one of JSON, PROTO, or ALL (default Proto)
  -experimental_write_units_first
    	When writing the kzip file, puts CU entries before files
  -ignore_duplicate_cus
    	Do not fail if we try to add the same CU twice
  -output string
    	Path to output kzip file
  -recursive
    	Recurisvely merge .kzip files from directories
  -rules value
    	VName rules to apply while merging (optional)
```

salguarnieri:
https://github.com/kythe/kythe/pull/6004 to fix the merge command problem

PR #6013, opened 2024-02-09
feat(typescript_indexer): support private fields 

nbeloglazov:
Adds support for private `#foo` properties: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties


nbeloglazov:
Adds support for private #foo properties: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

shahms (approved):

PR #6015, opened 2024-02-11
Set safe.directory='*' on `git` call

jingyuanliang:
This avoids the error:

```
fatal: detected dubious ownership in repository at '/tmp/workspace/gopath/src/golang.org/x/tools'
To add an exception for this directory, call:

	git config --global --add safe.directory /tmp/workspace/gopath/src/golang.org/x/tools
```

### DIFF
--- a/kythe/go/extractors/cmd/gotool/analyze_packages.sh
+++ b/kythe/go/extractors/cmd/gotool/analyze_packages.sh
@@ -92,7 +92,7 @@ fi
 # cd into the top-level git directory of our package and query git for the
 # commit timestamp.
 pushd "$(go env GOPATH)/src/${PACKAGE}"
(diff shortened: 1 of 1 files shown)


shahms:
Why is the directory being created with dubious ownership in the first place?

jingyuanliang:
It's because in the sample - https://github.com/kythe/kythe/blob/master/kythe/go/extractors/cmd/gotool/readme.md

`git clone` creates the directory as local user, then `docker run` without explicit `-u` runs as root user causing dubious ownership.

And if we use `docker run -u $(id -u):$(id -g)` instead, it fails with `docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec /usr/local/bin/analyze_packages.sh: permission denied: unknown.`

I think the easiest fix is to let git not complain.

shahms (approved):

PR #6028, opened 2024-02-17
feat(cxx_indexer): add an emit function for denotes edges

zrlk:


kythe/cxx/indexer/cxx/GraphObserver.h
```diff
@@ -835,6 +835,12 @@ class GraphObserver {
     return std::nullopt;
   }
 
+  /// \brief Records a denotes relationship as an edge in the graph.
+  /// \param DenoterId The identifier for the denoting node.
```
schroederc, line 839: Can you add a bit about the concrete vs abstract idea?

And a bit in the schema, now that I think about it?

schroederc (approved):

PR #6050, opened 2024-03-12
fix(cxx_extractor): allow empty PathEntry flag values

justbuchanan:
Without this change, leaving the `--per_file_canonicalization_policy` flag unset results in `[external/com_google_absl/absl/flags/internal/flag.cc : 598] RAW: Flag per_file_canonicalization_policy (from kythe/cxx/extractor/cxx_extractor_bazel_main.cc): string form of default value '' could not be parsed; error=missing @ delimiter between path and policy`

### DIFF
--- a/kythe/cxx/common/path_utils.cc
+++ b/kythe/cxx/common/path_utils.cc
@@ -314,6 +314,9 @@ std::string JoinPath(absl::string_view a, absl::string_view b) {
 
 bool AbslParseFlag(absl::string_view text, PathCanonicalizer::PathEntry* entry,
                    std::string* error) {
+  if (text.empty()) {
+    return true;
+  }
   size_t pos = text.find('@');
   if (pos == text.npos) {
     *error = "missing @ delimiter between path and policy";


kythe/cxx/common/path_utils.cc
```diff
@@ -314,6 +314,9 @@ std::string JoinPath(absl::string_view a, absl::string_view b) {
 
 bool AbslParseFlag(absl::string_view text, PathCanonicalizer::PathEntry* entry,
                    std::string* error) {
+  if (text.empty()) {
+    return true;
```
shahms, line 318: When returning `true` the `entry` needs to be set to something, likely just `*entry = {}`.

The `absl::StrSplit` in `vector<PathCanonicalizer::PathEntry>` below probably needs to `SkipEmpty` as a more comprehensive fix.

justbuchanan: :+1: changed to use `SkipEmpty` below and removed the `text.empty()` check.

Do we still need something like `*entries = {}` or will `entries` point to a default-constructed type?

shahms: I think omitting that is fine.  Abseil flags should support `std::optional<T>` flags these days to provide a reasonable default.

shahms (approved):

amorde:
Thank you for fixing this! I've been poking at using kythe for some projects I work on and I was regularly hitting this and didn't understand why.

PR #6057, opened 2024-03-13
chore: comment that exports list depends on code the extractor is run against

salguarnieri:


cushon:
> Depending on what code the extractor is run against, more items may need to be added to this list.

LGTM. If you wanted to get more specific, the issue is that if the code being run against does annotation processing, the extractor loads and executes those annotation processors during compilation, and that code could use internal APIs that require additional `--add-exports`. Overall this hopefully shouldn't be a super common pattern, though.

PR #6066, opened 2024-03-19
Replace instances of to_proto() method with proto.encode_text(..)

c-mita:
The to_proto and to_proto methods on struct are deprecated and will be
removed. The "json" and "proto" builtin modules should be used instead.

This replaces instances of `foo.to_proto()` with `proto.encode_text(foo)`.


shahms (approved):
Thanks!

PR #6067, opened 2024-03-19
feat: define a Language schema enum

schroederc:


jaysachs (approved):

update_generated_sources.sh
shahms, line 1: Can we not add random one-off scripts to the workspace root?

PR #6075, opened 2024-03-20
feat(testing): provide ability to save logs from indexers in verifier_test

nbeloglazov:
This PR enhances `verifier_test` in a few ways. 

1. Adds `KytheEntries.log` and `KytheEntries.name` fields. 
  `KytheEntries` is a provider created by bazel actions that execute indexers. The new `log` field contains path of a log file from the indexer execution. It allows to propagate log files to the `verifier_test` which later saves them in output directory.

2. Adds `KytheEntryProducerInfo.name` field.
  `KytheEntryProducerInfo` is a provider that contains shell script that executes indexers. `verifier_test` later runs these scripts as part of tests. With this new change logs from these scripts are stored in separate files (one per indexer execution) in test output folder. 

This change is a required to improve developer experience who write Kythe integration tests that involve multiple indexers. It's mostly targeting Google infra where files stored in TEST_UNDECLARED_OUTPUTS_DIR are displayed in test result UI.  For bazel users it might be a regression because bazel simply collects all files from TEST_UNDECLARED_OUTPUTS_DIR in a zip file that developer later has to inspect manually.


shahms:
```
All changes made by hooks:
diff --git a/tools/build_rules/verifier_test/verifier_test_impl.bzl b/tools/build_rules/verifier_test/verifier_test_impl.bzl
index 66dc9ef5a..09d3e2af6 100644
--- a/tools/build_rules/verifier_test/verifier_test_impl.bzl
+++ b/tools/build_rules/verifier_test/verifier_test_impl.bzl
@@ -57,7 +57,6 @@ def _verifier_test_impl(ctx):
         else:
             sources.append(src.files)
 
-
     # Check that all deps have unique names.
     unique_names = {}
     for dep in ctx.attr.deps:
```

nbeloglazov:
@shahms PTAL

shahms (approved):

PR #6090, opened 2024-04-04
Add `load()`s for the Bazel builtin java rules to `.bzl` files

hvadehra:
This is in preparation for the rules to be moved out of Bazel and into `@rules_java`


hvadehra:
btw, I would have liked to update the `rules_java` version as well, but since we dropped jdk20 support in the latest version, that required me to update toolchain registrations, and it seems things are not quite correctly setup here. For eg:

`bazel build --nobuild //kythe/javatests/...`  doesn't work with `--config=java20` (nor `java19`)

shahms (approved):
Agreed, updating toolchains can be challenging as is best left to its own PR.

PR #6094, opened 2024-04-08
feat(cxx_indexer): emit name nodes for flags

schroederc:


kythe/cxx/indexer/cxx/GoogleFlagsLibrarySupport.cc
```diff
@@ -172,28 +174,55 @@ static GraphObserver::NodeId NodeIdForFlag(GraphObserver& Observer,
...
+
+static void RecordFlagName(GraphObserver& Observer,
+                           const GraphObserver::NodeId& FlagNodeId,
+                           const clang::VarDecl* Decl,
+                           clang::SourceRange Range) {
+  const auto& Context = Decl->getASTContext();
+  const auto& SM = Context.getSourceManager();
+  const char* begin = SM.getCharacterData(Range.getBegin());
+  const char* end = SM.getCharacterData(Range.getEnd());
+  if (begin >= end) {
```
shahms, line 200: This is a potentially incorrect way of retrieving the source text for a given range as it assumes both locations reside within the same context and requires the range be a "character range" rather than a "token range" (the latter is generally the default interpretation of `clang::SourceRange`).

You likely want to use `clang::Lexer::getSourceText` for this, which handles all of the above properly (and explicitly documents the intended range interpretation).

```
bool Invalid = false;
llvm::StringRef FlagName = clang::Lexer::getSourceText(
    clang::CharSourceRange::getCharRange(Range),
    Context.getSourceManager(),
    Context.getLangOptions(),
    &Invalid);
if (Invalid) return;
...
```

kythe/cxx/indexer/cxx/GoogleFlagsLibrarySupport.cc
```diff
@@ -172,28 +174,55 @@ static GraphObserver::NodeId NodeIdForFlag(GraphObserver& Observer,
...
+    return;
+  }
+  llvm::StringRef FlagName(begin, end - begin);
+  if (auto FlagNameId =
+          NodeIdForFlagName(Observer, FlagNodeId, FlagName.str())) {
```
shahms, line 205: Avoid the unnecessary copy here by having `NodeIdForFlagName` accept `llvm::StringRef` or by using `NodeIdForFlagName(..., {FlagName.data(), FlagName.size()})`

kythe/cxx/indexer/cxx/GoogleFlagsLibrarySupport.cc
shahms: `clang::Lexer::getAsCharRange` turns a token-based `clang::SourceRange` into a char-based `clang::CharSourceRange` .  That is likely the source of the previous problems as, if this works correctly, the `clang::SourceRange` was already character-based, so reinterpreting it as a token range will frequently pick up extra junk on the end.

I would say that the use in `marked_source.cc` is potentially problematic, but the checks in `IsValidRange`  should be sufficient.

PR #6104, opened 2024-04-18
feat(cxx_indexer): emit symbols for extern C function defns

zrlk:
Symbols for extern "C" function definitions are emitted as name nodes in the symbol language. A function definition node is `named` its name node. The name node uses the corpus of the function it names. (The node kind, language, edge kind, and edge direction are all up for discussion.)

### DIFF
--- a/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
+++ b/kythe/cxx/indexer/cxx/IndexerASTHooks.cc
@@ -309,7 +309,7 @@ llvm::SmallVector<const clang::Decl*, 5> GetInitExprDecls(
   if (const auto* Decl = Type->getAsRecordDecl();
       Decl && (Decl = Decl->getDefinition())) {
     for (const auto* Field : Decl->fields()) {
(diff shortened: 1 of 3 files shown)


shahms (approved):

kythe/cxx/indexer/cxx/KytheGraphObserver.cc
```diff
@@ -1090,6 +1090,17 @@ void KytheGraphObserver::assignUsr(const NodeId& node, llvm::StringRef usr,
   recorder_->AddEdge(usr_vname, EdgeKindID::kClangUsr, node_vname);
 }
 
+void KytheGraphObserver::assignLinkName(const NodeId& node,
+                                        llvm::StringRef name) {
+  VNameRef node_vname = VNameRefFromNodeId(node);
+  VNameRef linkname_vname;
+  linkname_vname.set_language("symbol");
```
jaysachs, line 1097: That's a big generic bucket. "jvm" got its own name. How about "c_symbol" ?

PR #6108, opened 2024-04-19
feat(java_indexer): add names for native methods

zrlk:
This PR emits name nodes in the csymbol language for Java native methods according to the JNI spec at
https://docs.oracle.com/en/java/javase/11/docs/specs/jni/design.html .

Doing this for the JVM indexer is left to future work, mainly thanks to an annoying detail that makes it necessary to know whether a method has multiple native overloads to determine if it will be given a symbol without arguments.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -134,7 +134,8 @@ test:prepush --test_summary=terse --test_output=errors
 test:prepush --test_tag_filters=-manual,-broken,-arc-ignore,-docker
 
 # Include additional flags for use in compile_commands.json
-build:compile-commands --features=-layering_check --copt=-Wno-everything --copt=-Wno-error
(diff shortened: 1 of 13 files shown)


kythe/java/com/google/devtools/kythe/analyzers/jvm/JvmGraph.java
```diff
@@ -111,6 +144,61 @@ private static String methodSignature(
     return parentClass.qualifiedName + "." + methodName + methodType;
   }
 
+  private static String jniMangle(String s) {
```
shahms, line 147: Is this or an equivalent not provided by the JDK?

shahms (approved):

PR #6110, opened 2024-04-21
chore: build verifier_test deps in the exec configuration

salguarnieri:


jaysachs:
go/gcbrun

jaysachs (approved):

shahms:
It looks like the Java indexer emits both. I think the issue may be that verifier_test uses both if present, which is problematic.

shahms:
In fact, this PR breaks the *whole point* of `KytheEntryProviderInfo`, which was to build and run the indexers in the target configuration to make them easier to debug when writing tests.

PR #6130, opened 2024-08-22
chore(go): update go from 1.21 to 1.23 in extractor image

prattmic:
Go itself requires at least 1.22.6 for bootstrapping as of https://go.dev/cl/606156. Thus it is no longer possible to use the extractor image on the Go repo.

Update to 1.23.0, as that is the latest release.

As an aside, it is unclear to me why there is a mismatch in the base image (1.20) and the installed version (1.21). gcr.io/cloud-builders/go doesn't seem to have a 1.23 tag yet.

### DIFF
--- a/kythe/go/extractors/cmd/gotool/Dockerfile
+++ b/kythe/go/extractors/cmd/gotool/Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update && \
     apt-get clean
 
 # Get a recent go release, remove the current system one so we don't have multiple version problems, and install the new release.
-RUN curl -LO https://golang.org/dl/go1.21.6.linux-amd64.tar.gz && rm -rf /usr/local/go && tar -C /usr/local -xzf go1.21.6.linux-amd64.tar.gz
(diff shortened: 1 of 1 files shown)


prattmic:
> As an aside, it is unclear to me why there is a mismatch in the base image (1.20) and the installed version (1.21). gcr.io/cloud-builders/go doesn't seem to have a 1.23 tag yet.

Hm, is https://github.com/kythe/kythe/blob/master/kythe/release/cloudbuild/go-builder/Dockerfile the Dockerfile for `gcr.io/cloud-builders/go`?

shahms:
> > As an aside, it is unclear to me why there is a mismatch in the base image (1.20) and the installed version (1.21). gcr.io/cloud-builders/go doesn't seem to have a 1.23 tag yet.
> 
> Hm, is https://github.com/kythe/kythe/blob/master/kythe/release/cloudbuild/go-builder/Dockerfile the Dockerfile for `gcr.io/cloud-builders/go`?

No, that's the Dockerfile for `gcr.io/kythe-repo/go-builder`

lfolger (approved):

prattmic:
@shahms Thanks for the clarification!

PR #6133, opened 2024-08-25
feat: make a fully hermetic, ephemeral, and reproducible build

filmil:
Hello all. I'm not sure how useful -- if at all -- this is for y'all, but I thought I'd share. Feel free to disregard and close the PR if it is not of interest.

[This README][rr] shows what this is about.

[rr]: https://github.com/filmil/kythe/blob/nix/README.filmil.md

----

Makes the necessary changes to allow an ephemeral `nix` instance to be used for providing all the build prerequisites.

The result is a hermetic, ephemeral and reproducible ("her") build.

As a result, provided that you have `bazel` installed, it is possible to compile the entire kythe distribution with a one-liner in shell.  It still takes a preposterous time to complete, unless you have a 96-core machine like I do. In which case it's about 10 minutes.

If you are curious why this works, see:
https://hdlfactory.com/post/2024/04/20/nix-bazel-%EF%B8%8F/


tools/platforms/configs/rbe_default/cc/cc_toolchain_config.bzl
```diff
@@ -1238,7 +1238,7 @@ def _impl(ctx):
         flag_sets = [
             flag_set(
                 actions = [ACTION_NAMES.c_compile, ACTION_NAMES.cpp_compile],
-                flag_groups = [flag_group(flags = ["-Werror"])],
```
shahms, line 1241: This is autogenerated and will just be overwritten.

setup.bzl
```diff
@@ -66,14 +67,14 @@ def kythe_rule_repositories():
     )
 
     maybe(
-        http_archive,
+        git_repository,
```
shahms, line 70: Use `http_archive` to fetch a specific commit archive if necessary; `git_repository` is wastful of bandwidth and time as it unnecessarily clones the entire repository rather than a specific commit.

external.bzl
```diff
@@ -38,10 +38,10 @@ def _rule_dependencies():
     # Specifically define and register only Java toolchains we intend to support.
     remote_jdk17_repos()
     remote_jdk19_repos()
-    remote_jdk20_repos()
+    #remote_jdk20_repos()
```
shahms, line 41: AFAIK, Kythe still needs to support Java 20.

third_party/BUILD
```diff
@@ -41,10 +41,11 @@ filegroup(
...
     srcs = ["empty.cc"],
-    linkopts = select({
-        "@bazel_tools//src/conditions:darwin": [],
-        "//conditions:default": ["-luuid"],
-    }),
+    #linkopts = select({
+        #"@bazel_tools//src/conditions:darwin": [],
+        #"//conditions:default": ["-luuid"],
+    #}),
+    deps = [ "@libossp_uuid.dev//:lib" ],
```
shahms, line 48: Seems likely that this will break building on macOS?

WORKSPACE
```diff
@@ -12,6 +13,56 @@ load("//:setup.bzl", "kythe_rule_repositories")
...
+)
+
+# Configure the C++ toolchain
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_cc_configure")
+nixpkgs_cc_configure(
+    name = "nixpkgs_config_cc",
+    repository = "@nixpkgs",
+    attribute_path = "clang_14",
+    cc_lang = "c++",
+    cc_std = "c++11",
```
shahms, line 46: This needs to be at least C++17 and will presumably be C++20 at some point in the near future (I know Google was actively move that direction when I left).

WORKSPACE
```diff
@@ -12,6 +13,56 @@ load("//:setup.bzl", "kythe_rule_repositories")
...
+)
+
+# Configure Java toolchain.
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_java_configure")
+nixpkgs_java_configure(
+    attribute_path = "jdk17.home",
+    repository = "@nixpkgs",
+    toolchain = True,
+    toolchain_name = "nixpkgs_java",
+    toolchain_version = "17",
```
shahms, line 56: This will definitely need to support a newer Java version as Google was already using at least Java 20 when I left.

.bazelrc
```diff
@@ -75,10 +75,10 @@ build:java19 --tool_java_language_version=19
...
 build:java19 --tool_java_runtime_version=remotejdk_19
 
-build:java20 --java_language_version=20
-build:java20 --tool_java_language_version=20
-build:java20 --java_runtime_version=remotejdk_20
-build:java20 --tool_java_runtime_version=remotejdk_20
+#build:java20 --java_language_version=20
+#build:java20 --tool_java_language_version=20
+#build:java20 --java_runtime_version=remotejdk_20
+#build:java20 --tool_java_runtime_version=remotejdk_20
```
shahms, line 81: You're going to need to retain Java 20 support.

setup.bzl
filmil: I had some bizarre issues with the released bit - happy to rethink it now that things generally work.

external.bzl
filmil: Noted. I think this is because nix for some reason does not seem to have jdk20... not sure why, and maybe I just didn't quite get to the bottom of this.

third_party/BUILD
filmil: Yeah, I don't know what to do about macOS, since I don't have one to try it out, and building kythe on free infra exhausts my quota eventually.

WORKSPACE
filmil: c++11 was an attempt to resolve an issue that eventually got traced to a nixos bug. Could probably be removed.

.bazelrc
filmil: Indeed, except see above. I don't intend to merge this though.

Actually my high-level question here is whether y'all would care about an approach such as this.  Such high level feedback would be most welcome.

I worked out the "her" builds with bazel which enable doing some really cool stuff with bazel, since one longer needs to bazelize the world, and one is no longer required to start praying to the Nix gods.

Applying it to `kythe` hopefully shows that her approach has merit.

But if y'all don't think this matters, I'm happy to retreat to my cave and perhaps use the approach for other stuff.

WORKSPACE
filmil: Noted. I think we can go up to jdk22, but I haven't tried it. Getting all this to work was a challenge enough.

.bazelrc
shahms: As I'm not at Google anymore, I'm not in a position to say, but it seems like a beneficial change.

Hopefully someone on the current team will be able to engage with you on the change, though.

filmil:
On Sun, Aug 25, 2024 at 1:30 PM Shahms King ***@***.***>
wrote:

> As I'm not at Google anymore, I'm not in a position to say, but it seems
> like a beneficial change.
>
o7, kind sir, your contributions remain in our fond memory.

third_party/BUILD
filmil: Reverted.

WORKSPACE
filmil: I think that so long as nix supports jdk17, there should be no need to worry.  You need to worry if you depend on the JDK available on the target, but since we install our own hermetic one, it's actually a non-issue.

In other news, I wasn't able to get any other JDK to install hermetically. I'll see what is up about that.

filmil (commented):
Removed unneeded changes.

.bazelrc
filmil: Reverted.

WORKSPACE
filmil: Reverted.

external.bzl
filmil: Reverted.

setup.bzl
filmil: Reverted.

tools/platforms/configs/rbe_default/cc/cc_toolchain_config.bzl
filmil: Reverted.

WORKSPACE
filmil: Ok, now using JDK20.

lfolger:
We want to thank you for this effort and appreciate your contribution. Unfortunately, we don't have the capacity to look into this right now. We limit our engagement to critical bug fixes for now.

We might get back to this at a later point in time.

Sorry that I don't have a better answer for you.

filmil:
It's OK, I understand. Good luck, folks.

PR #6134, opened 2024-08-29
Adapt Go extractor to new -goroot flag

fmeum:
rules_go may pass the GOROOT via the `-goroot` flag instead of an environment variable since https://github.com/bazelbuild/rules_go/commit/6f206adcc9c2527a411a002669db1b7efafbeae8.


shahms:
@lfolger

lfolger (approved):
Seems reasonable to me.

PR #6137, opened 2024-09-11
Prepare for a change in an internal javac API

cushon:
SignatureGenerator was made a non-static inner class in this commit: https://github.com/openjdk/jdk/commit/cf0d9e0e521b495a96a3de7775b1b3ae14d20d80

This forks `BlockAnonymousSignatureGenerator` and `SignatureGenerator` into versions that are compatible with JDK 24+ and earlier JDK versions. Updating the BUILD files is future work.


sam-mccall:
Thank you, and sorry (again) for the delay

/gcbrun

cushon:
I don't seem to have access to the build logs for the failed actions, and am having trouble getting set up to build kythe locally. Do you have any pointers?

PR #6139, opened 2024-09-17
fix(typescript): Update getPropertyNameStr to always return a value

trevorade:
Resolves the error:

```
TS7030: Not all code paths return a value.
```

I'm assuming that the tsconfig tsc flags are not causing this error to trigger.

Compiling this code in our usage at Google results in the above errors for TS 5.6

### DIFF
--- a/kythe/typescript/indexer.ts
+++ b/kythe/typescript/indexer.ts
@@ -2037,6 +2037,7 @@ class Visitor {
         // case" is good enough for now.
         return name !== '__computed' ? name : undefined;
     }
+    return undefined;
   }
 
   /**


trevorade:
A bit of deja vu here but this was already done in:
https://github.com/kythe/kythe/pull/5957/commits/f9ccda0814d7f0cc82d3ffd0f040857b63fa2992

Unclear why it didn't stick.

trevorade:
Looking into this more, I see that this is error is occurring in TS 5.6 because `PropertyName` had `BigIntLiteral` which this switch does not handle.

So the proper fix is probably to add `case ts.SyntaxKind.BigIntLiteral:` after `case ts.SyntaxKind.NoSubstitutionTemplateLiteral:`. That can only be done though once this library updates to TS 5.6

trevorade:
I'm going to close this PR as it's not the proper fix.

When this repo moves to TS 5.6, this `switch` can be updated to handle `BigIntLiteral` which would be the proper fix.

PR #6140, opened 2024-09-23
docs: describe CU VFS inode semantics

sam-mccall:
CompilationUnit does not directly represent soft/hardlink information for its inputs. Currently a reasonable assumption is that files under different paths are unrelated, even if their content is the same.

For most languages this doesn't matter, but for the C preprocessor it does:
- the common `#pragma once` extension will allow both files to be parsed
- obscure cases of current-file-relative `#include` resolution

The "obvious" interpretation gives incorrect behavior in practice, and the C++ indexer treats same-content FileInputs as the same file to avoid this (see IndexVFS::IndexVFS in kythe/cxx/indexer/cxx/KytheVFS.cc).

This patch codifies this interpretation. I don't expect any indexers to actually change, as most languages don't care and C++ already does the right thing. But this provides important guidance to non-Kythe consumers of Kythe data.

### DIFF
--- a/kythe/proto/analysis.proto
+++ b/kythe/proto/analysis.proto
@@ -117,6 +117,10 @@ message CompilationUnit {
   // All files that might be touched in the course of this compilation.
   // Consumers of the CompilationUnit may not assume anything about the order
   // of the elements of this field.
+  //
+  // This specifies a virtual file system to be exposed to the compiler.
+  // FileInputs with equal digests are the same file, with e.g. the same inode.
+  // (This can affect the behavior of `#pragma once` in C, among other things).
   repeated FileInput required_input = 3;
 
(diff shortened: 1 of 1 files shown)


sam-mccall:
@kadircet hit subtle correctness issues with a C++ analysis based on KCD data.

After some discussion we landed on asking clients to use this model rather than complicating KCD's data structures to represent links directly. Google's internal build infra forces this interpretation for real compilations, and it seems workable. 

Kadir, does this spell out the contract in a way that makes sense?

kadircet:
thanks, yes this LGTM!

shahms:
https://github.com/kythe/kythe/pull/5927 specifically used those semantics to resolve issues around pragma once

amitcodescpp (approved):
LGTM

PR #6158, opened 2025-02-25
Upstreaming changes

varunmahajan:
These are some changes that were submitted in Google and are now fed back into the open source repository.


.bazelrc
```diff
@@ -56,10 +56,10 @@ query --@aspect_rules_ts//ts:default_to_tsc_transpiler
 common --incompatible_sandbox_hermetic_tmp
 
 # By default, compile for Java 11, run on Java 11
```
shahms, line 58: You probably want to update this comment as well.

varunmahajan: Done, thanks!

PR #6161, opened 2025-03-07
indexer/proto: correctly visit oneof

sluongng:
See
https://github.com/protocolbuffers/protobuf/blob/main/docs/implementing_proto3_presence.md#to-iterate-over-all-oneofs
for more info.

This helps silencing some error within simple proto3 files

```
syntax = "proto3";

package moo;

message Foo {
  optional int32 bar = 1;
}
```

which would yield
```
file_descriptor_walker.cc:262] Unexpected location vector [] while walking moo.proto
```

before this change.


sluongng:
Related https://github.com/kythe/kythe/commit/c23d4d5271cb22174328a6b77be6086c7dd55c22

varunmahajan (approved):

sluongng:
I just realized that I used the wrong example proto file. Fixed that in the description.

PR #6168, opened 2025-04-16
[extractors] Add an extractor for rust-project.json

### LINKED ISSUE #6167: Create a Rust extractor for rust-project.json files
It seems possible to convert rust-project.json files to .kzip for indexing by rust-analyzer-based indexers. Projects such as Fuchsia would be very interested in this, since we lost Kythe support when the old Rust extractor was deprecated.

wittrock:
Many projects using Rust with complicated build systems use rust-project.json to tell rust-analyzer about their source files. We can transform rust-project.json into a .kzip pretty trivially, without needing build system integration. This is simpler than integrating an extractor into every possible build system which large Rust projects use.

Add an extractor called rust_project_to_kzip which takes in a rust-project.json and outputs a .kzip. This kzip will have one CompilationUnit per Rust crate, and each CompilationUnit will include a copy of rust-project.json for the Kythe Rust indexer.

This extractor doesn't yet extract non-.rs files or the stdlib.

Test: added unit tests
Fixes #6167

### DIFF
--- a/kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
+++ b/kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
@@ -1,0 +1,603 @@
+/*
+ * Copyright 2025 The Kythe Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"kythe.io/kythe/go/platform/kzip"
+	"kythe.io/kythe/go/platform/vfs"
+	"kythe.io/kythe/go/util/vnameutil"
+	apb "kythe.io/kythe/proto/analysis_go_proto"
+	spb "kythe.io/kythe/proto/storage_go_proto"
+)
+
+// the central state for the rust-project extractor
+type extractor struct {
+	project             rustProject
+	projectRoot         string
+	corpus              string
+	rules               vnameutil.Rules
+	kzipWriter          kzipWriterInterface
+	requiredInputsCache *requiredInputsCache
+}
+
+// kzipWriterInterface defines the methods needed from a kzip writer.
+type kzipWriterInterface interface {
+	AddUnit(cu *apb.CompilationUnit, index *apb.IndexedCompilation_Index) (string, error)
+	AddFile(r io.Reader) (string, error)
+	Close() error // Add if Close is needed by the caller of writeCrate, though not writeCrate itself
+}
+
+// source is the set of directories which may have source files for this crate,
+// as well as directories which are explicitly excluded from containing necessary source files.
+type source struct {
+	IncludeDirs []string `json:"include_dirs"`
+	ExcludeDirs []string `json:"exclude_dirs"`
+}
+
+// dep is a dependency of a Crate
+type dep struct {
+	CrateId crateId `json:"crate"`
+	Name    string  `json:"name"`
+}
+
+type crate struct {
+	DisplayName       string   `json:"display_name,omitempty"`
+	RootModule        string   `json:"root_module"`
+	Edition           string   `json:"edition"`
+	Deps              []dep    `json:"deps"`
+	Cfg               []string `json:"cfg"`
+	CrateId           crateId  `json:"-"` // do not output
+	Label             string   `json:"label"`
+	Target            string   `json:"target"`
+	Source            source   `json:"source,omitempty"`
+	IsWorkspaceMember bool     `json:"is_workspace_member"`
+}
+
+type crateId uint32
+
+// A representation of rust-project.json.
+// Note: doesn't include all fields which exist in rust-project.json, only the ones we need to convert to a kzip.
+type rustProject struct {
+	Crates []crate `json:"crates"`
+}
+
+// Go doesn't let you use a struct with a slice member as a key to a map
+// So, define a struct which is of the type we want for a key, then give it a method to get a
+// valid key to a map out of it.
+type requiredInputMapKey string
+type requiredInputCacheKey struct {
+	includeDir   string
+	exclude_dirs []string
+}
+
+func (k *requiredInputCacheKey) mapKey() requiredInputMapKey {
+	sort.Slice(k.exclude_dirs, func(i, j int) bool {
+		return k.exclude_dirs[i] < k.exclude_dirs[j]
+	})
+	return requiredInputMapKey(strings.Join(append([]string{k.includeDir}, k.exclude_dirs...), "|"))
+}
+
+type requiredInputValue struct {
+	requiredInputs []*apb.CompilationUnit_FileInput
+	sourceFiles    []string
+}
+
+// requiredInputsCache stores the results of file discovery operations to avoid
+// redundant filesystem walks and kzip additions. It maps include directory paths
+// (relative to the project root) to the files found within them. This is
+// particularly useful when multiple crates share common dependency source directories.
+type requiredInputsCache struct {
+	cache map[requiredInputMapKey]*requiredInputValue
+}
+
+func newRequiredInputsCache() requiredInputsCache {
+	return requiredInputsCache{
+		cache: make(map[requiredInputMapKey]*requiredInputValue),
+	}
+}
+
+// Check that the project root exists and is valid.
+func ensureProjectRoot(projectRoot string) {
+	if info, err := os.Stat(projectRoot); os.IsNotExist(err) {
+		log.Fatalf("Project root does not exist: %s\n", projectRoot)
+	} else if err == nil && !info.IsDir() {
+		log.Fatalf("Project root is not a directory: %s\n", projectRoot)
+	} else if err != nil {
+		log.Fatalf("Error checking project root: %v\n", err)
+	}
+}
+
+// Parse a RustProject from a path on the disk, then make all paths within that RustProject relative to a projectRoot.
+func getProjectJSON(projectJSONPath string) rustProject {
+	projectJsonFile, err := os.Open(projectJSONPath)
+	if err != nil {
+		log.Fatalf("Error opening rust-project.json: %v\n", err)
+	}
+	defer projectJsonFile.Close()
+
+	var projectJson rustProject
+	decoder := json.NewDecoder(projectJsonFile)
+	err = decoder.Decode(&projectJson)
+	if err != nil {
+		log.Fatalf("Error decoding rust-project.json: %v\n", err)
+	}
+
+	return projectJson
+}
+
+// Make all paths in includeDirs, excludeDirs, and rootModule relative to the project root
+func (project *rustProject) relativizeProjectJson(projectRoot string) {
+	for i, crate := range project.Crates {
+		for j, includeDir := range crate.Source.IncludeDirs {
+			relativized, err := removeProjectRoot(includeDir, projectRoot)
+			if err != nil {
+				log.Printf("Error getting relative path between %s and %s, not relativizing:%v\n", includeDir, projectRoot, err)
+				continue
+			}
+
+			project.Crates[i].Source.IncludeDirs[j] = relativized
+		}
+
+		for j, excludeDir := range crate.Source.ExcludeDirs {
+			relativized, err := removeProjectRoot(excludeDir, projectRoot)
+			if err != nil {
+				log.Printf("Error getting relative path between %s and %s, not relativizing:%v\n", excludeDir, projectRoot, err)
+				continue
+			}
+			project.Crates[i].Source.ExcludeDirs[j] = relativized
+		}
+
+		relativized, err := removeProjectRoot(crate.RootModule, projectRoot)
+		if err != nil {
+			log.Printf("Error getting relative path between %s and %s, not relativizing:%v\n", crate.RootModule, projectRoot, err)
+			continue
+		}
+		project.Crates[i].RootModule = relativized
+
+	}
+}
+
+// Remove the project root from a path, returning the relative path.
+func removeProjectRoot(path string, projectRoot string) (string, error) {
+	newPath, err := filepath.Rel(projectRoot, path)
+	if err != nil {
+		log.Printf("Error getting relative path between %s and %s, not relativizing:%v\n", path, projectRoot, err)
+		return "", err
+	}
+	return newPath, nil
+}
+
+// Create a new kzip writer.
+func newKzipWriter(ctx context.Context, outputZipPath string) *kzip.Writer {
+	// Delete output zip if it already exists
+	_ = os.RemoveAll(outputZipPath)
+
+	out, err := vfs.Create(ctx, outputZipPath)
+	if err != nil {
+		log.Fatalf("Error creating output zip: %v\n", err)
+	}
+
+	kzipWriter, err := kzip.NewWriteCloser(out)
+	if err != nil {
+		log.Fatalf("Error creating kzip writer: %v\n", err)
+	}
+	return kzipWriter
+}
+
+// transitiveDepsForCrate performs a breadth-first search on the crate dependency graph
+// starting from the given `crate`. It uses the direct dependency relationships defined
+// in `crateDeps` to find all reachable crates.
+func transitiveDepsForCrate(crate crate, crateDeps map[crateId][]crateId) []crateId {
+	queue := []crateId{crate.CrateId}
+	visited := make(map[crateId]bool)
+	for len(queue) > 0 {
+		currentCrateId := queue[0]
+		queue = queue[1:] // Dequeue
+
+		// Iterate over the direct dependencies of the current crate
+		for _, dep := range crateDeps[currentCrateId] {
+			// If the dependency hasn't been visited, add it to the queue and mark it as visited
+			if _, ok := visited[dep]; !ok {
+				queue = append(queue, dep)
+				visited[dep] = true
+			}
+		}
+	}
+	transitiveDeps := make([]crateId, 0)
+	for k := range visited {
+		transitiveDeps = append(transitiveDeps, k)
+	}
+	sort.Slice(transitiveDeps, func(i, j int) bool {
+		return transitiveDeps[i] < transitiveDeps[j]
+	})
+
+	return transitiveDeps
+}
+
+// getTransitiveDeps returns a map of all crate IDs to transitive crate dependencies, including itself
+func getTransitiveDeps(crates []crate) [][]crateId {
+	crateDeps := make(map[crateId][]crateId)
+	for _, crate := range crates {
+		// Add direct dependencies
+		for _, dep := range crate.Deps {
+			crateDeps[crate.CrateId] = append(crateDeps[crate.CrateId], dep.CrateId)
+		}
+
+		// Add the crate itself as its own dependency
+		crateDeps[crate.CrateId] = append(crateDeps[crate.CrateId], crate.CrateId)
+	}
+
+	// get the transitive dependencies of each crate
+	transitiveDeps := make([][]crateId, len(crates))
+	for _, crate := range crates {
+		transitiveDepsForCrate := transitiveDepsForCrate(crate, crateDeps)
+		transitiveDeps[crate.CrateId] = transitiveDepsForCrate
+	}
+	return transitiveDeps
+}
+
+// getSourceDirs returns the set of included and excluded directories for all crates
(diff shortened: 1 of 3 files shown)


sam-mccall (commented):
Thanks John, and sorry about the delay. Easter and then I got sick...

My main high-level comments are:
- aggregating transitive sets of source dirs into a data structure, far away from where it's used, seems unnecessary (and I think gets exclude dirs wrong)
- I think passing the input `rust-project.json` through unmodified for every CU is a bit magic/risky, and would rather we synthesized one per CU that (explicitly) drops unused crates and (implicitly) drops unknown fields

Otherwise the approach looks sound to me, and I'll try it out with some existing projects.

(I haven't tried yet as I'm pretty sure e.g. the absence of implicit `include` of the `root_module` dir means things will not work with the rust-project.json generators I've been using)

kythe/go/extractors/rustproject/rust_project_to_kzip.go
```diff
@@ -0,0 +1,562 @@
...
+	if err != nil {
+		fmt.Printf("Error adding compilation unit to kzip: %v, crate %s, digest: %s\n", err, crate.Label, digest)
+		return err
+	}
+
+	return nil
+}
+
+func main() {
+	if len(os.Args) < 5 {
```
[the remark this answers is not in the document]
wittrock: Done.

kythe/go/extractors/rustproject/rust_project_to_kzip.go
```diff
@@ -0,0 +1,562 @@
...
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				return nil
+			}
+
+			// check that this file isn't in one of the dirs excluded by this crate
+			for exclude_dir := range sourceDirs.Exclude {
+				isSub, err := isSubdirectory(filepath.Join(projectRoot, exclude_dir), filepath.Dir(path))
```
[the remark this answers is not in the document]
wittrock: Yeah, this got a little crazy once I started thinking about the possible error cases. It's technically possible that excludes are symlinks or something like that, but you're right that it should really be rare. I'll make this simpler for now.

kythe/go/extractors/rustproject/rust_project_to_kzip.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+	return newPath
+}
+
+// Create a new kzip writer.
+func NewKzipWriter(ctx context.Context, outputZipPath string) *kzip.Writer {
+	kzip_options := kzip.WithEncoding(kzip.EncodingJSON)
+	out, err := vfs.Create(ctx, outputZipPath)
+	if err != nil {
+		fmt.Printf("Error creating output directory: %v\n", err)
```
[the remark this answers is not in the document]
wittrock: Done, and for all other occurrences in this file.

kythe/go/extractors/rustproject/rust_project_to_kzip.go
```diff
@@ -0,0 +1,562 @@
...
+			if include_dir != "" {
+				project.Crates[i].Source.IncludeDirs[j] = removeProjectRoot(include_dir, projectRoot)
+			}
+		}
+		for j, exclude_dir := range crate.Source.ExcludeDirs {
+			if exclude_dir != "" {
+				project.Crates[i].Source.ExcludeDirs[j] = removeProjectRoot(exclude_dir, projectRoot)
+			}
+		}
+		if project.Crates[i].RootModule != "" {
```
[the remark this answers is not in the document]
wittrock: Done.

kythe/go/extractors/rustproject/rust_project_to_kzip.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+
+	projectJSON.relativizeProjectJson(projectRoot)
+	return projectJSON
+}
+
+// Make all paths in include_dirs, exclude_dirs, and root_module relative to the project root
+func (project *rustProject) relativizeProjectJson(projectRoot string) {
+	for i, crate := range project.Crates {
+		for j, include_dir := range crate.Source.IncludeDirs {
```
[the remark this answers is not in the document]
wittrock: Sloppy!

kythe/go/extractors/rustproject/rust_project_to_kzip.go
```diff
@@ -0,0 +1,562 @@
...
+			}
+
+			// check that this file isn't in one of the dirs excluded by this crate
+			for exclude_dir := range sourceDirs.Exclude {
+				isSub, err := isSubdirectory(filepath.Join(projectRoot, exclude_dir), filepath.Dir(path))
+				if err != nil {
+					// we can't evaluate this exclude_dir for some reason. It might not exist. Bias towards including files.
+					continue
+				} else if isSub {
+					return nil
```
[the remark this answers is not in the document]
wittrock: Nice! Thanks.

wittrock (commented):
I believe I've addressed all your comments, but I've also broken the indexing pipeline somehow: my kzips look correct but the decorations service returns no decorations for expected source files. 

I'll continue to dig into that on Monday. It's likely some misconfiguration on my end.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	CrateId crateId `json:"crate"`
+	Name    string  `json:"name"`
+}
+
+type crate struct {
+	RootModule        string   `json:"root_module"`
+	Edition           string   `json:"edition"`
+	Deps              []dep    `json:"deps"`
+	Cfg               []string `json:"cfg"`
+	CompilerArgs      []string `json:"compiler_args"`
```
sam-mccall, line 74: I don't think this attribute exists, per https://rust-analyzer.github.io/book/non_cargo_based_projects.html

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	CrateId           crateId  `json:"crate_id"`
+	Label             string   `json:"label"`
+	Target            string   `json:"target"`
+	Source            source   `json:"source"`
```
sam-mccall, line 78: per the current definition, I believe if `source` is not set in the input data, this field will stay as default and you'll write `"source": {"include_dirs": [], "exclude_dirs": []}` - i.e. no sources!

You should leave it unset in this case instead. I think the easiest way is to make this an optional pointer:
```
Source *source `json:"source",omitempty`
```

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
sam-mccall: I suspect r-a may reject files with unknown fields, which might explain the lack of xrefs you're seeing.

The indexer results should have some clues if this is the case.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	return projectJson
+}
+
+// Make all paths in includeDirs, excludeDirs, and rootModule relative to the project root
+func (project *rustProject) relativizeProjectJson(projectRoot string) {
+	for i, crate := range project.Crates {
+		for j, includeDir := range crate.Source.IncludeDirs {
+			relativized, err := removeProjectRoot(includeDir, projectRoot)
+			if err != nil {
+				fmt.Printf("Error getting relative path between %s and %s, not relativizing:%v\n", includeDir, projectRoot, err)
```
sam-mccall, line 134: I think `log.Printf` is a bit better: it'll print to stderr by default, support redirection, may include timestamps etc

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+			}
+
+			shouldInclude, err := shouldIncludeFile(info, path, sourceDirs, e.projectRoot)
+			if err != nil || !shouldInclude {
+				return nil
+			}
+
+			relativePath, err := removeProjectRoot(path, e.projectRoot)
+			if err != nil {
+				return err
```
sam-mccall, line 316: we could consider making this fatalf unless/until there are known cases (in which case comment why it's ok to skip them).

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+	sort.Slice(transitiveDeps, func(i, j int) bool {
+		return transitiveDeps[i] < transitiveDeps[j]
+	})
+
+	return transitiveDeps
+}
+
+// getTransitiveDeps returns a map of all crate IDs to transitive crate dependencies
+func getTransitiveDeps(crates []crate) map[crateId][]crateId {
```
sam-mccall, line 218: the return value could include the crate itself.

This is a reasonable interpretation of "transitive dependencies" and would save duplicate code in `collectCrateSourcesImpl` and `synthesizeProjectJson`

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+func isSubdirectory(parent, sub string) bool {
+	return !strings.HasPrefix(sub, parent)
+}
+
+// whether we should include this file in the Compilation Unit based on the inclusions/exclusions in `sourceDirs`
+func shouldIncludeFile(info os.FileInfo, path string, sourceDirs source, projectRoot string) (bool, error) {
+	if info.IsDir() {
+		// check that this file isn't in one of the dirs excluded by this crate
+		for _, excludeDir := range sourceDirs.ExcludeDirs {
+			if isSubdirectory(filepath.Join(projectRoot, excludeDir), filepath.Dir(path)) {
```
sam-mccall, line 268: there's no need to check subdirectory-ness, an equality check on the cleaned path is enough
(you'll never descend into the subdirectory in the first place, due to SkipDir)

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+
+		dirRequiredInputs := []*apb.CompilationUnit_FileInput{}
+		err := vfs.Walk(ctx, filepath.Join(e.projectRoot, includeDir), func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			shouldInclude, err := shouldIncludeFile(info, path, sourceDirs, e.projectRoot)
+			if err != nil || !shouldInclude {
+				return nil
```
sam-mccall, line 311: log rather than silently returning error?
I suppose if there are lots of data files in these directories, maybe this would be spam.
Up to you.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+			continue
+		}
+
+		dirRequiredInputs := []*apb.CompilationUnit_FileInput{}
+		err := vfs.Walk(ctx, filepath.Join(e.projectRoot, includeDir), func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			shouldInclude, err := shouldIncludeFile(info, path, sourceDirs, e.projectRoot)
```
sam-mccall, line 309: nit: seems a little clearer to pass only sourceDirs.ExcludeDirs, to emphasize we've already handled IncludeDirs above

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+type collectCrateSourcesFunc func(ctx context.Context, e *extractor, sourceDirs source) ([]string, []*apb.CompilationUnit_FileInput, error)
+
+// getSourceFiles walks the file system based on the provided sourceDirs,
+// identifies relevant source files (currently only .rs files), adds them to the
+// kzip archive, and returns lists of their relative paths and FileInput protos.
+func collectCrateSourcesImpl(ctx context.Context, e *extractor, sourceDirs source) ([]string, []*apb.CompilationUnit_FileInput, error) {
+	var sourceFiles []string
+	var requiredInputs []*apb.CompilationUnit_FileInput
+
+	// cache required inputs for each include_dir
```
sam-mccall, line 293: it's possible to have different crates sharing the same include_dir but different exclude_dirs, in which case your cache is going to give the wrong answer. exclude_dirs could be part of the cache key.

Alternatively: collectCrateSourcesImpl could totally ignore exclude_dirs and include all transitive files! The extra files would be available in the VFS, but rust-analyzer would never look at them because `exclude_dirs` would still be set in `rust-project.json`. The limitation of this approach is if there's a huge file tree that's supposed to be excluded from all crates, we'll blow up the kzip size unneccesarily.

Up to you what to do here, TODO is also fine

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
sam-mccall, line 218: nit: return value here could simply be `[][]crateId`, the map is unnecessary since crate IDs are dense

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	project.Crates = append(project.Crates, currentCrate)
+
+	for _, dep := range transitiveDeps[crateId] {
+		// override IsWorkspaceMember
+		crate := crates[dep]
+		crate.IsWorkspaceMember = true
+		project.Crates = append(project.Crates, crate)
+	}
+	sort.Slice(project.Crates, func(i, j int) bool {
+		return project.Crates[i].CrateId < project.Crates[j].CrateId
```
sam-mccall, line 381: nit: I'd suggest sorting this list rather where the list of transitive deps is defined, that's less distance to plumb the required index over.
(e.g. if you implement sort.Interface then you can use external indexes and don't need to include CrateId at all)

alternatively, the BFS traversal order *is* deterministic so could be good enough. (sorting is a bit nicer though)

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	ExcludeDirs []string `json:"exclude_dirs"`
+}
+
+// dep is a dependency of a Crate
+type dep struct {
+	CrateId crateId `json:"crate"`
+	Name    string  `json:"name"`
+}
+
+type crate struct {
```
sam-mccall, line 69: I'd suggest also plumbing through the (optional) display name.
This isn't needed to analyze the code, but might show up in e.g. hovercards, and definitely shows up in indexer logs.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+
+	return bytes.NewReader(buf.Bytes())
+}
+
+// writeCrate processes a single Rust crate, determines its source files and all
+// required inputs (including transitive dependencies), constructs a Kythe
+// CompilationUnit protobuf, and adds the unit and its required files to the
+// kzip archive. It utilizes a cache (`requiredInputsBuilder`) to efficiently find
+// necessary files. The `rust-project.json` file, identified by its digest,
```
sam-mccall, line 401: "identified by its digest" looks stale.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+
+	// Synthesize a rust-project.json for this crate and add it to the CU
+	projectJsonDigest, err := e.kzipWriter.AddFile(getProjectBytes(synthesizeProjectJson(crate.CrateId, e.project.Crates, transitiveDeps)))
+	if err != nil {
+		fmt.Printf("Error adding project json to kzip: %v\n", err)
+		return err
+	}
+	projectJsonVname := &spb.VName{
+		Corpus:   e.corpus,
+		Language: "json",
```
sam-mccall, line 435: not required (and in fact files should not have languages)

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	})
+
+	compilationUnit := &apb.CompilationUnit{
+		VName: &spb.VName{
+			Corpus:   e.corpus,
+			Language: "rust",
+		},
+		RequiredInput: requiredInputs,
+		SourceFile:    crateFiles,
+		Argument:      crate.CompilerArgs,
```
sam-mccall, line 453: the Rust indexer doesn't look at `CompilationUnit.Argument`, and I don't think rust-project.json will give you anything sensible to put here anyway

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+	return nil
+}
+
+func main() {
+	projectJSONPath := flag.String("project_json", "", "Path to the rust-project.json file (required)")
+	outputZipPath := flag.String("output", "", "Path for the output kzip file (required)")
+	projectRoot := flag.String("root", "", "Directory to which all paths to source files are relative; the root directory to all possible source files. (required)")
+	corpus := flag.String("corpus", "", "Corpus name for VNames (required)")
+	vnamesJsonPath := flag.String("vnames_json_path", "", "Path to vnames.json (required)")
+	crateFilter := flag.Int("crate_filter", 0, "optional, a way to filter the output to a specific crate ID")
```
sam-mccall, line 471: nit: "a way to filter" is vague - if you'd like to keep this flag, be more specific?

(I'd suggest that "path to root module" might be a bit easier to describe and to use)

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+
+	// Many crates have the same source dirs they depend on.
+	// Caching the map of crate -> files speeds this program up by ~two orders of magnitude for the Fuchsia corpus.
+	requiredInputsCache := requiredInputsCache{
+		inputsCache:      make(map[string]([]*apb.CompilationUnit_FileInput)),
+		sourceFilesCache: make(map[string][]string),
+	}
+
+	// Delete output zip if it already exists
+	_ = os.RemoveAll(*outputZipPath)
```
sam-mccall, line 502: maybe move to newKzipWriter?

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+		if e.requiredInputsCache.inputsCache[includeDir] != nil {
+			requiredInputs = append(requiredInputs, e.requiredInputsCache.inputsCache[includeDir]...)
+			sourceFiles = append(sourceFiles, e.requiredInputsCache.sourceFilesCache[includeDir]...)
+
+			// we don't need to add this file to the kzip_writer, because we know it's already been written.
+			continue
+		}
+
+		dirRequiredInputs := []*apb.CompilationUnit_FileInput{}
+		err := vfs.Walk(ctx, filepath.Join(e.projectRoot, includeDir), func(path string, info os.FileInfo, err error) error {
```
sam-mccall, line 304: nit: I think the path joining should happen at the top of the enclosing loop, before the cache check.
Currently the `projectRoot` is part of the value computation, but not part of the cache key.
(in practice you're always calling with the same `e.projectRoot`, but...)

sam-mccall (commented):
Nice! LG with a few comments - most of these are minor/style and optional, up to you how to resolve.

let's do one more pass & then I'll merge

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
wittrock: Two problems in computer science: naming, cache invalidation, and off-by-one errors. 

Great catch. What a pain it is to use a struct with a slice member as a key to a map in Go! But I've done it.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
wittrock: We do practically have a bunch of these files in Fuchsia unfortunately, so I'd like to not log here.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -0,0 +1,552 @@
...
+				VName: vname,
+				Info: &apb.FileInfo{
+					Path:   relativePath,
+					Digest: digest,
+				},
+			})
+			return nil
+		})
+
+		if err != nil {
```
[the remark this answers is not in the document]
wittrock: You asked this last round and I forgot to respond! 

We do practically have a _ton_ of these cases in fuchsia, mostly where the file that's explicitly an include_dir doesn't exist. So it's a ton of logspam.  However, I've explicitly checked for that case and added a log here.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
wittrock: I think I've addressed this by keeping crate ID, which worked.

wittrock:
Thanks! All comments resolved, I'm going to make sure the pipeline works before we merge.

wittrock:
I've confirmed that I've got decorations back for my Fuchsia pipeline. It turned out to be three separate bugs. If you're happy, let's merge!

sam-mccall (approved):
ship it!

PR #6176, opened 2025-05-16
Add interface changes in KCD for introducing FetchCUSelector RPC.

aysinha1996:
This PR adds the interfaces changes in KCD dbs for introducing FetchCUSelector RPC.

### DIFF
--- a/kythe/go/platform/kcd/kcd.go
+++ b/kythe/go/platform/kcd/kcd.go
@@ -64,10 +64,10 @@ type Reader interface {
 	// If f returns an error, that error is returned from FilesExist.
 	FilesExist(_ context.Context, fileDigests []string, f func(string) error) error
 
-	// FetchCUSelector calls f with the digest of each known compilation matching filter.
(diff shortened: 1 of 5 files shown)


kythe/go/platform/kcd/kcd.go
```diff
@@ -63,6 +63,11 @@ type Reader interface {
 	// the store.
 	// If f returns an error, that error is returned from FilesExist.
 	FilesExist(_ context.Context, fileDigests []string, f func(string) error) error
+
+	// FetchCUSelector calls f with the digest of each known compilation matching filter.
```
creachadair, line 67: Does this need a new method? Could the existing `Find` filter be extended to support targets?

varunmahajan:
@aysinha1996 presubmit failed. Can you PTAL?

kythe/go/platform/kcd/kcd.go
aysinha1996: I will get back to this soon.

varunmahajan (approved):

PR #6177, opened 2025-05-20
[fix] Correct source_file inclusion in rust extractor

wittrock:
Previously, source file entries were polluted in the cache that speeds up directory traversal. Instead of putting only a single directory's source files in the cache, we were entering the source files for (potentially) the entire crate.

Fix: only store source files for the current directory, not the whole crate.

Test: proved that Fuchsia's corpus has no more errors like "source_file does not have a required_input entry"


sam-mccall (approved):
Thank you!

PR #6180, opened 2025-06-04
fix(cxx_extractor): Correct PathCleaner relativization for non-ancestor paths

wittrock:
The PathCleaner::Relativize method previously returned an absolute path if the path to be relativized was not a direct descendant of the cleaner's root path.

This change introduces a new helper function, ComputeRelativePath, which correctly calculates relative paths using '../' components when the target path is outside the base path's directory tree.

PathCleaner::Relativize now uses this new function, ensuring that it produces correct relative paths in such scenarios. For example, relativizing /a/b/c/d against /a/b/out/default now correctly yields ../../c/d.

The behavior of PathRealizer::Relativize remains unchanged as it continues to use the original TrimPathPrefix logic, satisfying the constraint that its existing tests should not be affected.

Unit tests for RelativizePath (which uses PathCleaner) have been updated and expanded to cover various new scenarios, including those highlighted in the issue. Tests for PathCanonicalizer that rely on PathRealizer or where PathCleaner's root encompasses the test paths continue to pass without modification.

Before this change, the value of compilation_unit.required_input[].info.path would be an absolute path if the out directory of a corpus didn't have the same path prefix as the source directory. That is, if the out directory was `/a/b/out/default` and the source directory was `/a/b/src`, we'd see `/a/b/src/foo.cc`, and not `../../src/foo.cc`. The Kythe spec demands that the path to a required_input's path is relative to the build directory, so the previous behavior was not compliant.

### DIFF
--- a/kythe/cxx/common/path_utils.cc
+++ b/kythe/cxx/common/path_utils.cc
@@ -19,12 +19,12 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <cerrno>
 #include <cstdlib>
 #include <cstring>
 #include <memory>
 #include <optional>
-#include <string>
 #include <utility>
 #include <variant>
 #include <vector>
@@ -138,6 +138,57 @@ constexpr struct VisitPattern {
 
 }  // namespace
 
+std::string ComputeRelativePath(absl::string_view target_path_sv,
+                                absl::string_view base_path_sv) {
+  // Pre-condition: Paths are absolute and cleaned.
+  if (target_path_sv.empty() || target_path_sv[0] != '/') {
+    return std::string(target_path_sv);
+  }
+  if (base_path_sv.empty() || base_path_sv[0] != '/') {
+    return std::string(target_path_sv);
+  }
+
+  if (target_path_sv == base_path_sv) {
+    return ".";
+  }
+
+  std::vector<absl::string_view> target_components = absl::StrSplit(
+      absl::StripPrefix(target_path_sv, "/"), '/', absl::SkipEmpty());
+  std::vector<absl::string_view> base_components = absl::StrSplit(
+      absl::StripPrefix(base_path_sv, "/"), '/', absl::SkipEmpty());
+
+  if (target_path_sv == "/") {
+    target_components.clear();
+  }
+  if (base_path_sv == "/") {
+    base_components.clear();
+  }
+
+  size_t common_prefix_len = 0;
+  size_t min_len = std::min(target_components.size(), base_components.size());
+  while (common_prefix_len < min_len &&
+         target_components[common_prefix_len] ==
+             base_components[common_prefix_len]) {
+    common_prefix_len++;
+  }
+
+  std::vector<std::string> relative_parts;
+
+  for (size_t i = common_prefix_len; i < base_components.size(); ++i) {
+    relative_parts.push_back("..");
+  }
+
+  for (size_t i = common_prefix_len; i < target_components.size(); ++i) {
+    relative_parts.push_back(std::string(target_components[i]));
+  }
+
+  if (relative_parts.empty()) {
+    return ".";
+  }
+
+  return absl::StrJoin(relative_parts, "/");
+}
+
 absl::StatusOr<PathCleaner> PathCleaner::Create(absl::string_view root) {
   if (absl::StatusOr<std::string> resolved = MakeCleanAbsolutePath(root);
       resolved.ok()) {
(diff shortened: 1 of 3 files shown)


shahms:
The spec should probably be clarified to remove that requirement. In practice, this will break a few things and cause sub-optimal behavior elsewhere.

1) [FindStableRoot](https://github.com/kythe/kythe/blob/7ebb77d84979f889e5a4c617b115e20e91a6b028/kythe/cxx/extractor/cxx_extractor.cc#L210) tries to find a short root which differs from the real root both to save space and to provide better isolation and caching (real build roots often contain unstable hashes which break caching).
2) As a result of the above, all of these paths will generally be wrong as they will contain an incorrect number of "../" components compared to the real root.
3) VNames are mapped based on the "relativized" source path. Prior to this, files with an absolute path would start with `/` which could be easily mapped to a consistent repository (or several, if it mattered). The default [simple_vnames.json](https://github.com/kythe/kythe/blob/7ebb77d84979f889e5a4c617b115e20e91a6b028/kythe/data/simple_vnames.json#L51) does so and would be broken by this.  Correct VName paths are much more important.
4) While less important, this will potentially cause an increase in overall CU size as non-build-root paths will be expanded to include a number of leading "../" components equivalent to the number of components in the build root.

As much as it would be nice for there to be a consistent definition and implementation for the info path, in practice it's often used as a direct map from "compiler path" to VName by indexers (although, the C++ indexer does not do so, it's not uncommon).

wittrock:
For what it's worth, I've verified that this change does make the indexing pipeline for Fuchsia work again. 

It seems that the actual error being kicked out by the indexing pipeline before my change is that the compiler invocation can't find the source file listed in the `argument` field in the CU, which looks like this:

```
required_input: [
      v_name: {
        corpus   : "fuchsia.googlesource.com/fuchsia//main"
        path     : "src/path/to/foo.cc"	
      }	# unit.required_input[886].v_name
      info: {	
        path  : "/absolute/path/src/path/to/foo.cc"
        digest: "<digest>"
      }	
]
source_file: [ "/absolute/path/src/path/to/foo.cc" ]
argument: [ "../../prebuilt/clang++", <snip>, "../../src/path/to/foo.cc"
```

We get an error like: `Errors during indexing: error reading ../../path/to/foo.cc: No such file or directory`

On a successful run (both using 0.0.53 and after this change), we have a compilation unit that looks like this:
```
required_input: [
      v_name: {
        corpus   : "fuchsia.googlesource.com/fuchsia//main"
        path     : "src/path/to/foo.cc"	
      }	# unit.required_input[886].v_name
      info: {	
        path  : "../../src/path/to/foo.cc"
        digest: "<digest>"
      }	
]
source_file: [ "../../src/path/to/foo.cc" ]
argument: [ "../../prebuilt/clang++", <snip>, "../../src/path/to/foo.cc"
```

I believe this is succeeding now because the path in the source_file and the FileInfo match the path in the `argument` line. I'm unfortunately not aware of a way to 'rebase' our compilation commands to absolute paths, but if there's a way to do so, I'd be happy to hear about it. 

Short of changing how we generate compile_commands, I think I need a way to convince kythe to output paths that match in both the argument and FileInfo.

shahms:
Can you include the `working_directory`? I suspect the issue is the one I alluded to with FindStableRoot above. The VFS should be able to find relative paths, that's one of the benefits of using it rather than a simple map. It just requires that there be enough parent directories as `..` components. 

I'm not sure what the best solution is, but one option might be preserving already-relative paths if they can't be both normalized and relative and having FindStableRoot preserve the working directory if any path begins with `../`

wittrock:
In all of the failing case (0.0.71), suceeding case (0.0.53) and case with my change, the `working_directory` in the CU is `/root`. 

I'll admit I'm confused about what FindStableRoot is supposed to do - from my reading of the function name and context it would seem like it tries to find a common ancestor for all of the various inputs, and then use that as the working directory. But then I read the code and it seems like it really just picks one of `/root`, `/build`, and `/kythe_cxx_extractor_root`. I also don't (yet!) understand how that gets used in constructing the VFS. 

Just in case you're asking about the CWD where we run the extractor, that's `/out` in the example I gave in my first comment, but in reality is just an absolute (somewhat randomized) path from our buildbots.

wittrock:
Asking some questions as well to make sure I don't use too many of your cycles waiting for replies from me. I'm very grateful for your time!

My understanding is that the way to get/keep indexing working for us is to have `source_file` and `FileInfo` paths that match the paths we have in the compiler command in `arguments`. Unfortunately that means those paths need to be of the form `../../foo.cc`, because that's how our compdb generation outputs them, and that's quite hard to change and has many dependents including our own build system.

> I'm not sure what the best solution is, but one option might be preserving already-relative paths if they can't be both normalized and relative 

By preserving already-relative paths, do you mean doing so in NormalizePath? I'm not sure what you mean by "both normalized and relative" though. I _think_  you mean if they don't share a common prefix with the root directory of PathCleaner? But I might be misunderstanding.

> having FindStableRoot preserve the working directory if any path begins with ../

This would mean preserving the absolute working directory path as the `working_directory` of the CU? For us that would be the `/out` directory, which means that paths in the indexer relative to that directory would indeed seem to work.

shahms:
On my phone, so pardon the brevity, but you understand correctly. I'm suggesting preserving the original if that path is already relative and doesn't share a common prefix with the working directory when made relative. "Normalized" in this context means without `..` or `.` components. 

`FindStableRoot` exists to do exactly that: replace the working directory with one which is "stable" in an attempt to minimize spurious differences between builds that can frequently occur in working directories (host names, hashes, user names, etc). In order to do so, it looks for the existing root in a number of places (to guard against compilations which are reliant on absolute paths). If that check fails, it just uses the original working directory. That check fails to account for the necessary number of components such that parent-traversal works reliably.

wittrock:
@shahms thanks so much. I've put up a change in #6182 which implements your suggestion. PTAL!

I'm going to close this one out.

PR #6182, opened 2025-06-06
fix(cxx_extractor): preserve upward relative paths in args

### LINKED ISSUE #6181: cxx_extractor: Upward-relative paths for source files not preserved
## Context
I work on the Fuchsia team, which has an open source kythe integration for cxx currently. I'm attempting to upgrade our kythe version in our infra from 0.0.53 (ancient!) to 0.0.71 so I can use the new Rust extractor which I landed recently. 

##  Previous behavior

Back in 0.0.53, there was no (or very little) path normalization for required_input.info.path. So if clang reported a path as `../../foo.cc`, it would get entered that way in `required_input[].info.path`. 

This worked out because the directory we run the cxx_extractor from is our build directory (which makes sense, because that's where our build commands expect to run, and all paths are relative to our build directory). 

The Fuchsia (and Chromium) tree is organized like:

```
   <root>/
  /       \ 
src/   out/
```

Most source files are in the `src` directory and the build happens in the`out` directory. This means that the path to source files from the build directory is `../src/path/to/foo.cc`. 

## New behavior

It seems that now, the PathCleaner requires that each path shares a common prefix with the CWD of the cxx_extractor, which for us is the build directory. If it doesn't, it [falls back](https://github.com/kythe/kythe/blob/7ebb77d84979f889e5a4c617b115e20e91a6b028/kythe/cxx/common/path_utils.cc#L154) to an absolute path as the value for `required_input.info.path`. 

We end up triggering this condition because our `root` in PathCleaner is `/out` and our `path` is `/src/path/to/foo`. The path doesn't have the root directory as a prefix, and so it shows up in the FileInfo as an absolute path. 

## The issue
It [seems](https://github.com/kythe/kythe/blob/7ebb77d84979f889e5a4c617b115e20e91a6b028/kythe/cxx/indexer/cxx/frontend.cc#L156) that FileInfo paths are used as the VFS paths for the indexer. 

So using the new version of Kythe, the VFS contains paths for required_input files at absolute paths (`/src/foo.cc`), while the compiler commands are looking for them at relative paths (`../src/foo.cc`, ). 

## Options
1. Obviously, the way I posted here basically restores the behavior of 0.0.53 (output relative paths to make the VFS paths in the indexer match the paths the compiler wants to look for). It seems this might have broader implications, and I don't want to land this if I'm going to break other clients. 
2. Add some option which tells PathCleaner not to fall back to absolute paths
3. Something else I haven't thought of

_Originally posted by @wittrock in https://github.com/kythe/kythe/issues/6180#issuecomment-2944959804_

wittrock:
For projects which use upward relative paths (`../<path>`), the existing PathCleaner logic was insufficient. Particularly, it would output absolute paths for inputs which were specified as ../<path>, which meant that later indexing would fail, as the `arguments` field likely referred to those inputs by their relative paths.

We address this with two fixes:

1. PathCleaner now preserves (does not modify) a relative path if it starts with ../
2. FindStableRoot now outputs the current working directory if any arguments start with ../

Many thanks to @shahms for his advice on this set of changes.

Test: Added unit tests, also verified manually that this gets the Fuchsia indexing pipeline working again
Fixes #6181

### DIFF
--- a/kythe/cxx/extractor/cxx_extractor.cc
+++ b/kythe/cxx/extractor/cxx_extractor.cc
@@ -205,30 +205,6 @@ class RequiredRoots {
   bool success_ = true;
 };
 
-/// \brief Finds a suitable stable root directory, if possible.
-/// Otherwise falls back to using the provided root.
-std::string FindStableRoot(
-    absl::string_view working_directory,
-    const RepeatedPtrField<std::string>& arguments,
-    const RepeatedPtrField<proto::CompilationUnit::FileInput>& required_input,
-    const proto::CxxCompilationUnitDetails& details) {
-  absl::ConsumeSuffix(&working_directory, "/");
-  for (absl::string_view arg : arguments) {
-    if (arg.find(working_directory) != arg.npos) {
-      LOG(WARNING) << "Using real working directory (" << working_directory
-                   << ") due to its inclusion in compiler argument: " << arg;
-      return std::string(working_directory);
-    }
-  }
-
-  RequiredRoots roots(working_directory);
-  roots.Update("required_input", required_input) &&
-      roots.Update("header_search_info", details.header_search_info().dir()) &&
-      roots.Update("system_header_prefix", details.system_header_prefix()) &&
-      roots.Update("stat_path", details.stat_path());
-  return roots.GetStableRoot();
-}
-
 google::protobuf::Any* FindMutableContext(
     kythe::proto::CompilationUnit::FileInput* file_input,
     kythe::proto::ContextDependentVersion* context) {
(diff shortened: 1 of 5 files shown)


kythe/cxx/extractor/cxx_extractor.cc
```diff
@@ -212,6 +212,13 @@ std::string FindStableRoot(
     const RepeatedPtrField<std::string>& arguments,
     const RepeatedPtrField<proto::CompilationUnit::FileInput>& required_input,
     const proto::CxxCompilationUnitDetails& details) {
+  for (const std::string& arg : arguments) {
+    if (absl::StartsWith(arg, "../")) {
+      LOG(WARNING) << "Using real working directory (" << working_directory
+                   << ") due to '..' prefix in argument: " << arg;
+      return std::string(working_directory);
+    }
+  }
```
shahms, line 221: While this is likely sufficient, the working directory check below uses `find` to remain conservative and argument format agnostic.  It wants to find both plain `<value>` and `--foo=value` style arguments.  `../` is more likely to occur in "harmless" places within arguments, though, so perhaps changing this to `StartWith("../") or `Contains("=../")` would provide a bit more protection.

Additionally, I would check that none of the `required_input[].info.path` fields start with `../` as those files may not get added to the VFS in a reasonable location without a matching number of parent directories.

wittrock: Done, added a test for the new behavior as well.

shahms: Thanks! Seems reasonable to me. @sam-mccall ?

PR #6187, opened 2025-06-30
Requires an update to Go v1.24

paritoshgoogle:
Requires an update to Go v1.24

### DIFF
--- a/external.bzl
+++ b/external.bzl
@@ -31,7 +31,7 @@ load("@rules_python//python:repositories.bzl", "py_repositories")
 
 def _rule_dependencies():
     go_rules_dependencies()
-    go_register_toolchains(version = "1.21.6")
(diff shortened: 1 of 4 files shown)


external.bzl
```diff
@@ -31,7 +31,7 @@ load("@rules_python//python:repositories.bzl", "py_repositories")
 
 def _rule_dependencies():
     go_rules_dependencies()
-    go_register_toolchains(version = "1.21.6")
+    go_register_toolchains(version = "1.24.6")
```
creachadair, line 34: Go 1.24.6 does not exist yet, the current release is 1.24.4 as of 29-Jun-2025.

paritoshgoogle: Thanks for catching. I have changed it to 1.24.4

PR #6200, opened 2025-09-09
Update xrefs for record accessor methods to their corresponding fields in Record Class.

vvk-singh:
Record fields are not present in record symbols if they belong to separate CU hence emit an anchor for record component and alias from record component anchor to record field when we encounter the CU defining the record. This way the record field accessors will redirect correctly to the record fields.


vvk-singh:
Failing because Java version doesn't yet support Records.

PR #6231, opened 2026-07-03
Fix: Rust extractor to filter out other crates' roots in source files.

MarcxJin:
The current extractor scans the input dir and grabs all the .rs files within that dir as the source files of a certain crate. As a result, if two crates reside in the same directory, they end up having the same source files, although a certain .rs file might not be used at all as a dependency of one of the craters.

In the indexer, the primary crate of that certain .rs file is selected alphabetically, meaning there are chances the wrong primary crate is selected, leading to xrefs missing.

E.g. 
"lib.rs", "cxx.rs", "util.rs" all resides in the same dir "dir".
target 1 has only "cxx.rs" (crate), input_dirs = ["dir"]
target 2 has "lib.rs"(crate), "util.rs" and depend on target 1,  input_dirs = ["dir"]
The current extractor now would create two crates, crate 1 is "cxx.rs" as primary, having "lib.rs", "cxx.rs", "util.rs" as source files. crate 2 is "lib.rs" as primary, having "lib.rs", "cxx.rs", "util.rs" as source files TOO.
Now the indexer checks "util.rs" and will consider crate 1 (alphabetically) as its crate, leading "util.rs" is not indexed as all, as when compiler works, it never visits "util.rs"


kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -485,6 +485,29 @@ func (e *extractor) writeCrate(ctx context.Context, crate crate, transitiveDeps
 		return err
 	}
 
+	// De-dupe the source files list
```
vvk-singh, line 488: NIT: End comments with a "."

Applicable everywhere.

kythe/go/extractors/cmd/rust/rust_project_to_kzip_test.go
```diff
@@ -1142,3 +1142,111 @@ func TestSynthesizeProjectJson(t *testing.T) {
 		})
 	}
 }
+
+func TestWriteCrate_RootModuleFiltering(t *testing.T) {
```
vvk-singh, line 1146: Please update TestWriteCrate() instead of creating a new test.

kythe/go/extractors/cmd/rust/rust_project_to_kzip.go
```diff
@@ -485,6 +485,29 @@ func (e *extractor) writeCrate(ctx context.Context, crate crate, transitiveDeps
...
+	var filteredCrateFiles []string
+	for _, file := range crateFiles {
+			// Never filter out the crate's own root module
+			if file == crate.RootModule {
+					filteredCrateFiles = append(filteredCrateFiles, file)
+					continue
+			}
+
+			isOtherRoot := false
+			for _, otherCrate := range e.project.Crates {
```
vvk-singh, line 498: Someone from the Rust team should be able to help us confirm if this logic is correct.

Since Sam reviewed the initial extractor code https://github.com/kythe/kythe/pull/6168.  @sam-mccall do you know who would be the best person in Rust team to help us here? Or if you can help us in confirming this?

sam-mccall:
Hi Marc,

I think the intended model here is that all files in the crate's input dirs are included as source files, even if some of them are in fact part of other crates. i.e. we actually list a superset of the files, and those supersets may overlap between crates.

I *think* this shouldn't cause a problem: when indexing a crate, the indexer should start at `RootModule` and read only the files that are included with `mod foo;`, ignoring other listed sources.

Can you explain where you see this behavior:

> In the indexer, the primary crate of that certain .rs file is selected alphabetically, meaning there are chances the wrong primary crate is selected, leading to xrefs missing.

And in particular what "primary crate" means?

(It would be useful to have a real example files used with the extractor to make this concrete: `rust-project.json` input, and `CompilationUnit` and `rust-project.json` output)

sam-mccall:
Apologies, you were totally right about this:

> Can you explain where you see this behavior:
> 
> > In the indexer, the primary crate of that certain .rs file is selected alphabetically, meaning there are chances the wrong primary crate is selected, leading to xrefs missing.
> 
> And in particular what "primary crate" means?

The behavior in question is in the indexer. I think it's clearer to change the indexer to not pick crates in such an arbitrary way: #6232 is the extractor side of that change.

PR #6232, opened 2026-07-08
rust extractor: place root module at the start of source_file

sam-mccall:
When multiple crates live in the same directory, they will have the
same set of source_files. We can't make this more precise as it involves
parsing and macro expansion.

The indexer needs to determine which crate we're trying to index.
We know what the root module is, so place it at the start of the
source_file list as a hint.


sam-mccall:
cc @MarcxJin - I think this is a somewhat cleaner fix than #6231.

+ @bmoxb as I send you the corresponding indexer change

+ @vvk-singh for Kythe approval

vvk-singh (approved):

PR #6234, opened 2026-07-15
rust: set is_workspace_member to indicate the crate being indexed

sam-mccall:
This gives an unambiguous signal of which crate is relevant to the
indexer. CompilationUnit.source_file is insufficient as presumed
sources may overlap if multiple crates are in the same directory.

This is an alternative to https://github.com/kythe/kythe/pull/6232,
but not reverting that yet until we have a solution that works.


sam-mccall:
cc @marcxjin

bmoxb (approved):

PR #6240, opened 2026-09-10
Fix data race on the fileSignatureChunk global variable in Riegeli writer.go

nickschwartz-jpg:
My team recently started using the Kythe Riegeli package. We received a Golang "Data race detector" (https://go.dev/doc/articles/race_detector) ticket regarding the "riegeli_test.go" file.

After investigating, it looks like the issue is around this `fileSignatureChunk` in `writer.go`. This variable gets initialized inside `init()`, and then `WriteTo()` modifies its `Header.DataHash` field in-place, which triggers a data race whenever multiple goroutines write Riegeli files concurrently.

So, this PR scraps the global-variable+init() strategy. We replace it with a fileSignatureChunk() factory that returns a fresh chunk every time, totally eliminating shared mutable state across writers.

With this change, riegeli_test.go no longer triggers the data race failure.


Aman50 (approved):

nickschwartz-jpg:
> Can you please update the PR description to convey the issue and how this fix resolves it? Thanks!

Done!

vvk-singh (approved):
LGTM

PR #6241, opened 2026-09-14
Update protobuf to v28.3 and Abseil to LTS 20250127.1.

Aman50:
Fixes C++ compilation on modern compilers (GCC 13+, Clang 19+) and pins checksummed releases. On modern toolchains, 52 targets fail to build on master with error: `unknown type name 'uintptr_t'`, due to missing transitive standard library headers in the old Abseil pin.

This is also a prerequisite for the rules_go upgrade that follows. rules_go v0.55.1 and later load `@com_google_protobuf//bazel/common:proto_common.bzl`, which does not exist before protobuf v27, so with protobuf at v25.2 the later commit fails at load time.

### Version Upgrades:

  * **com_google_absl: commit 71d553b1 -> LTS 20250127.1** Needed because the old commit does not include `<cstdint>` in `container_memory.h`, causing build failures on current compilers with `"unknown type name 'uintptr_t'"`. Also replaces an unchecksummed Git commit with a verified release tarball.
  * **com_google_protobuf: v25.2 -> v28.3** Updated to keep C++ protobuf in sync with modern Abseil. Kept at v28.3 because v29+ requires rules_java 7.x (Kythe pins 6.5.2).
  * **rules_cc: 0.0.9 -> 0.1.1** Required by Abseil LTS 20250127.1, which selects on `@rules_cc//cc/compiler:emscripten` (unsupported in 0.0.9).
  * **protobuf-matchers: commit 8edcd4f7 -> commit 79324778** Needed because Protobuf v28.3 changed the error collector API to `RecordError`; this commit adds upstream support for that API.

**What Changed:**

  * `external.bzl`: switch Abseil from `github_archive` to `http_archive` to point to the official release tarball URL (`github_archive` only supports raw git commit snapshots).
  * `ProtoLibrarySupport.cc`, `proto_extractor.cc`, `textproto/analyzer.cc`: update `ErrorCollector` overrides from `AddError` to `RecordError` taking `absl::string_view`, matching Protobuf v28.3 pure-virtual signatures.
  * `marked_source.cc`, `BUILD`: include `absl/log/vlog_is_on.h` and add the BUILD dep because `absl/log/log.h` no longer exports `VLOG_IS_ON`.
  * `JsonUtil.java`: register Gson adapter on `GeneratedMessage` instead of `GeneratedMessageV3` because Protobuf 28.x gencode extends `GeneratedMessage`. The old adapter matched nothing, breaking kzip proto deserialization.
[...]

PR #6242, opened 2026-09-14
Make cross-reference serving table ordering deterministic.

Aman50:
//kythe/go/serving/tools/testdata:write_tables_test fails once the Go SDK moves to 1.25 in the following commit. The failure is deterministic, not intermittent: with the tie-break reverted the test fails on every run.

### What Changed:

  * kythe/go/serving/pipeline/pipeline.go: tie-break on the target anchor's ticket in refLesser.Less. Anchors in different files can share a kind and have identical offsets. For ties, the comparator previously returned false in both directions, making output order depend on merge sort traversal.

### Verification:

  * bazel test --keep_going //... Result: 887 out of 887 tests pass (executed 7, 880 cached, 0 failed).
  
This PR is part 2 of the 4 PRs. See #6241 for the full series.


kythe/go/serving/pipeline/pipeline.go
```diff
@@ -607,6 +607,13 @@ func (refLesser) Less(a, b any) bool {
 		} else if x.TargetAnchor.Kind == y.TargetAnchor.Kind {
 			if x.TargetAnchor.Span.Start.ByteOffset == y.TargetAnchor.Span.Start.ByteOffset {
 				if x.TargetAnchor.Span.End.ByteOffset == y.TargetAnchor.Span.End.ByteOffset {
+					if x.TargetAnchor.SnippetSpan.End.ByteOffset == y.TargetAnchor.SnippetSpan.End.ByteOffset {
+						// Anchors in different files can share a kind and have
+						// identical offsets, so everything compared above can
+						// tie. Fall back to the anchor ticket to keep the
+						// ordering deterministic.
+						return x.TargetAnchor.Ticket < y.TargetAnchor.Ticket
```
vvk-singh, line 615: This seems like a new behaviour where we are also considering Ticket now.

Would making below comparison as `<=` instead of `<` would work ? If the sorting algo has changed we should be able to work with choosing the boundary conditions.

Aman50: Thank you for the feedback, @vvk-singh.

This behavior is consistent with the existing pattern being followed in this file.  As an example fragmentLesser.Less uses a very similar logic:

```
func (fragmentLesser) Less(a, b any) bool {
	x, y := a.(*decorationFragment), b.(*decorationFragment)
	if x.fileTicket == y.fileTicket {
		if len(x.decoration.Decoration) == 0 || len(y.decoration.Decoration) == 0 {
			return len(x.decoration.Decoration) == 0
		}
		return x.decoration.Decoration[0].Anchor.Ticket < y.decoration.Decoration[0].Anchor.Ticket
	}
	return x.fileTicket < y.fileTicket
}
```

edgeLesser.Less also uses a very similar logic. Summing it up, this logic is consistent with other comparators in this file.

amitcodescpp (commented):
LGTM

amitcodescpp (approved):
LGTM

PR #6244, opened 2026-09-14
Update rules_go to v0.60.0, gazelle to v0.48.0 and the Go SDK to 1.25.1.

Aman50:
Upgrades the Bazel Go toolchain from 1.21 to 1.25. Newer dependencies (including gRPC v1.83.2) declare go 1.25.0 in their go.mod, which causes Bazel's `go_repository` rules to fail fetching modules against older SDKs. Split out as an independent prerequisite to keep the gRPC security upgrade focused.

### Version Upgrades:

  * **io_bazel_rules_go: v0.41.0 -> v0.60.0** Required to support Go 1.25. Versions <= 0.55 pass a removed flag (`nocoverageredesign`) that crashes Go 1.25 compilation. Version 0.61 leaves `PATH` unset in `gazelle_binary` link actions, so `//tools/gazelle:gazelle` fails to link. Versions >= 0.62 require Bazel 8 (Kythe is on Bazel 7.1.0).
  * **bazel_gazelle: v0.32.0 -> v0.48.0** Required by rules_go 0.60; older gazelle relied on internal x/tools packages that were removed upstream.
  * **Go SDK: 1.21.6 -> 1.25.1** Hermetic toolchain version required to satisfy go >= 1.25 module requirements.
  * **platforms: 0.0.8 -> 1.1.0** Provides CPU target definitions required by rules_go 0.60.
  * **org_golang_x_tools: v0.7.0 -> v0.30.0** Brought in sync with rules_go 0.60 so existing workspace patches apply cleanly.
  * **protobuf-go: v1.31.0 -> v1.36.3** Provides protobuf edition features required by rules_go 0.60.
  * **golang/protobuf: v1.5.3 -> v1.5.4** Drops deprecated references removed by protobuf-go 1.36.
  * **doublestar/v4: v4.6.0 -> v4.9.1** Provides path-matching APIs required by gazelle 0.48.

**What Changed:**

  * `go.mod`: update `go` directive from `1.21` to `1.22` (required by `types.Unalias` in `emit.go`).
  * `external.bzl`: set `build_file_proto_mode = "disable_global"` on protobuf-go to prevent a cyclic build dependency between protoc-gen-go and protobuf.
  * `external.bzl`: define the `go_repository` wrapper locally rather than importing from `shims.bzl`, breaking a circular workspace load dependency introduced in newer rules_go.
  * `kythe/go/indexer/testdata/go_indexer_test.bzl`: update custom test rules to match rules_go 0.60's reorganized provider fields, and adjust test goal regex so standard Go 1.25 doc comment bullet lists are not mistaken for test assertions.
  * `kythe/go/indexer/emit.go`: handle type aliases in the indexer, preventing crashes on Go 1.22+ code where aliases are represented as dedicated types.
  * `third_party`: remove `gazelle-0.32.0-go_test_mode.patch` because the per-file test feature was adopted upstream in gazelle 0.48.
  * Regenerate checked-in Go proto stubs for protobuf-go v1.36.3 to satisfy Kythe sync tests.

**Note to Reviewers:**

  * 42 `*.pb.go` files in this PR are compiler-generated and can be skipped during review. They were regenerated for protobuf-go v1.36.3.
  * Files to focus on during review:
    - `go.mod`
    - `setup.bzl`, `external.bzl`
    - `kythe/go/indexer/testdata/go_indexer_test.bzl`
    - `kythe/go/indexer/emit.go`
    - `third_party/gazelle-0.32.0-go_test_mode.patch` (deleted)

**Verification:**

  * `bazel test --keep_going //...`
    Result: 887 out of 887 tests pass (0 failed).
  * `bazel test -c opt --stamp //... //kythe/docs/schema //kythe/release:release_test`
    Result: 888 out of 888 tests pass.
  * `go test -mod=readonly ./kythe/...`
    Result: PASS across all 184 Go packages.


kythe/go/indexer/testdata/go_indexer_test.bzl
```diff
@@ -46,12 +46,11 @@ def _emit_extractor_script(ctx, mode, script, output, srcs, deps, ipath, data, e
         for src in srcs
     ]
     for dep in deps:
-        gosrc = dep[GoSource]
-        path = gosrc.library.importpath
+        path = dep.importpath
```
vvk-singh, line 49: Won't this give all the deps instead of only `GoSource` related ones ?

kythe/go/indexer/emit.go
```diff
@@ -481,6 +481,11 @@ func (e *emitter) emitType(typ types.Type) *spb.VName {
 	}
 
 	switch typ := typ.(type) {
+	case *types.Alias:
+		// Go 1.22 and later materialize alias declarations as *types.Alias
+		// instead of resolving them eagerly.  Index the type the alias
+		// denotes; the alias name itself is bound by visitTypeSpec.
+		v = e.emitType(types.Unalias(typ))
```
vvk-singh, line 488: Why this? Why have we not choosen any other emit kind like emitTapp etc?

kythe/go/indexer/testdata/go_indexer_test.bzl
Aman50: I've updated the code to minimize the changes.

kythe/go/indexer/emit.go
Aman50: Without this case the indexer crashes on any file containing an alias.

As an example, kythe/go/indexer/testdata/types.go has a scenario:
````
type StringerA = fmt.Stringer
````

Earlier go/types used to store the type as interface type; after the Go version update, it now records a distinct `*types.Alias` object. `emitType` is a switch on the type kind and has no case for `Alias`, so it returns nil, which then gets dereferenced by writeEdge in emit.go:1592.

Why `types.Unalias`?
Because `type A = B` means A and B are the same type, so the edge should point at B's node. This matches what the indexer emitted before this upgrade, so the graph is unchanged.

Going with other node kinds like tapp would be wrong since a tapp is a constructor applied to arguments (`[]T`, `*T`, `map[K]V`). An alias has neither, and it would give A a different VName from B.

Comment (amitcodescpp) [reply kythe/go/indexer/emit.go]: Rather than adding a dedicated case *types.Alias: branch that recurses into e.emitType, we can unalias directly in the type switch header.
```
switch typ := types.Unalias(typ).(type) {
case *types.Named:
    ...
```

It would unwrap any alias or chain of aliases (type A = B = int) in a single step without extra recursion, and ensures every subsequent case branch receives the canonical underlying type.

Aman50: Nice catch, done.

amitcodescpp (approved):
LGTM

